PR #7, opened 2024-01-22
Removed that prefetch instructions can throw exceptions

francislaus:
I removed that prefetch instructions can throw exceptions. This is in line with the Zicbop rationale on prefetch instructions:

"A cache-block prefetch instruction is permitted to access the specified cache block whenever a load instruction,
store instruction, or instruction fetch is permitted to access the corresponding physical addresses. If access to the
cache block is not permitted, a cache-block prefetch instruction does not raise any exceptions and shall not
access any caches or memory. During address translation, the instruction does not check the accessed and dirty
bits and neither raises an exception nor sets the bits."

Furthermore, I added Zicbop as a prerequisite for the two `prefetch.w` instructions.

### DIFF
--- a/src/insns/prefetch.i.adoc
+++ b/src/insns/prefetch.i.adoc
@@ -38,8 +38,8 @@ likely to be accessed by an instruction fetch in the near future. The encoding
 is only valid if imm[4:0]=0. The authorising capability for this operation is
 `cs1`. This instruction does not throw any exceptions. However, following
 <<CHERI_SPEC>>, this instruction does not perform a prefetch if it is
-not authorized by `cs1`. This instruction does not issue a prefetch if one or
-more of the following conditions of `cs1` are met:
+not authorized by `cs1`. This instruction does not perform a memory access
+if one or more of the following conditions of `cs1` are met:
 * The tag is not set
 * The sealed bit is set
 * No bytes of the cache line requested is in bounds
--- a/src/insns/prefetch.i.adoc
+++ b/src/insns/prefetch.i.adoc
@@ -46,8 +46,6 @@ likely to be accessed by an instruction fetch in the near future. The encoding
 is only valid if imm[4:0]=0. The authorising capability for this operation is
 <<ddc>>.
 
-:prefetch_i:
-include::cbo_exceptions.adoc[]
 
 Prerequisites for PREFETCH.I.CAP::
 Zicbop, {cheri_base_ext_name}
--- a/src/insns/prefetch.i.adoc
+++ b/src/insns/prefetch.i.adoc
@@ -38,7 +38,7 @@ likely to be accessed by an instruction fetch in the near future. The encoding
 is only valid if imm[4:0]=0. The authorising capability for this operation is
 `cs1`. This instruction does not throw any exceptions. However, following
 <<CHERI_SPEC>>, this instruction does not perform a prefetch if it is
-not authorized by `cs1`. This instruction evaluates to a NOP if one or
+not authorized by `cs1`. This instruction does not issue a prefetch if one or
 more of the following conditions of `cs1` are met:
 * The tag is not set
 * The sealed bit is set
--- a/src/insns/prefetch.i.adoc
+++ b/src/insns/prefetch.i.adoc
@@ -38,7 +38,12 @@ likely to be accessed by an instruction fetch in the near future. The encoding
 is only valid if imm[4:0]=0. The authorising capability for this operation is
 `cs1`. This instruction does not throw any exceptions. However, following
 <<CHERI_SPEC>>, this instruction does not perform a prefetch if it is
-not authorized by `cs1`.
+not authorized by `cs1`. This instruction evaluates to a NOP if one or
+more of the following conditions of `cs1` are met:
+* The tag is not set
+* The sealed bit is set
+* No bytes of the cache line requested is in bounds
+* The X permission is not set
 
 Legacy Mode Description::
 A PREFETCH.I instruction indicates to hardware that the cache block whose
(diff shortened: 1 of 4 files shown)


andresag01:
@francislaus : Thanks for PR. This issue with prefetches was discussed extensively and it was decided that it is best to prevent software from arbitrarily execute prefetch instructions without at least some minimum requirements on the authorising capability.  The rationale for this is to reduce the attack surface through prefetches.

The discussion is summarised in this issue from the original CHERI technical report: https://github.com/CTSRD-CHERI/cheri-specification/issues/65

I agree that the prerequisites need fixing!

francislaus:
@andresag01 Thank you for your comment. I think there might have been a misunderstanding. The PR does not intend on allowing prefetches to execute if the authorising capability does not allow. I think the way to go in this case is to make them NOPs as suggested by Jessica.

However, what my PR intends on doing is to not throw exceptions in this case. That is exactly what the (Zicboc) RISC-V spec does. It does not allow access to the cache block, but it does also not raise exceptions. I think this would be sensible because pre-fetch instructions are HINTs and can be safely ignored. I think this PR is in line with the discussion you mentioned as well as the Zicboc extension.

It would probably be useful to explain the reasoning about this in the document. Do you think it would be sensible to add a few words explaining that?

PeterRugg:
@andresag01 Yes, I think the discussion, and the thrust of the conclusion in that issue, was to include the metadata to allow hardware to do checks and avoid side channels, not to throw exceptions. In the last reply to that thread, you said: "Do not start the prefetch if any of the checks above fails." Prefetch instructions aren't useful if they can throw exceptions.

PeterRugg:
Yes, @francislaus probably has comments due to his work on side channels, but we should have a section discussing what is allowed to happen even for traditional loads and stores. It doesn't (traditionally) belong in the architecture to discuss the allowed effects on microarchitectural state, but it seems appropriate to have some text, at least as a note, indicating that memory accesses where the checks fail should not escape to the memory hierarchy. It is particularly important for prefetches though because an attacker would be able to prefetch without having to squash the exception, presumably increasing the bandwidth considerably.

andresag01:
Agreed. The checks that we previously discussed are listed in the table that is being deleted; clearly these must not raise exceptions though. It would be good to keep a record of that somewhere in the text even if its a microarchitectural comment, which the RISC-V specs do have in some places.

tariqkurd-repo:
I agree 100% that prefetches must not throw exceptions, this was a mistake in the specification.

tariqkurd-repo:
@francislaus we should conditions under which execution the prefetch executes as a NOP (tag clear, sealed, no bytes in bounds in the cache line). Can you add this text?

francislaus:
@tariqkurd-repo Sure, I am just working on it.

francislaus:
@tariqkurd-repo It is all ready for the final approval now. I have added the conditions for the prefetch to not be issued as well as a very short intro text for our rationale on that. Credit to @PeterRugg for feedback.

andresag01:
@francislaus : Thanks for the change. Its looks good in my humble opinion. One minor comment I would make is that here:
> This instruction does not issue a prefetch if one or more of the following conditions of `cs1` are met

It might not be clear what the word "issue" means at the ISA level. I would tweak this to say something like the memory access is not "performed"...

francislaus:
@andresag01 Sounds sensible to me.

src/insns/prefetch.i.adoc
```diff
@@ -36,7 +36,14 @@ effective address is the sum of the base address specified in `cs1` and the
...
 is only valid if imm[4:0]=0. The authorising capability for this operation is
-`cs1`.
+`cs1`. This instruction does not throw any exceptions. However, following
+<<CHERI_SPEC>>, this instruction does not perform a prefetch if it is
+not authorized by `cs1`. This instruction does not perform a memory access
+if one or more of the following conditions of `cs1` are met:
+* The tag is not set
+* The sealed bit is set
+* No bytes of the cache line requested is in bounds
+* The X permission is not set
```
tariqkurd-repo, line 46: Should be <<x_perm>> to get the cross reference

src/insns/prefetch.r.adoc
```diff
@@ -36,7 +36,14 @@ effective address is the sum of the base address specified in `cs1` and the
...
 encoding is only valid if imm[4:0]=0. The authorising capability for this
-operation is `cs1`.
+operation is `cs1`. This instruction does not throw any exceptions. However,
+in following <<CHERI_SPEC>>, this instruction does not perform a prefetch
+if it is not authorized by `cs1`.  This instruction does not perform a memory
+access if one or more of the following conditions of `cs1` are met:
+* The tag is not set
+* The sealed bit is set
+* No bytes of the cache line requested is in bounds
+* The R permission is not set
```
tariqkurd-repo, line 46: Should be <<r_perm>> to get the cross reference

src/insns/prefetch.w.adoc
```diff
@@ -36,7 +36,14 @@ effective address is the sum of the base address specified in `cs1` and the
...
 encoding is only valid if imm[4:0]=0. The authorising capability for this
-operation is `cs1`.
+operation is `cs1`. This instruction does not throw any exceptions. However,
+following <<CHERI_SPEC>>, this instruction does not perform a prefetch if it
+is not authorized by `cs1`. This instruction does not perform a memory access
+if one or more of the following conditions of `cs1` are met:
+* The tag is not set
+* The sealed bit is set
+* No bytes of the cache line requested is in bounds
+* The W permission is not set
```
tariqkurd-repo, line 46: Should be <<w_perm>> to get the cross reference

arichardson (changes requested):
Thanks for fixing this - I agree that prefetch should not raise exceptions. LGTM with the references fixed.

PR #10, opened 2024-01-22
Clarify CJALR operation order

### LINKED ISSUE #9: Order of checks for CJALR & invalid address handling not specified
The spec for CJALR says

> The target capability may have Invalid address conversion performed and is then installed in pcc.

but it doesn't say whether invalid address conversion occurs before or after bounds checking (or the other checks, but those are unlikely to be affected by invalid address conversion).

IMO it would be more elegant to do it afterwards - it means most of the CJALR implementation doesn't need to care about invalid addresses; it's all neatly contained in the bit at the end where you install the cap in PCC. Also I imagine it would simplify implementations because there's no data dependency - you can do the bounds check without having to wait for invalid address conversion.

The security properties are the same, since invalid address conversion clears the tag if necessary. The only observable difference would be that in some cases you would get a tag violation on the next cycle, instead of a length violation on the current cycle. I think that's ok though.

Timmmm:
Clarify that length violations are checked on the address after it has undergone invalid address conversion.

Also fix a typo in the title.

Fixes #9

### DIFF
--- a/src/insns/cjalr_jalr_32bit.adoc
+++ b/src/insns/cjalr_jalr_32bit.adoc
@@ -6,7 +6,7 @@
 See <<JALR>>
 
 [#JALR,reftext="JALR"]
-==== CJAL, JALR
+==== CJALR, JALR
 
 Synopsis::
 Jump and link register
@@ -50,7 +50,7 @@ reported in the CAUSE field of <<mtval>> or <<stval>>:
 | Tag violation         |      | ✔     | `cs1` has tag set to 0
 | Seal violation        |      | ✔     | `cs1` is sealed and the immediate is not 0
 | Permission violation  |      | ✔     | `cs1` does not grant <<x_perm>>
-| Length violation      | ✔    | ✔     | Minimum length instruction is not within the target capability's bounds
+| Length violation      | ✔    | ✔     | Minimum length instruction is not within the target capability's bounds. This check uses the address after it has undergone xref:section_invalid_addr_conv[xrefstyle=short] but with the original bounds.
 |==============================================================================
 
 include::pcrel_debug_warning.adoc[]


src/insns/cjalr_jalr_32bit.adoc
```diff
@@ -24,11 +24,20 @@ Capability Mode Description::
...
-of the next instruction following the jump (<<pcc>> + 4) is sealed and written
-to `cd`.
+performed. The check for length violation is done next using the converted
+address and the original bounds.
+
+[NOTE]
+====
+The invalid address conversion representability check can be skipped because the
+length violation check is a stricter condition. The sealing check can also be
+skipped because the capability will always have been unsealed.
```
arichardson, line 36: I'm not sure about the sealing part? We still want to raise an error if `cs1` is a sentry and the offset is non-zero?

andresag01: Agreed, we would like to keep the sealing check

Timmmm: I meant the sealing check in the invalid address conversion procedure. I'll reword it to make that clearer.

Timmmm: I reworded it. Hopefully clearer.

src/insns/cjalr_jalr_32bit.adoc
```diff
@@ -24,11 +24,20 @@ Capability Mode Description::
...
-to `cd`.
+performed. The check for length violation is done next using the converted
+address and the original bounds.
+
+[NOTE]
+====
+When performing invalid address conversion the tag never needs to be cleared because
+the earlier length violation check is a stricter condition than the representability
+check, and the capability is always unsealed before invalid address conversion.
+====
```
andresag01, line 37: I don't fully agree with this note for a couple of reasons:
1. The tag is never cleared because the instruction excepts in all cases instead of changing PCC when something goes wrong
2. The capability is not _always_ unsealed. As @arichardson pointed out, there is an exception when the immediate is not 0 because you may be jumping to an instruction that is not the entry point.

I would prefer to remove this note altogether.

src/insns/cjalr_jalr_32bit.adoc
```diff
@@ -24,11 +24,20 @@ Capability Mode Description::
...
+
+The capability is then installed in <<pcc>>. The <<pcc>> of the next instruction
+following the jump (<<pcc>> + 4) is sealed and written to `cd`.
```
andresag01, line 40: Please note that that this test is wrapped under the `Synopsis::` header, but it is not currently indented properly because it is after the note and it is also missing the + at the end of the previous paragraph.

src/insns/cjalr_jalr_32bit.adoc
```diff
@@ -24,11 +24,20 @@ Capability Mode Description::
...
 target capability is obtained by unsealing `cs1` and incrementing its address by the
 sign-extended 12-bit immediate, and then setting the least-significant bit of the
-result to zero. The target capability may have
+result to zero. The target address may then have
 xref:section_invalid_addr_conv[xrefstyle=short]
-performed and is then installed in <<pcc>>. The <<pcc>>
-of the next instruction following the jump (<<pcc>> + 4) is sealed and written
-to `cd`.
+performed. The check for length violation is done next using the converted
+address and the original bounds.
```
andresag01, line 30: I think this text is better placed as a note in the `Exceptions::` section where the actual failure cases are highlighted.

andresag01 (changes requested):

src/insns/cjalr_jalr_32bit.adoc
Timmmm: I'll just remove this so we can get the important bit landed.

andresag01 (approved):

Issue #8, opened 2024-01-22
Build HTML and publish to Github pages


Timmmm:
It would be nice to extend the Makefile to support building the spec as HTML and publish the result via Github Pages.

tariqkurd-repo:
feel free!

arichardson:
This would be really nice to have! I'm not entierly sure what else needs to be done besides archiving the HTML, but I can see https://github.com/peaceiris/actions-gh-pages which should fix the deployment problem.

Timmmm:
Ok I got this to work on my fork. For some reason it deployed here: https://blog.timhutt.co.uk/riscv-cheri/

Hopefully it will deploy somewhere more sensible in this repo! I'll send a PR.

arichardson:
Seems to be deployed at https://riscv.github.io/riscv-cheri/ now :)

PR #12, opened 2024-01-22
Support building HTML output

Timmmm:
Add support for building the HTML version of the specification. By default `make` will still only build the PDF version; `make pdf` or `make html` will build one or the other, and `make all` will build both. `make all` is run in CI.

I also deduplicated the asciidoc commands a bit, and fixed ctrl-c for the docker builds.

See #8

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,8 +43,9 @@ CSVS	    = $(wildcard $(CSV_DIR)/*.csv)
 GEN_DIR     = $(SRC_DIR)/generated
 SCRIPTS_DIR = $(SRC_DIR)/scripts
 
-# Output file
+# Output files
 PDF_RESULT    := $(BUILD_DIR)/riscv-cheri.pdf
+HTML_RESULT   := $(BUILD_DIR)/riscv-cheri.html
 
 # Top asciidoc file of the document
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -79,11 +80,12 @@ GEN_SRC = $(GEN_DIR)/both_mode_insns_table_body.adoc               \
           $(GEN_DIR)/Zcheri_mode_insns_table_body.adoc             \
           $(GEN_DIR)/Zcheri_purecap_insns_table_body.adoc
 
-# Docker command
-DOCKER = docker run --rm -v $(PWD):/build -w /build $(DOCKER_IMAGE)
+# Docker command. `-it` is necessary so that ctrl-c works.
+DOCKER = docker run -it --rm -v $(PWD):/build -w /build $(DOCKER_IMAGE)
```
andresag01, line 84: I think `-it` stands for "interactive", which is probably why the CI is failing. Could you please take a look?

andresag01:
@Timmmm : Thanks for the changes to the Makefile. Looks good to me, but could you take a look at the failing CI?

andresag01:
@Timmmm : Thanks for the change! The CI now passes, so I will merge it.

PR #21, opened 2024-01-23
Deploy to Github pages on release

### LINKED ISSUE #8: Build HTML and publish to Github pages

Timmmm:
When a manual workflow is triggered this will also deploy the HTML to Github pages (if the relevant option is selected).

Fixes #8

### DIFF
--- a/.github/workflows/build-pdf.yml
+++ b/.github/workflows/build-pdf.yml
@@ -23,6 +23,11 @@ on:
         required: false
         type: boolean
         default: false
+      pages:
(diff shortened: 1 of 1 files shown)


Timmmm:
You also need to enable Pages in the project settings - set the source to Github Actions, and you probably want Enforce HTTPS.

image

I'm not sure exactly how Github determines the URL it gets deployed to. Hopefully it's something sane.

PR #22, opened 2024-01-23
representable range description, and some other minor clarifications

tariqkurd-repo:
Add note clarifying how the representable region is defined, and some other minor clarifications

### DIFF
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -509,7 +509,7 @@ bounds are formed of two or three sections:
 [#comp_addr_bounds,options=header,align="center"]
 |==============================================================================
 | Configuration  | Upper section | Middle Section | Lower section
-| EF=0, E varies | address[XLENMAX-1:E + MW] + ct | T[MW - 1:0] | {E{1'b0}}
+| EF=0, i.e. E>0 | address[XLENMAX-1:E + MW] + ct | T[MW - 1:0] | {E{1'b0}}
 | EF=1, i.e. E=0 | address[XLENMAX:MW] + ct     2+| T[MW - 1:0]
 |==============================================================================
 
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -349,13 +349,6 @@ Decoding the bounds:
 top:    t = { a[XLENMAX - 1:E + MW] + ct, T[MW - 1:0]    , {E{1'b0}} }
 base:   b = { a[XLENMAX - 1:E + MW] + cb, B[MW - 1:0]    , {E{1'b0}} }
 ```
-NOTE: The representable range
- is defined by the space beneath the bottom address bit (`a[E + MW]`)
- in the expression above. If the capability address changes causing that address bit
- to change, then the representable range is violated and so the tag will be cleared
- by an instruction such as <<CINCOFFSET>>. This is represented by a range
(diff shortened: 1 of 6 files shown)


src/cap-description.adoc
```diff
@@ -349,6 +349,12 @@ Decoding the bounds:
 top:    t = { a[XLENMAX - 1:E + MW] + ct, T[MW - 1:0]    , {E{1'b0}} }
 base:   b = { a[XLENMAX - 1:E + MW] + cb, B[MW - 1:0]    , {E{1'b0}} }
 ```
+NOTE: The representable range
+ is defined by the space beneath the bottom address bit (`a[E + MW]`)
+ in the expression above. If the capability address changes causing that address bit
+ to change, then the representable range is violated and so the tag will be cleared
+ by an instruction such as <<CINCOFFSET>>. This is represented by a range
+ of `s=2^E+MW^` in xref:cap_bounds_map[xrefstyle=short].
```
andresag01, line 357: @tariqkurd-repo : Isn't this similar to the description [here](https://github.com/riscv/riscv-cheri/pull/22/files#diff-2ac200119dffa8d1c8c860fa8f9d6614dc07cf4563a3cc1820cceb25152a12fbL497)?
```
NOTE: The encoding of the bounds depends upon the leading 1 of the address
which is used to determine the exponent. If the leading 1 of the address moves
then the bounds will need to be recalculated. Instructions like <<CINCOFFSET>>
and <<CSETADDR>> update the address field but do not recalculate the bounds.
Therefore, if the leading 1 moves relative to when the bounds were calculated
then the tag is cleared on the result as the encoding has been invalidated.
```

src/riscv-integration.adoc
```diff
@@ -1013,6 +1013,8 @@ Unlike machine and supervisor level CSRs, {cheri_base_ext_name} does not require
 The <<pcc>> is made visible in a CSR. This provides access to an
 <<infinite-cap>> capability while in debug mode without executing <<AUIPCC>>.
 
+<<pcc>> resets to the <<infinite-cap>> with the address field set to the core boot address.
```
andresag01, line 1016: Please note that the <<infinite-cap>> expands to the string `Infinity`, so this would sound better in the PDF if it was written like this:
```
<<pcc>> resets to the <<infinite-cap>> capability with the address ...
```

src/cap-description.adoc
tariqkurd-repo: yes - but actually that description is inaccurate which was my misunderstanding - the new description is accurate. I'll resolve the differences and push more commits.

src/cap-description.adoc
```diff
@@ -478,28 +477,60 @@ or 'root' capability.
...
+* Upper bits from the address
+* Middle bits from T and B decoded from the metadata
+* Lower bits are set to zero
+** This is only if there is an embedded exponent (EF=0)
+
+.Composition of address bounds
+[#comp_addr_bounds,options=header,align="center"]
+|==============================================================================
+| Configuration  | Upper section | Middle Section | Lower section
+| EF=0, E varies | address[XLENMAX-1:E + MW] + ct | T[MW - 1:0] | {E{1'b0}}
```
Timmmm, line 512: E varies -> `E>0`?

tariqkurd-repo: ok - will make a new commit

Issue #29, opened 2024-01-24
Document conflicts over floating point encodings


sorear:
C.FSD/C.FSDSP/C.FLD/C.FLDSP are part of Zcd, not Zca, and they conflict with the encoding chosen for C.SC on RV64, so (Zcheri_purecap & Zcd & RV64) is an illegal combination.  Zcmp and Zcmt use the same opcode and also cannot coexist with Zcheri_purecap & RV64.

C.FSW/C.FSWSP/C.FLW/C.FLWSP are part of Zcf, not Zca.

tariqkurd-repo:
Hi @sorear thanks for this - can you check the PR please

tariqkurd-repo:
PR merged

PR #35, opened 2024-01-24
correct prerequisite rules

tariqkurd-repo:
Addresses https://github.com/riscv/riscv-cheri/issues/29

### DIFF
--- a/src/insns/load_16bit_fp_dp.adoc
+++ b/src/insns/load_16bit_fp_dp.adoc
@@ -51,11 +51,11 @@ Standard floating point stack pointer relative load instructions, authorised by 
 
 include::load_exceptions.adoc[]
 
-Prerequisites for C.CFLD, C.CFLDSP::
-{c_cheri_base_ext_names}, and D
+Prerequisites for C.CFLD, C.CFLDSP (RV32 only)::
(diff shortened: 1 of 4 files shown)


sorear:
Zcd or D doesn't make much sense, Zcd implies D?

tariqkurd-repo:
>Zcd or D doesn't make much sense, Zcd implies D?

You can specify either. Zcd implies D, and D will infer C.FSDSP etc. if C or Zca are included even if Zcd isn't included.
It's weird but that's how it is.

jrtc27:
> > Zcd or D doesn't make much sense, Zcd implies D?
> 
> You can specify either. Zcd implies D, and D will infer C.FSDSP etc. if C or Zca are included even if Zcd isn't included. It's weird but that's how it is.

Which is sorear's point: that Zcd || D is entirely equivalent to D as a boolean expression given Zcd implies D. I think what you want here to make it *not* weird (other than due to the retroactive splitting of C) is Zcd or DC.

I don't think you're right that DZca infers Zcd though? LLVM doesn't do that, and I don't see why it would. Isn't the whole point of Zca to let you not have things like Zcd?

tariqkurd-repo:
Oh yes you're right, apologies.
Will fix tomorrow

Tariq

PR #36, opened 2024-01-24
Make permission bit order consistent

### LINKED ISSUE #25: Permission encoding order mismatch
For RV64 the ADP bits are encoded like this:

image

However `CGetPerms` and `CAndPerms` expect this order:

image

The table for encoding RV32 permissions uses the same order (reversed):

image

It would be much better if these were consistent!

Also did this question get resolved?

image

Timmmm:
Use the same ordering for permission bits in the RV64 encoding as in CAndPerms etc.

Fixes #25

### DIFF
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -121,8 +121,8 @@ permissions are granted as follows:
 
 * <<c_perm>>: bit 0 is set
 * <<w_perm>>: bit 1 is set
-* <<x_perm>>: bit 3 is set
 * <<r_perm>>: bits 3 or 2 are set
+* <<x_perm>>: bit 3 is set
 * <<asr_perm>>: bits 3 and 2 are set
 
 A {cap_rv64_perms_width}-bit vector encodes the permissions when XLENMAX=64. In
--- a/src/img/candperms_bit_field.edn
+++ b/src/img/candperms_bit_field.edn
@@ -13,9 +13,9 @@
 (draw-box "Reserved" {:span 11})
 (draw-box "ASR"  {:span 1})
 (draw-box "X"    {:span 1})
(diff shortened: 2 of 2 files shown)


src/cap-description.adoc
```diff
@@ -119,11 +119,11 @@ The encoding in xref:cap_perms_encoding32[xrefstyle=short] is chosen to
 facilitate hardware implementations. Therefore, it can be worked out if the
 permissions are granted as follows:
 
-* <<c_perm>>: bit 0 is set
+* <<r_perm>>: bits 3 or 0 are set
```
tariqkurd-repo, line 122: no - you've invalidated the encoding table above - you should have changed it on the CANDPERM page

tariqkurd-repo (changes requested):
don't do it this way

src/cap-description.adoc
Timmmm: Oops you're right. Ok should be fixed now. It was easier to change here than in CAndPerm. I assume there's no reason to prefer one specific ordering for the RV64 bits over another?

src/cap-description.adoc
```diff
@@ -119,10 +119,10 @@ The encoding in xref:cap_perms_encoding32[xrefstyle=short] is chosen to
 facilitate hardware implementations. Therefore, it can be worked out if the
 permissions are granted as follows:
 
-* <<c_perm>>: bit 0 is set
+* <<r_perm>>: bits 3 or 2 are set
```
Timmmm, line 122: These are the same as before now; I just reordered it so it matches the table columns and the RV64/CAndPerm order.

Timmmm:
Updated so that CAndPerm matches the RV64 encoding.

PR #38, opened 2024-01-24
Remove sole reference to embedded exponent

Timmmm:
"Embedded" and "internal" have very similar meanings in English so it's pretty confusing for them to mean opposite things in CHERI. Fortunately there was only one reference to "embedded" so it's easy enough to remove it.

### DIFF
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -225,7 +225,7 @@ address
 into the capability's address
 * *EF:* Exponent format flag indicating the encoding for T, B and E
     ** The exponent is stored in T and B if EF=0, so it is 'internal'
-    ** The exponent is zero if EF=1, so it is 'embedded'
+    ** The exponent is zero if EF=1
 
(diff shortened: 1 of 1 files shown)


andresag01 (changes requested):
@Timmmm : Thanks for the change. I agree with you. However, it looks like another PR was recently merged that also used the word `embedded`. Could you please rebase and fix that one as well? https://github.com/riscv/riscv-cheri/pull/22/files#diff-2ac200119dffa8d1c8c860fa8f9d6614dc07cf4563a3cc1820cceb25152a12fbR506

andresag01 (approved):
@Timmmm : Thanks for the changes!

Timmmm:
Np. Tbh I think "embedded" is a more appropriate word which is probably why it was used there by mistake! Could change it... though maybe not worth the effort.

andresag01:
"Internal" is the word that was used in the original CHERI spec, so it would be best to keep it for now to avoid confusion.

PR #45, opened 2024-01-25
fixed prerequisite rules for 16-bit fp dp load/store

tariqkurd-repo:
Hopefully the correct fix for https://github.com/riscv/riscv-cheri/issues/29 this time


andresag01:
Fixed adoc formatting in the last commit.

Issue #42, opened 2024-01-24
XLEN-bit results of CSR instructions must be sign extended to XLENMAX


sorear:
The current description in 6.3.4 says that they are zero extended, at variance with the base ISA.  This applies to both Zcheri_legacy as specified and to the simplified version.

jrtc27:
Probably it should be dropped entirely and just say the extension to XLENMAX follows the base specification

andresag01:
The PR has been merged, so closing this ticket.

arichardson:
> The PR has been merged, so closing this ticket.

If you add `fixes https://github.com/riscv/riscv-cheri/issues/42` (or closes) to the commit message, GitHub will do this automatically.

PR #46, opened 2024-01-25
Fix issue 42

tariqkurd-repo:
remove zero extension text and refer to unpriv spec instead

https://github.com/riscv/riscv-cheri/issues/42

### DIFF
--- a/src/riscv-legacy-integration.adoc
+++ b/src/riscv-legacy-integration.adoc
@@ -190,8 +190,8 @@ When the XLEN-bit alias is used by <<CSRRW>>:
 field.
     ** The tag and metadata are updated as specified in <<extended_CSR_writing>>.
 * Only XLEN bits are read from the capability address field, which are extended
-to XLENMAX bits according to cite:[riscv-unpriv-spec] and are then written
-to the destination *x* register.
+to XLENMAX bits according to cite:[riscv-priv-spec] _(3.1.6.2. Base ISA Control in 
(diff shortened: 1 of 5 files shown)


sorear:
[Not in citation given]

You're looking for "3.1.6.2. Base ISA Control in mstatus Register" in the privileged spec.

src/riscv-legacy-integration.adoc
```diff
@@ -189,8 +189,9 @@ When the XLEN-bit alias is used by <<CSRRW>>:
 * Only XLEN bits from the *x* source are written to the capability address
 field.
     ** The tag and metadata are updated as specified in <<extended_CSR_writing>>.
-* Only XLEN bits are read from the capability address field, which is zero
-extended to the destination *x* register.
+* Only XLEN bits are read from the capability address field, which are extended
+to XLENMAX bits according to cite:[riscv-unpriv-spec] and are then written
```
andresag01, line 193: @tariqkurd-repo: Thanks for the change. However, I think the relevant information is actually in the privileged spec. It would also be helpful to point out the Section 3.1.6.2 as @sorear said.

andresag01 (dismissed):

src/riscv-legacy-integration.adoc
tariqkurd-repo: I think it's unwise to point out an actual section number as that is subject to change over time.

andresag01:
Thanks!

PR #48, opened 2024-01-25
Start on a programmer's model for capabilities

sorear:
What we have here is a description of the rules governing CSETBOUNDS and CSETADDR which does not involve exponents (embedded, implied, or otherwise), mantissas, bitwise operations, Boolean lookup tables, arithmetic carry-out, or more than a passing mention of tops.

I anticipate reworking much of chapter 2 to clarify the various operations and reduce the conceptual complexity of explaining each part.  I wanted to post this early since there's a high potential for merge conflicts, and to find out if this needs to be coordinated with anyone and how much to break it apart.

### DIFF
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -7,16 +7,70 @@ space. There are two primary ISA variants, RV32I and RV64I, which provide
 32-bit and 64-bit address spaces respectively. The term XLEN refers to the
 width of an integer register in bits (either 32 or 64). The value of XLEN may
 change dynamically at run-time depending on the values written to CSRs, so we
-define XLENMAX to be widest XLEN that the implementation supports.
+define *XLENMAX* to be widest XLEN that the implementation supports.
 
+{cheri_base_ext_name} defines capabilities which act on virtual addresses of
+size XLENMAX, which will also be called *CXLEN* to indicate its link to
+capabilities.  {cheri_base_ext_name} capabilities occupy space in memory and
+registers equal to 2 * CXLEN bits, henceforth *CLEN*, and may be stored in
+memory only at addresses aligned to CLEN bits.  Provenance and integrity are
+provided by means of *tagged memory*: for every aligned CLEN-bit region of a
+register or memory that is permitted to hold a capability, there is a single
+extra bit, called the *tag*, which is 1 if the location currently holds a
+capability and 0 if it holds non-capability data.
 
-{cheri_base_ext_name} defines capabilities of size CLEN corresponding to 2 *
-XLENMAX without including the tag bit. The value of CLEN is always calculated
-based on XLENMAX regardless of the effective XLEN value.
+It is expected that "normal" memory will be able to hold capabilites, however
+memory that cannot hold capabilities may exist.  The details of this are
+determined by the environment.
+
+A tag, together with a capability or CLEN bits of non-capability data, is
+called a *capability value*.  A capability value with a tag of 1 will be
+considered interchangeable with the capability it holds; a capability value
+with a tag of 0 is an *untagged value*.  Capability values occupy CLEN+1 bits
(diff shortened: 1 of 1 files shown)


tariqkurd-repo:
are you a RISC-V member? If so you need to state your affiliation so we know who you are as you are currently anonymous.
If you are not a RISC-V member then you are not permitted to contribute directly to the specification.

jrtc27:
Stefan is a long-standing member of the community whose name features in multiple specifications

sorear:
Regardless of whether I contributed to the spec prior to RVI existing, I do have a current RVI membership e.g. https://lists.riscv.org/g/tech-unprivileged/message/629 .

src/cap-description.adoc
```diff
@@ -7,16 +7,70 @@ space. There are two primary ISA variants, RV32I and RV64I, which provide
...
 change dynamically at run-time depending on the values written to CSRs, so we
-define XLENMAX to be widest XLEN that the implementation supports.
-
-
-{cheri_base_ext_name} defines capabilities of size CLEN corresponding to 2 *
-XLENMAX without including the tag bit. The value of CLEN is always calculated
-based on XLENMAX regardless of the effective XLEN value.
+define *XLENMAX* to be widest XLEN that the implementation supports.
+
+{cheri_base_ext_name} defines capabilities which act on virtual addresses of
```
arichardson, line 12:
```suggestion
{cheri_base_ext_name} defines capabilities which act on addresses of
```
I'd drop the virtual here since we also want to support systems without virtual memory.

src/cap-description.adoc
```diff
@@ -37,6 +91,56 @@ All locations in registers or memory able to hold a capability are CLEN+1 bits
...
+half-open interval.
+
+Capability operations always treat addresses as unsigned.  A capability's valid
+region can span the gap between 2^CXLEN-1^-1 and 2^CXLEN-1^, but cannot wrap
+around from 2^CXLEN^-1 to 0.
+
+Subset and equality operations treat capability bounds as intervals, not as
+sets.  Two capabilities with length 0 and different bases are not equal and not
+subsets of each other.  A capability of length 0 can be constructed only inside
+an existing capability for an interval which surrounds the base.
```
arichardson, line 113: You can construct a zero-length capability for any address between base and top.
First increment to the desired in-bounds (or == top) address, and then CSETBOUNDS with length zero.

But maybe I am just misparsing "surrounds the base"

src/cap-description.adoc
```diff
@@ -37,6 +91,56 @@ All locations in registers or memory able to hold a capability are CLEN+1 bits
...
+
+The bounds and address of a capability are subject to alignment and
```
arichardson, line 115: I think it is very important to not representability, but I think the more detailed part with the table should be part of the bounds format section instead?

src/cap-description.adoc
```diff
@@ -7,16 +7,70 @@ space. There are two primary ISA variants, RV32I and RV64I, which provide
...
+register or memory that is permitted to hold a capability, there is a single
+extra bit, called the *tag*, which is 1 if the location currently holds a
+capability and 0 if it holds non-capability data.
+
+It is expected that "normal" memory will be able to hold capabilites, however
+memory that cannot hold capabilities may exist.  The details of this are
+determined by the environment.
+
+A tag, together with a capability or CLEN bits of non-capability data, is
+called a *capability value*.  A capability value with a tag of 1 will be
```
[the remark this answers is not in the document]
jrtc27: We've always used value to mean this, and it's the natural term to use. Arm's conflicting use is a bad idea that should be confined to the dustbin of history; it conflicts with the CHERI spec's terminology, is confusing, and makes it harder to refer to the whole capability. I am all for codifying "capability value" as being the whole capability.

tariqkurd-repo:

PR #49, opened 2024-01-25
Fix issue 42

tariqkurd-repo:


arichardson:
Sorry to be slow at responding today, but for changes like this where there are two fixup commits, can we please use squash and merge in the future? Otherwise the history gets quite messy. IMO merging only makes sense if there is logically consistent clean history where each commit makes sense.

tariqkurd-repo:
Hi Alex - yeah I'm not sure what happened with this one - I'm still getting to grips with this.

PR #53, opened 2024-01-25
Specify behaviour for reserved permission encoding

### LINKED ISSUE #47: Define behaviour for reserved RV32 permission encoding
The spec says

> Valid capabilities must not have the permissions field set to a reserved value according to [Table 3](https://blog.timhutt.co.uk/riscv-cheri/#cap_perms_encoding32) when XLENMAX=32.

I believe this can't ever happen for tagged capabilities but `CGetPerm` works for non-tagged capabilities. What should it return for this? Same question for `CAndPerm`.

Timmmm:
Fixes #47

### DIFF
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -144,8 +144,13 @@ permission.
 WARNING: #TODO: Confirm that we need a separate permissions format for 32-bit
 and 64-bit.#
 
-NOTE: Valid capabilities must not have the permissions field set to a reserved
-value according to xref:cap_perms_encoding32[xrefstyle=short] when XLENMAX=32.
+NOTE: When XLENMAX=32 there is a single reserved permission encoding (see
+xref:cap_perms_encoding32[xrefstyle=short]). It is not possible for a tagged
+capability to have this value since <<CANDPERM>> will never create it. It is
+possible for untagged capabilities to have it. <<CGETPERM>> will interpret it
+as if it were 0b0000 (no permissions). Future extensions may assign meaning to
+the reserved bit pattern, in which case <<CGETPERM>> is allowed to report a
+non-zero value.
 
 [#section_cap_sdp]
 ==== Software-Defined Permissions (SDP)
--- a/src/insns/cgetperm_32bit.adoc
+++ b/src/insns/cgetperm_32bit.adoc
@@ -15,9 +15,12 @@ Encoding::
 include::wavedrom/cgetperm.adoc[]
 
 Description::
-Converts the AP and SDP fields of capability `cs1`  into a bit field; one bit
+Converts the AP and SDP fields of capability `cs1` into a bit field; one bit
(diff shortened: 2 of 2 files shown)


arichardson (approved):
I think this makes a lot of sense especially if we consider forward compatibility.

andresag01 (commented):
@Timmmm @arichardson @tariqkurd-repo : What do you think would happen if executing CBuildCap to tag a cap that has permissions set to a reserved value? Options include:
1. Set the permissions of the output cap to 0 (i.e. no permissions)
2. Set tag 0 for the output tag

I would prefer (2)!

jrtc27:
Any kind of reserved input should always be untagged in the output. Anything else sounds like trouble.

src/cap-description.adoc
```diff
@@ -145,8 +145,11 @@ permission.
 WARNING: #TODO: Confirm that we need a separate permissions format for 32-bit
 and 64-bit.#
 
-NOTE: Valid capabilities must not have the permissions field set to a reserved
-value according to xref:cap_perms_encoding32[xrefstyle=short] when XLENMAX=32.
+NOTE: When XLENMAX=32 there is a single reserved permission encoding (see
+xref:cap_perms_encoding32[xrefstyle=short]). It is not possible for a tagged
+capability to have this value since <<CANDPERM>> will never create it. It is
+possible for untagged capabilities to have it. <<CGETPERM>> will interpret it
+as if it were 0b0000 (no permissions).
```
jrtc27, line 152: The way this is worded forbids an extension later assigning meaning to it (yes, this is non-normative text, but you still don't want it to be contradictory)

Timmmm: Good point. Would it be ok if I change it to  `It is currently not possible...  <<CGETPERM>> will currently interpret it...  This may change in future.`.

arichardson: Maybe something like this still allows for future extensibility:

```
If the permission encoding has reserved bit patterns (e.g. the format for XLENMAX=32 with {cheri_base_ext_name}
xref:cap_perms_encoding32[xrefstyle=short]) then it is not possible for a tagged
capability to have this value since <<CANDPERM>> will never create it. However, it is
possible for untagged capabilities to have it in which case <<CGETPERM>> will interpret it
as if it were 0b0000 (no permissions). Future extensions may assign meaning to these bit patterns for which <<CANDPERM>> is allowed to report a non-zero value.
```

Timmmm: Updated based on @arichardson's suggestion.

arichardson (approved):
Lgtm, @jrtc27 are you happy with this wording?

jrtc27 (approved):

andresag01 (approved):

andresag01 (commented):
It looks good to me, but we still need a comment about CBuildCap. Perhaps its best to do that as a separate PR.

PR #61, opened 2024-01-26
improve wording about about tag

axel-h:


arichardson (approved):
This definitely makes it clearer.

Not necessarily for this PR, but I think tag is such an overloaded word (and lately a lot of people think of e.g. MTE), so maybe we should be explicit here and refer to it as something like "validity tag"/"valid bit"?

andresag01 (approved):

PR #66, opened 2024-01-26
Misaligned LC/SC is a fatal error

### LINKED ISSUE #59: Exception policy for misaligned LC/SC is odd
The base ISA defines "load misaligned" and "store misaligned" as exception causes distinct from access faults to allow them to be quickly dispatched to an emulation handler, and has language (in privileged §3.6.3.3 Alignment) saying that in at least one case misaligned exceptions should only be generated if emulation is intended, and access-faults should be generated otherwise.

LC/SC (and compressed and capability-mode versions thereof) are defined to always check alignment and throw exceptions.  Do we intend for those instructions to be emulated, and appear to succeed in U-mode?

If yes, the spec should be loosened so that implementations supporting misaligned access for integer types can execute misaligned LC/SC without an exception.  (Naturally, any such operation must not create a tagged capability from misaligned data or allow a tagged capability to be partially overwritten without untagging.)

If no, the exception should be changed to an access-fault.

Do we expect compilers to use LC and SC for inline small struct copies of unknown alignment?

sorear:
Make it clear in the spec that this overrides Zicclsm and should not be emulated.

Resolves #59.

### DIFF
--- a/src/riscv-integration.adoc
+++ b/src/riscv-integration.adoc
@@ -161,6 +161,20 @@ Capability load and store instructions also cause load or store/AMO address
 misaligned exceptions if the address is not naturally aligned to a CLEN
 boundary.
 
+Misaligned capability loads and stores are errors.  Implementations must
+generate exceptions for misaligned capability loads and stores even if they
+allow misaligned integer loads and stores to complete normally.  Execution
+environments must report misaligned capability loads and stores as errors
+and not attempt to emulate them using byte access.  The Zicclsm extension
+does not affect capability loads and stores.  Software which uses capability
+loads and stores to copy data other than capabilities must ensure that
+addresses are aligned.
+
+NOTE: Since there is only one tag per CLEN bit block in memory, it is not
+possible to represent a capability value complete with its tag at an
+address not aligned to CLEN.  To transfer CLEN unaligned bits without a tag,
+use integer loads and stores.
+
 For loads, the tag of the capability loaded from memory is cleared if the
 authorising capability does not grant permission to read capabilities (i.e.
 both <<r_perm>> and <<c_perm>> must be set in AP). For stores, the tag of the


src/riscv-integration.adoc
```diff
@@ -161,6 +161,15 @@ Capability load and store instructions also cause load or store/AMO address
 misaligned exceptions if the address is not naturally aligned to a CLEN
 boundary.
 
+Misaligned capability loads and stores are fatal errors.  Implementations must
```
arichardson, line 164: "fatal" might be a bit strong? To me that reads as "system is broken must reset CPU", maybe the second sentence is sufficient?

src/riscv-integration.adoc
```diff
@@ -161,6 +161,15 @@ Capability load and store instructions also cause load or store/AMO address
...
+generate exceptions for misaligned capability loads and stores even if they
+allow misaligned integer loads and stores to complete normally.  Execution
+environments must report misaligned capability loads and stores as errors
+and not attempt to emulate them using byte access.  The Zicclsm extension
```
arichardson, line 168: I wonder if we want an explanatory note here as well along the lines of "Capability loads must also fetch the associated tag bit and there is only one tag bit per capability-size memory region, so a misaligned access that straddles multiple tag bit regions cannot be emulated in software."

arichardson (commented):
Thanks, this looks good to me. Just two comments:

src/riscv-integration.adoc
sorear: Not fully faithfully, but CHERI does a lot of "do something close enough, and then clear the tag bit" and I thought for a while that applied here too.

src/riscv-integration.adoc
sorear: Do you have any preferences for how to describe this in a way that applies to both hardware implementations of the privileged architecture (where the second sentence is sufficient), and things more like qemu user where exceptions aren't meaningful and the environment generates SIGBUS instead?

The relevant terminology from the unprivileged specification §1.6 is "contained trap" (control transfer to a handler within the same conceptual scope, e.g. xtvec or an installed signal handler) or "fatal trap" (terminates execution), but there's a confusing amount of leakage from the privileged spec into that section.

andresag01: Is it sufficient to say that implementations must "trap" for misaligned capability loads and stores etc? Trap is ideal because it is sufficiently high-level for the purposes of this section yet a definition is given in the RISC-V privileged spec.

src/riscv-integration.adoc
andresag01: I agree with @arichardson: it would be helpful to add a note for the rationale behind this paragraph

src/riscv-integration.adoc
tariqkurd-repo:
```
Fatal Trap
The trap represents a fatal failure and causes the execution environment to terminate execution.
Examples include failing a virtual-memory page-protection check or allowing a watchdog timer to
expire. Each EEI should define how execution is terminated and reported to an external
environment
```
This does seem appropriate but we haven't used the terms contained trap, fatal trap etc. anywhere in the doc. To start using those terms we'd need another section mapping the CHERI exceptions onto them.

src/riscv-integration.adoc
```diff
@@ -161,6 +161,20 @@ Capability load and store instructions also cause load or store/AMO address
...
+does not affect capability loads and stores.  Software which uses capability
+loads and stores to copy data other than capabilities must ensure that
+addresses are aligned.
+
+NOTE: Since there is only one tag per CLEN bit block in memory, it is not
+possible to represeent a capability value complete with its tag at an
```
arichardson, line 174:
```suggestion
possible to represent a capability value complete with its tag at an
```

src/riscv-integration.adoc
```diff
@@ -161,6 +161,20 @@ Capability load and store instructions also cause load or store/AMO address
...
+address not aligned to CLEN.  To transfer CLEN bits without a tag, use
```
arichardson, line 175:
```suggestion
address not aligned to CLEN.  To transfer unaligned CLEN bits without a tag, use
```
Maybe add this qualification here since capability loads without C permission can be used to transfer without the tag.

arichardson (approved):
Thanks for the clarifications. One typo and a suggested clarification otherwise ready to merge.

arichardson (approved):
Will give others a day to comment on this and if there are no further comments I'll merge it tomorrow.

andresag01 (approved):

PR #68, opened 2024-01-27
Sync with template repository to add pre-commit checks and CI

arichardson:
In the future this will make it easier to add more pre-commit checks to ensure consistency.


.github/workflows/build-pdf.yml
```diff
@@ -39,57 +40,57 @@ jobs:
...
-    - name: Build Files
-      run: make all
-      env:
-        VERSION: v${{ github.event.inputs.version }}
-        REVMARK: ${{ github.event.inputs.revision_mark }}
+      - name: Build Files
+        run: make all
+        env:
+          VERSION: v${{ github.event.inputs.version }}
+          REVMARK: ${{ github.event.inputs.revision_mark }}
```
andresag01, line 57: This part, which pre-dates the PR, make me a little sad because the CI build that ends in the artifacts has the following version string in the title page: `Version v, 2024-01-27:`. In other words, VERSION and REVMARK end up being empty here. Is it possible to fix this somehow?

readme.adoc
```diff
@@ -8,7 +8,7 @@ This work is licensed under a Creative Commons Attribution 4.0 International Lic
```
andresag01, line 8: It would be helpful to enhance with an explanation (or link) on how to use pre-commit.

src/csv/CHERI_CSR.csv
```diff
@@ -30,4 +30,4 @@ direct write if address didn't change","✔","","","✔","✔","Zcmt","Jump Vect
 "stdc","0x163","","","S","SRW, <<asr_perm>>","<<null-cap>>","","","","","","✔","","S-mode","Supervisor Trap Data Capability (scratch register)","","","","","","","","","","","","","","","","","","","","",""
 "ddc","0x416","","","U","URW","<<infinite-cap>>","","","","","","✔","","none","User Default Data Capability","","","","","","","","","","","","","","","","","","","","",""
 "pcc","0xcb0","","","U","URO","<<infinite-cap>>
-(address = boot address)","","","✔","","","✔","✔","none","User Program Counter Capability (to allow reading in legacy mode)","","","","","","","","","","","","","","","","","","","","",""
\ No newline at end of file
+(address = boot address)","","","✔","","","✔","✔","none","User Program Counter Capability (to allow reading in legacy mode)","","","","","","","","","","","","","","","","","","","","",""
```
andresag01, line 33: These files are downloaded from a spreadsheet in Google Drive. Did you modify the spreadsheet or the CSVs?

.github/workflows/build-pdf.yml
arichardson: I'm sure we can get a default from the the makefile. I'll look into this today

readme.adoc
arichardson: Agreed, I will submit a pr to the templates repo

src/csv/CHERI_CSR.csv
arichardson: I just ran pre-commit so it edited the checked in csvs. I can exclude these from the check or modify the download step to add a newline

andresag01: Thanks! Either that, modify the download or modify the spreadsheet. Otherwise, we will continuously get this change when we trigger an update of the CSV.

.github/workflows/build-pdf.yml
arichardson: Since this is unrelated to this PR I've file https://github.com/riscv/riscv-cheri/issues/72 and plan to address this separately.

readme.adoc
arichardson: https://github.com/riscv/docs-spec-template/pull/32. I'll pull this in once the upstream PR has been accepted.

Issue #58, opened 2024-01-26
pcc metadata after cm.j(al)t


sorear:
The jump table contains addresses, which raises the question of how the new pcc capability is derived from the address in the jump table.  I think the most useful option is to take the bounds from jvtc, so that it can be used as a subroutine library shared between many code capabilities.

tariqkurd-repo:
`jvtc` is intended to cover the bounds of the table to prevent access to it, which doesn't restrict its usage to that, but it doesn't seem appropriate to use it in this way.

The intention of the extension is to reduce code size, and so doubling the size of the table by making every entry a cap greatly reduces the benefit, this is something which has been mentioned before.

It would be possible to have another cap which provides the bounds, maybe with a bit in the LSB of each table entry to say whether to take PCC bounds or bounds of this other register.

sorear:
How did you get from "The jump table contains addresses" and "take the bounds from jvtc" to "the opener wants to redefine the jump table to make every entry a cap and take the bounds from capabilities stored in the jump table"?

andresag01:
The _jump via table_ instructions calculate the target address like this: `jvtc.address+index*XLEN/8`. So the question is where the new PCC's metadata is coming from -- this is not clear to me looking at the current spec. There are 2 options as far as I can see:

1. The bounds come from PCC, so the "new" PCC i.e. the target capability of the jump is (excluding legalization, etc):
```
newPCC = setCapAddr(PCC, jvtc.address+index*XLEN/8)
```
2. The bounds come from JVTC, so the "new" PCC is:
```
newPCC = setCapAddr(JVTC, jvtc.address+index*XLEN/8)
```
I think @sorear is advocating for (2). Right?

IMHO there are arguments for either of the options above. @tariqkurd-repo: In the existing Zcmt trials, how does the toolchain deal with the table? Where is it normally placed relative to the code? Of course, this will not consider CHERI, but at least is a bit more info.

@tariqkurd-repo: A other thoughts about Section 8.11 in the CHERI spec (i.e. Zcmt):
* It would be good to indicate that jvmtc needs to be Almighty if someone wants to make it look like they are running regular RISC-V. We have similar notes for PCC and DDC elsewhere
* Why must JVTC have X-perm? Are you assuming (2) above as well? In the case of (1) above, I would have expected JVTC to have only R-perm

sorear:
[...]

PR #76, opened 2024-01-30
clarify PCC bounds after zcmt instruction

tariqkurd-repo:


andresag01:
Fixes https://github.com/riscv/riscv-cheri/issues/58

andresag01 (commented):
Thanks for the changes. It would still be good to clearly state somewhere that the target address of the jump table instruction takes its metadata from PCC.

PR #77, opened 2024-01-30
github workflow: update action

axel-h:
Avoid CI warning that node.js 16 is deprecated.

See https://github.com/riscv/riscv-cheri/actions/runs/7714759013: `Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/upload-artifact@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.`

### DIFF
--- a/.github/workflows/build-pdf.yml
+++ b/.github/workflows/build-pdf.yml
@@ -66,7 +66,7 @@ jobs:
 
       # Upload the built PDF and HTML files as a single artifact
       - name: Upload Build Artifacts
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: Build Artifacts
           path: |


jrtc27:
Can we instead just sync with riscv/docs-spec-template again, which does include this fix among a bunch of other changes, rather than repeating the same fixes ourselves as the template?

axel-h:
Sure, makes sense. I'm not an expert on the processes here. This PR just came along with the same changes on a few other repos to get rid of the warning.

arichardson:
In my latest PR I did not merge with the templates repo since all of its history was squashed into the first commit of this repo. I could do a merge with --allow-unrelated-histories, so that future updates are simpler. @jrtc27 what do you think is the right approach? Should we continue diffing the files to the template and pulling over patches or do the merge?

jrtc27:
> In my latest PR I did not merge with the templates repo since all of its history was squashed into the first commit of this repo.

Thanks GitHub...

> I could do a merge with --allow-unrelated-histories, so that future updates are simpler. @jrtc27 what do you think is the right approach? Should we continue diffing the files to the template and pulling over patches or do the merge?

Merging seems like the best approach going forwards, other than GitHub PRs not handling merge commits...

arichardson:
> > In my latest PR I did not merge with the templates repo since all of its history was squashed into the first commit of this repo.
> 
> Thanks GitHub...
> 
> > I could do a merge with --allow-unrelated-histories, so that future updates are simpler. @jrtc27 what do you think is the right approach? Should we continue diffing the files to the template and pulling over patches or do the merge?
> 
> Merging seems like the best approach going forwards, other than GitHub PRs not handling merge commits... although given people seem to be using "Create a merge commit" for PRs I guess it might actually work ok?

Sounds good, I'll give this a try now.

axel-h:
Closing this PR, because #79 will handle this.

Issue #39, opened 2024-01-24
Proposal to simplify (and fix) Zcheri_legacy


sorear:
Zcheri_legacy, to my understanding, serves two purposes:

1. Allowing capability-naive software stacks to run on capability-capable hardware.
2. Supporting the execution of hybrid software.

I argue that, as currently defined, it doesn't really work for either case.

1. Legacy mode indeed allows valid instructions to execute in the same way they would without Zcheri_purecap.  However, stacks will include untrusted code, which can compromise a naive supervisor by modifying ddc prior to a trap.  Zcheri_legacy has no mechanism to prevent untrusted code from modifying ddc other than lowering XLEN, which affects the naive code that can be run.

2. Since ISAv9's [LS][TYPE].CAP instructions are not supported in Zcheri_purecap, the only way to access data through a capability in Legacy mode is to perform a csrrw on ddc between each memory access.  This is highly likely to be a serializing instruction.  

   Not a dealbreaker, except that Zcheri_mode is Pareto-better.  It's shorter - 2 bytes for C.CMODESWITCH instead of 4 for CSRRW, and can be omitted entirely between accesses to different capabilities.  It can be made faster easily - CMODESWITCH affects no state other than pcc, allowing it to be executed during decode in typical implementations without a pipeline bubble.  It's simpler to compile, since there is no need to allocate the ddc register.  It simplifies the hardware - assuming Capability mode execution in higher privileges, there is no need for the mtdc, dddc, stdc, or pcc registers, only add one instruction and it doesn't even need backend handling.

We could, in principle, address problem 1 by adding a second bit to xenvcfg which functions the same as lowering XLEN, but that isn't a simplification and doesn't address problem 2.  Instead, let's simplify Zcheri_legacy by focusing exclusively on use case 1, and leaving use case 2 to Zcheri_mode.  The intent is that new-Zcheri_legacy + Zcheri_mode will be simpler and faster in all cases than old-Zcheri_legacy alone.

---

Define three execution modes.

In _Legacy mode_, all instructions defined by Zcheri_purecap are "reserved" in the language of the base ISA spec and cause an illegal-instruction exception if [Sstrict](https://lists.riscv.org/g/tech-unprivileged/message/640) is claimed.  All instructions not defined by Zcheri_purecap function as described for Legacy mode, using ddc instead of explicit argument capabilities to authorize access.  No capability-related state is accessible except for the implicit use of ddc.

[...]

PR #81, opened 2024-01-31
capability mode enables and register access controls for Zcheri_legacy

sorear:
A CHERI mode enable for M-mode allows Zcheri_legacy to run full legacy software stacks, including firmware soon after reset.

This also removes the behavioral difference between Zcheri_legacy and Zcheri_mode in terms of the instruction set in effect after reset, making Zcheri_mode a true extension of Zcheri_legacy.

CHERI register access disables for S-mode and U-mode allow Zcheri_legacy to prevent cross-domain interference and covert channels within a legacy environment.

These two are the strictly additive part of #39.


src/riscv-legacy-integration.adoc
```diff
@@ -345,6 +370,12 @@ The CHERI Mode Enable (CME) bit controls whether less privileged levels (e.g.
 S-mode and U-mode) execute in Capability or Legacy mode. When CME=1, the
 CHERI execution mode is Capability. When CME=0, the mode is Legacy.
 
+The CHERI Register Enable (CRE) bit controls whether less privileged levels can
```
arichardson, line 373: I wonder if this would fit better into the state enable extension since conceptually this is about making the additional state accessible? E.g. as a bit in `mstateen`/`sstateen`?

sorear: Potentially, but once there's consensus on the approach here the bit is just going to get deleted by phase 2 so I'd rather not focus on where it lives until then.

tariqkurd-repo:
What exactly is the difference between Xenvcfg.CME and Xenvcfg.CRE?
Is your intention that for the current privilege level/ring:

* if CME = 0 then it's legacy mode so memory checking is authorised by DDC, but capabilities _can_ be manipulated, DDC can be accessed etc.
* if CRE = 0 then no CHERI instructions can be executed, so capabilities _cannot_ be manipulated, DDC can't be updated etc.

so if CME=CRE=0 it really is a pure RISC-V machine - for that privilege level

do you intend CME to override CRE, so if CME=1 then CRE is don't care?

We've been back and forth many times over whether we need these extra enable bits (I was generally in favour of them, others less so), and where they should live (state-emable or envcfg). I'm not so sure that the location is very important, as all these things will be adjusted by the ARC in due course.

Given that we already say that CME=CRE=0 is what you get if XLEN < XLENMAX then adding CRE is not describing a new state in the machine, it's making an existing state more useful.

what's you ropinion @andresag01 ?

andresag01:
@sorear : Thanks for the proposal. I generally agree with the idea in this PR, but there are a couple of things that are unclear to me:

* Should M-mode have its own CME and CRE bits?
* Does the CHERI Execution mode control what the new CHERI instructions and the re-mapped instructions in CHERI do?

IMHO, it should be like this:

* CME:
    - Makes the re-mapped instructions whatever they were in standard RISC-V (e.g. like the compressed FP) when in Legacy
    - Change all the relevant instruction operands to integer values when in Legacy mode
* The CRE bit:
    - Forces CHERI mode to be Legacy
    - Disables access to CHERI CSRs
    - Disables access to CHERI instructions

What do you think about this?

src/riscv-legacy-integration.adoc
```diff
@@ -53,8 +53,9 @@ The CHERI execution mode is key in providing backwards compatibility with the
 base RISC-V ISA. RISC-V software is able to execute unchanged in
```
andresag01, line 53: I think that's true, except for the debugger (which I guess is not technically the hart's software) because under the current spec Debug mode would be running in Capability mode even if all CME bits are 0.

tariqkurd-repo: there are no CME/CRE bits which affect debug mode

src/riscv-legacy-integration.adoc
tariqkurd-repo: mstateen/stateen does seem more sensible, but I agree with @sorear that anything we allocate will be moved later, so maybe it's not important where they go

sorear:
@tariqkurd-repo 

> do you intend CME to override CRE, so if CME=1 then CRE is don't care?

I intended CRE=0 to override CME, on the theory that CRE is an access control but CME just affects instruction decoding.

> Given that we already say that CME=CRE=0 is what you get if the current XLEN < XLENMAX then adding CRE is not describing a new state in the machine, it's making an existing state more useful.

It's a new mode from the point of view of fixed XLEN hardware.

> Something else to add is that we can always allow some of these enables to be hardwired in an implementation, e.g. in mseccfg so that it isn't required to implement the full set of different modes.

There are three modes: capability (CME=1 CRE=1), transitional (CME=0 CRE=1), legacy (CRE=0 CME=x).  Ignore transitional until I have a chance to post part 2.  I don't think it makes sense to hardwire enables in Zcheri_legacy because if you only want capability mode, you should do Zcheri_purecap, and if you only want legacy mode you shouldn't do Zcheri at all.

@andresag01

> Should M-mode have its own CME and CRE bits?

M-mode needs its own CME bit in order to allow running legacy M-mode firmware.  M-mode's CRE bit is hardwired to 1 because there is no higher privilege mode that needs to control dataflow in and out of M-mode capability registers.

> IMHO, it should be like this:

Unless I'm missing or poorly communicated something, that's how this proposal works.

tariqkurd-repo (commented):
I like this PR in general.

I think that "CHERI Register Enable" should be "CHERI State Enable" because, just like othe rstate enable bits, it's controlling access to registers, CSRs and instructions.

With that renaming I thin it makes it clear that it should use a bit in `Xstateen`.

I know all this will change when the ARC get their hands on it, but I think the description more easily fits with the RISC-V ethos that way.

Other than that LGTM

sorear:
With xstateen as currently defined, you can set it to ~0 or omit the extension if you don't care about covert channels; CHERI State Enable would be the first xstateen bit that allows U-mode to mount denial of service attacks against S-mode if it is set prematurely of other software support.  I'm hesitating for that reason.

Does the rename block this being resolved?

arichardson:
> With xstateen as currently defined, you can set it to ~0 or omit the extension if you don't care about covert channels; CHERI State Enable would be the first xstateen bit that allows U-mode to mount denial of service attacks against S-mode if it is set prematurely of other software support. I'm hesitating for that reason.
> 

I agree that we have to ensure CHERI is disabled for lower rings on boot due to DDC not being banked. If your kernel turns on all unknown extensions for lower privilege levels (even ones it doesn't know about), then that's just asking for trouble and I don't think that's something we have to prevent in this spec. Being able to hardwire it to all zeroes is more problematic, but we could just require the CHERI bit for lower privilege levels must be zero on reset?

jrtc27:
> > With xstateen as currently defined, you can set it to ~0 or omit the extension if you don't care about covert channels; CHERI State Enable would be the first xstateen bit that allows U-mode to mount denial of service attacks against S-mode if it is set prematurely of other software support. I'm hesitating for that reason.
> 
> I agree that we have to ensure CHERI is disabled for lower rings on boot due to DDC not being banked. If your kernel turns on all unknown extensions for lower privilege levels (even ones it doesn't know about), then that's just asking for trouble and I don't think that's something we have to prevent in this spec. Being able to hardwire it to all zeroes is more problematic, but we could just require the CHERI bit for lower privilege levels must be zero on reset?

See https://github.com/riscv/riscv-state-enable/issues/12. This was a stupid choice in the RISC-V spec that wasn't really thought through properly IMO. Arm have the concept of RES0 and RES1 bits in their system registers for this reason, but in RISC-V we're pretending everything can be RES1, which is clearly a bad idea.

arichardson:
> > > With xstateen as currently defined, you can set it to ~0 or omit the extension if you don't care about covert channels; CHERI State Enable would be the first xstateen bit that allows U-mode to mount denial of service attacks against S-mode if it is set prematurely of other software support. I'm hesitating for that reason.
> > 
> > 
> > I agree that we have to ensure CHERI is disabled for lower rings on boot due to DDC not being banked. If your kernel turns on all unknown extensions for lower privilege levels (even ones it doesn't know about), then that's just asking for trouble and I don't think that's something we have to prevent in this spec. Being able to hardwire it to all zeroes is more problematic, but we could just require the CHERI bit for lower privilege levels must be zero on reset?
> 
> See [riscv/riscv-state-enable#12](https://github.com/riscv/riscv-state-enable/issues/12). This was a stupid choice in the RISC-V spec that wasn't really thought through properly IMO. Arm have the concept of RES0 and RES1 bits in their system registers for this reason, but in RISC-V we're pretending everything can be RES1, which is clearly a bad idea.

Ah thanks for pointing me at that issue. It sounds like we do need to go for xenvcfg instead of stateenable if that is recommended.

The more invasive alternative would be to have DDC be per-ring (like with Morello) in which case enabling CHERI for a lower privilege level should not have any effect (other than the lower privilege level being able to shoot itself in the foot more effectively). But that change would require more discussion, so I am happy with the bit allocation as specified in this PR.

andresag01 (approved):

tariqkurd-repo:
Yes - the stateenable description assume that no privilege can highjack the privilege above. So as Alex says without multiple DDCs then this seems like the way to go.

tariqkurd-repo (approved):

src/riscv-legacy-integration.adoc
tariqkurd-repo: accepting that we don't currently have banked DDC let's go with this proposal for now

PR #82, opened 2024-02-01
Cbuildcap and ctestsubset do not use DDC

### LINKED ISSUE #18: Cbuildcap and ctestsubset and the use of DDC
Both cbuilcap and ctestsubset use DDC as a parameter if the cs1 == 0. However, both instructions only have Zcheri_purecap as their prerequisite. DDC is a register that is only defined in Zcheri_legacy, if I understand the spec correctly. Thus, there exists a case where the specified behaviour of the instruction is not possible.

These are the two straightforward solutions that came into my mind.

1. Cbuildcap and ctestsubset use c0 if cs1 == 0
2. DDC is defined in Zcheri_purecap

There probably are other solutions. Given the two I named, the first one seems way more clean. This would not optimally use the encoding space of the register index for cs1, but it would prevent DDC from being defined in purecap where it is not needed.

francislaus:
This should close #18 

I decided to not make cs1==0 RESERVED because it is a valid input.

### DIFF
--- a/src/insns/cbuildcap_32bit.adoc
+++ b/src/insns/cbuildcap_32bit.adoc
@@ -3,6 +3,10 @@
 [#CBUILDCAP,reftext="CBUILDCAP"]
 ==== CBUILDCAP
 
+ifdef::cheri_v9_annotations[]
+NOTE: *CHERI v9 Note:* cbuildcap does not use ddc if cs1==0
+endif::[]
+
 include::new_encoding_note.adoc[]
 
 Synopsis::
@@ -20,7 +24,8 @@ sealed, `cs1` 's permissions and bounds are equal or a superset of `cs2` 's,
 `cs2` 's bounds are not malformed (see
 xref:section_cap_malformed[xrefstyle=short]), and all reserved bits in `cs2` 's
 metadata are 0. <<CBUILDCAP>> is typically used alongside <<CSETHIGH>> to build
-capabilities from integer values.
+capabilities from integer values. The case `cs1 == 0` trivially leads to the tag
+of `cd` being stripped.
 
 Prerequisites::
 {cheri_base_ext_name}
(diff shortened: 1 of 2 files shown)


src/insns/cbuildcap_32bit.adoc
```diff
@@ -20,19 +24,20 @@ sealed, `cs1` 's permissions and bounds are equal or a superset of `cs2` 's,
 `cs2` 's bounds are not malformed (see
 xref:section_cap_malformed[xrefstyle=short]), and all reserved bits in `cs2` 's
 metadata are 0. <<CBUILDCAP>> is typically used alongside <<CSETHIGH>> to build
-capabilities from integer values.
+capabilities from integer values. The case `cs1 == 0` trivially leads to the tag
```
arichardson, line 27: Should this be the canonical ccleartag? If so could add a note suggesting this case is optimized in the uarch

francislaus: I think so. This would add a pseudo instruction `ccleartag cd, cs1` which evaluates to `cbuildcap cd, c0, cs1`

francislaus:
I had a longer discussion with @PeterRugg and he explained that cbuildcap also clears the otype if it is successful. In the current form cbuildcap can be used for tag clearing, but this would change if the spec got more otypes later.

francislaus:
What was the reasoning for not having a `ccleartag` instruction? @tariqkurd-repo

andresag01:
@francislaus: The expectation was `ccleartag` is not a common instruction, so it was not included in the current proposal to save encoding space. It could be added if there is a strong use-case for it.

francislaus:
@tariqkurd-repo What were your thoughts about `ccleartag`?

arichardson (approved):
I'm happy with this being merged as is and I've filed https://github.com/riscv/riscv-cheri/issues/94 for the CClearTag question.

andresag01 (approved):

src/insns/cbuildcap_32bit.adoc
andresag01: I'll resolve this for merge since @arichardson created that as a separate issue.

Issue #44, opened 2024-01-25
Debug GPR access


sorear:
The debug spec allows GPRs to be accessed through the Access Register abstract command, and non-GPRs either through abstract commands or by running program buffer fragments.  Access Register can use an access size of 32, 64, or 128, but 65 and 129 are not options; how does a debugger read or write capabilities in GPRs, or restore capabilities in GPRs after using the GPR for a program buffer fragment?

System bus access also needs to be extended for CLEN-bit access, but it is an optional feature.

tariqkurd-repo:
The intention is that program buffer code is used to read the integer value, then `cgethigh` then `cgettag` to read the fields of the CLEN registers, so that the debug interface doesn't change.

System bus as specified will work without changing the width, but may not be able to monitor tags.

sorear:
That would work if a split capability register file were used, but with our merged file, you cannot do a `cgethigh` or `cgettag` without overwriting a register and you can't overwrite a register until you've read both its metadata and its tag.

tariqkurd-repo:
yes - so we want to stuck with the standard debug specification - so I believe the documented scheme works correctly.

sorear:
dscratchN is described as "The debugger can make no assumptions about the contents of these registers between commands", so if the debugger cannot save a GPR in dscratchN between abstract commands.  It might be possible to use a single abstract command to completely deconstruct a GPR into dataN registers, but this would definitely go over the 2 instruction mandatory minimum size of the program buffer.  Moreover, both dscratch registers are explicitly defined as optional both here and in the debug spec.

jrtc27:
There needs to be an extension to abstract commands (in fact our cores _only_ support that), but it's not hard to envisage how that might look.

sorear:
I was briefly considering proposing a "DGETHIGH" instruction that deconstructs a capability while preserving the tag, and can also be used to construct Infinity capabilities while avoiding the can of worms that is pc/ddc access in debug mode, but this doesn't help with the reverse problem of building a capability in a single GPR while not touching the other 30 GPRs.

I would rather avoid mandating dscratch[01]c and sufficient program buffer and data register sizes to send a program using csrr/csethigh/cbuildcap, as that would involve hundreds of bits of new state.

Given that the Access Register abstract command is required for GPRs anyway, I think the most direct route is to allocate 64 abstract register numbers for "HIGH(xN)" and "TAG(xN)"?

sorear:
[...]

PR #88, opened 2024-02-02
Document minimum resources for Sdext operation

sorear:
The baseline strategy for accessing capabilities in integer registers requires 4 instructions in the program buffer, 1 XLEN's worth of data register shadows, 1 dscratch register, and pc access, all of which are optional in the debug spec.  Document that Zcheri needs them.

(#44, but that was tagged as an enhancement so don't close it automatically.)


src/debug-integration.adoc
```diff
@@ -5,6 +5,69 @@ This section describes changes to integrate the Sdext ISA and
...
+* All harts which support {cheri_base_ext_name} must provide
+  `hartinfo.nscratch` of at least 1 and implement the <<dscratch0c>> register.
+* All harts which support {cheri_base_ext_name} must provide
+  `hartinfo.datasize` of at least 1 if `hartinfo.dataaccess` is 0,
+  `hartinfo.datasize` at least `XLENMAX/32` otherwise.
+* The program buffer must be implemented, with `abstractcs.progbufsize` of at
+  least 4 if `dmstatus.impebreak` is 1, or at least 5 if `dmstatus.impebreak`
+  is 0.
+* The <<pcc>> must be readable using an <<AUIPCC>> instruction while executing
+  from the program buffer.
```
andresag01, line 21: Why is this necessary?

sorear: Every argument I can think of for forbidding reads of pcc via AUIPCC applies just as strongly to csrr of pcc, which I why I tried to move away from using pcc at all as the sole source of Infinity in #44.

sorear: Split that out into #90 to make this a smaller change.

andresag01: Thanks for splitting that out. Can we remove this bullet-point altogether? Reading the debug spec, support for CSR instructions is not optional when executing code from the program buffer.

sorear: §4.1

> 12. Instructions that depend on the value of the PC (e.g. auipc) may act as illegal instructions.

I think that "csrr rd, pcc" is an "instruction that depends on the value of the PC".

andresag01: This PR was merged, so there is no longer a PCC CSR: https://github.com/riscv/riscv-cheri/pull/113.

andresag01:
@sorear : Could you please update this PR? The instruction names have changes and the PCC is no longer accessible through a CSR. Thanks!

andresag01 (approved):

andresag01:
@arichardson : I am happy with the changes, but it would be good to have an additional reviewer. Could you please have a look? Thanks!

arichardson (approved):

PR #98, opened 2024-02-08
Document that DDC/DDDC are address pointers and can be compressed

### LINKED ISSUE #50: DDC / DDDC should be address registers, so don't need to be full width in RV64
As metioned in https://github.com/riscv/riscv-cheri/issues/41 to separate it out from the other ideas

andresag01:
Fixes #50


tariqkurd-repo (approved):
LGTM

tariqkurd-repo:
@arichardson merging this now I want to make another update to the CSR spreadsheet, let us know if you find any problems with it and we can make another PR to fix them

Issue #43, opened 2024-01-24
Reset values of capability CSRs


sorear:
The base privileged architecture does not specify reset values for CSRs if M-mode code can ensure that they are not used until they have been written with a value.  While defining the initial values of the capability system is important for lower modes, M-mode is likely to run before RAM initialization and erasing tags in RAM is unlikely to be performed in a way transparent to the earliest M-mode code, so M-mode code which runs immediately after reset must already be prepared to deal with garbage capabilities.

For consistency with the base ISA, remove reset values from capability CSRs that can be safely initialized by M-mode at the same time as RAM.

This applies to mscratchc, sscratchc, mepcc, sepcc, stvecc, jvtc, ddc, stdc, mtdc straightforwardly as they are not used prior to the execution of CSR instructions or of a MRET.  It also applies to mtvecc, since M-mode code is responsible for not enabling interrupts or doing anything that could generate an exception until a trap handler is installed.  (This may be controversial since somebody decided to file CVE-2021-1104.  If there is overwhelming demand for a mtvecc reset value, it should be Null not Infinity to ensure that trap loops are harmless; pcc provides the initial Infinity capability and clearly must have a reset value.)

tariqkurd-repo:
Interesting information - especially the link to the CVE.
On a purecap machine the reset values can be avoided, except for PCC, as you say, which needs to be the source of infinity.
On a legacy machine it must run existing RISC-V code, without the code ever knowing that it is on a CHERI machine. This only works if the reset value of MTVECC is infinity, otherwise when the boot code writes to MTVEC then it may not be able to successfully update the address field.

Any capabililty width CSRs which are non reset _must_ reset the tag to zero, so (e.g.) mscratchc could _only_ have the tag bit reset.

sorear:
Even with Zcheri_legacy implemented, M-mode always starts in Capability mode (since MXL resets to the widest supported XLEN and xenvcfg CSRs only affect modes lower than x), so I don't think it's possible for cap-supporting hardware to run a cap-naive M-mode.  I wouldn't personally oppose an effort to make cap-naive M-mode work, although past comments (riscv/riscv-isa-manual#820, riscv/riscv-isa-manual#817, possibly others) suggest this is considered a non-goal.

andresag01:
[...]

PR #101, opened 2024-02-09
replace null-cap CSR reset values with a tag clear (fixes issue 43)

tariqkurd-repo:
All CSRs which used to reset to NULL_CAP now just have a tag clear with unspecified data.
All CSRs which reset to infinity are unchanged


arichardson (commented):
This looks reasonable to me. @jrtc27 @sorear do you agree?

Commit message should include "Fixes https://github.com/riscv/riscv-cheri/issues/43"?

sorear:
I think that we will eventually need an ARC position on whether "reset values" apply before or after the execution of an implementation-dependent BootROM and where the appropriate place for compatibility requirements is.

Within the assumptions that have been made previously on this tracker, LGTM.

tariqkurd-repo:
https://github.com/riscv/riscv-cheri/pull/101 still has JVTC as infinity but fixes everything else to do with reset values.

src/riscv-integration.adoc
```diff
@@ -545,7 +545,9 @@ include::img/mscratchreg.edn[]
 ==== Machine Scratch Register Capability (mscratchc)
 
 The <<mscratchc>> register is an extension to <<mscratch>> that is able to hold
-a capability. Its reset value is the <<null-cap>> capability.
+a capability.
+
+{TAG_RESET_CSR}
```
andresag01, line 550: Could you please also fix the text of these other CSRs?
* mepcc sepcc: Currently states that the reset value is NULL
* sscratchc

andresag01 (changes requested):

src/riscv-integration.adoc
tariqkurd-repo: yes ok - thanks for checking

PR #103, opened 2024-02-09
Added note to cbuildcap and reserved cs1 == 0

### LINKED ISSUE #94: Canonical tag clearing pseudo
ISAv9 had a dedicated CClearTag instruction that writes back the input register without a tag. This is not hugely common. One place I've seen it is in the CheriBSD kernel does for a tag-stripping memcpy since it predates tag-clearing semantics for loads without permit_load_cap.

With #82 we could treat `cbuildcap cd, c0, cs2` as the canonical `ccleartag cd, cs1` (previously I believe the best we could do was the two-instruction `cgethigh cd, cs1; csethigh cd, cs1, cd;`.

1) Should we add a note to buildcap that it might be a good idea for a high-performance CHERI core to special-case `cbuildcap cd, c0, cs2` in the decoder to be a tag-clearing move?

2) Should there be an assembly pseudo for ccleartag that maps to `cbuildcap cd, c0, cs2`?

francislaus:
This should close #94 .

### DIFF
--- a/src/insns/cbuildcap_32bit.adoc
+++ b/src/insns/cbuildcap_32bit.adoc
@@ -24,8 +24,12 @@ sealed, `cs1` 's permissions and bounds are equal or a superset of `cs2` 's,
 `cs2` 's bounds are not malformed (see
 xref:section_cap_malformed[xrefstyle=short]), and all reserved bits in `cs2` 's
 metadata are 0. <<CBUILDCAP>> is typically used alongside <<CSETHIGH>> to build
-capabilities from integer values. The case `cs1 == 0` trivially leads to the tag
-of `cd` being stripped.
+capabilities from integer values.
+
+NOTE: The case `cs1 == 0` trivially leads to the tag of `cd` being stripped.
+We opted to not make this a pseudo instruction to clear the tag of capability.
+`cbuildcap cd, c0, cs2` will behave differently in the case of otypes being
+added. Thus we reserve the encoding `cs1 == 0` for now.
 
 Prerequisites::
 {cheri_base_ext_name}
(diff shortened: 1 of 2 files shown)


tariqkurd-repo (approved):

src/insns/cbuildcap_32bit.adoc
```diff
@@ -24,8 +24,12 @@ sealed, `cs1` 's permissions and bounds are equal or a superset of `cs2` 's,
...
 xref:section_cap_malformed[xrefstyle=short]), and all reserved bits in `cs2` 's
 metadata are 0. <<CBUILDCAP>> is typically used alongside <<CSETHIGH>> to build
-capabilities from integer values. The case `cs1 == 0` trivially leads to the tag
-of `cd` being stripped.
+capabilities from integer values.
+
+NOTE: The case `cs1 == 0` trivially leads to the tag of `cd` being stripped.
+We opted to not make this a pseudo instruction to clear the tag of capability.
+`cbuildcap cd, c0, cs2` will behave differently in the case of otypes being
+added. Thus we reserve the encoding `cs1 == 0` for now.
```
andresag01, line 32: Thanks for the change! Just a couple of notes:
* Please avoid writing sentences in the first person (i.e. I, we, etc). It is not common to do this in specifications.
* It would be good to remove mention to "otype" since this is not defined anywhere in this spec. I would suggest instead writing "future extensions" or something like that.

andresag01 (changes requested):

src/insns/cbuildcap_32bit.adoc
jrtc27: NOTE: Although currently this will set the tag to 0 and leave the metadata otherwise unchanged when `cs1` is `c0`, this may change in future extensions, and so software should not assume this.

perhaps?

Note it's not really a reserved encoding; it doesn't trap and it's guaranteed to do exactly what you asked it to do.

andresag01: I would be happy with that note @jrtc27, but I would change the "should not" to "must not"

jrtc27: It is legitimate for software to assume it if it can guarantee (or require) the lack of such extensions, e.g. in an embedded environment where you know the exact targeted hardware

arichardson (approved):
Happy with this change once the suggested edits have been made.

andresag01 (approved):

PR #104, opened 2024-02-09
Rename AUIPCC, CJ* and fix some other minor documentation issues

tariqkurd-repo:


src/insns/jal_32bit.adoc
```diff
@@ -49,19 +38,13 @@ address is not within the bounds of the <<pcc>>. In this case, _CHERI jump or
...
 {cheri_legacy_ext_name}
 
-CJAL Operation::
-+
---
-TODO
---
-
-JAL Operation TODO #where's the target check?# ::
+Operation::
```
arichardson, line 47: Might still need separate semantics here, but we can deal with that later.

arichardson (approved):

PR #96, opened 2024-02-08
CSEALENTRY changes

francislaus:
I renamed the `CSEAL` instruction to `CSEALENTRY` and changed the semantics such that a seal entry operation on an already sealed capability will lead to the tag bit of `cd` being set to `0`. Furthermore, I extended the text describing the sealed bit in order to more closely reflect its actual meaning. This should close #83 .

PR #105, opened 2024-02-09
Csealentry changes

tariqkurd-repo:
replacement for https://github.com/riscv/riscv-cheri/pull/96 as I broke the history


andresag01 (approved):

arichardson (changes requested):
Think we should keep #96 ?

tariqkurd-repo:
I broke https://github.com/riscv/riscv-cheri/pull/96 by messing up the git history - this PR has taken the correct commits from it as a replacement.

arichardson:
> I broke #96 by messing up the git history - this PR has taken the correct commits from it as a replacement.

I guess that works too, but @francislaus could have force pushed a correct history to #96 to make it sensible again. We could have also used squash and merge to not include the PR history in the main repositoyr and instead just one curated commit.

PR #108, opened 2024-02-12
Allow CSRRS/C to read CLEN-wide values

### LINKED ISSUE #92: No way to achieve `CSRR` on a capability
The RISC-V spec says that `CSRR rd, csr` is encoded as `CSRRS rd, csr, x0`. Under the current CHERI-RISC-V spec, `CSRRS` can only read the address, not the metadata. `CSRRW` has a special case for `rd=x0` but none for `rs1=x0`, so unless I'm missing something, it's not possible to read the full value of a capability CSR without overwriting its contents?

andresag01:
Fixes #92

### DIFF
--- a/src/insns/csrr_32bit.adoc
+++ b/src/insns/csrr_32bit.adoc
@@ -52,11 +52,11 @@ These are standard RISC-V CSR instructions with extended functionality for
 accessing CLEN-wide CSRs, such as <<mtvec>>/<<mtvecc>>, which can be
 accessed through either the RISC-V address or the capability address alias.
 +
-Unlike <<CSRRW>>, these instructions only update the address field when writing
-capability CSRs regardless of the CSR's alias used. The final address to write
-to the capability CSR is determined as defined by RISC-V for these
-instructions. The metadata and tag are updated as defined in
-xref:extended_CSR_writing[xrefstyle=short].
+Unlike <<CSRRW>>, these instructions only update the address field and the tag
+as defined in xref:extended_CSR_writing[xrefstyle=short] when writing
+capability CSRs regardless of the CSR alias used. The final address to
+write to the capability CSR is determined as defined by RISC-V for these
+instructions.
 +
 See xref:aliased_CSRs[xrefstyle=short] for a list of CLEN-wide CSRs and
 xref:extended_CSR_writing[xrefstyle=short] for the action taken on writing an XLEN-wide value to each one.
--- a/src/insns/csrr_32bit.adoc
+++ b/src/insns/csrr_32bit.adoc
@@ -30,25 +30,33 @@ endif::[]
 Synopsis::
 CSR access (CSRRWI, CSRRS, CSRRSI, CSRRC, CSRRCI) 32-bit encodings
 
-Register Source Mnemonics::
-`csrr[s|c] rd, rs1, csr`
+Mnemonics for accessing capability CSRs at CLEN-wide aliases::
+`csrrs  cd, rs1, csr` +
+`csrrc  cd, rs1, csr` +
+`csrrwi cd, imm, csr` +
+`csrrsi cd, imm, csr` +
+`csrrci cd, imm, csr`
 
-Immediate Source Mnemonics::
-`csrr[w|s|c]i rd, imm, csr`
+Mnemonics for accessing XLEN-wide CSRs or capability CSRs at XLEN-wide aliases::
+`csrrs  rd, rs1, csr` +
+`csrrc  rd, rs1, csr` +
+`csrrwi rd, imm, csr` +
+`csrrsi rd, imm, csr` +
+`csrrci rd, imm, csr`
 
 Encoding::
 include::wavedrom/csr-instr.adoc[]
 
 Description::
 These are standard RISC-V CSR instructions with extended functionality for
-accessing CLEN-wide CSRs, such as <<mtvec>>/<<mtvecc>> which can be
+accessing CLEN-wide CSRs, such as <<mtvec>>/<<mtvecc>>, which can be
 accessed through either the RISC-V address or the capability address alias.
 +
-Unlike <<CSRRW>> these instruction perform the same update to CLEN-wide CSRs
- to either the XLEN or CLEN-wide alias as they only every perform an XLEN-wide update.
- Where a CLEN-wide CSR is updated, through either alias, the final address is determined
- as defined by RISC-V for these instructions. The metadata and tag are updated
- as defined in xref:extended_CSR_writing[xrefstyle=short].
+Unlike <<CSRRW>>, these instructions only update the address field when writing
+capability CSRs regardless of the CSR's alias used. The final address to write
+to the capability CSR is determined as defined by RISC-V for these
+instructions. The metadata and tag are updated as defined in
+xref:extended_CSR_writing[xrefstyle=short].
 +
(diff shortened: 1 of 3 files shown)


andresag01:
@PeterRugg: I tried a fix as proposed [here](https://github.com/riscv/riscv-cheri/issues/92#issuecomment-1936105294).

src/riscv-integration.adoc
```diff
@@ -325,24 +325,31 @@ NOTE: *CHERI v9 Note:* CSpecialRW is removed. Its role is assumed by
...
-
-{non-csrrw-and} specify *x* registers and so only access the address field
-of the capability when specifying a capability CSR such as <<mtvecc>>.
-They calculate the final address using the standard
-RISC-V behaviour (set bits, clear bits etc.) and that final address is
-updated in the capability. The update typically uses the semantics of a <<CSETADDR>>
-instruction which clears the tag if the capability is sealed, or if the
-updated address is not representable. <<extended_CSR_writing>> shows the
+accessed is of capability size then the capability's tag, metadata and address
+are all accessed atomically.
```
tariqkurd-repo, line 329: please don't reformat - or can it be excluded from the diff somehow? there's no change here

src/riscv-integration.adoc
```diff
@@ -325,24 +325,31 @@ NOTE: *CHERI v9 Note:* CSpecialRW is removed. Its role is assumed by
...
+
+When the <<CSRRW>> instruction is accessing a capability width CSR, then the
+source and destination operands are *c* registers and it atomically swaps the
+values in the whole CSR with the CLEN width register operand.
```
tariqkurd-repo, line 333: please don't reformat - or can it be excluded from the diff somehow? there's no change here

PeterRugg:
Thanks for sorting this! Looks good, and the text is a lot clearer than the previous text. Is it worth adding a note about no write happening, and thus no representable/sealed check happening, when the rs1 index is zero (as discussed in the issue)? It's surely the most sensible interpretation at the moment, but could be worth making it explicit?

andresag01:
@tariqkurd-repo @PeterRugg : I've made an explicit note about the no-write on CSRRS/CSRRC when rs1=x0

tariqkurd-repo (approved):

src/insns/csrr_32bit.adoc
```diff
@@ -30,31 +30,49 @@ endif::[]
...
+Unlike <<CSRRW>>, these instructions only update the address field when writing
+capability CSRs regardless of the CSR's alias used. The final address to write
+to the capability CSR is determined as defined by RISC-V for these
+instructions. The metadata and tag are updated as defined in
+xref:extended_CSR_writing[xrefstyle=short].
 +
 See xref:aliased_CSRs[xrefstyle=short] for a list of CLEN-wide CSRs and
 xref:extended_CSR_writing[xrefstyle=short] for the action taken on writing an XLEN-wide value to each one.
 +
+If `cd` is `c0` (or `rd` is `x0`), then <<CSRRWI>> shall not read the CSR and
```
arichardson, line 64: Should we add a note here that`csrr cd, csr` maps to csrs to match the base ISA?

arichardson (approved):
This LGTM. Would be nice to have a note on csrr but not necessarily required in this pr.

arichardson:
> This LGTM. Would be nice to have a note on csrr but not necessarily required in this pr.

@andresag01 just to clarify this, I'm happy for you to rebase-and-merge this now. The note can be added later.

src/insns/csrr_32bit.adoc
andresag01: I made a note about csrr and csrw.

arichardson (approved):

src/insns/csrr_32bit.adoc
```diff
@@ -30,31 +30,52 @@ endif::[]
 Synopsis::
 CSR access (CSRRWI, CSRRS, CSRRSI, CSRRC, CSRRCI) 32-bit encodings
 
-Register Source Mnemonics::
-`csrr[s|c] rd, rs1, csr`
-
-Immediate Source Mnemonics::
-`csrr[w|s|c]i rd, imm, csr`
+Mnemonics for accessing capability CSRs at CLEN-wide aliases::
```
Timmmm, line 33: Presumably this actually works for all CSR numbers, you just get null cap metadata if you use a non CLEN-wide CSR? Or would it give an assembly-time error?

src/insns/csrr_32bit.adoc
```diff
@@ -30,31 +30,52 @@ endif::[]
...
+accessing CLEN-wide CSRs, such as <<mtvec>>/<<mtvecc>>, which can be
 accessed through either the RISC-V address or the capability address alias.
 +
-Unlike <<CSRRW>> these instruction perform the same update to CLEN-wide CSRs
- to either the XLEN or CLEN-wide alias as they only every perform an XLEN-wide update.
- Where a CLEN-wide CSR is updated, through either alias, the final address is determined
- as defined by RISC-V for these instructions. The metadata and tag are updated
- as defined in xref:extended_CSR_writing[xrefstyle=short].
+Unlike <<CSRRW>>, these instructions only update the address field when writing
+capability CSRs regardless of the CSR's alias used. The final address to write
```
Timmmm, line 56: I think removing `'s` would be more grammatically correct.

src/insns/csrr_32bit.adoc
```diff
@@ -30,31 +30,52 @@ endif::[]
...
+to the capability CSR is determined as defined by RISC-V for these
+instructions. The metadata and tag are updated as defined in
```
Timmmm, line 58: This is quite confusing because you say "these instructions only update the address field ... The metadata and tag are update..."

Also is the metadata updated? I think only the tag can be changed here?

src/insns/csrr_32bit.adoc
andresag01: The instruction opcode is the same in _all_ cases. The idea is instead that the machine infers what's the size of the access based on the CSR number that you are accessing. For example, if you access mtvec, an XLEN value, then your assembly would print *x* registers, but if you access mtvec*c*, a CLEN value at a different CSR number, then your assembly would print *c* registers.

src/insns/csrr_32bit.adoc
andresag01: Fixed

src/insns/csrr_32bit.adoc
andresag01: Fixed

andresag01:
@Timmmm : Could you please review again and resolve the conversation if you are happy with the changes? Thanks!

src/insns/csrr_32bit.adoc
Timmmm: Yeah that's what I mean, it seems a little weird to say "this one is for CLEN-wide aliases" when it would actually work for all CSRs.

andresag01: This is just about the assembly instead the actual encoding

Timmmm: Ok. I can't resolve this conversation because Github's "Resolve conversation" button only seems to appear during a solar eclipse, but consider it resolved!

Timmmm: Ah only the person that opened the PR or people with Write access to the repo can resolve conversations.

tariqkurd-repo:
any objections to merging this one? @arichardson maybe you can do the honours

Issue #85, opened 2024-02-02
Checking whether PCC is sealed


francislaus:
The spec says that for each instruction PCC must be checked for a set of conditions. One is whether PCC is not sealed. I believe that none of the instructions in the current form of the spec can bring a sealed capability into PCC. If we can convince ourselves of this invariant being true at all times, we could drop this architecturally enforced checked on each instruction. This might be a property that can be confirmed by Sail proofs.

andresag01:
There is a contrived case, which might actually happen in the current spec, when the hart goes to debug mode and the debugger writes a sealed capability to `dpcc` then executes `dret`. In this case, the spec does not require that `dpcc` is unsealed before writing to `pcc`.

But I think you are right: it should not be possible to write a sealed capability to `pcc`. I think we should:
* Tidy up the debug part of this
* Remove that check from `pcc` and add a note explaining the rationale for not including that check

Hopefully there is a way to prove that `pcc` will never be "sealed", if not, then it would be prudent to keep the check in.

tariqkurd-repo:
`dpcc`,  `mepcc` and `sepcc` all unseal on execution
so i think Franz is right - pcc can never be sealed.

PR #112, opened 2024-02-14
Clarify that sealing check is not needed on pcc

andresag01:
Fixes https://github.com/riscv/riscv-cheri/issues/85

### DIFF
--- a/src/riscv-integration.adoc
+++ b/src/riscv-integration.adoc
@@ -63,7 +63,7 @@ failing check causes a CHERI exception.
 NOTE: Operations that update <<pcc>>, such as changing privilege or executing
 jump instructions, unseal capabilities prior to writing. Therefore,
 implementations do not need to check that that <<pcc>> is unsealed when
-executing each instruction.
+executing each instruction. However, this property has not yet been formally verified and may not hold if additional CHERI extensions beyond {cheri_base_ext_name} are implemented.
 
 [#section_cap_instructions]
 === Capability Instructions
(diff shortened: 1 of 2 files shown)


tariqkurd-repo (approved):

src/riscv-integration.adoc
```diff
@@ -56,10 +56,14 @@ executed in addition to the checks already required by the base RISC-V ISA. A
 failing check causes a CHERI exception.
 
 * The tag must be set
-* The capability must not be sealed
 * The capability must grant execute permission
 * All bytes of the instruction must be in bounds
 
+NOTE: Operations that update <<pcc>>, such as changing privilege or executing
+jump instructions, unseal capabilities prior to writing. Therefore,
```
francislaus, line 63: While I think we all believe that PCC can never be sealed, we have no proof of this. I am not sure whether it is the right way forward to claim that the sealing check can be omitted without having a proof. Checking that PCC can never be sealed seems to be a perfect case for a Sail proof. I would suggest to leave the check in until we have a proof and change the note such that it states our beliefs.

arichardson: I agree, while I'm pretty sure this cannot happen, I'd leave the requirement above and just add this note.

andresag01: No problem. I left the note and put the check back in.

arichardson (approved):
LGTM with some cautioning wording. @francislaus what do you think?

francislaus:
LGTM with the suggestions from @arichardson.

francislaus (approved):

PR #113, opened 2024-02-14
Fix issue 90

tariqkurd-repo:
Changes are:

1. pcc is no longer accessible as a CSR
2. the pcc value is UNSPECIFIED in debug mode, and ASR permission is implicitly granted
3. add dinfc CSR as a debug mode only source of infinity
4. remove ASR permission from all debug mode CSRs as it's implicitly granted
5. remove "M-mode" as a prerequisite for (e.g.) MTVECC as M-mode must always be present

### DIFF
--- a/src/csv/CHERI_CSR.csv
+++ b/src/csv/CHERI_CSR.csv
@@ -8,11 +8,11 @@ direct write if address didn't change","✔","","✔","","✔","✔","Sdext","De
 Always update the CSR with <<SCADDR>> even if the address didn't change, including the MODE field in the address for simplicity.
 Vector range check ^*^ if vectored mode is programmed.","Apply <<section_invalid_addr_conv>>.
 Always update the CSR with <<SCADDR>> even if the address didn't change, including the MODE field in the address for simplicity.
-Vector range check ^*^ if vectored mode is programmed.","✔","","","","✔","✔","None","Machine Trap-Vector Base-Address Capability","","","","","","","","","","","","","","","","","","","","",""
-"mscratchc","0x760","mscratch","0x340","M","MRW, <<asr_perm>>","tag=0, otherwise undefined","Update the CSR using <<SCADDR>>.","direct write","","","","✔","✔","✔","None","Machine Scratch Capability","","","","","","","","","","","","","","","","","","","","",""
+Vector range check ^*^ if vectored mode is programmed.","✔","","","","✔","✔","M-mode","Machine Trap-Vector Base-Address Capability","","","","","","","","","","","","","","","","","","","","",""
+"mscratchc","0x760","mscratch","0x340","M","MRW, <<asr_perm>>","tag=0, otherwise undefined","Update the CSR using <<SCADDR>>.","direct write","","","","✔","✔","✔","M-mode","Machine Scratch Capability","","","","","","","","","","","","","","","","","","","","",""
 "mepcc","0x761","mepc","0x341","M","MRW, <<asr_perm>>","<<infinite-cap>>","Apply <<section_invalid_addr_conv>>.
 Always update the CSR with <<SCADDR>> even if the address didn't change.","Apply <<section_invalid_addr_conv>> and update the CSR with the result if the address changed,
-direct write if address didn't change","✔","","✔","","✔","✔","None","Machine Exception Program Counter Capability","","","","","","","","","","","","","","","","","","","","",""
+direct write if address didn't change","✔","","✔","","✔","✔","M-mode","Machine Exception Program Counter Capability","","","","","","","","","","","","","","","","","","","","",""
 "stvecc","0x505","stvec","0x105","S","SRW, <<asr_perm>>","<<infinite-cap>>","Apply <<section_invalid_addr_conv>>.
 Always update the CSR with <<SCADDR>> even if the address didn't change, including the MODE field in the address for simplicity.
 Vector range check ^*^ if vectored mode is programmed.","Apply <<section_invalid_addr_conv>>.
@@ -26,7 +26,7 @@ direct write if address didn't change","✔","","✔","","✔","✔","S-mode","S
 Always update the CSR with <<SCADDR>> even if the address didn't change.","Apply <<section_invalid_addr_conv>> and update the CSR with the result if the address changed,
 direct write if address didn't change","✔","","","","✔","✔","Zcmt","Jump Vector Table Capability","","","","","","","","","","","","","","","","","","","","",""
 "dddc","0x7bc","","","D","DRW","tag=0, otherwise undefined","","","","✔","","","✔","","Sdext","Debug Default Data Capabilty (saved/restored on debug mode entry/exit)","","","","","","","","","","","","","","","","","","","","",""
-"mtdc","0x74c","","","M","MRW, <<asr_perm>>","tag=0, otherwise undefined","","","","","","","✔","","None","Machine Trap Data Capability (scratch register)","","","","","","","","","","","","","","","","","","","","",""
+"mtdc","0x74c","","","M","MRW, <<asr_perm>>","tag=0, otherwise undefined","","","","","","","✔","","M-mode","Machine Trap Data Capability (scratch register)","","","","","","","","","","","","","","","","","","","","",""
 "stdc","0x163","","","S","SRW, <<asr_perm>>","tag=0, otherwise undefined","","","","","","","✔","","S-mode","Supervisor Trap Data Capability (scratch register)","","","","","","","","","","","","","","","","","","","","",""
 "ddc","0x416","","","U","URW","<<infinite-cap>>","","","","✔","","","✔","","none","User Default Data Capability","","","","","","","","","","","","","","","","","","","","",""
 "dinfc","0x7bd","","","D","DRW","<<infinite-cap>>","","","","","","✔","✔","✔","Sdext","Source of <<infinite-cap>> capability in debug mode, writes are ignored","","","","","","","","","","","","","","","","","","","","",""
--- a/src/csv/CHERI_CSR.csv
+++ b/src/csv/CHERI_CSR.csv
@@ -29,4 +29,4 @@ direct write if address didn't change","✔","","","","✔","✔","Zcmt","Jump V
 "mtdc","0x74c","","","M","MRW, <<asr_perm>>","tag=0, otherwise undefined","","","","","","","✔","","None","Machine Trap Data Capability (scratch register)","","","","","","","","","","","","","","","","","","","","",""
 "stdc","0x163","","","S","SRW, <<asr_perm>>","tag=0, otherwise undefined","","","","","","","✔","","S-mode","Supervisor Trap Data Capability (scratch register)","","","","","","","","","","","","","","","","","","","","",""
 "ddc","0x416","","","U","URW","<<infinite-cap>>","","","","✔","","","✔","","none","User Default Data Capability","","","","","","","","","","","","","","","","","","","","",""
-"dinfc","0x7bc","","","D","DRW","<<infinite-cap>>","","","","","","✔","✔","✔","Sdext","Source of <<infinite-cap>> capability in debug mode, writes are ignored","","","","","","","","","","","","","","","","","","","","",""
+"dinfc","0x7bd","","","D","DRW","<<infinite-cap>>","","","","","","✔","✔","✔","Sdext","Source of <<infinite-cap>> capability in debug mode, writes are ignored","","","","","","","","","","","","","","","","","","","","",""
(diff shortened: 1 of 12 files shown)


src/insns/pcrel_debug_warning.adoc
```diff
@@ -1,2 +1,4 @@
 NOTE: The instructions on this page are either PC relative or may update the
-<<pcc>>. Therefore an implementation may make them illegal in debug mode.
+<<pcc>>. Therefore an implementation may make them illegal in debug mode. If
+not then the value of the <<pcc>> in debug mode is UNSPECIFIED by this
```
arichardson, line 3:
```suggestion
they are supported then the value of the <<pcc>> in debug mode is UNSPECIFIED by this
```

arichardson: I will need to read the debug spec to ensure that whatever we say here matches the debug spec, but maybe we say PCC has almighty permissions/bounds and whatever address mandated by the debug spec?

src/debug-integration.adoc
```diff
@@ -107,3 +106,13 @@ The <<dscratch1c>> register is a CLEN-bit plus tag bit extension to
 
 .Debug scratch 1 capability register
 include::img/dscratch1creg.edn[]
+
+[#dinfc,reftext="dinfc"]
+==== Debug Infinite Capability Register (dinfc)
+
+The <<dinfc>> register is a CLEN-bit plus tag bit CSR only accessible in debug mode.
+It returns the value of the <<infinite-cap>> capability. It is read/write but with no
```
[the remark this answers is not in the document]
tariqkurd-repo: probably not - and there are many other similar CSRs which are read/write with no writeable fields. The point is regularity - all other Debug mode CSRs are R/W - and having this one cause an exception on writing is irregular (I've already been moaned at for adding custom CSRs in an irregular way - so it's much easier to keep the whole thing consistent).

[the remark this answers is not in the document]
tariqkurd-repo: @Timmmm you complained about irregularity in CSR permissions before  - what's your opinion?

src/debug-integration.adoc
```diff
@@ -50,15 +50,14 @@ Upon entry to debug mode, cite:[riscv-debug-spec], does not specify how to
 update the PC, and says PC relative instructions may be illegal. This concept
 is extended to include any instruction which updates <<pcc>>.
 
+As a result, the value of <<pcc>> is UNSPECIFIED in debug mode, and has no
+architectural effect if PC relative instructions are illegal. Therefore
+<<asr_perm>> is implicitly granted for access to all CSRs in debug mode.
```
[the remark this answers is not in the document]
tariqkurd-repo: >`but that does not mean it is unspecified`

that's exactly what it means, they haven't specified it, but also they haven't written that they haven't specified it they just haven't bothered to mention it.

In contrast, we are saying that we haven't specified it, and admitting to that. That's the only difference.

src/debug-integration.adoc
```diff
@@ -50,15 +50,14 @@ Upon entry to debug mode, cite:[riscv-debug-spec], does not specify how to
...
+
 <<dpcc>> (and consequently <<dpc>>) are updated with the
 capability in <<pcc>> whose address field is set to the address of the next
-instruction to be executed as described in cite:[riscv-debug-spec].
-
-Additionally, the <<pcc>> is updated as follows:
-
-* All metadata is set to the <<infinite-cap>> capability
-  ** The <<pcc>> may be used as a source of the <<infinite-cap>> capability in
-debug mode to allow other capabilities to be created and written into memory.
```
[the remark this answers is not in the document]
tariqkurd-repo: no because the pc is `unspecified`

src/debug-integration.adoc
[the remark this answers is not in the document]
Timmmm: I think if you look at this table then all debug-mode-only registers *have* to be read/write, so I think this is the only correct option.

image

src/csv/CHERI_CSR.csv
```diff
@@ -25,9 +25,8 @@ direct write if address didn't change","✔","","✔","","✔","✔","S-mode","S
...
+"mtdc","0x74c","","","M","MRW, <<asr_perm>>","tag=0, otherwise undefined","","","","","","","✔","","None","Machine Trap Data Capability (scratch register)","","","","","","","","","","","","","","","","","","","","",""
 "stdc","0x163","","","S","SRW, <<asr_perm>>","tag=0, otherwise undefined","","","","","","","✔","","S-mode","Supervisor Trap Data Capability (scratch register)","","","","","","","","","","","","","","","","","","","","",""
 "ddc","0x416","","","U","URW","<<infinite-cap>>","","","","✔","","","✔","","none","User Default Data Capability","","","","","","","","","","","","","","","","","","","","",""
-"pcc","0xcb0","","","U","URO","<<infinite-cap>>
-(address = boot address)","","","✔","","","","✔","✔","none","User Program Counter Capability (to allow reading in legacy mode)","","","","","","","","","","","","","","","","","","","","",""
+"dinfc","0x7bc","","","D","DRW","<<infinite-cap>>","","","","","","✔","✔","✔","Sdext","Source of <<infinite-cap>> capability in debug mode, writes are ignored","","","","","","","","","","","","","","","","","","","","",""
```
Timmmm, line 32: Is this meant to be 0x7be? Looks like `dddc` already uses 0x7bc?

src/debug-integration.adoc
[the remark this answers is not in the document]
tariqkurd-repo: spec updated, please have a look

src/debug-integration.adoc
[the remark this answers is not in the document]
tariqkurd-repo: spec updated please have a look

src/csv/CHERI_CSR.csv
tariqkurd-repo: it was supposed to be 0x7bd, fixed now

src/debug-integration.adoc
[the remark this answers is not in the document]
tariqkurd-repo: I've made it R/W for future expansion now, so that if multiple roots are necessary then this can cover it

src/csv/CHERI_CSR.csv
Timmmm: Oh yeah I forgot the alphabet :-D

src/debug-integration.adoc
```diff
@@ -107,3 +110,22 @@ The <<dscratch1c>> register is a CLEN-bit plus tag bit extension to
...
+[#dinfc,reftext="dinfc"]
+==== Debug Infinite Capability Register (dinfc)
+
+The <<dinfc>> register is a CLEN-bit plus tag bit CSR only accessible in debug mode.
+
+The reset value is the <<infinite-cap>> capability. The M field (if
+{cheri_mode_ext_name} is implemented) represents the current mode and
+will be only updated if <<MODESW>> is executed.
+
+<<dinfc>> is read/write but with no writeable fields, and so writes are
```
arichardson, line 123: Wouldn't it be simpler to declare it is read-only and if needed relaxing that restriction in the future? I'm not sure I see the need for writing even with multiple formats.

src/debug-integration.adoc
```diff
@@ -107,3 +110,22 @@ The <<dscratch1c>> register is a CLEN-bit plus tag bit extension to
...
+ignored.
+
+NOTE: A future version of this specification may add writeable fields to allow creation
```
arichardson, line 126: If there are multiple roots in the future couldn't we just add new debug-"infinite code cap, infinite data cap" CSRs and have this one be one of them? Writing to this CSR seems a bit odd since it's then stateful and we have to keep track of that state while in debug mode?

arichardson (commented):
Not sure about the dinfc R/W designation otherwise LGTM.

src/debug-integration.adoc
Timmmm: It can't be read-only because all the debug-only CSRs are also in the read/write CSRs section.

Does make me wonder why they bothered with the RW/RO distinction at all tbh.

src/debug-integration.adoc
Timmmm: The issue is there's only space for 16 debug-only CSRs and with this change 14 of them are used.

src/debug-integration.adoc
```diff
@@ -48,17 +67,20 @@ include::img/dpccreg.edn[]
 
 Upon entry to debug mode, cite:[riscv-debug-spec], does not specify how to
 update the PC, and says PC relative instructions may be illegal. This concept
-is extended to include any instruction which updates <<pcc>>.
+is extended to include any instruction which updates <<pcc>>, which refers to
```
sorear, line 70:
```suggestion
is extended to include any instruction which reads or updates <<pcc>>, which refers to
```

AUIPC doesn't update pcc

src/debug-integration.adoc
tariqkurd-repo: I think it's more problematic to change teh type of the register and I'm not aware of that ever happening. The difference is trivial really - either you return an error to the debug host on write or you ignore the write. Also the RISC-V debug arch. doesn't include _any_ read only debug mode CSRs, so having it as DRW fits much better without being irregular.

src/debug-integration.adoc
sorear: dcsr, dpc, dscratch0, dscratch1, dddc, dpcc, dscratch0c, dscratch1c, what are the 5 I'm missing?

tariqkurd-repo: One identified problem here was if we introduced an otype later, or something else, for compartmentalization. Would we need to redefine this CSR? The most basic idea was to have this as a CSR where the tag is _always_ set and so the software can write whatever it likes and have it tagged. This idea then extended to different root, or anything else that we may need in the future.

The other reason not to add more CSRs is that there are only 16 available. There are four standard ones: 
image

and we've added an alias to three of them, so that's 7. 
We've also added `dddc` and `dinfc`, making 9 so we're quite close to filling the whole space, and then there's no provision for adding any more without some kind of banking scheme.

So I'd really like to get the "make any cap" functionality based on a single debug CSR if possible.

src/debug-integration.adoc
```diff
@@ -11,6 +11,25 @@ When executing code due to an abstract command, the hart stays in debug mode
...
 
+==== CHERI Mode when in Debug Mode
+
+When entering debug mode, the CHERI mode is always the same, and may be
+switchable depending on which extensions have been implemented.
+
+If {cheri_mode_ext_name} is implemented:
+
+* the core enters Legacy Mode when entering debug mode,
+ and the mode can be switched using <<MODESW>>.
```
sorear, line 22: Why is this in this PR?  It seems unrelated to non-monotonic capability construction.

This will be nontrivial to implement for Rocket-type debuggers with a Debug ROM that contains instructions in a fixed ISA.

src/debug-integration.adoc
arichardson: Fair enough, happy to keep it RW with all written bits ignored.

src/debug-integration.adoc
tariqkurd-repo: because otherwise there's no way of switching to capability mode in debug mode. Previously MODESW was potentially illegal in debug mode.

jrtc27: I continue be of the belief that, just as DXLEN is fixed at the maximum supported, so too should debug mode on a CHERI system always be capability mode rather than legacy mode

tariqkurd-repo: Is it not reasonable on a legacy system that the debugger starts in legacy mode and then switches into cap mode? For software compatibility?

jrtc27: Then you make legacy the primary mode, which the spec is //not// doing in other places.

jrtc27: (and everything that wants to use capability mode, e.g. to read a capability from a CSR, has to waste precious instruction buffer space switching to capability mode)

arichardson (approved):

sorear (approved):
I still think that a Rocket-style debug ROM implementation will require D mode to always execute using the capability ISA, but that can be discussed as a separate issue.

We're no longer violating the debug spec's own rules on PC access, so approve.

src/debug-integration.adoc
```diff
@@ -1,17 +1,20 @@
...
+ for CHERI-RISC-V so change is likely.
+
+==== Debug Mode
 
 When executing code due to an abstract command, the hart stays in debug mode
 and the rules outlined in Section 4.1 of cite:[riscv-debug-spec]
 apply.
 
-=== Core Debug Registers
+==== Core Debug Registers
```
andresag01, line 17: I think the headers within this section are a little off.

andresag01 (commented):
Just a small formatting change. Otherwise looks good to me.

src/debug-integration.adoc
tariqkurd-repo: yeah - there aren't enough levels to make it work - I think it's good enough

src/debug-integration.adoc
tariqkurd-repo: cap mode is the default now, and MODESW is optionally supported

PR #115, opened 2024-02-15
Fix bit number ranges for capability bounds encodings

veselypeta:
I noticed that the bit ranges for bounds were incorrectly specified in some cases.


andresag01 (commented):
The bit ranges on T/B in these images are not incorrect. Its just that the ranges indicate the bits from T and B that are encoded in the capability given that their fully expanded size is MW.

In RV32, for example, MW=10 and the T bits directly in the capability's metadata are TE=T[1:0] and T[7:2], so the final, exapnded T is `T= {T[9:8], T[7:2], TE}`

veselypeta:
My misunderstanding :+1: I had assumed that these were bit ranges

Issue #56, opened 2024-01-26
the auipcc representability challenge


sorear:
Call a segment a sequence of one of more functions and data objects packed together and accessible using a single value of pcc.  Data references and function pointer (sentry) construction from the functions in the segment to the other contents of the segment will use sequences of AUIPCC and CINCOFFSETIMM or load/store instructions, mimicking the use of auipc and addi or load/store sequences in the base ISA.

AUIPCC constructs a capability whose address is within [-2047,2048] of the symbol address.  Assuming that the symbol address is strictly within or at one end of the segment, the intermediate capability constructed by AUIPCC will have an address between 2047 before the segment base and 2048 after the segment end.

Under RV64, this is always representable, because all capabilities have a representable range that covers at least 2048 before and 8193 after their bounds.  Under RV32, the representability depends on the segment size.  If the segment is strictly less than 2048 bytes, then the only AUIPCC instructions which appear will have an immediate of 0, which returns the pcc unchanged and is always representable.  If the segment is 4096 bytes or larger, it will have a representable range covering 2048 before and after and any intermediate value for any address inside the segment will be representable.

The problematic case for RV32 is segments of [2048,4095] bytes. Within that size range, addresses at least 2049 bytes after the end are guaranteed representable, but only 1024 before the base.  If an AUIPCC near the end of the segment is used for a symbol near the beginning, the AUIPCC value can fall off the representable region, resulting in an untagged capability.

Plausible solutions:

* Change AUIPCC immediate decoding for RV32 Capability mode to produce imm[11], presumably from insn[30].
   Pro: provides access to the entire representable region of segments of any size.  Simple access semantics.
   Con: Notable change from base ISA behavior.  Restricts segment sizes to 1GiB.
* Add a S2048PCC instruction, to cover the one missed case (access 2048 to 4096 bytes before pcc uses S2048PCC and is monotonic).
   Pro: Accesses the entire segment with minimal opcode space impact.  No change to immediate encodings.
   Con: Brittle.  Any change to the capability design which affects representable regions may require adding or redefining instructions.
* Pad the beginning of segments in the problematic [2048,4095] size range with 1024 bytes of unused data, or enough to reach a safe size of 4096.
   Pro: No ISA changes.  No relocation changes.
[...]

PR #116, opened 2024-02-15
Fix the auipcc reachability challenge

PeterRugg:
This is Jon's proposal to change the specification according to Lawrence's suggestion, addressing https://github.com/riscv/riscv-cheri/issues/56.

Ideally we wouldn't merge it until we implement it, at least in our cheri-cap-lib, to confirm it doesn't affect timing too much to do a MW-bit comparison rather than a 3-bit comparison on these paths, and to make sure it fixes the issue.

### DIFF
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -392,7 +392,7 @@ R  = B - 2^MW-2^
 .Calculation of top address correction
 [#cap_encoding_ct,options=header,cols="^1,^1,^1",width="40%",align="center"]
 |==============================================================================
-| A < R    | T < R    | c
+| A < R    | T < R    | c~t~
 | false    | false    | 0
 | false    | true     | +1
 | true     | false    | -1
@@ -402,7 +402,7 @@ R  = B - 2^MW-2^
 .Calculation of base address correction
 [#cap_encoding_cb,options=header,cols="^1,^1,^1",width="40%",align="center"]
 |==============================================================================
-| A    < R | B    < R | c
+| A    < R | B    < R | c~b~
 | false    | false    | 0
 | false    | true     | +1
 | true     | false    | -1
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -386,7 +386,7 @@ xref:cap_encoding_cb[xrefstyle=short].
 
 ```
 A = a[E + MW - 1:E]
-R  = B - 2^MW-3^
+R  = B - 2^MW-2^
 ```
 
 .Calculation of top address correction
@@ -431,7 +431,7 @@ _representable_ region while maintaining the original bounds. This is enabled by
 defining a lower boundary R from the out-of-bounds values that allows us to
 disambiguate the location of the bounds with respect to an out-of-bounds address.
 R is caluculated
-relative to the base by subtracting 2^MW-3^ from B.
+relative to the base by subtracting 2^MW-2^ from B.
 If B, T or _a_[E + MW - 1:E] is less than R, it is inferred that they lie in the
 2^E+MW^ aligned region above R labelled space~U~ in
 xref:cap_bounds_map[xrefstyle=short] and the corrections _c~t~_ and _c~b~_ are
@@ -568,7 +568,7 @@ xref:cap_bounds_map[xrefstyle=short].
 
 The gap between the object bounds and the bound of the representable range
 is always guaranteed
-to be at least 1/4 of `s`. This is represented by `R = B - 2^MW-3^` in
+to be at least 1/4 of `s`. This is represented by `R = B - 2^MW-2^` in
 xref:section_cap_encoding[xrefstyle=short].
 This gives useful guarantees, such that if an executed instruction is in
 <<pcc>> bounds, then it is also guaranteed that the next linear instruction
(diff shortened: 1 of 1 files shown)


src/cap-description.adoc
```diff
@@ -385,16 +385,14 @@ definitions in xref:cap_encoding_ct[xrefstyle=short] and
...
-R  = Bc - 1
+A = a[E + MW - 1:E]
+R  = B - 2^MW-2^
 ```
 
 .Calculation of top address correction
 [#cap_encoding_ct,options=header,cols="^1,^1,^1",width="40%",align="center"]
-| A~c~ < R | T~c~ < R | c~t~
+| A < R    | T < R    | c
```
andresag01, line 395: This one is still ct to define the variable above for top

src/cap-description.adoc
```diff
@@ -404,7 +402,7 @@ R  = Bc - 1
 .Calculation of base address correction
 [#cap_encoding_cb,options=header,cols="^1,^1,^1",width="40%",align="center"]
 |==============================================================================
-| A~c~ < R | B~c~ < R | c~b~
+| A    < R | B    < R | c
```
andresag01, line 405: This one is still cb to define the variable above for base

src/cap-description.adoc
PeterRugg: Quite right. Fixed

src/cap-description.adoc
PeterRugg: Fixed

arichardson:
@LawrenceEsswood does this look correct to you?

src/cap-description.adoc
```diff
@@ -429,15 +427,18 @@ and a carry out is implied if `T[MW - 3:0] < B[MW - 3:0]` since it is
...
-computed accordingly. The overall effect is that at least 2^E+MW^/8 bytes below
-the base address and 2^E+MW^/4 bytes above the top address can roam
-out-of-bounds while still allowing the bounds to be correctly decoded.
+computed accordingly. The overall effect is that the address can roam
+2^E+MW^/4 bytes below
+the base address and at least 2^E+MW^/4 bytes above the top address
+while still allowing the bounds to be correctly decoded.
 
 [#cap_bounds_map]
 .Memory address bounds encoded within a capability
```
Timmmm, line 444: Btw this diagram will become out of date with this change.

sorear:
Are we intentionally doing this for both RV64 and RV32 or should it be RV32 only?

Assuming that the cost is negligible, I don't have a problem with the change as such, although for the reasons discussed on the mailing list I now have a weak preference for redefining auipcc in such a way as to be compatible with any representability definition.

sorear:
This semantic change is very annoying for the CSR architecture.  Pre this PR, the low 7 or 12 bits of a capability's address can be changed without representability checks, so you don't need representability checking for writes to jvt, xepc or xtvec (assuming that vectored mode is not implemented or is implemented with an appropriate alignment requirement).  This might also affect the implicit bit zero clear in jalr, and the fuzzy block addressing used by CMOs.

jrtc27:
Writes to CSRs change, yes, and we’re aware of this potential concern. I don’t believe any of the other issues you raise are real though; if the capability in use becomes unrepresentable it will be out of bounds of the original capability, and since bounds checks in instructions are done with the original bounds you will take a bounds fault deterministically.

LawrenceEsswood:
> @LawrenceEsswood does this look correct to you?

Yeah, this is what I had proposed. Although I have nothing to add about how it complicates writes to legacy CSRs, I had not considered that would cause issues.

PeterRugg:
Here's an attempt to implement this in Sail, by the way. https://github.com/CTSRD-CHERI/sail-cheri-riscv/tree/auipcc-representable-fix . To prove we cover all the cases, we probably want to bring https://github.com/CTSRD-CHERI/sail-cheri-riscv-verif/blob/master/cap_properties.sail up to date and add a property.

src/cap-description.adoc
tariqkurd-repo: I think the diagram is still correct, this just changes how far the representable space can extended past the bounds in each direction.

tariqkurd-repo:
Are there any opjections to merging this one? 
I'd like to see the proofs run to prove that the problem is solved, but I think that this should be accepted into the project while this happens.

arichardson:
@PeterRugg agreed to run the proofs in the TG meeting today, so we should hopefully be able to merge this PR soon.

sorear:
Are the meetings being organized somewhere other than the SIG CHERI lists.riscv.org mailing list?

jrtc27:
There is a CHERI SIG / TG meeting slot every other week, present in the RISC-V technical meetings shared calendar.

tariqkurd-repo:
Fix https://github.com/riscv/riscv-cheri/issues/56

PRugg-Cap:
FWIW, I think I have proved for RV32 and RV64 that this proposed fix, as implemented in https://github.com/CTSRD-CHERI/sail-cheri-riscv/tree/auipcc-representable-fix, fixes the problem, in the sense that, for all encoded, executable (unsealed, in-bounds, ...) capabilities, and all in-bounds target addresses within the reachable range of AUIPCC and CJALR, the AUIPCC will not break representability and the CJALR will succeed and take you to the intended target address.

The property finds the known counterexample for RV32 without the fix, then proves the property once the fix is applied. RV64 goes through fine (albeit slowly!) with or without the fix. There are some issues with making the property and proof flow public that I'll try to get sorted out. I also had to hack in some minor changes to Sail's SMT backend, so take with a pinch of salt!

PR #119, opened 2024-02-15
Capabilities are little endian

### LINKED ISSUE #69: Define capability endianness
An implementation of Zcheri where endianness is switchable at runtime must use a fixed endianness for capabilities, since RISC-V requires byte sequences* to remain invariant across endian switches, and if an invariant byte sequence were reinterpreted with the opposite endianness it would result in a catastrophic loss of provenance.

Such an implementation must present a capability endianness opposite its integer endianness in one mode, so a requirement that capability endianness equal integer endianness is unsupportable in the general RISC-V setting.  Allowing the capability endianness to be implementation defined separate from the integer endianness would contradict the requirement of an architectural capability representation that can be inspected in byte dumps.

As such: Require that capability representation be little-endian on all systems, or only define Zcheri for fixed little-endian systems.

* The precise wording of the requirement is that a byte written at some address will remain readable at this address.  Narrowly, this means that byte invariance is only required within a CLEN-bit region if the tag is zero, and an implementation could allow the capability endianness to appear to vary by reversing the bytes of CLEN-bit regions on loads and stores conditionally on a set tag.  This is unlikely to be microarchitecturally reasonable.

sorear:
Possible alternatives include requiring little endian for Zcheri_purecap, allowing bi-endian implementations to use an implementation defined endianness, and rules-lawyering the unprivileged ISA byte invariance requirement to only apply to CLEN-bit regions with the tag cleared.  This is the only option that is consistent with the architectural capability representation and how we handle variable XLEN.

Resolves #69.

### DIFF
--- a/src/riscv-integration.adoc
+++ b/src/riscv-integration.adoc
@@ -472,7 +472,9 @@ both {cheri_base_ext_name} and {cheri_legacy_ext_name}.
 
 The *mstatus* and *mstatush* registers operate as described in
 cite:[riscv-priv-spec] except for the SXL and UXL fields that control the
-value of XLEN for S-mode and U-mode, respectively.
+value of XLEN for S-mode and U-mode, respectively, and the MBE, SBE, and UBE
+fields that control the memory system endianness for M-mode, S-mode,
+and U-mode, respectively.
 
 The encoding of the SXL and UXL fields is the same as the MXL field of *misa*,
 shown in xref:misa_mxl_field[xrefstyle=short]. Only 1 and 2 are supported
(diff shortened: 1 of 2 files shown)


jrtc27:
Or we just take the same approach as XLEN and say you only get CHERI state/instructions/etc in the native endianness, which is little-endian in the base privileged ISA if supported (given that's the well-defined reset state for any system that's not exclusively big-endian). It's not like big-endian RISC-V is a thing anyone actually cares about... (see the lack of any toolchain support beyond some very basic defines of riscv*be in GCC).

arichardson (approved):
I'm happy with either this or requiring them to be whatever the reset endianness is for a bi-endian system (I assume this would always be little)?

tariqkurd-repo:
I think it seems rasonable to go with the reset endianness - otherwise big endian systems will get confusing (if anyone actually builds one....)

sorear:
I agree, reset makes more sense and we can reuse the XLEN behavior.

andresag01 (approved):

sorear:
It looks like when this was merged, GitHub automatically replaced the (correct because it was edited) commit message with the (wrong) PR title and text of the original post in the PR. For #131 the commit messages I wrote were lost entirely.

Is this an intentional workflow decision or something that we're otherwise stuck with? I spent a fair amount of effort on atomic commits and meaningful commit messages for several of the other PRs and would rather it not be wasted.

arichardson:
The default is squash and merge in this repository, but if there is more than one commit I make sure to use the rebase and merge option (unless it's a messy history with fixup commits). My workaround so far has generally been to sync the PR message with the commit message so that the GitHub default remains sensible.

PR #120, opened 2024-02-15
Add IntelliJ editing plugin support

arichardson:
Move attribute definitions to a separate .adoc file, so that they
can be read by the IntelliJ plugin. 
This makes it easier for me to edit the files and will not have
any functional changes. Opening this as a pull request to ensure
I don't conflict with anyone currently editing the asciidoc attributes.


andresag01 (changes requested):
Thanks for the changes. However, the build is getting a new warning (see [here](https://github.com/riscv/riscv-cheri/actions/runs/7923827199/job/21634332761?pr=120#step:6:144)) and the title page of the PDF is not displaying the right logos. Could you please take a look? Thanks!
Screenshot from 2024-02-16 10-43-16

tariqkurd-repo:
Hi Alex - any idea why the template isn't showing in the built PDF?

arichardson:
Looks like the missing image is due to the newline between the title and the `include[]` statement. Did not realize asciidoc was that whitespace sensitive.

andresag01 (approved):
The generated PDF has the expected layout, so I'll approve.

PR #123, opened 2024-02-20
Remove XLENMAX

### LINKED ISSUE #67: Remove XLENMAX
The ARC is [considering to make MXLEN constant](https://lists.riscv.org/g/tech-unprivileged/message/646).  If this happens, XLENMAX will become equal to MXLEN in all cases and we can simplify our spec by removing the Zcheri-specific term.

sorear:
As of riscv/riscv-isa-manual@86ff22d6 MXLEN is read-only and reflects the widest XLEN of the machine, so use that instead.

As of riscv/riscv-isa-manual@0d7621d9 MXLEN >= SXLEN >= UXLEN is required by the base privileged architecture, so it does not need to be restated here.

Resolves #67.

### DIFF
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -10,6 +10,10 @@ change dynamically at run-time depending on the values written to CSRs, so we
 define capability behavior in terms of MXLEN, which is the value of XLEN used
 in machine mode and the widest XLEN the implementation supports.
 
+NOTE: {cheri_base_ext_name} assumes a version of the privileged architecture
+which defines MXLEN as constant and requires higher privilege modes to have at
+least the same XLEN as lower privilege modes; these changes are present in the
+current draft and expected to be part of privileged architecture 1.13.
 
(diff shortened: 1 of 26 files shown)


arichardson (approved):
Thanks! This looks good to me, will give others time to take a look and then merge it.

Only comment I have is whether we need to add a note that we assume privspec 1.13 with those changes as our baseline?

andresag01 (approved):
It looks good, but do we need to wait for those changes in the RISC-V privileged spec to be formally released before merging this?

arichardson:
> It looks good, but do we need to wait for those changes in the RISC-V privileged spec to be formally released before merging this?

This PR explicitly references the privspec draft version, so I don't think we need to wait until it has been released (I assume that will happen before this spec is close to ratification).

andresag01:
@sorear : Thanks for your contribution! I rebased the PR and prepared it for merge since it was behind main.

PR #124, opened 2024-02-21
Redundant pcc unseal text

tariqkurd-repo:
remove redundant exception check for the PCC being sealed


francislaus (changes requested):
This does not match the result of the discussion in #112 if I understand your PR correctly.

tariqkurd-repo:
@francislaus this PR is saying that there can't be an exception caused by the PCC being sealed, so I've removed that from the exception table

andresag01 (approved):

src/riscv-integration.adoc
```diff
@@ -1090,6 +1090,8 @@ NOTE: `auth_cap` is <<ddc>> for Legacy mode and `cs1` for Capability Mode
 | capability stores                              | 6                    | N/A                     | N/A                        |capability alignment  | Misaligned capability store
 |=========================================================================================
 
+^1^ this check is architecturally required, but is not required in an implementation
```
arichardson, line 1093:
```suggestion
^1^ This check is architecturally required, but is impossible to encounter so may not required in an implementation.
```

Maybe like this?

arichardson (approved):
LGTM with the footnote changes to a full sentence.

PR #126, opened 2024-02-23
fix-issue-122: Add rules around programming CRE and CME bits, and variable XLEN

tariqkurd-repo:


src/riscv-legacy-integration.adoc
```diff
@@ -250,13 +250,33 @@ cause illegal instruction exceptions
 addresses (xref:csr-numbers-section[xrefstyle=short]) cause illegal
 instruction exceptions
 * All allowed instructions execute as if the CHERI execution mode is Legacy.
-The CME bits in <<mseccfg>>, <<menvcfg>>, and <<senvcfg>> have no effect whilst
+The CME bits in <<menvcfg>> and <<senvcfg>> are zero whilst
```
andresag01, line 253: This could use a bit of clarification. 
* If senvcfg.CRE=0 then senvcfg.CME=0
* If menvcfg.CRE=0 then menvcfg.CME=0 && senvcfg.CRE=0 && senvcfg.CME=0

andresag01 (commented):
Thanks for the changes. Could you also please clarify in the description of xenvcfg what is the relationship between the CME bits in different privileges? Thanks!

src/riscv-legacy-integration.adoc
tariqkurd-repo: that's in the table already

src/riscv-legacy-integration.adoc
```diff
@@ -250,13 +250,33 @@ cause illegal instruction exceptions
...
 CHERI register access is disabled.
 
 CHERI register access is disabled if XLEN in the current mode is less than
 XLENMAX or if CRE active at the current mode (<<menvcfg>>.CRE for S-mode or
 <<senvcfg>>.CRE for U-mode) is 0.
 
+<<menvcfg>>.CRE, <<menvcfg>>.CME, <<senvcfg>>.CRE and <<senvcfg>>.CME form a
+single WARL field. This allows higher privilege software to restrict lower
```
sorear, line 261: WARL fields must be written atomically, so it makes very little sense to use WARL to describe state split between multiple registers.

Every existing xenvcfg and xstateen field is described using the language of "if X is zero, Y is read-only zero".  The CHERI configuration registers should follow that convention.

I'm expecting to dramatically change the CRE factorization as part of the second work unit of #39, once I have a change and description I'm happy with.  I feel like I've wasted your time by failing to communicate that there is no purpose in working to improve the transitional CRE.

jrtc27: Per-mode CRE, whatever it’s called, needs to exist, so I’m not sure what you think needs redesigning there. Before you go ahead and spend time writing up a complete proposal, perhaps you should sketch it out, as I worry we’d then lose something we need.

sorear: Right now my #39 efforts are split into four pieces:

* #81 
* Remove hybrid ABI support from Zcheri_legacy by deleting all CME bits and tying them to the CRE bits.  This allows a large number of downstream simplifications to the ISA.
* Alternate Zcheri_mode which only tracks the mode with the pcc and xepc, implicitly assuming that all capabilities used as function pointers contain capability ISA code.
* Rework the descriptions of CLEN-bit CSR writes to make it clear that you can implement them with only one SCADDR unit, NFCI

jrtc27: Whether or not CME exists, xenvcfg.CRE still affects yenvcfg.CRE for x >_priv y just as it affects yenvcfg.CME for x >=_priv y, so I don't think it's much wasted work to extend the definition to cover CME.

andresag01 (approved):

src/riscv-legacy-integration.adoc
tariqkurd-repo: on that basis are you happy for me to merge this one @jrtc27 ?

src/riscv-legacy-integration.adoc
```diff
@@ -250,13 +250,33 @@ cause illegal instruction exceptions
...
+xref:xenvcfg-warl-field[xrefstyle=short].
+
+[[xenvcfg-warl-field]]
+.Xenvcfg joint WARL field
+[%autowidth,float="center",align="center",cols="<,<,<,<,<",options="header"]
+|===
+|<<menvcfg>>.CRE|<<menvcfg>>.CME|<<senvcfg>>.CRE|<<senvcfg>>.CME|Comment
+|0              |0              |0              |0              |<<menvcfg>>.CRE=0 disables all other state
+|1              |0/1            |0              |0              |<<senvcfg>>.CRE=0 disables <<senvcfg>>.CME
+|1              |0/1            |1              |0/1            |both levels of CME can be programmed
```
arichardson, line 273: I find this table slightly difficult to parse, maybe the following makes the implications of setting a given bit more obvious?
```suggestion
|0              | RO 0          | RO 0          | RO 0          |<<menvcfg>>.CRE=0 disables all other state
|1              |0/1            |0              | RO 0          |<<senvcfg>>.CRE=0 disables <<senvcfg>>.CME
|1              |0/1            |1              |0/1            |both levels of CME can be programmed
```

src/riscv-legacy-integration.adoc
jrtc27: Yes; this looks like what we discussed

Issue #127, opened 2024-02-23
CHERI exception priority table incomplete


tariqkurd-repo:
The CHERI exception priority table shows PCC checks and load/store checks but ti doesn't include the equivalent of illegal instruction checks e.g. ASR fault on CSR access

PR #128, opened 2024-02-23
add ASR permission and target out of bounds exception priority

tariqkurd-repo:
fix [issue ](https://github.com/riscv/riscv-cheri/issues/127)


src/riscv-integration.adoc
```diff
@@ -764,10 +764,14 @@ Instruction access fault
...
 | .>|2 +
 0 +
+*{cheri_excep_mcause}* +
+*{cheri_excep_mcause}* +
 8,9,11 +
 3 +
 3 .<|Illegal instruction +
 Instruction address misaligned +
+*CHERI <<ASR-permission>> fault* +
+*CHERI branch/jump target out of bounds fault* +
```
andresag01, line 774: I agree that the ASR permission might be missing, but the jump/branch failures are already below. Note that a few rows below it says:

> *Prior to address translation for an explicit memory access or jump:*
*CHERI fault*

andresag01 (changes requested):

andresag01:
@tariqkurd-repo : I added more information to the table and shifted around the entries a little. Here is the new proposed table.
Screenshot from 2024-02-26 16-49-16
What do you think?

PR #130, opened 2024-02-25
descope and simplification of Zcheri_legacy

sorear:
This is the main part of #39.

What we lose: You can no longer manipulate the capability environment from Legacy mode without either an environment call or Zcheri_mode. S-mode must run at least partially in Capability mode to support capability execution in U-mode.

What we gain: Cuts three quarters of the architectural state required to support Zcheri_legacy beyond that of Zcheri_purecap. Saves 20 CSR allocations (13 currently plus 7 required to extend the current scheme to support H and Srnmi), likely more with extensions not yet defined. Returns a full minor opcode to greenfield state.


sorear:
> In other words, there are no longer alias addresses to access CLEN CSRs. Is this correct @sorear?

Yes.

> As you pointed out, code running in Legacy mode is no longer able to access capability-wide CSRs which is why this idea was discarded when we last discussed it. @sorear : Why do you think that this drawback is not be a problem?

The individual commit messages have fairly detailed rationales specific to the commits.

In short, Legacy mode is used for purely non-capability code, Capability mode is used for anything that does nontrivial things with capabilities, and at the boundary between them you need an environment call anyway to reconfigure xenvcfg so do the capability CSR reconfiguration on the Capability mode side of the boundary.

Is there a specific use case you'd like me to address?

arichardson:
The "hybrid" model can be rather useful where you use some capability instructions and you use them frequently enough that trapping would be a significant cost.
For example the CheriBSD kernel can be used in such a mode (and is currently still the default for CheriBSD on Morello I believe).

I think merging the CSR namespace and maybe also removing the extra scratch registers are valuable changes, but the hybrid programming model is rather important if you want to support gradual adoptions.

LawrenceEsswood:
Deleting Xtdc does cause some issues for (at least trivially) hybrid-izing kernels. 

A hybrid kernel needs to allow userspace to mess with DDC, which means it needs to restore its own before it can do any spills / loads, so that typically will consume a capability width scratch register. It _also_ typically needs a register to hold an address to spill / restore. This cannot also be DDC as there are no instructions that use the address of DDC as base address.

Without two scratch registers, once DDC has been swapped, no registers are left (apart from possibly the zero register) to use as an address. Any existing kernels that had been using scratch to hold a stack pointer or something like it to spill to would be in trouble. A kernel may also try to use a `lui + add` sequence to load an address to spill to. This also breaks without two scratch registers.

The only real option without two registers is to only write purecap sequences for such a kernel (just to spill a few registers), and then switch to legacy mode for the rest of the kernel. Possibly this is the solution, but then mode switches need to be in included in anything that claims to support legacy.

sorear:
@arichardson If you have a kernel that's mostly legacy code, but also needs some capability-aware code to handle traps from a capability-aware userspace, and can't afford the cost of hypercalls to switch in and out of capability mode on privilege switches, that sounds like you should be using Zcheri_mode. Luckily, with #133 the marginal cost of Zcheri_mode is only 4 bits of architectural state, compared to the minimum 236 this PR saves over the baseline.

I hope that there is no intention of supporting "Legacy mode with writes to ddc around every memory access" as a compilation target for CHERI C/C++?

@LawrenceEsswood this matches my analysis for hybrid kernels. To me the use case for _not_ supporting Zcheri_mode is that you're running a purecap kernel and at least one purecap application, but you have to deal with some legacy applications and/or legacy boot stages. I don't see purecap UEFI happening any time soon. In this situation, changes to senvcfg.CME only happen on context switches or execve, and changes to menvcfg.CME happen a bounded number of times and a hypercall is fine for them.

jrtc27:
FWIW, purecap EDK2 is being worked on at the moment.

arichardson:
@sorear I have tried to include some of the simplifications in #162 and am personally in favour of dropping the separate CSR namespace if we can make the mode switching instruction mandatory (#161)

arichardson:
We discussed this in the CHERI TG earlier today and believe that merging of CSR space and removal of JALR.MODE are changes that can be merged immediately.

@sorear I plan to pull out those commits from this PR as a separate PR if you don't mind.

arichardson:
@sorear Thanks for all of these changes! I've split out #168 and #167 from this PR which leaves the xtdc removal. CME is removed in #162, zcheri_mode is removed in #161 . Please let me know what you think of these individual changes

PRugg-Cap:
I believe these have all been separately implemented except for https://github.com/riscv/riscv-cheri/pull/130/commits/4e48ffef6db7bf91b89be98027b21fd7cf86a0de. I think we currently plan to keep the *tdc registers.

tariqkurd-repo:
can we close the PR then @PRugg-Cap ?

andresag01:
There has not been activity in this PR for sometime and some of the ideas here are already merged into `main` via other PRs, so I will close this for now. Please feel free to create/re-open the PR if you'd like to propose further changes.

PR #133, opened 2024-02-25
Zcheri_mode "loose" redefinition

sorear:
Meet the functional requirements of a hybrid ABI with 4 bits of architectural state and no use of omnipresent capability metadata.  Independent of but strongly synergistic with #130.


sorear:
> My understanding of this proposed change is:

Correct.

> Then, if I understand correctly, the mode can only be changed in the following ways:

Taking a trap at any privilege also changes the mode, as does `JALR.MODE`.

> The idea of keeping the CHERI mode in the capability's encoding is nice because then all the information you need to jump to some code (i.e. permissions, address, bounds, mode, etc) is in one place and its inherently synchronized by-design.

I agree completely about the benefits of having as much information as possible relevant to the secure interpretation of a capability in the capability itself.

We're already trading off capability semantic completeness with capability metadata size by not including an ASID in the metadata. I suspect that if you decided to spend a bit to catch errors, a bit of ASID would help more than a capability mode bit, which will virtually always be 1 (since legacy code does not use capabilities for its function pointers and return addresses, the _only_ places legacy mode code capabilities appear are in capability mode to legacy mode switch code and trap code).

Wide availability of a feature is its own kind of niceness. That's the direction I'm hoping to move this in.

> For example, if you wanted to context switch a process, you'd need to retrieve the pcc, then the mode and then store them both separately.

The saved mode is in `sstatus`, which has been part of the context switch on Linux and seL4 since the beginning and appears to also be in the CheriBSD switch code.  A context switch without `sstatus` will break quite a few extensions.

> Also, if you wanted to have truly hybrid code (e.g. because your program is Capability but there is this only library that's Legacy), then you'd need to do something MODESW followed by jump to call the library and then MODESW back when you return which is inconvenient.

You already need nontrivial codegen to spill all call-saved capability registers, reestablish `csp` as a capability after the call, and switch to a stack covered by `ddc` if that wasn't a given.  Two `MODESW` fit right in.

andresag01:
This PR has not seen activity for several months, so I will close it for now. Please rebase, update and re-open it if you'd like to pursue this change although it would best to discuss it in the TG call or mailing list first.

Also, we cannot remove instructions like SCMODE because the ISA must provide mechanisms to get/set all fields in the capability without having to manually decode the metadata and then use `cbld`.

PR #136, opened 2024-02-27
Invert subset/superset for CBLD

Timmmm:
Very minor but I think this way round is easier to thing about (you can only build a subset of the authorising cap), and more importantly it matches the sense of SCSS.

### DIFF
--- a/src/insns/cbld_32bit.adoc
+++ b/src/insns/cbld_32bit.adoc
@@ -22,7 +22,7 @@ include::wavedrom/cbld.adoc[]
 
 Description::
 Copy `cs2` to `cd` and set the tag to 1 if `cs1.tag` is set, `cs1` is not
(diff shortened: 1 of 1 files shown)


tariqkurd-repo:
I think that the specification is correct because:

CBLD (CBUILDCAP)  has a check that cs2 is a subset of cs1
- cs2 is authorised by cs1 in this case, so it's copied to the output with cs1's tag

SCSS (CTESTSUBSET) has a check that cs2 is a subset of cs1

so that they match. @Timmmm ?

arichardson (approved):
I agree it's slightly easier to read and since it doesn't change the semantics I'm happy for this to merged.

Timmmm:
> so that they match

Yeah, this doesn't change the semantics (which were already correct); it's just rewording a bit to make it clearer / more regular.

PR #137, opened 2024-02-27
fix C.MODESW encoding so it doesn't overlap C.SUBW

tariqkurd-repo:
I've changed this to the intended encoding, which is already in riscv-opcodes.
The previous one was a typo and conflicted with c.subw


tariqkurd-repo:
https://github.com/riscv/riscv-opcodes/pull/214 is the reference. I had incorrectly typed in the encoding into this doc, which this PR fixed.

PR #141, opened 2024-03-11
Corner cases for narrow XLENs

sorear:
The address behavior is dictated by the base ISA and the desire to keep Legacy and Capability mode as close as possible. This just defines metadata and CHERI exception behavior in some cases that aren't covered by Zcheri_purecap.

---

I think there are currently five semantically distinct places where representability checks are needed:

1. SCADDR instruction
2. Writes of capability CSRs through XLEN-bit aliases, using bitwise instructions, or after legalisation
3. Predetermining truncation results for explicit writes to variable width capability CSRs
4. Predetermining truncation results for writes to xepcc as a result of traps
5. Truncation of pcc on trap return

Cases 3-5 are only possible if multiple XLEN values are supported. 3-4 are only possible if there are multiple SXLEN or (for 3 only) Zcmt is implemented. 3 can share significant logic with 2 since it operates on the same input capability. Depending on the pipeline design it may be possible to combine 1 with 2, 1 with 5, or 3 with 4.

### DIFF
--- a/src/riscv-legacy-integration.adoc
+++ b/src/riscv-legacy-integration.adoc
@@ -287,6 +287,73 @@ NOTE: Disabling CHERI register access prevents a low-privileged Legacy mode
 from interfering with the correct operation of higher-privileged Legacy modes
 that do not perform <<ddc>> switches on trap entry and return.
 
+=== CHERI execution with XLEN less than XLENMAX
+
+While XLEN is less than XLENMAX CHERI register access is disabled as described
+in xref:section_cheri_disable[xrefstyle=short]. Base ISA instructions still
+need to perform capability checks on XLEN-wide addresses.
+
+==== Addresses in pcc
+
+The `pc` width is the current XLEN. The address field of <<pcc>> is the `pc`
+sign extended to XLENMAX bits at all times. All changes to the `pc`, including
+jumps, conditional branches, trap returns, and sequential execution, sign
+extend `pc` to set the <<pcc>> address, and a <<pcc>> address which is not sign
+extended is an impossible state.
+
+[NOTE]
+====
+Sign extension of `pc` into `__x__epc` is a requirement imposed by the base
+ISA.
+
+Sequential execution across 2^XLEN^/2 or from 2^XLEN^ to 0 when XLEN is less
+than XLENMAX and <<pcc>> does not hold an infinite-bounds capability has
+different effects depending on whether the addressing boundary coincides with
+an instruction boundary. A single instruction which spans the address boundary
+will be treated as a length violation and is reported as a CHERI exception with
+`__x__epcc` the starting address of the instruction. If an instruction ends
+exactly at the addressing boundary, the instruction will complete successfully
+but the <<pcc>> becomes unrepresentable afterward, and a tag violation CHERI
+exception is taken on the next instruction unless an interrupt or higher
+priority synchronous exception is taken instead; in all cases `__x__epcc` will
+have an address of 0 or 2^XLEN^/2 and a tag of zero.
+====
+
+==== Bounds checking
+
+Each byte address of a load, store, or instruction fetch is individually sign
+extended from XLEN bits to XLENMAX bits prior to bounds checking.
+
+NOTE: A load, store, or instruction fetch which spans 2^XLEN^/2 or wraps from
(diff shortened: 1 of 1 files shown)


src/riscv-legacy-integration.adoc
```diff
@@ -287,6 +287,72 @@ NOTE: Disabling CHERI register access prevents a low-privileged Legacy mode
...
+for all XLENMAX is less than 32.
+
+==== CSR width modulation
+
+Address CSRs have a width defined by the base ISA which may be variable,
+depending on the implementation's supported XLEN and mode combinations. The
+corresponding extended CSRs have a variable width address field. If the width
+changes, the address field is truncated to the new width and zero extended to
+XLENMAX bits. The CSR's tag will be cleared at the time of the width change if
+the new address is not representable.
```
jrtc27, line 339: or the CSR is sealed and the original and modulated addresses differ

arichardson (commented):
Thanks this looks good to me, will merge if @jrtc27 is happy.

src/riscv-legacy-integration.adoc
```diff
@@ -287,6 +287,73 @@ NOTE: Disabling CHERI register access prevents a low-privileged Legacy mode
...
+than XLENMAX and <<pcc>> does not hold an infinite-bounds capability has
+different effects depending on whether the addressing boundary coincides with
+an instruction boundary. A single instruction which spans the address boundary
+will be treated as a length violation and is reported as a CHERI exception with
+`__x__epcc` the starting address of the instruction. If an instruction ends
+exactly at the addressing boundary, the instruction will complete successfully
+but the <<pcc>> becomes unrepresentable afterward, and a tag violation CHERI
+exception is taken on the next instruction unless an interrupt or higher
+priority synchronous exception is taken instead; in all cases `__x__epcc` will
+have an address of 0 or 2^XLEN^/2 and a tag of zero.
```
jrtc27, line 319: It won't always be unrepresentable, e.g. if PCC's bounds are `[0, (2^XLEN)/2)` and you attempt to cross the middle of the address space?

```suggestion
but the <<pcc>> becomes out of bounds or unrepresentable afterward, and a length or tag violation CHERI
exception is taken on the next instruction unless an interrupt or higher
priority synchronous exception is taken instead; in all cases `__x__epcc` will
have an address of 0 or 2^XLEN^/2 and, in the unrepresentable case, a tag of zero.
```

Also that "all cases" doesn't apply to the "unless an interrupt or ..." case.

src/riscv-legacy-integration.adoc
```diff
@@ -287,6 +287,73 @@ NOTE: Disabling CHERI register access prevents a low-privileged Legacy mode
...
+====
+
+==== Bounds checking
+
+Each byte address of a load, store, or instruction fetch is individually sign
+extended from XLEN bits to XLENMAX bits prior to bounds checking.
```
jrtc27, line 325: Hm, the Supervisor spec says:

```
If UXLEN \(<\) SXLEN, user-mode instruction-fetch addresses and load and store effective addresses are taken modulo \(2^{\text{UXLEN}}\). For example, when UXLEN=32 and SXLEN=64, user-mode memory accesses reference the lowest 4 GiB of the address space.
```

Should we not do the same here? Otherwise this lets you access memory that is in reality outside the bounds of the capability in use. Though I note this is at odds with:

```
Similarly, pc bits above XLEN are ignored, and when the pc is written, it is sign-extended to fill the widest supported XLEN.
```

So an OS needs to know to mask those bits off in xEPC when figuring out what actual address was being fetched from (but not xTVAL). Which is particularly silly when switching to RV32 then back to RV64 in the more privileged mode discards said upper bits of xEPC. This sounds to me like an oversight in the spec and that PC should have been regarded like a CSR not a GPR (or at least in the cases where it is used for CSRs, as opposed to AUIPC and JALR where I can understand the desire to have it pre-sign-extended, but that's a hardware implementation optimisation and already specified by virtue of writes to GPRs sign-extending the value).

I wonder if there's appetite for making that saner in the base spec. Otherwise we probably need some ludicrous "PCC's address will be sign-extended, as visible when put into xEPCC, but gets masked when the bounds check is performed, so both need to be representable (though only the latter in bounds)" requirement.

PR #142, opened 2024-03-11
Misaligned tweaks

sorear:
The only functional change is the second commit which makes the relative priority of CHERI exceptions wrt misalignment consistent with access faults and page faults.

### DIFF
--- a/src/riscv-integration.adoc
+++ b/src/riscv-integration.adoc
@@ -779,12 +779,12 @@ Load/store/AMO address breakpoint
 *PCC <<asr_perm>> clear* +
 *Branch/jump target address checks (tag, execute permissions and bounds)*
 
+| .>|4,6 .<|Optionally: +
+Load/store/AMO address misaligned
 | .>|*4,6* +
 *{cheri_excep_mcause}* .<|*Prior to address translation for an explicit memory access:* +
 *Load/store/AMO capability address misaligned* +
 *CHERI fault due to capability checks (tag, permissions and bounds)*
-| .>|4,6 .<|Optionally: +
-Load/store/AMO address misaligned
 | .>|13, 15, 5, 7 .<|During address translation for an explicit memory access: +
 First encountered page fault or access fault
 | .>|5,7 .<|With physical address for an explicit memory access: +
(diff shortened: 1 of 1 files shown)


src/riscv-integration.adoc
```diff
@@ -1101,6 +1103,10 @@ NOTE: `auth_cap` is <<ddc>> for Legacy mode and `cs1` for Capability Mode
 
 ^1^ This check is architecturally required, but is impossible to encounter so may not required in an implementation.
 
+^2^ Accesses that are not naturally aligned can wrap around the address space
+from 2^XLEN^ to 0. A capability's bounds include both 0 and 2^XLEN^-1 if and
+only if they are the same as the bounds of <<infinite-cap>>.
```
andresag01, line 1108: I think there is an edge case when XLENMAX=64 but the effective XLEN=32 because you can have a capability (that is not Infinite) whose bounds cover the first (2^32)-1 and memory accesses wrap at 2^32 as required by RISC-V.

src/riscv-integration.adoc
```diff
@@ -774,15 +774,17 @@ Environment call +
...
+| .>| *{cheri_excep_mcause}* +
+*{cheri_excep_mcause}* .<| *CHERI faults due to:* +
 *PCC <<asr_perm>> clear* +
 *Branch/jump target address checks (tag, execute permissions and bounds)*
 
-| .>|*{cheri_excep_mcause}* .<|*Prior to address translation for an explicit memory access:* +
-*Load/store/AMO capability address misaligned* +
-*CHERI fault due to capability checks (tag, permissions and bounds)*
 | .>|4,6 .<|Optionally: +
 Load/store/AMO address misaligned
```
andresag01, line 783: The motivation for putting misaligned faults after CHERI is to avoid splitting standard RISC-V exception causes in 2 blocks i.e. we now have
* RISC-V exception (misaligned)
* CHERI exceptions
* Other RISC-V exceptions

This is not ideal for some hardware implementations where all the RISC-V load/store/amo exceptions (including misaligned) come from one hardware block while the CHERI exceptions come from another block. In the original case, this is not a problem because the two hardware blocks produce RISC-V and CHERI exceptions separately and then we have a 2-way priority (if !CHERI_exception then RISC-V_exception else CHERI_exception).

My preference is to revert this change.

andresag01 (changes requested):

src/riscv-integration.adoc
jrtc27: Alignment faults must be lower priority than CHERI faults. Otherwise you have to reimplement the entirety of the CHERI checks in your misaligned access emulation routine. Which people will 100% screw up.

andresag01: Thanks for the comment @jrtc27! I was unaware of those existing checks. Seems that's an even stronger reason than what I pointed out ifrom a hardware implementation perspective!

src/riscv-integration.adoc
tariqkurd-repo: I think the RISC-V spec is ambiguous here which Allen Baum tried to clear up on the mailing list: https://lists.riscv.org/g/tech-privileged/message/1116. The conclusion seems to be that setting out of range bits will cause an access fault,and so wrapping behaviour is not expected.

andresag01:
There has not been activity in this PR for some time, so I will close it. Please re-open, address the comments and rebase the PR if you'd like to pursue this changes although there are very good reasons for keeping the priority of exception causes as currently written in the CHERI RISC-V spec.

PR #150, opened 2024-03-21
C.SC and C.SCSP imm should be 64-bit aligned (RV32)

James-Williams:
C.SC and C.SCSP (RV32) store a 64-bit capability and should encode a 64-bit aligned offset.

C.LC (RV32) encodes `offset[7:3]`
C.LCSP (RV32) encodes `offset[8:3]`

C.SC (RV32) should likewise encode `offset[7:3]`
C.SCSP (RV32) should likewise encode `offset[8:3]`

This would match the RV64 RVC encodings which encode a 128-bit aligned offset for C.LC, CLCSP, C.SC, C.SCSP.


tariqkurd-repo:
Thaks for spotting this.

PR #155, opened 2024-03-25
Define a new 32-bit permissions format 

arichardson:
This is based on suggestions by @LawrenceEsswood and is chosen such that
it allows for future extensibility with recursively mutable permissions
as well as dataflow extensions. We also encode the mode bit more
efficiently now (only for executable capabilities), so we will be able
to drop the separate Zcheri_mode extension to simplify the ISA.

Co-authored-by: Lawrence Esswood <lesswood@google.com>


LawrenceEsswood:
Put this comment in the wrong place:

The idea behind having a separate reset permission and ASR+RWXC cap was that when we grew sealing/non-sealing the first would get those permissions and the second wouldn't.

If we decide that separate capability roots are OK, we would not need either of the reset ones. If we like the idea of a single root capability, we still only need one (not two) reset patterns. It should be mode legacy.

src/cap-description.adoc
```diff
@@ -93,36 +93,53 @@ Therefore, it is only possible to encode a subset of all combinations.
 |==============================================================================
```
andresag01, line 93: This table and the previous paragraph state that the RV32 encoding is 4 bits although it is 5 bits with this proposed encoding.

src/cap-description.adoc
```diff
@@ -93,36 +93,53 @@ Therefore, it is only possible to encode a subset of all combinations.
...
+| 0b00000   |   |   |   |   |     |      | No permissions
+| 0b00001   | ✔ |   |   |   |     |      | Data RO
+| 0b00010   | ✔ | ✔ | ✔ | ✔ |  ✔  |  Leg | Almighty (reset)
+| 0b00011   | ✔ | ✔ | ✔ | ✔ |  ✔  | Pure | Almighty (purecap)
+| 0b00100   |   | ✔ |   |   |     |      | Data WO
+| 0b00101   7+| reserved
+| 0b00110   | ✔ | ✔ |   |   |     |      | Data RW
+| 0b00111   7+| reserved
+| 0b01000   | ✔ | ✔ | ✔ | ✔ |  ✔  |  Leg | Execute + ASR
+| 0b01001   | ✔ | ✔ | ✔ | ✔ |  ✔  | Pure | Execute + ASR
```
andresag01, line 108: Is this correct? These 2 entries are described as "execute + ASR" but the other columns are the same as the almighty cap. Shouldn't this just have X and ASR being granted?

src/cap-description.adoc
```diff
@@ -93,36 +93,53 @@ Therefore, it is only possible to encode a subset of all combinations.
...
+| 0b10010   7+| reserved
+| 0b10011   7+| reserved
+| 0b10100   7+| reserved
+| 0b10101   7+| reserved
+| 0b10110   7+| reserved
+| 0b10111   7+| reserved
+| 0b11000   7+| reserved
+| 0b11001   7+| reserved
+| 0b11010   7+| reserved
+| 0b11011   | ✔ |   | ✔ |   |     |      |
```
andresag01, line 126: The description for this one should be "Cap RO"

src/cap-description.adoc
```diff
@@ -93,36 +93,53 @@ Therefore, it is only possible to encode a subset of all combinations.
...
+| 0b11100   7+| reserved
+| 0b11101   7+| reserved
+| 0b11110   7+| reserved
+| 0b11111   | ✔ | ✔ | ✔ |   |     |      |
```
andresag01, line 130: The description for this one should be "Data + Cap RW

src/cap-description.adoc
```diff
@@ -93,36 +93,53 @@ Therefore, it is only possible to encode a subset of all combinations.
...
+| 0b00110   | ✔ | ✔ |   |   |     |      | Data RW
+| 0b00111   7+| reserved
+| 0b01000   | ✔ | ✔ | ✔ | ✔ |  ✔  |  Leg | Execute + ASR
+| 0b01001   | ✔ | ✔ | ✔ | ✔ |  ✔  | Pure | Execute + ASR
+| 0b01010   | ✔ |   | ✔ | ✔ |     |  Leg | Execute + Cap RO
+| 0b01011   | ✔ |   | ✔ | ✔ |     | Pure | Execute + Cap RO
+| 0b01100   | ✔ | ✔ | ✔ | ✔ |     |  Leg | Execute + Cap RW
+| 0b01101   | ✔ | ✔ | ✔ | ✔ |     | Pure | Execute + Cap RW
+| 0b01110   | ✔ | ✔ |   | ✔ |     |  Leg | Execute + Data RW
+| 0b01111   | ✔ | ✔ |   | ✔ |     | Pure | Execute + Data RW
```
andresag01, line 114: Is "Execute + Data RO" something that would ever be needed?

andresag01 (changes requested):
I believe I understand the general idea of the change. I noted a few questions about the allocated encodings -- it looks like the scheme is rather light on ASR encodings although its hard to tell which ones will be needed.

src/cap-description.adoc
arichardson: When discussing this with @LawrenceEsswood we decided that `Execute + Cap RO` should cover that case and it's not worth burning another encoding for it unless there is a compelling use-case.

arichardson:
Latest versions drops the separate almighty capability encoding. This was intended to allow a single root once we have sealing
capabilities. However, it makes the encoding a bit awkward and we don't
have the same constraint with future sealing caps that they need to be
derived from a single root as we do for RW+RX capabilities. Those two
roots can still be subset from the Almighty ASR capability.

src/cap-description.adoc
arichardson: Duplicate removed in new version.

src/cap-description.adoc
```diff
@@ -92,37 +92,56 @@ Therefore, it is only possible to encode a subset of all combinations.
...
 permissions are granted as follows:
 
-* <<c_perm>>: bit 0 is set
-* <<w_perm>>: bit 1 is set
-* <<r_perm>>: bits 3 or 2 are set
-* <<x_perm>>: bit 3 is set
-* <<asr_perm>>: bits 3 and 2 are set
+* <<c_perm>>: `P[4] OR (P[3] AND P[1:0] != 0b11)`
+* <<w_perm>>: `P[2] OR P[4:1] == 0b0100`
+* <<r_perm>>: `P[4:0] != 0b00000 AND P[4:0] != 0b00100`
```
arichardson, line 142: Could also write this as `P[4:3] != 0b00 OR P[1:0] != 0b00`

andresag01:
A couple of questions:
1. We have "Execute + Cap RO", isn't a "Execute + Data RO" also useful?
2. In the proposed scheme, ASRs are only granted if you have "almighty permissions" (i.e. X, C, R, W). Aren't there any scenarios where we would like to grant ASR (and clearly X too) although only, say Data RW or Data RO?

tariqkurd-repo (changes requested):
I've made non-normative changes to fix some errors and make the table easier to read.
Splitting the table into quadrants has made it easier for me to see what's going on.
I think it makes sense to:

1. keep quadrant 0 as it is
2. move quadrant 3 (cap load/store) into quadrant 1, and share the same encodings for P[2:0] for the populated cases
3. move quadrant 1 (executable) maybe into quadrant 3
4. leave quadrant 2 reserved

that way quadrant 2 could be seen as an extensions of 0,1 or 3 later on as more encodings are added.

I think point 2. is important - the RO/RW/WO encodings shoudl be the same for each quadrant - with just the quadrant number determining whether cap permission is present.

I'm happy to make these changes if you agree

arichardson:
> I've made non-normative changes to fix some errors and make the table easier to read. Splitting the table into quadrants has made it easier for me to see what's going on. I think it makes sense to:
> 
> 1. keep quadrant 0 as it is
> 2. move quadrant 3 (cap load/store) into quadrant 1, and share the same encodings for P[2:0] for the populated cases
> 3. move quadrant 1 (executable) maybe into quadrant 3
> 4. leave quadrant 2 reserved
> 
> that way quadrant 2 could be seen as an extensions of 0,1 or 3 later on as more encodings are added.
> 
> I think point 2. is important - the RO/RW/WO encodings should be the same for each quadrant - with just the quadrant number determining whether cap permission is present.
> 
> I'm happy to make these changes if you agree
> 
> I've also added a proposal to the ACPERM page about clearing permissions resulting in illegal combinations. I think it makes sense that in general targeting an impossible permission set means all permissions are cleared, but, clearing X also requires clearning ASR and may then result in a useful combination. Although maybe in this case the software should also clear ASR anyway, and then if the software doesn't do so then you get zero permissions.
> 
> Thoughts?

- As noted in the SIG/TG call, capability memory is in quadrant 3 since it will also occupy quadrant 2 once Transitive and Local/Global extensions are defined.
- I've added a new table explaining the allocation of permissions more clearly, since we don't have WO cap memory we can't quite have them line up. In quadrant 0, I allocated bit 2 for write (to match quadrant 2+3) and bit 0 for read (no equivalent in quadrant 2+3 since read is implicitly included). I could use bit 1 for read instead if you prefer?
- Regarding ACPERM, I don't think we should fall back to zero but instead the closest possible encoding that is a subset of the requested perms since otherwise stripping X from the infinite-cap would result in a zero perms capability but what we really want is cap memory with all permissions.

src/cap-description.adoc
```diff
@@ -92,37 +92,77 @@ Therefore, it is only possible to encode a subset of all combinations.
...
-* <<r_perm>>: bits 3 or 2 are set
-* <<x_perm>>: bit 3 is set
-* <<asr_perm>>: bits 3 and 2 are set
+The mapping of bit values to decoded permissons in <<cap_perms_encoding32>> is based on the following encoding strategy:
+[#cap_perms_encoding_description32,width="100%",options=header,cols="^2,^1,^1,^1,6",align="center"]
+|==============================================================================
+| Bit 4&3 | Bit 2    | Bit 1      |  Bit 0     | Notes
+5+<|_Data & capability memory access (Quandrant 2+3)_
+|   Q=0b11 | Write 2+| Reserved=0b11          <| Reserved bits for future extensions must be 1
+|   Q=0b01 3+| Reserved=0b111                 <| Reserved bits for future extensions must be 1
```
andresag01, line 156: Is this correct? The subtitle says quadrants 2+3 but Q=0b11 and Q=0b01 (i.e. 3 and 1)

andresag01 (approved):
Happy with the change. I would remove a large commented out block from the MR though.

arichardson:
> Happy with the change. I would remove a large commented out block from the MR though.

Agreed will remove this commented table now that the contents is included in the main table. Will do one more proof read as well and then remove the Draft: label.

src/cap-description.adoc
```diff
@@ -85,44 +85,81 @@ superfluous when the capability does not grant either <<r_perm>> or <<w_perm>>.
...
-| 0b1000   | ✔ |   |   | ✔ |
-| 0b1001   | ✔ |   | ✔ | ✔ |
-| 0b1010   | ✔ | ✔ |   | ✔ |
-| 0b1011   | ✔ | ✔ | ✔ | ✔ |
-| 0b1100   | ✔ |   |   | ✔ | ✔
-| 0b1101   | ✔ |   | ✔ | ✔ | ✔
-| 0b1110   | ✔ | ✔ |   | ✔ | ✔
-| 0b1111   | ✔ | ✔ | ✔ | ✔ | ✔
+| Encoding[2:0] | R | W | C | X | ASR | Mode | Notes
+8+| *Quadrant 0: Non-capability data read/write*
```
LawrenceEsswood, line 108: Nit: As we add non-memory permissions to this quad this description will become inaccurate.

tariqkurd-repo (approved):

src/cap-description.adoc
```diff
@@ -85,44 +85,81 @@ superfluous when the capability does not grant either <<r_perm>> or <<w_perm>>.
...
 .Permissions widths depending on XLENMAX
-[#perms_bit_width,options=header,align="center",width="55%"]
+[#perms_bit_width,options=header,align="center",width="75%",cols="^1,^3,5"]
 |==============================================================================
-^| XLENMAX ^| Permissions width
-^| 32      ^| {cap_rv32_perms_width}
-^| 64      ^| {cap_rv64_perms_width}
+^| XLENMAX ^| Permissions width      ^| Comment
+^| 32      ^| {cap_rv32_perms_width} ^| Encodes some combinations of 5 permission bits and the Mode bits
+^| 64      ^| {cap_rv64_perms_width} ^| Separate bits for each permission, and the Mode bit is not included
```
LawrenceEsswood, line 92: I would like to align RV64 and RV32 so that, at least for the base ISA, they have the same format.

This would mean that as people start writing software for CHERI it can be transportable between the two. Even for RV64, saving bits is also a great idea. There are plenty of good ideas floating about on how to use extra bits: MTE-like colours, N-bit local/global etc. It seems a shame to waste them just because we were never forced to think about it until RV32.

It will hopefully also simplify the spec somewhat to have fewer differences.

src/riscv-mode-integration.adoc
```diff
@@ -29,9 +29,8 @@ include::img/cap-encoding-xlen32-mode.edn[]
 [#cap_encoding_xlen64_mode]
 include::img/cap-encoding-xlen64-mode.edn[]
 
-{cheri_mode_ext_name} allows the M bit to be set to 1 when the capability does
-not grant <<x_perm>>. In this case, the M bit is superfluous, so the encoding
-may be used to support additional features in future extensions.
+For RV64, {cheri_mode_ext_name} allows the M bit to be set to 1 when the capability does not grant <<x_perm>>.
+In this case, the M bit is superfluous, so the encoding may be used to support additional features in future extensions.
```
LawrenceEsswood, line 33: If we unified permissions between RV32 and RV64 these peculiarities could also go away.

tariqkurd-repo: this is true - but I worry that we will be unnecessarily limiting the options for RV64 where there is ample encoding space, so we shoudl allow more flexibility.
We just need to define the quirkly cases.

tariqkurd-repo: https://github.com/riscv/riscv-cheri/issues/164

LawrenceEsswood: I think if there were such motivating use cases, they should be equally applicable to RV32 and we could hopefully argue for their existence now.

I can of course see us missing something. But there is also nothing stopping us from adding in extensions to add permission combinations back to RV64 (or in RV32, if not too many) if we discover they were needed. It is worse to take them back.

But I still think the default position should be to save the bits, and to align base specs for software trans-portability.

arichardson: We discussed that briefly in yesterdays SIG/TG meeting and I don't think we need to constrain RV64 (yet). For RV32 we have no choice, but for RV64 we might as well allow these permission combinations since adding them back later is a lot more difficult than defining a new restricted encoding once we add new features that want to make use of those bits.

carlrobshaw: I agree with @LawrenceEsswood here. We now have a different behavior for acperm depending on XLENMAX, from an instruction that was meant to abstract the permissions handling from the permissions encoding. We could be in the position where we will have to write different permissions code for 32 and 64 bit CHERI (and accidentally using that code written for 64 bit on a 32 bit system could result in unexpected behavior), or else write trans-portable code that will restrict the permission combinations to those of the 32 bit degenerative case anyway.

Do we really want to have diverging base CHERI permission behavior between 32 bit and 64 bit at this stage?

jrtc27: Do we want it? No. But it's inevitable given the constraints. RV64 should not be restricted in its combinations, but RV32 has to be due to encoding space limitations.

src/cap-description.adoc
arichardson: True, we will need to update it once that happens.

src/cap-description.adoc
arichardson: Conceptually I agree that having them the same would be ideal, but I'd argue we want RV32 to be like RV64 and allow all combinations. Since that is not feasible due to the limited number of bits available I believe we should just accept that they will be different.

Issue #165, opened 2024-04-02
remove JALR.MODE


tariqkurd-repo:
now that we have MODESW allowing easy mode switches, we can remove JALR.MODE.
As discussed in the TG meeting on 2nd April.

PR #161, opened 2024-04-01
Merge Zcheri_mode and Zcheri_legacy

arichardson:
The main reason for a separate Zcheri_mode extensions is that it freed
up a bit from the 32-bit capability format. With a new 32-bit
permissions encoding (https://github.com/riscv/riscv-cheri/pull/155), we
need less than a full bit for the mode since it is only required for
executable capabilities rather than fully orthogonal. Removing the
separate Zcheri_mode extension simplifies the architecture and will
allow for further simplification of the CME architectural state.

Co-authored-by: Tariq Kurd <tariq.kurd@codasip.com>
Co-authored-by: Andres Amaya Garcia <andres.amaya@codasip.com>

PR #167, opened 2024-04-02
Remove JALR.MODE

arichardson:
This is split out from #130, I only made a minor change to the pull request description:

In Capability mode this can be expressed by a three-instruction sequence of auipc, scaddr, and jalr. While this is not precisely the same (the link register is a capability not an address) it behaves almost exactly the same.
Purecap ABIs are expected to use capabilities, possibly sealed, as function pointers, and relative vtables and PIC switch tables do not benefit from jalr.mode over an instruction sequence using cadd only.

In Legacy mode with Zcheri_mode, switch to Capability mode before adjusting pcc metadata and switch back afterward. If the legacy code could have been exited without a trap, capability-aware code within the same pcc region must already have existed, otherwise the trap return can be used for the mode switch.

It is no longer possible without Zcheri_mode to transfer control to a sentry in Legacy mode.
However, #161 merges Zcheri_mode into Zcheri_legacy, so this is no longer a concern.


andresag01:
Fixes #165

andresag01 (approved):

tariqkurd-repo (commented):
LGTM

Issue #156, opened 2024-03-25
Sdtrig exception priorities


tomaird:
The `Sdtrig` specification (RISC-V Debug Spec) describes a list of synchronous exception priorities. 

The CHERI spec makes no reference to how the new CHERI exceptions fit into this priority list.

e.g. We have a `mcontrol6` trigger configured to match on load addresses, with timing "before", and `action` set to raise a breakpoint exception. Then we do a load with an `auth_cap` without load permissions (or no tag etc.). Does this fire the trigger and raise a breakpoint exception, or the CHERI exception?

I'm not sure if there are any other interactions between CHERI and Sdtrig that need considering too.

tariqkurd-repo:
Breakpoints must always be higher priority otherwise debug wouldn't be possible.
This should be in the doc...... that was the intention.....

andresag01:
@tomaird is right in that we have the exception priorities but these do not include the Debug exception priorities. We just need to add the equivalent table in the debug integration chapter of the CHERI spec -- Its essentially the same table we already have with a few more rows added in

PR #171, opened 2024-04-03
Add table with exception priority with triggers

andresag01:
Fixes https://github.com/riscv/riscv-cheri/issues/156

This change adds a new section under Zcheri_purecap with a table indicating where the trigger exceptions fit alongside CHERI and other exceptions. The table added here is mostly a merge between the regular exception priority table in 3.7.9 (mcause description) and the trigger exception priority table 5.2 from the RISC-V debug spec.

### DIFF
--- a/src/riscv-cheri.adoc
+++ b/src/riscv-cheri.adoc
@@ -44,6 +44,8 @@ include::riscv-integration.adoc[]
 
 include::debug-integration.adoc[]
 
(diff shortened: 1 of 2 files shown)


tariqkurd-repo (approved):
I think this table is useful but strictly redundant. Let's keep it for now though, the ARC will remove it I'm sure.

PR #172, opened 2024-04-03
Fix broken variable resolution in title

andresag01:
Adoc has trouble resolving variables in titles. This change fixes that in one instance which was throwing a build warning

### DIFF
--- a/src/tables.adoc
+++ b/src/tables.adoc
@@ -1,6 +1,6 @@
 == Extension summary
 
-=== {lr_sc_bh_ext_name}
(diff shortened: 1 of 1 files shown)


tariqkurd-repo:
the build has other warnings doens't it? It's better to keep the variable IMHO

andresag01:
> the build has other warnings doesn't it? It's better to keep the variable IMHO

It has another warning for a different reason. If we do not change this then the title will simply print the variable name `{lr_sc_bh_ext_name}`

Issue #173, opened 2024-04-03
Precisely define RV32 ACPERM state transitions


arichardson:
For RV32, some ACPERM operations result in a non-encodable permissions set.
We should include a table (or a state transition diagram like CherIoT) to precisely define which result you get.

Issue #164, opened 2024-04-02
behaviour of RV64 M bit with X=0


tariqkurd-repo:
Currently we have this text:

>For RV64, the M bit can be set to 1 when the capability does not grant <<x_perm>>.
In this case, the M bit is superfluous, so the encoding may be used to support additional features in future extensions

I wonder if we should define that if X is zero, then M must also be zero for the tag to be set.
We could specify that ACPERM will always clear M if it is clearing X. The same is true for ASR - it is superfluous / don't care if X=0.

I think that we shouldn't allow a tagged cap with X=0 and either M=1 or ASR=1, to cleanly reserve these cases for future expansion, and so we should update the ACPERM specification to reflect this.

arichardson:
As you mentioned, I think we can just ignore it in the same way as other "dependent permission bits" for RV64 (ASR without X). The argument for clearing the bit is that it would make RV64 behaviour a bit more similar to RV32 but I don't really know whether that is worthwhile.

Maybe we should re-encode permissions for RV64 as well, but I think keeping them orthogonal is fine since there is little need to save as many bits as there is for RV32.

PR #175, opened 2024-04-04
Permission transitions

tariqkurd-repo:
fix https://github.com/riscv/riscv-cheri/issues/173
fix https://github.com/riscv/riscv-cheri/issues/164

### DIFF
--- a/src/insns/acperm_32bit.adoc
+++ b/src/insns/acperm_32bit.adoc
@@ -32,12 +32,31 @@ output capability has its tag set to 0 if `cs1` is sealed.
 +
 The AP field is not able to encode all combinations of permissions when MXLEN=32, therefore removing a permission may yield a capability which cannot be encoded.
 Valid combinations are shown in <<cap_perms_encoding32>>.
-If removing permissions would result in architectural permissions which cannot be encoded, additional architectural permissions will be removed until an encodable subset of the requested permissions has been found.
-For example, with executable capabilities, where <<x_perm>> is set, clearing <<x_perm>> also clears <<asr_perm>> and the Mode becomes irrelevant.
-As a result,the capability is converted from executable to data.
-Similarly, if <<c_perm>>, <<r_perm>> and <<w_perm>> are currently set, attempting to remove <<r_perm>> results will result in a capability with only <<w_perm>> permissions.
++
+Additionally some combinations of permissions are not useful when XLENMAX=64.
+These cases are defined to return useful minimal sets of permissions in each case, which may be no permissions.
++
+NOTE: Future extensions may allow more combinations of permissions, especially for XLENMAX=64.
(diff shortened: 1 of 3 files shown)


arichardson (changes requested):
Thanks, I think this addresses almost all transitions, but I believe we have not defined what happens when you try to remove C from `Execute + Data & Cap RO`. Since that is not representable we also end up removing X. Similar for inifinite-cap, removing C also removes ASR.

arichardson:
I also believe we can't remove R from capabilities with C and have to also drop C.

src/insns/acperm_32bit.adoc
```diff
@@ -32,12 +32,35 @@ output capability has its tag set to 0 if `cs1` is sealed.
...
+. <<x_perm>> cannot be set without <<r_perm>> being set.
+.. Therefore removing <<r_perm>> also requires removing <<x_perm>>.
+. If removing <<c_perm>> with <<x_perm>> set cannot be represented, then also remove <<x_perm>>.
+.. This effects the _Execute + Data & Cap RO_ case from <<cap_perms_encoding32>>.
+
+NOTE: Failing to follow the rules above will clear *all* permissions.
+
+The XLENMAX=64 additional rule is:
+
+. Clearing <<x_perm>> also clears the <<m_bit>>
```
arichardson, line 61: This is also true for RV32, we just can't "directly read the M bit"

arichardson (approved):
Thanks this looks good to me now, but I'd like @LawrenceEsswood to have a look as well before merging since the format is based on his proposal.

LawrenceEsswood:
I don't believe the list is exhaustive for RV32. It seems to at least be missing what happens when you try make a cap with ASR but missing any other permission.

Also, while defining all unrepresentable permissions sets to transition to no permissions does address https://github.com/riscv/riscv-cheri/issues/173, I was hoping that we would take a less surprising approach and instead transition to the nearest useful cap. 

cap-description basically states what I was hoping. I think it can be described succinctly for RV32 as (applying these rules in order):

If you do not have ALL permissions, clear ASR
If you do not have R, clear C and X
If you have X and R, but not C, clear X

src/insns/acperm_32bit.adoc
tariqkurd-repo: fixed

tariqkurd-repo:
please have another look @LawrenceEsswood

andresag01:
@tariqkurd-repo: I replaced instances of XLENMAX with MXLEN and rebased on top of main because the branch was behind.

andresag01 (approved):

LawrenceEsswood:
> please have another look @LawrenceEsswood

Too late(!), but yes, LGTM

Issue #180, opened 2024-04-04
SH4ADD is RV64 only


andresag01:
The instruction listing for SH4ADD needs to indicate that the instruction is RV64 only and not RV32.

PR #181, opened 2024-04-05
make SH4ADD/SH4ADD.UW RV64 only

tariqkurd-repo:
Fix https://github.com/riscv/riscv-cheri/issues/180


src/insns/sh4add_32bit.adoc
```diff
@@ -36,10 +36,10 @@ Legacy Mode Description::
 Increment `rs2` by `rs1` shifted left by 4 bit positions.
 
 Prerequisites for Capability Mode::
-{cheri_base_ext_name}, Zba
+RV64, cheri_base_ext_name}, Zba
```
andresag01, line 39: There is a typo here, so the variable will not expand

PRugg-Cap:
Oops, race with https://github.com/riscv/riscv-cheri/pull/182. We both have some other fixes though. Note that "RV64" isn't listed as a prerequisite for some other RV64-only instructions, so I didn't think it went there, but happy to fix the others.

andresag01 (approved):

src/insns/sh4add_32bit.adoc
andresag01: Fixed

Issue #149, opened 2024-03-21
Top needs to be 66 bits


PRugg-Cap:
(Using RV64 for clarity, but obviously 34 bits in RV32 etc.)

There's currently a case where base > top, reminding me that Cambridge's "underivable" branch needed to move to 66-bit arithmetic for top: https://github.com/CTSRD-CHERI/sail-cheri-riscv/tree/underivable.

This is to address the fairly specific issue where a cap has:
 - E = max_E (52), meaning the top two bits of B and the (reconstituted) top two bits of T "poke out" over the top of the 64-bit address space. (The only tagged cap that can exist with E=52 has B=T=0 and is the almighty cap).
 - B is zero in the top two bits (otherwise it would be malformed) and non-zero in the remaining bits.
 - T is less than B, making the top two bits of T: 00 + 1 (implied MSB) + 1 (T < B), and setting bit 65.

If the 66-bit arithmetic is too painful for hardware, we may also be able to fix this in the malformed check, but that would be the first case (I think) where you can become malformed without twiddling the bits directly using `CSetHigh` or round-tripping through memory. The above capability can be constructed (untagged) simply by setting the address of a cap to 2^64-1 and requesting length 2^64-1. Changing it in the malformed check would also make the malformed check dependent on the full mantissa, whereas up to now it only depended on the exponent and top two bits of B.

`CBuildCap` currently has a check that `requestedBase <= requestedTop` (https://github.com/CTSRD-CHERI/sail-cheri-riscv/blob/6e3613a2c46fb809e526b55c5c72acb041194ab8/src/cheri_insts.sail#L803), which should be able to go away with this fix. We'd want to prove this first though!

tariqkurd-repo:
>and requesting length -1

is this possible? the `scbnds` instruction page doesn't say anything about the sign of `rs2` - shouldn't it be an unsigned operand?

tariqkurd-repo:
Can we special case E=52 to avoid all this? Are there's only one valid cap with that exponent?

PRugg-Cap:
By "special case" I guess you mean "make malformed"? The malformed check would depend on the the mantissa bits then, but I guess only on whether they're the specific allowed value rather than on a subtract.

PRugg-Cap:
It still means you can go malformed fairly easily with common instructions, but I doubt that's a real problem.

tariqkurd-repo:
>The only tagged cap that can exist with E=52 has B=T=0 and is the almighty cap

[...]

PR #184, opened 2024-04-05
Fix malformed check to avoid 65-bit top overflow

PRugg-Cap:
Hopefully addresses https://github.com/riscv/riscv-cheri/issues/149, but I'd like to do some more checking before this gets merged!

### DIFF
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -594,8 +594,8 @@ capability because its bounds cannot be correctly decoded. The following check
 indicates whether a capability is malformed.
 
 ```
-malformedMSB =  (E == CAP_MAX_E     && B != 0)
-             || (E == CAP_MAX_E - 1 && B[MW - 1]        != 0)
+malformedMSB =  (E == CAP_MAX_E     && B         != 0)
+             || (E == CAP_MAX_E - 1 && B[MW - 1] != 0)
 malformedLSB =  (E  < 0)
 malformed    =  !EF && (malformedMSB || malformedLSB)
 ```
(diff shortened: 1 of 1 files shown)


tariqkurd-repo (approved):
nice and simple

arichardson:
Does this also mean we can no longer observe BASE > TOP for all tagged capabilities?

PRugg-Cap:
> Does this also mean we can no longer observe BASE > TOP for all tagged capabilities?

Depends what you mean... This means that in 65-bits, BASE <= TOP, tagged or untagged. For tagged capabilities, that was (hopefully!) already true, even in a 64-bit space, with the exception of 2^64 itself. Note though that there isn't a CGetTop (GCTop?), so software would have to do base + length itself, with length saturating at uint_max.

arichardson:
> Depends what you mean... This means that in 65-bits, BASE <= TOP, tagged or untagged. For tagged capabilities, that was (hopefully!) already true, even in a 64-bit space, with the exception of 2^64 itself. Note though that there isn't a CGetTop (GCTop?), so software would have to do base + length itself, with length saturating at uint_max.

Yes I mean in 65-bit space. I vaguely remember seeing cases in the cheri-compressed-cap library where use of the high bits would infer a top < base due to wrapping, so we couldn't assert that the decompressed top > decompressed base in all cases.

PRugg-Cap:
@arichardson Yes, that's exactly the problem this is to fix.

PRugg-Cap:
As with https://github.com/riscv/riscv-cheri/pull/116#issuecomment-2047932173 I've proved that the decoded bounds from encoded capabilities always have `base <= top`. It finds the problem before this fix and proves the property after this fix. Same caveats apply, including that I've had to change the SMT backend.

This should be good to merge.

PRugg-Cap:
... Now with better whitespace.

PR #187, opened 2024-04-08
Added thread identifier (TID)

### LINKED ISSUE #179: Add Secure Thread ID CSR for compartmentalisation
A placeholder that we need to add compartmentalisation support as soon as possible, probably by adding a Secure Thread ID CSR.

@francislaus

francislaus:
Identifying threads is essential for CHERI compartmentalisation.

This will close #179


src/riscv-integration.adoc
```diff
@@ -973,11 +976,29 @@ is the cause of the fault. The possible CHERI types and causes are encoded as
...
+
 === Unprivileged CSRs
 
 Unlike machine and supervisor level CSRs, {cheri_base_ext_name} does not require
 <<pcc>> to grant <<asr_perm>> to access unprivileged CSRs.
 
+[#utid,reftext="utid"]
+==== User Thread Identifier (utid)
+
+The <<utid>> register is an SXLEN-bit read-write register. It is a read-only
```
jrtc27, line 996: Needs to be UXLEN

src/riscv-integration.adoc
```diff
@@ -1102,3 +1123,17 @@ written to these CSRs unless the specification explicitly states that the CSR
...
+Compartmentalization seeks to separate the privileges between different protection units, e.g., two or more libraries.
+Code can be separated by sentries, which allow for giving out code capabilities to untrusted code where the untrusted code can only call code capability, but not modify it.
+Sentries can be called from different threads and thus there needs to be a way of identifying the current thread.
+While identifying the current thread can be done by privileged code, e.g., the kernel, the implied performance overhead of this is not bearable for CHERI systems with many compartments.
+
+The RISC-V ABI knows a _thread pointer (tp)_ register, which is not usuable for the purpose of reliably identifying the current thread because the tp register is a general purpose register and can be changed arbitrarily by untrusted code.
+Therefore, this specification offers two additional CSRs that facilitate a trusted source for the thread ID.
+The supervisor thread identifier (STID) register is readable and writeable from the supervisor mode allowing to change the thread ID on a context switch.
+The user thread identifier (UTID) exposes the current value of STID as a read-only copy.
```
jrtc27, line 1139: This only helps user-level compartmentalisation

src/riscv-integration.adoc
```diff
@@ -1102,3 +1123,17 @@ written to these CSRs unless the specification explicitly states that the CSR
...
+[#CHERI_COMP,reftext="CHERI Compartmentalization]
+=== CHERI Compartmentalization
+
+This section describes how this specification enables support for compartmentalization for CHERI systems.
+Compartmentalization seeks to separate the privileges between different protection units, e.g., two or more libraries.
+Code can be separated by sentries, which allow for giving out code capabilities to untrusted code where the untrusted code can only call code capability, but not modify it.
+Sentries can be called from different threads and thus there needs to be a way of identifying the current thread.
+While identifying the current thread can be done by privileged code, e.g., the kernel, the implied performance overhead of this is not bearable for CHERI systems with many compartments.
+
+The RISC-V ABI knows a _thread pointer (tp)_ register, which is not usuable for the purpose of reliably identifying the current thread because the tp register is a general purpose register and can be changed arbitrarily by untrusted code.
```
tariqkurd-repo, line 1136: `knows` should maybe be `includes`.

tariqkurd-repo:
I've added them into my riscv-opcodes PR: https://github.com/riscv/riscv-opcodes/pull/237

0xC80, "utid"
0x580, "stid"

and I suggest they are both non-reset

francislaus:
> I've added them into my riscv-opcodes PR: [riscv/riscv-opcodes#237](https://github.com/riscv/riscv-opcodes/pull/237)
> 
> 0xC80, "utid" 0x580, "stid"
> 
> and I suggest they are both non-reset

The allocation looks good to me. I am not sure what you mean by "non-reset". Do you mean: undefined on reset as `sscratchc` for example?

src/riscv-integration.adoc
```diff
@@ -973,11 +976,29 @@ is the cause of the fault. The possible CHERI types and causes are encoded as
 shown in xref:mtval-cheri-type[xrefstyle=short] and
 xref:mtval-cheri-causes[xrefstyle=short] respectively.
 
+[#stid,reftext="stid"]
+==== Supervisor Thread Identifier (stid)
+
+The <<stid>> register is an SXLEN-bit read-write register. It is used to
```
arichardson, line 982: Does it need ASR to be written?

src/riscv-integration.adoc
```diff
@@ -973,11 +976,29 @@ is the cause of the fault. The possible CHERI types and causes are encoded as
...
+
 === Unprivileged CSRs
 
 Unlike machine and supervisor level CSRs, {cheri_base_ext_name} does not require
 <<pcc>> to grant <<asr_perm>> to access unprivileged CSRs.
 
+[#utid,reftext="utid"]
+==== User Thread Identifier (utid)
+
+The <<utid>> register is an UXLEN-bit read-write register. It is a read-only
```
arichardson, line 996:
```suggestion
The <<utid>> register is an UXLEN-bit read-only register. It is a read-only
```
Do we really want these to be the same or have a separate user one that can only be written by S-mode?

arichardson (commented):
Would it make sense to have these be CLEN registers for added flexibility? Would allow S-mode to write a sealed cap into the register that can be unsealed by trusted user-mode code?

You can still use them as integer thread IDs, but now it's just a SW choice rather than a HW restriction.

src/riscv-integration.adoc
francislaus: That would indicate that we commonly expect that there is not a 1-1 mapping of user threads to kernel threads. Once we have user space threading, the user space thread scheduler can provide a solution, e.g., giving out a read-only capability that contains the current user space thread ID.

francislaus:
> Would it make sense to have these be CLEN registers for added flexibility? Would allow S-mode to write a sealed cap into the register that can be unsealed by trusted user-mode code?
> 
> You can still use them as integer thread IDs, but now it's just a SW choice rather than a HW restriction.

This is a potential optimisation discussed in a bigger PR to the cheri-specification repo (https://github.com/CTSRD-CHERI/cheri-specification/pull/123). Once you enter multiple levels of user-space compartmentalisation, the thread ID capability register would need to be checked every time whether it holds the correct capability. If not, the current compartmentalisation level needs to retrieve the correct capability, which would then work via the original TID, which you then do not have. Thus, you would need both the capability and the integer thread ID.

tariqkurd-repo:
>The allocation looks good to me. I am not sure what you mean by "non-reset". Do you mean: undefined on reset as sscratchc for example? @tariqkurd-repo

yes - undefined on reset

src/riscv-integration.adoc
```diff
@@ -973,11 +976,30 @@ is the cause of the fault. The possible CHERI types and causes are encoded as
...
 
+[#stid,reftext="stid"]
+==== Supervisor Thread Identifier (stid)
+
+The <<stid>> register is an SXLEN-bit read-write register. It is used to
+identify the current thread. Supervisor mode code can write to it if it
+has <<asr_perm>>.
+
+.Supervisor thread identifier register
+include::img/stidreg.edn[]
```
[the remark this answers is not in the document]
carlrobshaw: It is, but I don't think we have enough information yet to determine what a smaller, useful bit width could be - and I don't think we want to have a situation where the same software stack isn't portable across a similar class of cores. In the case of the ASID, it's an optimization (cores can have 0 length ASID) but for TID it's fundamental to operation. I'd just stick with SXLEN for now.

src/riscv-integration.adoc
carlrobshaw: "is not usuable" should be "is not usable"

src/riscv-integration.adoc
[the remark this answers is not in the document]
francislaus: I agree with Carl. We would need a differently sized register for each class of cores  as I assume that the number of currently executed threads differs substantially between classes of cores.

src/riscv-integration.adoc
```diff
@@ -1102,3 +1124,17 @@ written to these CSRs unless the specification explicitly states that the CSR
...
+Code can be separated by sentries, which allow for giving out code capabilities to untrusted code where the untrusted code can only call code capability, but not modify it.
+Sentries can be called from different threads and thus there needs to be a way of identifying the current thread.
+While identifying the current thread can be done by privileged code, e.g., the kernel, the implied performance overhead of this is not bearable for CHERI systems with many compartments.
+
+The RISC-V ABI includes a _thread pointer (tp)_ register, which is not usuable for the purpose of reliably identifying the current thread because the tp register is a general purpose register and can be changed arbitrarily by untrusted code.
+Therefore, this specification offers two additional CSRs that facilitate a trusted source for the thread ID.
+The supervisor thread identifier (STID) register is readable and writeable with <<asr_perm>> from the supervisor mode allowing to change the thread ID on a context switch.
+The user thread identifier (UTID) exposes the current value of STID as a read-only copy.
```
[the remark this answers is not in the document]
francislaus: @andresag01:

A compartment transition stays within a software thread. Usually, there is a trampoline in between a compartment transition in order to facilitate building up a data capability for the called compartment. A call from "CompA" to "CompB" looks as follows:

CompA -> Trampoline -> CompB

After "CompB" has finished the task it will return to "CompA", which also goes through trampoline:

CompB -> Trampoline -> CompA

All transitions are using sentries. The trampoline needs to have access to a trusted stack in order to find its own information as well as finding the data capability of the compartment it is transitioning to. The TID is needed in order to identify which trusted stack is the correct one. The same trampoline code can be invoked from different threads. Imagine two threads doing exactly the same thing, e.g., making a call from "CompA" to "CompB". The trampoline needs to use a different trusted stack for each thread it is invoked from.

The user code does not need to change the thread ID - that is why `utid` is read-only. The kernel changes the `stid` register whenever it changes the software thread running on a hardware thread.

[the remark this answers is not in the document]
andresag01: Ok, thanks I see what you mean. My description is a little confused regarding the meaning of "compartment" and "thread".

In previous discussion, you mentioned that the thread scheduling could be done in user or supervisor mode. If so, then this stid/utid arrangement is only for supervisor mode. Correct? Did I misunderstand something?

src/riscv-integration.adoc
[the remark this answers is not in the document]
andresag01: The specific size would not be an architectural requirement. The OS could easily discover how many thread IDs are available and deal with it when it runs out.

src/riscv-integration.adoc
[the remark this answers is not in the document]
francislaus: There can be multiple levels of thread scheduling. When there is user-level threading, there will also be supervisor-level threading. `stid` and `utid` allow for the thread ID to be communicated from supervisor mode to user mode. We currently believe that there is a way to provide a thread id within a privilege domain via read-only capabilities.

src/riscv-integration.adoc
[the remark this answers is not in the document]
francislaus: @andresag01 Do you have data that would suggest a certain minimal size for the thread ID registers?

tariqkurd-repo:
How about making this into a separate extension so we can mark it as work-in-progress like the PTE extension?

That way we can merge something sooner without accepting it into the base architecture.

src/riscv-integration.adoc
```diff
@@ -1102,3 +1143,17 @@ written to these CSRs unless the specification explicitly states that the CSR
 behaves otherwise (see <<mepcc>> and <<sepcc>>). Also note that <<pcc>> can
 be written with a <<JALR>> instruction in Capability Mode which automatically unseal the capability _before_
 the invalid address conversion above.
+
+[#CHERI_COMP,reftext="CHERI Compartmentalization]
+=== CHERI Compartmentalization
+
+This section describes how this specification enables support for compartmentalization for CHERI systems.
+Compartmentalization seeks to separate the privileges between different protection units, e.g., two or more libraries.
+Code can be separated by sentries, which allow for giving out code capabilities to untrusted code where the untrusted code can only call code capability, but not modify it.
```
PeterRugg, line 1152: "call code capability" -> "call the code capability"

PeterRugg (approved):

PeterRugg:
Looks good! (A few refs for UTID and STID possibly missing in the "CHERI Compartmentalization" section, but that's a very minor nit)

arichardson (changes requested):
Since the precise semantics are still experimental please factor it out to a separate extension.

src/riscv-integration.adoc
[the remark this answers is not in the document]
arichardson: There should be no need to reduce the number of bits here, it's a single 32/64-bit CSR and making the length variable would make it harder for the OS to use arbitrary thread IDs.

src/riscv-integration.adoc
```diff
@@ -973,11 +994,31 @@ is the cause of the fault. The possible CHERI types and causes are encoded as
 shown in xref:mtval-cheri-type[xrefstyle=short] and
 xref:mtval-cheri-causes[xrefstyle=short] respectively.
 
+[#stid,reftext="stid"]
+==== Supervisor Thread Identifier (stid)
+
+The <<stid>> register is an SXLEN-bit read-write register. It is used to
+identify the current thread. The reset value of this register is
```
arichardson, line 1001:
```suggestion
identify the current user-mode thread. The reset value of this register is
```
Since the value is shared we should be explicit that it's for U mode. If S-mode wants to reuse the same ID that's fine too, but maybe it should be something like `sutid` to indicate that it's the writable version of `utid`?

src/attributes.adoc
```diff
@@ -48,6 +48,8 @@ endif::[]
 :cheri_pte_ext_name:       Zcheri_pte
 // Extension for CHERI capabilities in vector registers
 :cheri_vectorcap_ext_name: Zcheri_vectorcap
+// Extension for CHERI thread identification
+:cheri_tid_ext_name: Zcheri_tid
```
andresag01, line 52: Thanks for factoring this out into another extension. However, I thought that we were hoping this wouldn't be a CHERI-specific instruction, so perhaps it doesn't need the `cheri_` prefix? Also, note that `_` is an illegal character in extension names, so could you remove that too? (We are supposed to fix this in the other extensions but haven't got around to it yet!)

francislaus: Sure, will fix. It is a CHERI-dependent extension because we are introducing capabilities. `utidc` has CLEN.

src/tid-ext.adoc
```diff
@@ -43,9 +43,10 @@ include::img/stidreg.edn[]
 [#stidc,reftext="stidc"]
 ==== Supervisor Thread Identifier Capability (stidc)
 
-The <<stidc>> register is an MXLEN-bit read-write register. It is used to
-identify the current thread. On reset the tag will be set
-to 0 and the remainder of the data is UNSPECIFIED.
+The <<stidc>> register is an MXLEN-bit read-write capability register.
```
arichardson, line 46:
```suggestion
The <<stidc>> register is an CLEN-bit read-write capability register.
```

tariqkurd-repo (approved):

francislaus:
@jrtc27 @tariqkurd-repo @andresag01 @PeterRugg @arichardson

I have factored this out into a separate extension. Please let me know what you think.

src/tid-ext.adoc
```diff
@@ -0,0 +1,104 @@
+== "Zstid Extension for Thread Identification
+
+{tid_ext_name} is an optional extension to the RISC-V base ISA.
+Implementations that support {cheri_base_ext_name} and {tid_ext_name}.
+define a variant of the CHERI ISA that allows for software compartmentalization
+of CHERI programs.
```
arichardson, line 6: I'd argue it's possible without as well, so maybe quantify that with "improved"/"more efficient"?

src/tid-ext.adoc
```diff
@@ -0,0 +1,104 @@
...
+
+The <<stid>> register is an SXLEN-bit read-write register. It is used to
+identify the current thread. The reset value of this register is
+UNSPECIFIED.
+
+.Supervisor thread identifier register
+include::img/stidreg.edn[]
+
+[#stidc,reftext="stidc"]
+==== Supervisor Thread Identifier Capability (stidc)
```
arichardson, line 44: I'd move the capability versions below the integer ones and add a sentence above along the lines of 

`When {cheri_base_ext} is implemented, the {tid_ext_name} CSRs are extended as follows:`

arichardson (commented):
This overall looks good to me now - just two comments I'd like to see addressed before merging.

Also it might make sense to update the PR description with a bit more detail for the default squash+merge commit message.

francislaus:
@arichardson Thank you, I have addressed your comments.

src/introduction.adoc
```diff
@@ -95,6 +95,8 @@ virtual-memory
 {cheri_vectorcap_ext_name}:: CHERI extension for the RISC-V Vector (V)
 extension. It adds support for storing CHERI capabilities in
 vector registers, intended for vectorised memory copying
+{tid_ext_name}:: Extension for supporting thread identifiers. This extenstions
```
arichardson, line 98:
```suggestion
{tid_ext_name}:: Extension for supporting thread identifiers. This extension
```

src/introduction.adoc
```diff
@@ -95,6 +95,8 @@ virtual-memory
...
+allowing for software compartmentalization on CHERI systems.
```
arichardson, line 99:
```suggestion
improves software compartmentalization on CHERI systems.
```

arichardson (approved):
Minor spelling issue otherwise LGTM

PR #188, opened 2024-04-08
Change CSR/jump/branch invalid address handling

### LINKED ISSUE #139: Specify how invalid address conversion is handled for jumps.
For JAL in legacy mode it says:

> the target address which is written to the [pcc](https://riscv.github.io/riscv-cheri/#pcc)'s address field

(`the` should probably be removed btw)

It says a very similar thing for conditional branches.

It doesn't really specify how invalid address handling should be applied here. Suggestion from Jamie was that the (non-capability) target address undergoes invalid address conversion. PCC is not affected if invalid address conversion changes the target address.

andresag01:
* Unify the invalid address handling for both CSR and control transfer instructions
* The new invalid address handling clears the tag if the capability being written is not Infinity. Also, invalid address conversion is now performed before anything else

Fixes #139

### DIFF
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -491,7 +491,8 @@ That is, invert the most significant bit of _t_ if the decoded length of the
 capability is larger than E.
 
 NOTE: A capability has infinite bounds if E=CAP_MAX_E and it is not
-malformed (see xref:section_cap_malformed[xrefstyle=short]).
+malformed (see xref:section_cap_malformed[xrefstyle=short]); this check is
+equivalent to _b_=0 and _t_&#8804;2^MXLEN^.
 
 [#section_cap_malformed]
 ===== Malformed Capability Bounds
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -475,7 +475,8 @@ while still allowing the bounds to be correctly decoded.
 .Memory address bounds encoded within a capability
 image::cap-bounds-map.png[width=80%,align=center]
 
-A capability whose bounds cover the entire address space has 0 base and top
+A capability has _infinite_ bounds if its bounds cover the entire address space
+such that the base address is 0 and top
 equals 2^MXLEN^, i.e. _t_ is a MXLEN + 1 bit value. However, _b_ is a
 MXLEN bit value and the size mismatch introduces additional complications
 when decoding, so the following condition is required to correct _t_ for
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -476,8 +476,8 @@ while still allowing the bounds to be correctly decoded.
 image::cap-bounds-map.png[width=80%,align=center]
 
 A capability has _infinite_ bounds if its bounds cover the entire address space
-such that the base address is 0 and top
-equals 2^MXLEN^, i.e. _t_ is a MXLEN + 1 bit value. However, _b_ is a
+such that the base address _b_=0 and the top address _t_&#8804;2^MXLEN^,
+i.e. _t_ is a MXLEN + 1 bit value. However, _b_ is a
 MXLEN bit value and the size mismatch introduces additional complications
 when decoding, so the following condition is required to correct _t_ for
 capabilities whose <<section_cap_representable_check>> wraps the edge of the address
@@ -489,6 +489,9 @@ if ( (E < (CAP_MAX_E - 1)) & (t[MXLEN: MXLEN - 1] - b[MXLEN - 1] > 1) )
 ```
 That is, invert the most significant bit of _t_ if the decoded length of the
 capability is larger than E.
+
+NOTE: A capability has infinite bounds if E=CAP_MAX_E and it is not
+malformed (see xref:section_cap_malformed[xrefstyle=short]).
 
 [#section_cap_malformed]
 ===== Malformed Capability Bounds
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -476,7 +476,7 @@ while still allowing the bounds to be correctly decoded.
 image::cap-bounds-map.png[width=80%,align=center]
 
 A capability has _infinite_ bounds if its bounds cover the entire address space
-such that the base address _b_=0 and the top address _t_&#8804;2^MXLEN^,
+such that the base address _b_=0 and the top address _t_&#8805;2^MXLEN^,
 i.e. _t_ is a MXLEN + 1 bit value. However, _b_ is a
 MXLEN bit value and the size mismatch introduces additional complications
 when decoding, so the following condition is required to correct _t_ for
@@ -492,7 +492,7 @@ capability is larger than E.
 
 NOTE: A capability has infinite bounds if E=CAP_MAX_E and it is not
 malformed (see xref:section_cap_malformed[xrefstyle=short]); this check is
-equivalent to _b_=0 and _t_&#8804;2^MXLEN^.
+equivalent to _b_=0 and _t_&#8805;2^MXLEN^.
 
 [#section_cap_malformed]
 ===== Malformed Capability Bounds
--- a/src/riscv-integration.adoc
+++ b/src/riscv-integration.adoc
@@ -1087,7 +1087,7 @@ and the bounds encoding depends on the address value, so implementations must
 not convert invalid addresses to other arbitrary invalid addresses in an
 unrestricted manner.
 The following procedure must be used instead when writing a capability A to
-a CSR:
+a CSR that cannot hold all invalid addresses:
 
 . If A's address cannot be held, then convert it to another address that the
 register can hold.
@@ -1097,7 +1097,8 @@ xref:section_cap_encoding[xrefstyle=short]), then A's tag is set to 0.
 <<mtvecc>>, <<mepcc>>, etc.
 
 The following procedure must be used instead when jumping or branching to a
-capability A. Note that A is the <<pcc>> when executing direct jumps and branches
+capability A if the <<pcc>> cannot hold all invalid invalid addresses. Note
+that A is the <<pcc>> when executing direct jumps and branches
 or the CHERI execution mode is Legacy (see
 xref:section_cheri_legacy_ext[xrefstyle=short]):
 
--- a/src/riscv-integration.adoc
+++ b/src/riscv-integration.adoc
@@ -679,7 +679,7 @@ _Reserved_
 |Priority |Exc.Code |Description
 |_Highest_ |3 |Instruction address breakpoint
 | .>|*{cheri_excep_mcause}* .<|*Prior to instruction address translation:* +
-*CHERI fault due to PCC checks (tag, execute permission and bounds)*
+*CHERI fault due to PCC checks (tag, execute permission, invalid address and bounds)*
 | .>|12, 1 .<|During instruction address translation: +
 First encountered page fault or access fault
 | .>|1 .<|With physical address for instruction: +
@@ -697,11 +697,11 @@ Load/store/AMO address breakpoint
 
 | .>| *{cheri_excep_mcause}* .<| *CHERI faults due to:* +
 *PCC <<asr_perm>> clear* +
-*Branch/jump target address checks (tag, execute permissions and bounds)*
+*Branch/jump target address checks (tag, execute permissions, invalid address and bounds)*
 
 | .>|*{cheri_excep_mcause}* .<|*Prior to address translation for an explicit memory access:* +
 *Load/store/AMO capability address misaligned* +
-*CHERI fault due to capability checks (tag, permissions and bounds)*
+*CHERI fault due to capability checks (tag, permissions, invalid address and bounds)*
 | .>|4,6 .<|Optionally: +
 Load/store/AMO address misaligned
 | .>|13, 15, 5, 7 .<|During address translation for an explicit memory access: +
@@ -763,7 +763,8 @@ xref:mtval-cheri-causes[xrefstyle=short] respectively.
 | 1                | Seal violation
 | 2                | Permission violation
 | 3                | Length violation
-| 4-15             | Reserved
+| 4                | Invalid address violation
+| 5-15             | Reserved
 |==============================================================================
 
 [#supervisor-level-csrs-section]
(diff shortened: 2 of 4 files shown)


andresag01:
FYI: @PeterRugg @Timmmm

src/riscv-integration.adoc
```diff
@@ -1078,27 +1078,41 @@ memory addresses must match for the address to be valid:
...
+a CSR:
 
-. If A's address cannot be held then convert it to another address that the CSR can
-hold
-. If conversion _was_ required, then A's tag is cleared if A is
-sealed or if the new address is not representable -- this is equivalent to the
-semantics of <<SCADDR>>
+. If A's address cannot be held, then convert it to another address that the
+register can hold.
+. If A's address is invalid and A's bounds do not equal the bounds of the
```
tariqkurd-repo, line 1094: Infinite capability should be an xref, in all references in the text.

it's worth spelling out how to make the bounds match - either here or in the xref - all we're really saying is that EXP is 0 representing maximum bounds. There's no need to look at the T and B fields, for example (once @PRugg-Cap has merged the 66-bit arithmetic PR)

it's also worth making that a clear statement _first_ that all CSRs or registers which store an address, where the full bit width isn't stored in an RV64 VM core, must always clear the tag if the address is invalid _and_ the bounds aren't infinity.

the rest of the text is fine, but having that statement first will make it clearer.

PRugg-Cap: Yeah, I'd be tempted to reword "do not equal the bounds of the Infinite capability" to "does not have infinite bounds", and link to a section somewhere that defines that in text as `base <= 0, top >= 2**xlenmax`, with a note saying that for tagged (and in fact non-malformed) capabilities, this is equivalent to checking EXP is `exp_max` or I guess 0 now as Tariq says.

jrtc27:
I do not like that we're polluting the architecture with this inconsistency just to support what is supposed to be an optimisation, one which is even less fruitful for CHERI given you still have to preserve all the metadata.

tariqkurd-repo:
> I do not like that we're polluting the architecture with this inconsistency just to support what is supposed to be an optimisation, one which is even less fruitful for CHERI given you still have to preserve all the metadata.

It's an implementation choice - no-one has to do any of this - they can keep full 64-bit VAs everywhere. The idea behind this is to try and minimise the hardware area. It's still perfectly valid to ignore all of this, but we are keen on minimising the area of the implementation.

src/riscv-integration.adoc
andresag01: I addressed these comments. Could you please take another look? Thanks!

src/riscv-integration.adoc
```diff
@@ -1078,27 +1078,43 @@ memory addresses must match for the address to be valid:
...
-be written with a <<JALR>> instruction in Capability Mode which automatically unseal the capability _before_
-the invalid address conversion above.
+The following procedure must be used instead when jumping or branching to a
+capability A. Note that A is the <<pcc>> when executing direct jumps and branches
+or the CHERI execution mode is Legacy (see
+xref:section_cheri_legacy_ext[xrefstyle=short]):
+
+. Calculate the effective target address T of the jump or branch as required by
+the instruction's behavior.
+. If T is invalid and A does not have infinite bounds (see
```
tariqkurd-repo, line 1106: this needs to be predicated on pcc not storing all possible invalid addresses. There's no reason to clear the tag if pcc _can_ store all 64 address bits.

tariqkurd-repo: so that this behaviour is only enabled when pcc and the other address registers are narrow

jamie-melling:
Maybe a naive question, but why does this not apply to data operations like loads and stores?

src/riscv-integration.adoc
andresag01: I will clarify that in the text just above this bullet list

andresag01: Fixed

src/riscv-integration.adoc
```diff
@@ -1078,27 +1078,44 @@ memory addresses must match for the address to be valid:
...
 . Write the final (potentially modified) version of capability A to the CSR e.g.
 <<mtvecc>>, <<mepcc>>, etc.
 
-This implies that sealed capabilities will always get their tags cleared when
-written to these CSRs unless the specification explicitly states that the CSR
-behaves otherwise (see <<mepcc>> and <<sepcc>>). Also note that <<pcc>> can
-be written with a <<JALR>> instruction in Capability Mode which automatically unseal the capability _before_
-the invalid address conversion above.
+The following procedure must be used instead when jumping or branching to a
+capability A if the <<pcc>> cannot hold all invalid invalid addresses. Note
```
tariqkurd-repo, line 1100: repeated word - invalid invalid

src/riscv-integration.adoc
```diff
@@ -1078,27 +1078,44 @@ memory addresses must match for the address to be valid:
...
+a CSR that cannot hold all invalid addresses:
 
-. If A's address cannot be held then convert it to another address that the CSR can
-hold
-. If conversion _was_ required, then A's tag is cleared if A is
-sealed or if the new address is not representable -- this is equivalent to the
-semantics of <<SCADDR>>
+. If A's address cannot be held, then convert it to another address that the
+register can hold.
+. If A's address is invalid and A does not have infinite bounds (see
```
tariqkurd-repo, line 1094: the first and second point are the same one - it cannot be held if it's invalid -it's the same concept - so better to combine these two

andresag01: Fixed

src/riscv-integration.adoc
andresag01: Fixed

tariqkurd-repo (approved):

src/riscv-integration.adoc
tariqkurd-repo: is this ok now @PRugg-Cap ?

src/cap-description.adoc
```diff
@@ -468,6 +469,8 @@ if ( (E < (CAP_MAX_E - 1)) & (t[MXLEN: MXLEN - 1] - b[MXLEN - 1] > 1) )
 That is, invert the most significant bit of _t_ if the decoded length of the
 capability is larger than E.
 
+NOTE: Capabilities with E=CAP_MAX_E have infinite bounds.
+
```
andresag01, line 473: @PRugg-Cap @PeterRugg @tariqkurd-repo : I modified the definition of "infinite bounds". Could you please confirm that this note and the previous paragraph are as expected? Thanks!

PRugg-Cap: Looks fixed, thanks! For some reason this preview is showing me something stale, but the most recent thing I saw that includes the malformed note looks right.

tariqkurd-repo:
Summary from a separate discussion about this tricky problem:

1. We don't want to generate a capability with a changed address, without having to perform additional checks which is not a good use of logic gates. If JALRing to an invalid address, and the bounds include invalid addresses, then technically we need to do the bounds comparison before _and_ after the _optional_ address conversion. This not a good use of logic gates and power.
2. The problem can only arise if the OS has set bounds including an invalid address range

The proposal is to add a new CHERI fault type, usable for both load/store derefencing and jumps/branches as all can attempt to access invalid addresses.

If the bounds are _not_ infinity (so the legacy case isn't broken) _and_ the calculated address is invalid _then_ take a new type of CHERI fault - to be defined.
The intention is to also allow a narrower bounds check, only covering the necessary address bits for Sv39, Sv48 etc. which will be a big saving over full 64-bit comparators every time.

So... when calculating an address
1. If it's invalid and the bounds aren't infinity - take a new type of CHERI fault
2. If it's valid, check the bounds - which is lower priority to allow the narrower comparators

Therefore we're no longer spending logic gates and power on checking the out of range address bits on the bounds checks.

It means that JALRs and branches to invalid addresses will trap on the jump/branch (unless the bounds are infinity for the legacy case)

PRugg-Cap:
Just to clarify exception precedence: it should be Tag, Seal, Perm, Optional Invalid Addr, Bounds? I don't think that priority is actually listed anywhere in the spec currently, though of course the sail will disambiguate. It does look like the encodings go in that order though, so I wonder if the Bounds violation encoding wants to get bumped up by one?

src/cap-description.adoc
```diff
@@ -475,8 +475,9 @@ while still allowing the bounds to be correctly decoded.
 .Memory address bounds encoded within a capability
 image::cap-bounds-map.png[width=80%,align=center]
 
-A capability whose bounds cover the entire address space has 0 base and top
-equals 2^MXLEN^, i.e. _t_ is a MXLEN + 1 bit value. However, _b_ is a
+A capability has _infinite_ bounds if its bounds cover the entire address space
+such that the base address _b_=0 and the top address _t_&#8804;2^MXLEN^,
```
Timmmm, line 479: I think `&#8804;` is `<=`. Maybe `&geq;` works and is more readable?

src/cap-description.adoc
```diff
@@ -489,6 +490,9 @@ if ( (E < (CAP_MAX_E - 1)) & (t[MXLEN: MXLEN - 1] - b[MXLEN - 1] > 1) )
 That is, invert the most significant bit of _t_ if the decoded length of the
 capability is larger than E.
 
+NOTE: A capability has infinite bounds if E=CAP_MAX_E and it is not
```
Timmmm, line 493: Maybe it's worth being more explicit about the relationship between `t>=2^MXLEN` and `E=CAP_MAX_E`. Are they exactly equivalent? Something like:

```
NOTE: Checking for infinite bounds via `b= 0` and `t >= 2^MXLEN` is exactly equivalent to checking `E=CAP_MAX_E` and that the capability is not malformed.
```

(if that is the case)

src/riscv-integration.adoc
```diff
@@ -1078,27 +1078,42 @@ memory addresses must match for the address to be valid:
 * For Sv48, bits [63:48] must equal bit 47
 * For Sv57, bits [63:57] must equal bit 56
 
-RISC-V permits that some CSRs, such as <<mtvec>> and <<mepc>> (see
-xref:CSR_exevectors[xrefstyle=short]), need not be able to hold all possible
-invalid addresses. Prior to writing these CSRs, implementations may convert an
+RISC-V permits that some CSRs holding addresses, such as <<mtvec>> and <<mepc>>
+(see xref:CSR_exevectors[xrefstyle=short]) and consequently <<pcc>>, need not be able to hold all possible
```
Timmmm, line 1082: Minor nit: `consequently` is a bit weird here. It isn't really a consequence of `mtvec` or `mepc` which is what it sounds like.

Maybe:

```
RISC-V permits that some registers that only hold addresses, such as <<mtvec>> and <<mepc>> (see ...) and also <<pcc>>, need not ...
```

Timmmm (commented):
Does the `Action taken on writing to extended CSRs.` table need to be updated?

src/cap-description.adoc
PRugg-Cap: Yep!
`(inCap.E == cap_max_E & not(isCapMalformed(inCap))) == (base == 0 & top >= 2^(sizeof(xlen)))`

Timmmm: Maybe a nice property to try and prove btw!

PRugg-Cap: Yes, sorry, should have been clearer: that's a property that SMT was happy with.

Timmmm: Oh right, nice! Definitely worth pointing that out in the spec then I think.

tariqkurd-repo:
> Just to clarify exception precedence: it should be Tag, Seal, Perm, Optional Invalid Addr, Bounds? I don't think that priority is actually listed anywhere in the spec currently, though of course the sail will disambiguate. It does look like the encodings go in that order though, so I wonder if the Bounds violation encoding wants to get bumped up by one?

bounds needs to be _below_ optional invalid address check, because otherwise we need a full 64-bit bounds check instead of a much narrower one for Sv48/Sv39. So keeping the order you said will give a smaller and less power hungry cheri core.

src/cap-description.adoc
andresag01: Fixed

src/cap-description.adoc
andresag01: Sorry, I tried but asciidoc complained, so I left it as is. Its very minor, so I will resolve this comment.

src/riscv-integration.adoc
andresag01: Done!

andresag01:
> Does the `Action taken on writing to extended CSRs.` table need to be updated?

I don't think there is anything to change there. The invalid address handling changed, but we still need the csetaddr checks, etc.

src/cap-description.adoc
Timmmm: But I mean the symbol is wrong isn't it? Don't you mean `>=` not `<=`?

src/riscv-integration.adoc
```diff
@@ -763,7 +763,8 @@ xref:mtval-cheri-causes[xrefstyle=short] respectively.
 | 1                | Seal violation
 | 2                | Permission violation
 | 3                | Length violation
-| 4-15             | Reserved
+| 4                | Invalid address violation
```
PRugg-Cap, line 766: This is what I was trying to say in https://github.com/riscv/riscv-cheri/pull/188#issuecomment-2051486224: up till now the encodings have mirrored the priorities, so should Length be 4 and Invalid address be 3? It seems likely future extensions would be unable to maintain this property though, so maybe it's not worth bothering with. Still, it could be worth having a table of priorities?

src/riscv-integration.adoc
```diff
@@ -1078,27 +1079,75 @@ memory addresses must match for the address to be valid:
...
+
+The following procedure must be used when jumping or branching to a target
+capability A if the <<pcc>> cannot hold all invalid addresses. Note
+that the target capability A is the <<pcc>> when executing direct jumps and
+branches or the CHERI execution mode is Legacy (see
+xref:section_cheri_legacy_ext[xrefstyle=short]):
+
+. Calculate the effective target address T of the jump or branch as required by
+the instruction's behavior.
+. If T is invalid and A does not have infinite bounds (see
```
PRugg-Cap, line 1122: Not sure how to add it here concisely, but this happens after tag/seal/perm checks but before bounds check.

PRugg-Cap (approved):

src/riscv-integration.adoc
andresag01: Fixed

src/riscv-integration.adoc
andresag01: I added a list of priorities in the section where the codes are defined.

src/cap-description.adoc
andresag01: Ah right! Sorry, I did not realize you meant that the sign needs to change. Fixed! However, the `&ge;` code didn't work.

Timmmm (approved):

src/riscv-integration.adoc
```diff
@@ -1078,27 +1087,79 @@ memory addresses must match for the address to be valid:
...
-The following procedure must be used instead when writing a capability A to
-these CSRs:
-
-. If A's address cannot be held then convert it to another address that the CSR can
-hold
-. If conversion _was_ required, then A's tag is cleared if A is
-sealed or if the new address is not representable -- this is equivalent to the
-semantics of <<SCADDR>>
+RISC-V permits that CSRs holding addresses, such as <<mtvec>> and <<mepc>>
+(see xref:CSR_exevectors[xrefstyle=short]) as well as <<pcc>>, need not
```
arichardson, line 1091: Should this refer to pc rather than pcc?

src/riscv-integration.adoc
```diff
@@ -1078,27 +1087,79 @@ memory addresses must match for the address to be valid:
...
+However, the bounds encoding of capabilities in {cheri_base_ext_name} depends
+on the address value, so implementations must not convert invalid addresses to
+other arbitrary invalid address in an unrestricted manner. The remainder of
+this section describes how invalid address handling must be supported in
+{cheri_base_ext_name} when accessing CSRs, branching and jumping, and
+accessing memory.
+
+===== Accessing CSRs
+
+The following procedure must be used instead when executing instructions, such
```
arichardson, line 1111:
```suggestion
The following procedure must be used when executing instructions, such
```

arichardson (approved):
I'm happy with these semantics and I think we should merge it once @jrtc27 has taken another look.

src/riscv-integration.adoc
andresag01: Ah yes, you are right. I will fix it!

andresag01: Fixed!

PR #190, opened 2024-04-08
S mode registers should be of SXLEN width

francislaus:


andresag01 (changes requested):
These bit widths were correct. The capability-sized registers are always CLEN width which is 2*MXLEN. We use MXLEN because it is the maximum XLEN supported by the machine and it is guaranteed to not change.

Recall that SXLEN changes depending the value of `mstatus.SXL`, so we cannot define the width of a cap based on that because if software changed `mstatus.SXL` dynamically then what would happen to the capability encoding?

andresag01:
Closing because the existing diagrams in the spec are correct.

PR #197, opened 2024-04-11
Define AP field for Infinite on purecap and legacy exts

andresag01:
The MXLEN=32 encoding for the AP field was recently changed. However, the following remained unclear:
1. The encoding of the AP field for MXLEN=32 when Zcheri_legacy is or is not supported
2. The encoding of the Infinite capability for MXLEN=32.

For (1), this change ensures that the AP encodings for Mode=Legacy are "reserved" in valid capabilities (ie. we cannot have a valid tag with an encoding that has Mode=Legacy).

For (2), the tables and text are updated such that the AP encoding for the Infinite Capability is:

* 9 when Zcheri_purecap is supported but Zcheri_legacy is **not**
* 8 when Zcheri_purecap and Zcheri_legacy are **both** supported


src/cap-description.adoc
```diff
@@ -188,6 +188,11 @@ reserved values as if it were 0b00000 (no permissions). Future extensions may as
 meanings to the reserved bit patterns, in which case <<GCPERM>> is allowed to report a
 non-zero value.
 
+NOTE: When MXLEN=32 the hart's CHERI execution mode is always Capability (see
+xref:m_bit[xrefstyle=short]), so the encodings for Legacy Mode are reserved,
+and must not appear in valid capabilities, if only {cheri_base_ext_name} is
+supported.
+
```
tariqkurd-repo, line 195: why is this here? It should start with "if only Zcheri_purecap is supported" to make the intent clearer, but do we need to say this? Surely this text is elsewhere, when we instructroduce Zcheri_purecap / Zcheri_legacy?

src/riscv-legacy-integration.adoc
```diff
@@ -55,26 +55,30 @@ top is 2^MXLEN^
...
+xref:cap_encoding_xlen64_mode[xrefstyle=short]; M=0 indicates Legacy Mode and
+M=1 indicates Capability Mode.
+
+The current CHERI execution mode is given by the M field of the <<pcc>> and the
+CRE bits in <<mseccfg>>, <<menvcfg>>, and <<senvcfg>> as follows:
+
+* The mode is Capability when the M field of the <<pcc>> is Capability and the effective
 CRE=1 for the current privilege level
-* The mode is Legacy when the effective CRE=0 for the current privilege level or the M bit of the <<pcc>> is 0
+* The mode is Legacy when the effective CRE=0 for the current privilege level or the M field of the <<pcc>> is Legacy
```
tariqkurd-repo, line 70: use the "M-bit" xref for consistency

src/cap-description.adoc
andresag01: I could not find this kind of text elsewhere. The closest is perhaps this from the Zcheri_legacy chapter:
> NOTE: The implication is that the CHERI execution mode is always Capability on
implementations that support {cheri_base_ext_name}, but not
{cheri_legacy_ext_name}.

I think the main point here is to indicate that the AP encodings for Legacy are "reserved" when only Zcheri_purecap is supported.

Happy to make the text clearer once we agree whether its needed and where it should go.

src/riscv-legacy-integration.adoc
```diff
@@ -55,26 +55,30 @@ top is 2^MXLEN^
...
 
 .Capability encoding when MXLEN=64 and {cheri_legacy_ext_name} is supported
 [#cap_encoding_xlen64_mode]
 include::img/cap-encoding-xlen64-mode.edn[]
 
-For RV64, the M bit can be set to 1 when the capability does not grant <<x_perm>>.
-In this case, the M bit is superfluous, so the encoding may be used to support additional features in future extensions.
+When MXLEN=32, the M field is Legacy Mode in the <<infinite-cap>> capability,
+so its AP field is 8. The <<null-cap>> capability does not grant <<x_perm>>, so
```
tariqkurd-repo, line 77: what is this talking about? Above you say it must be 9 for infinity and 8 isn't an option but now it's 8?

tariqkurd-repo: Need a new line before null-cap

src/riscv-legacy-integration.adoc
```diff
@@ -55,26 +55,30 @@ top is 2^MXLEN^
...
+the M field is superfluous in this case.
 
-The M bit is 0 in both the <<null-cap>> and <<infinite-cap>> capabilities.
+When MXLEN=64, the M bit can be set to 1 when the capability does not grant <<x_perm>>.
+In this case, the M bit is superfluous, so the encoding may be used to support additional features in future extensions. The M bit is 0 in both the <<null-cap>> and <<infinite-cap>> capabilities.
```
tariqkurd-repo, line 81: M-bit should be a xref

src/riscv-legacy-integration.adoc
andresag01: According to the encoding table:
* 8 = all permissions granted + Legacy mode
* 9 = all permissions granted + Capability mode

In the text above with 9, its really related to Zcheri_purecap because the CHERI execution mode is not defined there, so it makes sense that your infinite cap is also Capability mode.

In this text with 8, its the definition of Zcheri_legacy where the CHERI execution mode is already defined. The agreement so far is that a machine supporting Zcheri_legacy has Infinite cap with Legacy mode hence 8.

src/riscv-legacy-integration.adoc
tariqkurd-repo: I'm now totally confused. It would be so much easier to have one table for inifite and null-cap instead of trying to add information in this way - or just replicate the tables here.

Is infinite cap really in legacy mode? Does the infinite cap support both modes? 

it seems to me that the value doesn't change - there's no M-bit on a purecap core, and for legacy it's set to zero so there's no difference? what am I missing?

tariqkurd-repo (approved):

src/riscv-legacy-integration.adoc
```diff
@@ -55,26 +56,27 @@ top is 2^MXLEN^
...
-capabilities. In RV32, the mode is encoded in the permissions field as
-described in xref:cap_permissions_encoding[xrefstyle=short]. In RV64, a new
-mode bit field is added to the capability format as shown in
-xref:cap_encoding_xlen64_mode[xrefstyle=short]. The current CHERI execution
-mode is given by the M bit of the <<pcc>> and the CRE bits in <<mseccfg>>,
-<<menvcfg>>, and <<senvcfg>> as follows:
+{cheri_legacy_ext_name} adds a new CHERI execution mode field (M) to
+capabilities within the AP field (see xref:section_cap_perms[xrefstyle=short]).
+When MXLEN=32, the mode is encoded in bit 0 of quadrant 1 from the AP field.
+When MXLEN=64, the mode is encoded in bit 0 of the AP fielda new M bit field is
```
arichardson, line 62:
```suggestion
When MXLEN=64, the mode is encoded in bit 0 of the AP field: a new M bit field is
```

src/riscv-legacy-integration.adoc
```diff
@@ -55,26 +56,27 @@ top is 2^MXLEN^
...
+* The mode is Legacy when the effective CRE=0 for the current privilege level or the M field of the <<pcc>> is Legacy
 
-For RV64, the M bit can be set to 1 when the capability does not grant <<x_perm>>.
-In this case, the M bit is superfluous, so the encoding may be used to support additional features in future extensions.
+When MXLEN=32, the M field is Legacy Mode in the <<infinite-cap>> capability,
+so its AP field is 8. The <<null-cap>> capability does not grant <<x_perm>>, so
+the M field is superfluous in this case.
 
-The M bit is 0 in both the <<null-cap>> and <<infinite-cap>> capabilities.
+When MXLEN=64, the M bit can be set to 1 when the capability does not grant <<x_perm>>.
```
arichardson, line 78: I'd suggest putting the MXLEN=64 paragraph first since I was confused why we are talking about the null cap when reading it in order, but it made sense after reading the 64-bit one.

arichardson (commented):
Minor comments otherwise good to go.

src/riscv-legacy-integration.adoc
andresag01: The NULL part only applies to MXLEN=32, so its really a continuation of the same idea. I will change the punctuation to make it clearer.

arichardson (approved):

PR #203, opened 2024-04-16
Make CSRR{S|W} read CLEN bits

### LINKED ISSUE #201: CSR[W|S|C]I and CSR[S|C] inconsistency of operations
There is a contradiction in the spec regarding what should be done for CSR operations that are not CSRRW.

Section 3.5.1. states that non-CSRRW operations will perform a CLEN read and an XLEN write for purecap machines.

Section 5.4.4. states that non-CSRRW operations will perform XLEN read and writes operations to the CSR regardless of the mode for Legacy machines.

Section 7.4.6., the mnemonics state that non-CSRRW operations will perform a CLEN read and an XLEN write in capability mode and XLEN read and writes in Legacy mode. 

If the approach mentioned in section 5.4.4. is chosen, it leaves no way for a CLEN CSRR to be performed (CSRR[C|S] rs1==x0 in capability mode)

andresag01:
Remove outdated text claiming that CSRR{S|W} always read XLEN-sized data. This is incorrect as discussed in https://github.com/riscv/riscv-cheri/issues/92 and the behaviour was changed to ensure that CSRR{S|W} reads CLEN data while in Capability mode. This change simply clears up some of the old text.

Fixes #201

### DIFF
--- a/src/riscv-legacy-integration.adoc
+++ b/src/riscv-legacy-integration.adoc
@@ -219,12 +219,9 @@ When an extended CSR is used with another CSR instruction (<<CSRRWI>>,
 (set bits, clear bits etc).
 * The final address is updated as specified in <<extended_CSR_writing>> for an
 XLEN write.
-* XLEN bits are read from the capability address field and written to an output
-*x* register.
-
-There is _no distinction_ between Legacy and Capability Mode in this case -
-XLEN access is always performed, and the assembly syntax always uses *x*
-registers.
+* In Legacy Mode, XLEN bits are read from the capability address field and
+written to an output *x* register. In Capability Mode, CLEN bits are read from
+the CSR and written to an output *c* register.
 
 All CSR instructions cause CHERI exceptions if the <<pcc>> does not grant
 <<asr_perm>> and the CSR accessed is not user-mode accessible.


src/riscv-legacy-integration.adoc
```diff
@@ -220,11 +220,8 @@ When an extended CSR is used with another CSR instruction (<<CSRRWI>>,
 * The final address is updated as specified in <<extended_CSR_writing>> for an
 XLEN write.
 * XLEN bits are read from the capability address field and written to an output
```
tariqkurd-repo, line 222: CLEN bits are read, which is what we're trying to fix

src/riscv-legacy-integration.adoc
```diff
@@ -220,11 +220,8 @@ When an extended CSR is used with another CSR instruction (<<CSRRWI>>,
...
-*x* register.
-
-There is _no distinction_ between Legacy and Capability Mode in this case -
-XLEN access is always performed, and the assembly syntax always uses *x*
-registers.
+*x* register if the mode is Capability. In Legacy mode, CLEN bits are read from
```
tariqkurd-repo, line 223: elsewhere I capitalise legacy Mode. it would be nicer to say

"... in Capability Mode" "...in Legacy Mode"

src/riscv-legacy-integration.adoc
andresag01: Yes, sorry. I inverted the sense of what I wanted to say. Fixed now.

src/riscv-legacy-integration.adoc
andresag01: Fixed

andresag01:
@tariqkurd-repo : I addressed your changes. Please take another look

tariqkurd-repo (approved):

PR #204, opened 2024-04-16
Chapter 1 minor revision

marnovandermaas:
Here are some minor changes I made while I was reviewing chapter 1. I'll add follow up PRs as I make myself through the document.


arichardson (approved):
Thanks!

PR #208, opened 2024-04-19
Fix bit indices of envcfg registers

PRugg-Cap:
Removal of CME bits left it ambiguous which bit was CRE.


PRugg-Cap:
@arichardson  seems to have consciously widened the undefined region above CRE in one of the fields, which is what makes me think the intent was to leave CRE as 28.

arichardson:
The bits are arbitrary, I think the rational is that we used 28 (0x1_C_) for the exception code and using C for CHERI also makes sense here, so yes it should be 28.

PR #212, opened 2024-04-22
Introduce a Load Mutable permission

arichardson:
This permission exists in Morello and CHERIoT and is useful for sharing a
recursively read-only version of a data structure:
any capability loaded via an authority without Load Mutable permission will
have the Write and Load Mutable permissions cleared.

While it may look like this new feature limits forwarding of capability
loads, the resulting transformation can be deferred until inspection and
thus allows forwarding a single "must clear LM+W permission" bit.

Compared to CHERIoT and Morello, the 32-bit encoding format introduces one
additional restriction to allow for future extensibility: `ACPERM` clearing
`LM` also clears the `W` permission. This restriction does not apply to
the 64-bit encoding.

Co-authored-by: Lawrence Esswood <lesswood@google.com>
Co-authored-by: Andrés Amaya Garcia <andres.amaya@codasip.com>


src/cap-description.adoc
```diff
@@ -118,6 +118,11 @@ memory if the authorising capability also grants <<w_perm>>.
 [#x_perm,reftext="X-permission"]
 Execute Permission (X):: Allow instruction execution.
 
+[#lm_perm,reftext="LM-permission"]
+Load Mutable Permission (LM):: Allow loads of capabilities with <<w_perm>> via a read-only authorizing capability one.
```
tariqkurd-repo, line 122:
```suggestion
Load Mutable Permission (LM):: Allow loads of capabilities with <<w_perm>> via a read-only authorizing capability.
```

src/insns/load_32bit_cap.adoc
```diff
@@ -22,14 +22,14 @@ Encoding::
 include::wavedrom/loadcap.adoc[]
 
 Capability Mode Description::
-Load a CLEN+1 bit value from memory and writes it to `cd`. The capability in `cs1` authorizes the operation. The effective address of the memory access is obtained by adding the address of `cs1` to the sign-extended 12-bit offset. The tag value written to `cd` is 0 if the tag of the memory location loaded is 0 or `cs1` does not grant <<c_perm>>.
+Load a CLEN+1 bit value from memory and writes it to `cd`. The capability in `cs1` authorizes the operation. The effective address of the memory access is obtained by adding the address of `cs1` to the sign-extended 12-bit offset.
```
tariqkurd-repo, line 25:
```suggestion
Load a CLEN+1 bit value from memory and write it to `cd`. The capability in `cs1` authorizes the operation. The effective address of the memory access is obtained by adding the address of `cs1` to the sign-extended 12-bit offset.
```

src/insns/load_tag_perms.adoc
```diff
@@ -0,0 +1,10 @@
+Resulting value of `cd`::
+The tag value written to `cd` is 0 if the tag of the memory location loaded is
+0 or the authorizing capability (<<ddc>> or `cs1`) does not grant <<c_perm>>.
++
+If the authorizing capability does not grant <<lm_perm>>, and the tag of `cd` is 1 and `cd` is not sealed, then and implicit <<ACPERM>> clearing <<w_perm>> and <<lm_perm>> is performed to obtain the final permissions on `cd`.
```
tariqkurd-repo, line 5:
```suggestion
If the authorizing capability does not grant <<lm_perm>>, and the tag of `cd` is 1 and `cd` is not sealed, then an implicit <<ACPERM>> clearing <<w_perm>> and <<lm_perm>> is performed to obtain the final permissions on `cd`.
```

src/insns/load_tag_perms.adoc
```diff
@@ -0,0 +1,10 @@
...
+
+NOTE: Missing <<lm_perm>> does not affect untagged values since this could result in surprising bit patterns when copying non-capability data.
+Similarly, sealed capabilities are not modified as they are not directly derefenceable.
```
tariqkurd-repo, line 8:
```suggestion
Similarly, sealed capabilities are not modified as they are not directly dereferenceable.
```

src/insns/load_tag_perms.adoc
```diff
@@ -0,0 +1,10 @@
...
+
+NOTE: The implicit <<ACPERM>> does not introduce a new dependency on the loaded data since the actual masking of permissions is only needed once the loaded capability is dereferenced or the metadata bits are inspected using <<GCPERM>> or <<GCHI>>.
```
andresag01, line 10: I disagree with this statement. The ACPERM does introduce a dependency on the loaded data, but it seems that there are microarchitectural tricks that could be implemented to defer the operation a little.

andresag01 (commented):
Would it make sense to put LM in a separate extension? Perhaps together with local/global.

src/insns/load_tag_perms.adoc
tariqkurd-repo: yes - the point is that the transformation can be deferred until the data is consumed, it doens't need to be done with the load data is produced. This could be clearer.

src/cap-description.adoc
```diff
@@ -118,6 +118,11 @@ memory if the authorising capability also grants <<w_perm>>.
...
+Clearing this permission allows sharing a read-only version of e.g. a tree structure without making a copy.
+If this permission is not set, capability loads authorized by this capability will have resulting permissions as if an implicit <<ACPERM>> clearing <<w_perm>> and <<lm_perm>> was executed.
```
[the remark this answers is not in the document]
tariqkurd-repo: I think it's a useful way of describing it, it's just the same as describing the CSR updates as using SCADDR, so that the full semantics are described - however - it's not accurate because ACPERM clears the tag if the cap is selaed but in this case the cap is _not modified_ if sealed. Also ACPERM still operates if the tag=0, and this doesn't do anything.

I've suggested a change to the text above accordingly.

src/cap-description.adoc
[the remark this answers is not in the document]
arichardson: That should be removed. We could imagine a future encoding optimization that only does it for RO capabilities, but I think for now keeping orthogonal makes more sense.

LawrenceEsswood (commented):
Partially bikeshedding, partially a suggestion for subtly different semantics:

I think this should be an "ElevateWrite" rather than "LoadWritable" permission.

These are the same if a capability has both / neither permissions. The distinction is what should happen when loading via a capability that does have  "ElevateWrite" but DOES have "Write". I would like to allow loading writeable capability via such a capability (and this would make no sense if the permission was called "LoadWriteable").

The reason for this is that as we add new kinds of elevation ( "ElevateGlobal" and "ElevateStoreLocal" for instance) I would like to have a single "Elevate" bit in the encoding that sets all of these. It should be possible to create a capability that is "deeply read-only" or "deeply local" or both. To do so one could clear all the Elevate bits, but then independently control the W / Local bit. This makes it important that lacking ElevateWrite but having Write allows loading the "Write" permission.

The only real difference for the use case of making structures deeply read-only is that software needs to clear both W and this new permission.

src/cap-description.adoc
arichardson, line 122:
```suggestion
Load Mutable Permission (LM):: Allow loads of capabilities with <<w_perm>>.
```

src/cap-description.adoc
```diff
@@ -118,6 +118,11 @@ memory if the authorising capability also grants <<w_perm>>.
 [#x_perm,reftext="X-permission"]
 Execute Permission (X):: Allow instruction execution.
 
+[#lm_perm,reftext="LM-permission"]
+Load Mutable Permission (LM):: Allow loads of capabilities with <<w_perm>> via a read-only authorizing capability.
```
arichardson, line 122:
```suggestion
Load Mutable Permission (LM):: Allow loads of capabilities with <<w_perm>>.
```

src/insns/load_tag_perms.adoc
arichardson: Agreed, will try to rephrase this

andresag01:
@LawrenceEsswood:

> The distinction is what should happen when loading via a capability that does have "ElevateWrite" but DOES have "Write".

Did you mean this instead?

> via a capability that does **not** have "ElevateWrite" but DOES have "Write".

My reading of your comment is that you would like to replace the proposed LoadMutable bit with an Elevate bit similar to the C permission. i.e. Elevate is associated with Write, but it could also be associated with Local which is analogous to R+C and W+C. Is this correct?

andresag01:
@arichardson: Any thoughts on @LawrenceEsswood's proposal?

arichardson:
> @arichardson: Any thoughts on @LawrenceEsswood's proposal?

I am planning to update the PR based on that proposal.

src/img/cap-encoding-xlen64.edn
```diff
@@ -11,7 +11,7 @@
 (draw-box "Reserved"    {:span 4})
 (draw-box "SDP"         {:span 2})
 (draw-box "M"           {:span 1})
-(draw-box "AP"          {:span 3})
+(draw-box "AP"          {:span 4})
```
arichardson, line 14: TODO: adjust the column headers

arichardson (commented):
Sorry for the long delay here, updated to use "Elevate" semantics.

src/img/acperm_bit_field.edn
```diff
@@ -6,11 +6,12 @@
...
 (def right-margin 100)
 (def boxes-per-row 32)
-(draw-column-headers {:height 20 :font-size 18 :labels (reverse ["0" "1" "2" "3" "4" "" "" "" "" "" "" "" "" "" "" "" "16" "" "" "SDPLEN+15" "" "" "" "" "" "" "" "" "" "" "" "XLEN-1"])})
+(draw-column-headers {:height 20 :font-size 18 :labels (reverse ["0" "1" "2" "3" "4" "5" "" "" "" "" "" "" "" "" "" "" "16" "" "" "SDPLEN+15" "" "" "" "" "" "" "" "" "" "" "" "XLEN-1"])})
 
 (draw-box "Reserved" {:span 12})
 (draw-box "SDP"  {:span 4})
-(draw-box "Reserved" {:span 11})
+(draw-box "Reserved" {:span 10})
+(draw-box "LM"  {:span 1})
```
andresag01, line 14: Should this be EM rather than LM?

src/insns/atomic_exceptions.adoc
```diff
@@ -6,6 +6,8 @@ Requires <<r_perm>> and <<w_perm>> in the authorising capability.
 +
 If <<c_perm>> is not granted then store the memory tag as zero, and load `cd.tag` as zero.
 +
+If <<em_perm>> and <<w_perm>> are both unset and `cd.tag` is set and the `cd` is not sealed, then an implicit <<ACPERM>> clearing <<w_perm>> and <<em_perm>> is performed to obtain the final permissions on `cd` (see <<LC>>).
```
andresag01, line 9: It is worth clarifying what happens if the capability is unsealed. I saw a note about it in an instruction listing, but we should write this more clearly in places like this. My guess was that we just clear the tag of the loaded cap, which wrong.

andresag01 (changes requested):
Thank you for the changes! I think it is looking good, but we should make some clarifications.

src/insns/load_tag_perms.adoc
```diff
@@ -0,0 +1,11 @@
+Resulting value of `cd`::
+The tag value written to `cd` is 0 if the tag of the memory location loaded is
+0 or the authorizing capability (<<ddc>> or `cs1`) does not grant <<c_perm>>.
```
andresag01, line 3: This text is repeated in some of the instruction listings (e.g. LC)

src/insns/load_tag_perms.adoc
```diff
@@ -0,0 +1,11 @@
...
++
+If the authorizing capability has both <<em_perm>> and <<w_perm>> unset, and the tag of `cd` is 1 and `cd` is not sealed, then an implicit <<ACPERM>> clearing <<w_perm>> and <<em_perm>> is performed to obtain the final permissions on `cd`.
+
+NOTE: Missing <<em_perm>> does not affect untagged values since this could result in surprising bit patterns when copying non-capability data.
+Similarly, sealed capabilities are not modified as they are not directly dereferenceable.
+
+NOTE: The implicit <<ACPERM>> does not introduce a new dependency on the loaded data since the actual masking of permissions can be deferred until the loaded capability is dereferenced or the metadata bits are inspected using <<GCPERM>> or <<GCHI>>.
+This is similar to the final value of the tag bit when the authorizing capability does not have <<c_perm>>.
```
andresag01, line 11: I disagree with this statement and would remove/modify it. The key difference is that the machine blindly clears the tag bit of the loaded cap when c_perm is clear; there is no need to decode/encode anything. However, em_perm does require decoding the permissions of the loaded cap, clearing some permissions, and encoding the permissions again which is not as straight forward because there are some restrictions on what permissions can be set/cleared and the RV32 permissions encoding is not a bitmap.

src/insns/load_tag_perms.adoc
```diff
@@ -0,0 +1,11 @@
+Resulting value of `cd`::
+The tag value written to `cd` is 0 if the tag of the memory location loaded is
+0 or the authorizing capability (<<ddc>> or `cs1`) does not grant <<c_perm>>.
++
+If the authorizing capability has both <<em_perm>> and <<w_perm>> unset, and the tag of `cd` is 1 and `cd` is not sealed, then an implicit <<ACPERM>> clearing <<w_perm>> and <<em_perm>> is performed to obtain the final permissions on `cd`.
+
+NOTE: Missing <<em_perm>> does not affect untagged values since this could result in surprising bit patterns when copying non-capability data.
+Similarly, sealed capabilities are not modified as they are not directly dereferenceable.
+
+NOTE: The implicit <<ACPERM>> does not introduce a new dependency on the loaded data since the actual masking of permissions can be deferred until the loaded capability is dereferenced or the metadata bits are inspected using <<GCPERM>> or <<GCHI>>.
```
andresag01, line 10: I think it does introduce a dependency, and that might well be what some microarchitectures do for simplicity. Its just that some other microarchitectures can work around this dependency using the hinted trick here.

andresag01:
@LawrenceEsswood : Is the new proposal aligned with what you had in mind?

LawrenceEsswood (commented):
This LGTM. One note (and I know there is a discussion elsewhere on exactly this), but is it clear from this text what order tag squashing and write stripping occur in if a capability does not have C or EM? This matters as untagged capabilities do not get W stripped.

src/img/acperm_bit_field.edn
arichardson: Yes, fixed. Will change as part of #325

src/cap-description.adoc
```diff
@@ -118,6 +118,12 @@ memory if the authorising capability also grants <<w_perm>>.
 [#x_perm,reftext="X-permission"]
 Execute Permission (X):: Allow instruction execution.
 
+[#em_perm,reftext="EM-permission"]
+Elevate Mutable Permission (EM):: Allow loads of capabilities with <<w_perm>> via read-only authorizing capabilities.
+Clearing this permission and the <<w_perm>> allows sharing a read-only version of e.g. a tree structure without making a copy.
+If neither this permission nor <<w_perm>> is set and the loaded capability is tagged and not sealed, capability loads authorized by this capability will have <<w_perm>> and <<em_perm>> removed.
+The rules specified by <<ACPERM>> are followed which means that additional permissions may also be removed as a result.
```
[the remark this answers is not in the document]
arichardson: Thanks, used your suggested wording.

src/insns/load_tag_perms.adoc
arichardson: Do you mean I should factor this out?

src/insns/load_tag_perms.adoc
arichardson: Dropped this part.

src/insns/load_tag_perms.adoc
arichardson: Reworded, hopefully better now?

src/insns/load_tag_perms.adoc
arichardson: Ah I see what you meant. Fixed now.

arichardson:
@nwf @davidchisnall it would be great to get your opinion on this as well. Behaviour is not quite the same as load mutable so it would be good to confirm that all important use cases for load mutable can be covered.

andresag01 (approved):

src/cap-description.adoc
```diff
@@ -118,6 +118,12 @@ memory if the authorising capability also grants <<w_perm>>.
 [#x_perm,reftext="X-permission"]
 Execute Permission (X):: Allow instruction execution.
 
+[#em_perm,reftext="EM-permission"]
+Elevate Mutable Permission (EM):: Allow preserving the <<w_perm>> of capabilities loaded from memory when the load is authorized via a read-only capability (with <<r_perm>> but no <<w_perm>>).
+If a capability grants <<r_perm>>, but neither <<w_perm>> nor <<ew_perm>>, then a capability loaded via this authorizing capability will have <<w_perm>> and <<em_perm>> removed provided that the loaded capability has its tag set and is not sealed; loaded capabilities that are sealed or untagged do not have their permissions changed.
```
tariqkurd-repo, line 123:
```suggestion
If a capability grants <<r_perm>>, but neither <<w_perm>> nor <<em_perm>>, then a capability loaded via this authorizing capability will have <<w_perm>> and <<em_perm>> removed provided that the loaded capability has its tag set and is not sealed; loaded capabilities that are sealed or untagged do not have their permissions changed.
```

src/cap-description.adoc
```diff
@@ -118,6 +118,12 @@ memory if the authorising capability also grants <<w_perm>>.
...
+The rules specified by <<ACPERM>>> are followed when <<w_perm>> and <<em_perm>> are removed, so additional permissions may also be removed.
```
tariqkurd-repo, line 124:
```suggestion
The rules specified by <<ACPERM>> are followed when <<w_perm>> and <<em_perm>> are removed, so additional permissions may also be removed.
```

src/cap-description.adoc
```diff
@@ -155,34 +161,38 @@ Quadrant 1 encodes permissions for executable capabilities and the <<m_bit>>.
...
+| 6-7   | ✔ | ✔ |   |   | ✔ |     | Mode^1^  | Execute + Data RW
+9+| *Quadrant 2: Capability data read/write (without <<em_perm>>)*
+9+| bit[2] - write. R and C implicitly granted.
+9+| _Reserved bits for future extensions must be 1 so they are implicitly granted_
+| 0-2   8+| reserved
+| 3       | ✔ |   | ✔ |   |   |     | N/A | Data & Cap RO
+| 4-6   8+| reserved
+| 7       | ✔ | ✔ | ✔ |   |   |     | N/A | Data & Cap RW
+9+| *Quadrant 3: Capability data read/write*
+9+| [2] - write. R and C implicitly granted.
```
tariqkurd-repo, line 190:
```suggestion
9+| [2] - W. R and C implicitly granted.
```

src/cap-description.adoc
```diff
@@ -155,34 +161,38 @@ Quadrant 1 encodes permissions for executable capabilities and the <<m_bit>>.
...
-8+| *Quadrant 3: Capability data read/write*
-8+| [2] - write. R and C implicitly granted.
-8+| _Reserved bits for future extensions must be 1 so they are implicitly granted_
-| 0-2   7+| reserved
-| 3       | ✔ |   | ✔ |   |     | N/A | Data & Cap RO
-| 4-6   7+| reserved
-| 7       | ✔ | ✔ | ✔ |   |     | N/A | Data & Cap RW
+| Encoding[2:0] | R | W | C | EM | X | ASR | Mode^1^ | Notes
+9+| *Quadrant 0: Non-capability data read/write*
+9+| bit[2] - write, bit[1] - reserved (0), bit[0] - read
```
tariqkurd-repo, line 168:
```suggestion
9+| bit[2] - W, bit[1] - reserved (0), bit[0] - R
```

src/cap-description.adoc
```diff
@@ -155,34 +161,38 @@ Quadrant 1 encodes permissions for executable capabilities and the <<m_bit>>.
...
+| 6-7   | ✔ | ✔ |   |   | ✔ |     | Mode^1^  | Execute + Data RW
+9+| *Quadrant 2: Capability data read/write (without <<em_perm>>)*
+9+| bit[2] - W. R and C implicitly granted.
+9+| _Reserved bits for future extensions must be 1 so they are implicitly granted_
+| 0-2   8+| reserved
+| 3       | ✔ |   | ✔ |   |   |     | N/A | Data & Cap RO
+| 4-6   8+| reserved
+| 7       | ✔ | ✔ | ✔ |   |   |     | N/A | Data & Cap RW
+9+| *Quadrant 3: Capability data read/write*
+9+| [2] - W. R and C implicitly granted.
```
tariqkurd-repo:
```suggestion
9+| [2] - W. R, C and EM implicitly granted.
```

src/cap-description.adoc
```diff
@@ -155,34 +161,38 @@ Quadrant 1 encodes permissions for executable capabilities and the <<m_bit>>.
...
+| 6-7   | ✔ | ✔ |   |   | ✔ |     | Mode^1^  | Execute + Data RW
+9+| *Quadrant 2: Capability data read/write (without <<em_perm>>)*
+9+| bit[2] - W. R and C implicitly granted.
+9+| _Reserved bits for future extensions must be 1 so they are implicitly granted_
+| 0-2   8+| reserved
+| 3       | ✔ |   | ✔ |   |   |     | N/A | Data & Cap RO
+| 4-6   8+| reserved
+| 7       | ✔ | ✔ | ✔ |   |   |     | N/A | Data & Cap RW
+9+| *Quadrant 3: Capability data read/write*
+9+| [2] - W. R, C and EM implicitly granted.
```
tariqkurd-repo, line 190:
```suggestion
9+| bit[2] - W. R, C and EM implicitly granted.
```

src/insns/atomic_exceptions.adoc
```diff
@@ -1,11 +1,13 @@
 Permissions::
 ifdef::cap_atomic[]
-Requires the authorising capability to be tagged and not sealed.
+Requires the authorizing capability to be tagged and not sealed.
 +
 Requires <<r_perm>> and <<w_perm>> in the authorising capability.
 +
 If <<c_perm>> is not granted then store the memory tag as zero, and load `cd.tag` as zero.
 +
+If the authorizing capability has both <<w_perm>> and <<em_perm>> unset,  `cd.tag` is 1 and `cd` is not sealed, then an implicit <<ACPERM>> clearing <<w_perm>> and <<em_perm>> is performed to obtain the final permissions on `cd` (see <<LC>>)..
```
tariqkurd-repo, line 9:
```suggestion
If the authorizing capability has both <<w_perm>> and <<em_perm>> unset,  `cd.tag` is 1 and `cd` is not sealed, then an implicit <<ACPERM>> clearing <<w_perm>> and <<em_perm>> is performed to obtain the final permissions on `cd` (see <<LC>>).
```

tariqkurd-repo:
I've done some cleanup of the text (no spec changes).
Is the intention for _all_ cores to have this? So we'll need to have this on the application cores as well as embedded? Or should this be a separate optional extension?

In many respects it's easier if it's always included, but it may effect the critical path on load data.

src/insns/load_tag_perms.adoc
```diff
@@ -0,0 +1,10 @@
+Resulting value of `cd`::
+The tag value written to `cd` is 0 if the tag of the memory location loaded is
+0 or the authorizing capability (<<ddc>> or `cs1`) does not grant <<c_perm>>.
++
+If the authorizing capability has both <<em_perm>> and <<w_perm>> unset, and the tag of `cd` is 1 and `cd` is not sealed, then an implicit <<ACPERM>> clearing <<w_perm>> and <<em_perm>> is performed to obtain the final permissions on `cd`.
```
tariqkurd-repo, line 5:
```suggestion
If the authorizing capability has both <<em_perm>> and <<w_perm>> unset, `cd.tag` is 1 and `cd` is not sealed, then an implicit <<ACPERM>> clearing <<w_perm>> and <<em_perm>> is performed to obtain the final permissions on `cd`.
```

nwf-msr:
Assuming I understand the proposal correctly, the `W` and `EM` bits now meaningfully describe three states for `C`-bearing capabilities:
1. `!W !EM`: a read-only view of transitively read-only views
2. `!W EM`: a read-only view of possibly read-write views (in C terms, a pointer to a constant structure containing pointers to non-constant data)
3. `W ?EM`: a read-write view of possibly read-write views; `EM` does not influence the authority of this capability, but...

Because `EM` is a separate architectural bit from `W`, and, AFAICT, `W` does not architecturally imply `EM` [0], it is possible (up to encoding permission compression) to construct a `W !EM` capability, which is a strange thing to have: the rights to mutate (transitively reachable) memory, which may be attenuated to be a deeply read-only view but which may not be attenuated to the shallow-immutable `!W EM` form.  I'm wary of allowing such strange things to come to exist in the system.

[0] We've never, I think, had that kind of "upwards" implication, where having a permission implies having -- and being unable to clear -- another.  Big-CHERI had the full permission powerset lattice and all of CHERIoT's implications (AFAIR; I may be wrong) are "downwards", in that clearing one permission implies clearing others, or are exclusivity of permissions encoded in the different roots.

rmn30:
I think the idea of a single Elevate bit combining what CHERIoT calls load global and load mutable is interesting. @nwf-msr 's concern about `W & !E` could probably be addressed in the permission encoding by forbidding that case and recovering a point in the encoding space for something else. 

We should also address the question of why `!W & E` is useful, i.e. why not make all read-only capabilities deeply immutable then we might not even need the E permission? On CHERIoT this is currently necessary to allow constant pointers to mutable data to be stored in the read-only data section accessible via PCC (which is read only). For example:

```c
int i; // goes in compartment rw global section
int* const p = &i; // goes in compartment relro section accessed via PCC
```

If `p` had to go in the rw globals section we'd lose enforcement of the `const`. I'm not sure how critical this is. There may be other uses for read-only capabilities to writable capabilities but I'm not aware of them.

I am still thinking about what happens when we add global to the mix. There are currently supported cases in CHERIoT that can't be supported by a single elevate bit so I'm trying work out how important they are.

davidchisnall:
> There may be other uses for read-only capabilities to writable capabilities but I'm not aware of them.

Read-only pointers to mutable data is quite a common idiom.  For example, a read-only struct that contains a pointer to a buffer that you can write into is a common way of writing request-response protocols.

nwf-msr:
Having thought about this further, I think the way I'd phrase this suggestion is "the Elevate permission avoids ANDing some permissions of the authority into the load result".  That is, generalizing to a world where we had `E`-levate apply to both `W`-rite and `G`-lobal, with an `auth`-ority, a `load`-ed cap from memory, and the `res`-ult cap to be placed in a register, assuming the `res`-ult is tagged, we'd have:

```
res.E = load.E & (load.is_sealed | auth.E)
res.W = load.W & (load.is_sealed | auth.E | auth.W)
res.G = load.G & (                 auth.E | auth.G)
```

By contrast, CHERIoT, with its two recursive permissions, under the same assumptions, looks like:

```
res.LM = load.LM & (load.is_sealed | auth.LM)
res.W  = load.W  & (load.is_sealed | auth.LM)
res.LG = load.LG & (load.is_sealed | auth.LG)
res.G  = load.G  & (                 auth.LG)
```

(`G` is not mediated by `is_sealed` on CHERIoT and I've taken the liberty of assuming the same for Zcheri, above; see https://github.com/microsoft/cheriot-sail/issues/14.  I'm also ignoring some subtleties of preserving untagged capability-width bit patterns; see https://github.com/microsoft/cheriot-sail/issues/13)

David's example,
> For example, a read-only struct that contains a pointer to a buffer that you can write into is a common way of writing request-response protocols.
would manifest, using CHERIoT terminology, as a `!G !LG !W LM` capability: a deep-no-capture (`!G !LG`), shallow-constant (`!W`) reference to a writable pointer (`LM`), or as `G !LG !W LM`, should one wish to permit shallow capture.  Neither of these can be expressed with the one `E` bit.

I continue to worry that the `G W !E` form (and friends) are strange things to be able to express, since, absent special handling, they do not strip `W` or `G` when used as the load authority, and yet cannot be attenuated except to forms that would strip bits when so used.

Separately, and not to bike-shed too much, I don't particularly like the name "Elevate" for this and might suggest "Preserve" instead.  Note that some permissions seemingly must be exempted from Elevate/Preserve processing, notably `X`, since we wish to have non-executable pointers to function pointers, and processing `R` is not particularly meaningful, since `auth` must have `R` set for us to even be here in the first place.

Thoughts?

LawrenceEsswood:
> Note that some permissions seemingly must be exempted from Elevate/Preserve processing, notably X, since we wish to have non-executable pointers to function pointers

It would seem to me that there are three permissions that concern data flow: Write, Global, and StoreLocal[0]. I think that we want deep properties to be describable for these. Which is why no LoadX (or LoadR, or LoadAccessSystemReg, for that matter).

Two of these have already had a deep property bit described in prior art: Write and Global, paired with LoadMutable and LoadGlobal. Notably, I would argue LoadStoreLocal is the third missing deep property. The "Deep you can't stash local things here". The lack of this makes no difference where the stack / switcher is just a flat buffer. But, in general, this is the last piece of the control flow policy if I wish to share a copy of, say, a linked list that could otherwise StoreLocal capabilities, but have a non-stashable copy to delegate so I don't have to search it when doing revocation.

To help understand this proposal, I think maybe it might help to untangle two distinct choices that were made here: to move from having LoadMutable/LoadGlobal/LoadStoreLocal to ElevateMutable/ElevateGlobal/ElevateStoreLocal, and then to merge ElevateMutable/ElevateGlobal/ElevateStoreLocal into a single E bit in the encoding scheme (so you can have all or nothing).

The first change is not actually the thing I think people are having trouble with in the examples above, even though it is a bit weird, I agree, that you can describe these R/W structures that become read-only after the first read-only edge they hit. The difference between LoadQ and ElevateQ is that LoadQ unequivocally allows you to gain Q via a capability, whilst ElavateQ allows you to gain a Q >= the Q of the authorising capability.

On the difference between what they can describe, to use LoadMutable vs ElevateMutable as an example: both describe the _exact_ same shallow properties when the bit is set. On the other hand, if the bits are clear, Elevate struggles to describe the policy that the first edge is writable, but then all others from there on are deeply immutable (sort of opposite to David's). I don't imagine a huge call for it, but, such a policy can easily be described (if desired) with a second capability to the root of this structure with the different policy for the root but without the C permission. The "Why" for this change is twofold: first, while the difference is not so big for single bit permissions, for N bit permissions (such as the N bit generalisation of local/global) the ">" relation of elevate becomes a very efficient single-bit way to describe what level a structure should deeply be loaded at without needing another N bits to describe this (because this level is just level of the root pointer). The second is to facilitate another change...

...The second change is to merge all the Elevate bits into one. Purely to save bits. Merging all LoadX bits into one would not really be feasible, because we want Read/Write structures with properties like deep non-capture. And we would want capturable objects that  are deeply read-only. However, I would argue, that merging all the Elevate permissions into one _is_ feasible. It allows R/W structures that are deeply non-capture by having the W bit, and capturable objects that are deeply read-only by setting the G bit.

I agree read-only capabilities to read/write capabilities are important to consider.  The example Wes gives actually highlights a ramification of the _second_ change, not the first, I would argue. First, had we wanted the object without the deep non-capture part, then it would have been `LG !W LM` in CheriOT lingo, or `LG & !W & E` in this proposal. Both ElevateMutable/LoadMutable have identical behaviour when they are equal to 1.  The issue arises in the examples because ElevateMutable has been conflated with ElevateGlobal and we wanted a deeply non-capture object,which forced a deep property w.r.t. read/write, and we also wanted the first edge to be read-only.

Now, I don't think this is a big problem, and I think in a broader sense this problem also existed before but possibly didn't crop up as much. As to why it existed before, I think there is a similar "opt out" problem with just mutability alone, forgetting capture for a moment. The read-only capability to writable capability works fine if passed as an argument and I can elect to make it LG or not. But what if it is contained deep somewhere within another structure that is otherwise made transitively read-only in a way I don't control? I think programmers already have to think about whether their construct works if viewed under a transitive read-only graph. If they desire some edges to be writable _anyway_, even if reached by transitive read-only, then they need some way to do so. The answer is typically using sealing as such "opt out" to the deep property (although second capability to structure can also be an option in some cases). 

In this proposal, if a programmer wishes to make read-only objects that can load writable ones (and so is explicitly interacting with CHERI, this is not legacy C as casting through const does not imply CHERI read only), _even as a part of a structure that has some deep policy set_, then they need to seal subsequent edges. They can always make the unsealing capability freely available if any code is allowed to do so. They already needed to make this decision on CheriOT w.rt. to their structure being a part of an otherwise deeply read-only one. The new added complexity is that they also hit this case when their structure is shared as a part of a deeply local one as well, because there is no independent way of turning deep policies on/off wr.t. to Local and Write.

Writing this has made me wonder, purely as an optimisation if needing sealing in this case would ever be a codesize / performance concern, whether or not we want a dedicated type that behaves exactly like unsealed apart from being considered sealed for the purpose of these squashes.

And finally, to bike shedding (this got too long, sorry): I would also be happy with something like "amplify". But, in my mind at least, this is the permission that allows gaining a more powerful privilege than the capability used to gain it (with respect to a few choice permissions). I think this is just mentally defaulting in a different direction: that one should not gain power unless one has special permission to, as opposed to "Preserve" which implies you were due your authority, unless denied.

[0] well, arguably R and C too. But there is no point discussing deep properties for these as lacking either terminates the structure there anyway.

nwf-msr:
> [...] LoadStoreLocal is the third missing deep property. [...]

That could be interesting, but A) I'm not entirely convinced, since a `!LoadStoreLocal`-mediated view would prevent, for example, rearrangement of an existing graph structure containing local edges, and B) surely that's a separate proposal we can leave for later.

> To help understand this proposal, I think maybe it might help to untangle [...]

Yes, decoupling those is a good idea.

> The first change is not actually the thing I think people are having trouble with in the examples above, even though it is a bit weird, I agree, that you can describe these R/W structures that become read-only after the first read-only edge they hit. The difference between LoadQ and ElevateQ is that LoadQ unequivocally allows you to gain Q via a capability, whilst ElavateQ allows you to gain a Q >= the Q of the authorising capability.

Neither of these claims are quite accurate: an `ElevateQ`-bearing capability does not always load `Q`-bearing capabilities; this claim glosses over the load target *actually bearing Q*.  See below for more pedagogy.

> On the difference between what they can describe, to use LoadMutable vs ElevateMutable as an example: both describe the _exact_ same shallow properties when the bit is set.

Right, because when either is set, we're back in a baseline capability system without view mediation.

> On the other hand, if the bits are clear, Elevate struggles to describe the policy that the first edge is writable, but then all others from there on are deeply immutable (sort of opposite to David's).

That's true and problematic, but at least my real objection to `Q/ElevateQ` is that `Q/LoadQ` has, IMHO, completely sensible semantics for all four settings:

* `!Q !LQ` -- deeply not-`Q`; attenuates only to itself
* `!Q LQ` -- shallowly not-`Q` but `Q`-preserving thereafter; attenuates to itself or `!Q !LQ`
* `Q !LQ` -- shallowly `Q` but not-`Q` thereafter; attenuates to itself or `!Q !LQ`
* `Q LQ` -- `Q` and deeply `Q`-preserving; attenuates to any.

while `Q/ElevateQ` has sensible semantics for only three:

* `!Q !EQ` -- deeply not `Q`; attenuates only to itself
* `!Q EQ` -- shallowly not-`Q`, but `Q`-preserving thereafter; attenuates to itself or `!Q !EQ`
* `Q EQ` -- `Q` and deeply `Q`-preserving; attenuates to any.

The odd one out is its `Q !EQ` state, which behaves as "`Q` and deeply `Q`-preserving", that is, identically to `Q EQ`, and yet attenuates only to itself or `!Q !EQ`; that is, it cannot be attenuated to `!Q EQ`.  We could, like KeyKOS (see far below) take the easy way out and enforce that `Q -> EQ`, but that has consequences for the merged-`E` proposal (see slightly closer below).

> I don't imagine a huge call for it, but, such a policy can easily be described (if desired) with a second capability to the root of this structure with the different policy for the root but without the C permission.

I do not believe that is correct.  A `C W !LM` capability authorizes the storing of tagged capabilities within, while a `C !W` * `!C W` pair would not -- the first cannot store at all, and the second cannot store set tags.

> The "Why" for this change is twofold: first, while the difference is not so big for single bit permissions, for N bit permissions (such as the N bit generalisation of local/global) the ">" relation of elevate [...]

This feels like a separable consideration.  Despite my use of `G` in the above examples -- yes, to argue against the "merged E" future proposal below -- AFAIK Zcheri is currently not proposing anything that looks like the `G`/`SLC` mechanism.  I'm going to be good and avoid dumping my nerd-sniped thoughts here.

> [...] merging all the Elevate permissions into one _is_ feasible. [...]

At the cost of expressiveness (the loss of the 4th set of semantics) and a more complicated attenuation story (since I don't think we would want to take the easy way out, enforcing `!E -> (!W & !G & !...)` ?).

> [...]

> And finally, to bike shedding (this got too long, sorry): I would also be happy with something like "amplify". But, in my mind at least, this is the permission that allows gaining a more powerful privilege than the capability used to gain it (with respect to a few choice permissions). I think this is just mentally defaulting in a different direction: that one should not gain power unless one has special permission to, as opposed to "Preserve" which implies you were due your authority, unless denied.

"Amplify" specifically has precedent in capability systems going back to HYDRA/C.mmp, wherein it means "Given a capability, _c_, of a type for which I am the type owner, derive a capability holding possibly more permissions than _c_."  While there is no direct analogy in CHERI, it's fair to think of this as a kind of unseal operation -- _c_ is an attenuated reference to an object of a particular type and amplification combines _c_ and a capability witnessing type ownership to generate a (most likely) higher-privileged capability.

FWIW and AFAICT, historically, capability systems always meant "load" as "load whatever's there".  KeyKOS "sensory keys" were the first to do otherwise, baking an attenuating membrane into the capability system's primitives rather than implementing one in trusted software.  (These were, in the above notation, `!W !LM` capabilities, and the system could not express `W !LM`, neatly side-stepping the central issue here.  EROS had some odd behavior for its writable weak keys, but Coyotos dropped it for the KeyKOS design again.)  I believe CHERI's "LC/SC", CHERIoT's "MC", and Zcheri's "C" permissions h  I would like our nomenclature to reflect that recursive permissions are membranous attenuation; I think "Load" conveys that just fine, but "Elevate" and "Attenuate" do not.

Whew.

LawrenceEsswood:
> Neither of these claims are quite accurate: an ElevateQ-bearing capability does not always load Q-bearing capabilities; this claim glosses over the load target actually bearing Q. See below for more pedagogy.

You knew what I meant 😂. Counter pedantry: does "unequivocally allows you to gain" actually mean "always loads"? Just because something is _allowed_ doesn't mean you can actually always do it, just that you are not being stopped by that mechanism. I, for example, unequivocally allow you to fly to the moon.

>  surely that's a separate proposal we can leave for later.

I think there is certainly a lot of the "why" here that is about things we have no intention of specifying (yet) and so might be seen as premature. The issue is that once the bits are spent and the spec set in concrete, I think it's going to be very awkward to claw back. I am not convinced that future CHERI extensions should just respecify the format when all other resources are exhausted. They would break all the CHERI software that will (hopefully!) be written to run on hardware conforming to this spec and be out there in the wild. The number of bits we have really is swiftly running out, and `LG` rather than `EG` probably blocks us from ever having more local/global than 1-bit. And if we do end up adding `EG` later on, we can't ever merge it with `LW` the way we could with `EW`.  `LoadStoreLocal` would actually cost no extra bits if added in a future extension with merged `E`s. Not so if we needed an extra `LSL`. Even the single bit that this currently saves is very valuable IMO. The thing is, if it turns out that merging `EG` and `EW` in representation micro-architecturally is a bad idea, that actually is possible to take back without breaking software (because we could expose them architecturally as two separate bits, and if suddenly more combinations become possible this is a superset of what is supported before).

> I do not believe that is correct. A `C W !LM` capability authorizes the storing of tagged capabilities within, while a `C !W` * `!C W` pair would not -- the first cannot store at all, and the second cannot store set tags.

You are correct there. The pair should have ben `C W !R` (can write capabilities but not read anything) and `C !W R EW` (cannot write and is deeply non-write). Although I will note that we have very few `W !R` combinations available, but I digress.

> that is, it cannot be attenuated to !Q EQ

Yeah, that would need sealing the !Q EQ capability and then unsealing it immediately on load. Again, if this really is the only sticking point (and it would be a sticking point only for performance IMO), we could dedicate a type to that behaviour.

> At the cost of expressiveness (the loss of the 4th set of semantics) and a more complicated attenuation story (since I don't think we would want to take the easy way out, enforcing !E -> (!W & !G & !...) ?).

Certainly at that cost, but no such thing as free lunch if we want to leave ourselves open to more expressiveness in the future! Is this 4th set of semantics really worth locking out all the innovation a bit could allow in the future?

More bikeshedding: what about deep vs shallow? With the bit=1 the privilege of the capability only applies to one level, and with bit=0 it is a deep property, so this could also be called the shallow bit.

PRugg-Cap:
Trying to revive this to make progress and summarise pros and cons (likely angering everyone involved, so sorry about that!):
1. Elevate forces you to conflate deepness between e.g. "mutability" and "localness". There may be cases where CHERIoT wants one but not the other, i.e. you might want "deep" immutability but only "shallow" localness, or vice versa. (- for Elevate)
2. Elevate saves encoding bits by sharing the deepness across the different permission bits. (+ for Elevate)
3. "permitLoadMutable" and friends allow differentiating the first edge from the deeper accesses, whereas Elevate only allows, for example, non permitLoadMutable if the reference is itself not writable. (- for Elevate)
4. Elevate allows encoding that some properties are "deep" without actually applying it to any properties, which may be confusing. (- for Elevate)
5. Elevate generalises to non-bit fields, like a numerical localness value. (+ for Elevate)

Seems like there are 3 plausible options for how to proceed:
- The current proposal: make Elevate the exposed architectural bit and encode it that way. If (4) is a problem, then don't have a mapping in the compressed format for that combination. This would preclude CHERIoT doing its current behaviour "underneath", so only go down this route if it turns out (1) and (3) are always workaround-able.
- The compromise: make "permitLoadMutable" etc the exposed format, but encode it as an Elevate bit. That keeps the advantage (2). (5) can be addressed with something more elevate-like in the exposed format when needed, without adding any new encoding bits. This is for people who want to enable CHERIoT as it is, but aren't convinced (1) and (3) are problems for less specialised applications.
- Full CHERIoT: Encode and expose the bits individually, thinking carefully about which combinations make sense.

Please do correct if anything here isn't quite right. It sounds like a lot of people are trying to follow along!

LawrenceEsswood:
Thanks Pete for driving this along.

I think (1) is only a problem if there is only a single architectural E bit. This proposal is, for now, just an EM. If (when) we introduce local/global we should have a second ElevateLoadGlobal architecturally (but, yes, with a choice for the base encoding to map them to the same bit). CheriOT would be free to have separate encoding bits, if the encoding is (I think we are leaning this way), seperate from the architectural view.

(4) It's similar to how PermitLoadGlobal makes no sense if you don't have C. Architecturally you could always express redundant things, but I don't think that's a problem, per se, just an artefact of making the architectural view as orthogonal as possible.

(3) Is, in my mind, the true downside to this architectural view. Even with separate E bits, the semantics are just slightly different.

Might I suggest that your first option should be: expose multiple architectural E bits, encoding them as one in the base encoding. (4) happens to not be allowed by this encoding. (1) is a downside of the base encoding. CheriOT is allowed to choose a different encoding if (1) is an issue. (3) would still be an issue.

On the second option: LX == EX || X, (X being any permissions here). So, we certainly can expose an LX architecturally even if EX is underneath. However, EX still "leaks". An LX capability loads a given capability differently based on whether it was LX because of the EX or the X. I think it is therefore more natural to expose EX. Programmers can still write `(getperm(Cap) & (EX | X)) != 0` if they wish to infer LX.

nwf:
Since I fear that I have already belabored my point(s), I will endeavor to be brief.

- I think https://github.com/riscv/riscv-cheri/pull/212#issuecomment-2271422064 omitted summarizing the bizarre attenuation of Elevate, unless that's what was meant by (4).  In any case, this would seem to strongly separate Load* and Elevate* as architectural approaches, not just matters of encoding.

- On point (5), specifically, there's no reason to believe that Load* wouldn't generalize beyond bits.  Booleans are just a particular lattice, and Load* uses only lattice operations (well, and a type pun).  That is, we can readily generalize `&` to `min` and `load.is_sealed | x` to `if load.is_sealed then Top else x` (making the pun explicit) in the formulae above, to give one example.

LawrenceEsswood:
@nwf On (5), the statement is not that LQ cannot be generalised for an N bit Q permission (both LQ and EQ make andperm into minperm as you suggest), just that LQ requires N bits, where as EQ is always just one (borrowing the N bits from the Q permission to select precisely what value should be min-ed with).

For example, if you N bit local/global you could make your structure deeply non-capture at level Y just by setting local to Y, and stripping E.

nwf-msr:
Fair enough.  Having absolutely no experience to draw on, I can't say that I object to using an Elevate perspective for such multi-level classification schemes.  The bizarre attenuation properties are still there, but maybe they're less bizarre in that domain in practice, I dunno.  Restricting to the 1-bit classification scheme that is Local/Global, it'd manifest as "here's a global capability that can load globals but can't be attenuated to a local capability that can load globals", which still makes me nervous.  On the other side, I'm not sure that we have any examples of the things EG couldn't express -- `G !LG MC R` -- in the system so far; @davidchisnall?

I continue to think that LoadMutable is the more expressive and more worked (CHERIoT, Morello) approach specifically for its domain.  I do not feel compelled by the possibility of fusing multiple Elevate bits together in capability encodings, and so don't think that overrides my preference, since I think we will want to vary those independently, especially as the constituent dimensions become more elaborate (like multi-level classification), but that is, admittedly, just a guess.

arichardson:
I had a discussion with @nwf and @LawrenceEsswood last week and have updated the PR based on this discussion.

The latest version is back to `Load Mutable` as the architectural permission (matching Morello and CHERIoT). The one difference is that the 32-bit encoding scheme effectively encodes this as an "elevate" bit to allow efficiently encoding the future Local/Global extension.

Essentially, the only difference compared to Morello/CHERIoT is that ACPERM clearing LM will also clear the W permission. RV32 cannot represent a data structure where the first level is capability W+!LM, but any later level is !W&!LM (i.e. you can store caps to the root buffer but not load writable ones, and any level of indirection is deeply read-only). Instead the entire data structure will have to be deeply read-only.

Since CHERIoT only relaxes this encoding-specific ACPERM restriction (and matches RV64), the CHERIoT LM is fully compatible with this specification.

@davidchisnall @nwf: I believe this pull request should address another one of the few outstanding incompatibilities.

davidchisnall:
Sounds good to me, thank you!

nwf-msr (approved):
LGTM

src/insns/acperm_32bit.adoc
```diff
@@ -51,13 +51,16 @@ The common rules are:
...
+. <<lm_perm>> cannot be set without <<c_perm>> being set
+.. Clear <<lm_perm>> unless <<c_perm>> is set.
 
 NOTE: The combination of <<x_perm>> clear and <<m_bit>> set is reserved for future extensions.
 
 The MXLEN=32 additional rules are:
 
 . Clear <<asr_perm>> unless _all_ other permissions are set
 . Clear <<c_perm>> and <<x_perm>> if <<r_perm>> is not set
+. Clear <<lm_perm>> unless both <<c_perm>> and <<w_perm>> are set.
```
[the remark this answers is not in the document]
arichardson: Ah yes, this was meant to be clear LM unless c is set.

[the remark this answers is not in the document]
arichardson: Force-pushed a new commit that fixes this

arichardson:
Thanks for the review, will merge this once @andresag01 / @tariqkurd-repo / @jrtc27 have also taken a look.

src/cap-description.adoc
```diff
@@ -118,6 +118,12 @@ memory if the authorising capability also grants <<w_perm>>.
 [#x_perm,reftext="X-permission"]
 Execute Permission (X):: Allow instruction execution.
 
+[#lm_perm,reftext="LM-permission"]
+Load Mutable Permission (LM):: Allow preserving the <<w_perm>> of capabilities loaded from memory.
+If a capability grants <<r_perm>> and <<c_perm>>, but no <<lm_perm>>, then a capability loaded via this authorizing capability will have <<w_perm>> and <<lm_perm>> removed provided that the loaded capability has its tag set and is not sealed; loaded capabilities that are sealed or untagged do not have their permissions changed.
+The rules specified by <<ACPERM>>> are followed when <<w_perm>> and <<lm_perm>> are removed, so additional permissions may also be removed.
```
andresag01, line 124:
```suggestion
The rules specified by <<ACPERM>> are followed when <<w_perm>> and <<lm_perm>> are removed, so additional permissions may also be removed.
```

src/insns/atomic_exceptions.adoc
```diff
@@ -1,11 +1,13 @@
 Permissions::
 ifdef::cap_atomic[]
-Requires the authorising capability to be tagged and not sealed.
+Requires the authorizing capability to be tagged and not sealed.
 +
 Requires <<r_perm>> and <<w_perm>> in the authorising capability.
 +
 If <<c_perm>> is not granted then store the memory tag as zero, and load `cd.tag` as zero.
 +
+If the authorizing capability does not grant <<lm_perm>>, and the tag of `cd` is 1 and `cd` is not sealed, then an implicit <<ACPERM>> clearing <<w_perm>> and <<lm_perm>> is performed to obtain the final permissions on `cd` (see <<LC>>)..
```
andresag01, line 9:
```suggestion
If the authorizing capability does not grant <<lm_perm>>, and the tag of `cd` is 1 and `cd` is not sealed, then an implicit <<ACPERM>> clearing <<w_perm>> and <<lm_perm>> is performed to obtain the final permissions on `cd` (see <<LC>>).
```

src/insns/atomic_exceptions.adoc
```diff
@@ -1,11 +1,13 @@
...
++
 (_This tag clearing behaviour may become a data dependent exception in future._)
```
andresag01, line 11: The _this_ in this sentence is a reference to:
> If <<c_perm>> is not granted then store the memory tag as zero, and load `cd.tag` as zero.

So the sentence should really move up, or even better, remove it altogether! It sounds like its still unclear what the behaviour must be here.

andresag01, line 11:
```suggestion
```

src/insns/load_tag_perms.adoc
```diff
@@ -0,0 +1,10 @@
+Resulting value of `cd`::
+The tag value written to `cd` is 0 if the tag of the memory location loaded is
+0 or the authorizing capability (<<ddc>> or `cs1`) does not grant <<c_perm>>.
++
+If the authorizing capability does not grant <<lm_perm>>, and the tag of `cd` is 1 and `cd` is not sealed, then an implicit <<ACPERM>> clearing <<w_perm>> and <<lm_perm>> is performed to obtain the final permissions on `cd`
```
andresag01, line 5:
```suggestion
If the authorizing capability does not grant <<lm_perm>>, and the tag of `cd` is 1 and `cd` is not sealed, then an implicit <<ACPERM>> clearing <<w_perm>> and <<lm_perm>> is performed to obtain the final permissions on `cd`.
```

andresag01 (approved):
Only found a few typos, but otherwise good to go.

tariqkurd-repo (approved):

src/insns/atomic_exceptions.adoc
tariqkurd-repo: I agree with Andres - what's the motivation of keeping the option of an exception?

tariqkurd-repo: I think the highlighted sentence reads ok.....

arichardson: Agreed, this dates back to the initial commit. Will remove this in a separate change.

Issue #213, opened 2024-04-22
CMO exceptions don't look quite right


Timmmm:
The CMO/prefetch exceptions are:

```
ifdef::cbo_clean_flush[]
| Permission violation  | It does not grant <<w_perm>> and <<r_perm>>
| Length violation      | At least one byte accessed is within the bounds
endif::cbo_clean_flush[]

ifdef::cbo_inval[]
| Permission violation  | It does not grant <<w_perm>>, <<r_perm>> or <<asr_perm>>
| Length violation      | At least one byte accessed is outside the bounds
endif::[]

ifdef::prefetch_i[]
| Permission violation  | It does not grant <<x_perm>>
| Length violation      | At least one byte accessed is within the bounds
endif::[]

ifdef::prefetch_r[]
| Permission violation  | It does not grant <<r_perm>>
| Length violation      | At least one byte accessed is within the bounds
endif::[]

ifdef::prefetch_w[]
| Permission violation  | It does not grant <<w_perm>>
| Length violation      | At least one byte accessed is within the bounds
endif::[]
```

1. The prefetch instructions shouldn't ever raise exceptions because they're just hints. The (non-cheri) spec says:

> A cache-block prefetch instruction is permitted to access the specified cache block whenever a load instruction,
store instruction, or instruction fetch is permitted to access the corresponding physical addresses. **If access to the
cache block is not permitted, a cache-block prefetch instruction does not raise any exceptions** and shall not
access any caches or memory

This makes sense because they're just `or.i` instructions semantically. In fact do they even need to be mentioned in this spec at all?

2. Shouldn't `cbo.clean` and `cbo.inval` have the same `Length violation`? Pretty sure they should both be `At least one byte accessed is outside the bounds`.

3. The CBO spec actually says you only need read *OR* write permission, not both:

> A cache-block management instruction is permitted to access the specified cache block whenever a load
instruction or store instruction is permitted to access the corresponding physical addresses. If neither a load
instruction nor store instruction is permitted to access the physical addresses, but an instruction fetch is
permitted to access the physical addresses, whether a cache-block management instruction is permitted to access
the cache block is UNSPECIFIED. (yeay)

jrtc27:
1. Yes because it matters for microarchitectural side channels
2. No. Clean is safe, inval is not, because the latter throws away data in the cache line, acting sort of like a write to the whole line with whatever possibly stale data is in DRAM.
[...]

PR #221, opened 2024-04-26
Add note that INVAL is dangerous and should be FLUSH

tariqkurd-repo:
and fix the bounds check text for INVAL

### DIFF
--- a/src/insns/cbo_exceptions.adoc
+++ b/src/insns/cbo_exceptions.adoc
@@ -9,6 +9,11 @@ CBO.INVAL performs cache block flushes instead of
 invalidations for less privileged modes. The instruction checks shown in the
 table below remain unchanged regardless of the value of CBIE and the privilege
 mode.
+
+NOTE: A valid implementation of CBO.INVAL is to _always_ execute it as CBO.FLUSH.
+ This is recommended for CHERI systems as invalidate _could_ reveal stale capabilities
+ in memory which the current execution environment should not have access to.
+
 endif::[]
 
 [%autowidth,options=header,align=center]
@@ -19,13 +24,12 @@ endif::[]
 
 ifdef::cbo_clean_flush[]
 | Permission violation  | It does not grant <<w_perm>> and <<r_perm>>
-| Length violation      | None of the bytes accessed are within the bounds
 endif::cbo_clean_flush[]
 
 ifdef::cbo_inval[]
 | Permission violation  | It does not grant <<w_perm>>, <<r_perm>> or <<asr_perm>>
-| Length violation      | At least one byte accessed is outside the bounds
 endif::[]
+| Length violation      | None of the bytes accessed are within the bounds
 |==============================================================================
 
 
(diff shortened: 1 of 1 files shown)


tariqkurd-repo:
thsi is intended to fix https://github.com/riscv/riscv-cheri/issues/213

src/insns/cbo_exceptions.adoc
```diff
@@ -9,6 +9,11 @@ CBO.INVAL performs cache block flushes instead of
 invalidations for less privileged modes. The instruction checks shown in the
 table below remain unchanged regardless of the value of CBIE and the privilege
 mode.
+
+NOTE: A valid implementation of CBO.INVAL is to _always_ execute it as CBO.FLUSH.
+ This is recommended for CHERI systems as invalidate _could_ reveal stale capabilities
```
jrtc27, line 14: I don't know if we want to go that far. Invalidate still leaks secrets in non-CHERI. I'd just want to highlight the potential to reveal capabilities from past memory contents that you wouldn't otherwise have access to as justification for why it needs ASR + write perms. INVAL is already FLUSH for userspace normally, per the CSR field that lets you enforce that.

tariqkurd-repo: ok - so do you think we need this note at all? Or feel free to suggest a modification which weakens it.

jrtc27: How about:

```NOTE: Invalidating a cache block can re-expose capabilities previously stored to it after the most recent flush, not just secret values. As such, CBO.INVAL has stricter checks on its use than CBO.FLUSH, and should only be made available to, and used by, sufficiently-trusted software. Untrusted software should use CBO.FLUSH instead.```

File: src/insns/cbo_exceptions.adoc [hunk 2]
```diff
@@ -19,13 +24,12 @@ endif::[]
...
 ifdef::cbo_clean_flush[]
 | Permission violation  | It does not grant <<w_perm>> and <<r_perm>>
-| Length violation      | None of the bytes accessed are within the bounds
 endif::cbo_clean_flush[]
 
 ifdef::cbo_inval[]
 | Permission violation  | It does not grant <<w_perm>>, <<r_perm>> or <<asr_perm>>
-| Length violation      | At least one byte accessed is outside the bounds
 endif::[]
+| Length violation      | None of the bytes accessed are within the bounds
```
arichardson, line 32: I think inval should still have the same check as zero, i.e. full cache line. Personally, I'd like it to require almighty but since I can make it behave as flush if I control the software stack that may not be necessary.

src/insns/cbo_exceptions.adoc
arichardson: I like that suggested wording.

src/insns/cbo_exceptions.adoc
andresag01: Agreed, it would be best to require that all bytes are within bounds for CBO.INVAL just as we do for CBO.ZERO.

arichardson: This was not resolved?

jrtc27: Ping. This is still wrong.

jrtc27: https://github.com/riscv/riscv-cheri/pull/607

Issue #218, opened 2024-04-25
Clarification on CRE's effect on CSR accesses


jamie-melling:
I think that section 5.7. Disable CHERI Registers needs more information regarding its effects on CSR accesses.  All that is mentioned is that:

>Executing CSR instructions accessing any capability wide CSR addresses (Section 3.6) cause illegal instruction exceptions

There is no mention of DDC, MTDC, STDC.
Furthermore, without mention of those registers it leaves this question: if mseccfg.CRE is 0, does this mean any accesses to the XLEN address field of mtvec, mepc, mscratch cause illegal instruction exceptions?

The statement in the spec does not distinguish between RISC-V CSR extended by Zpurecap and the new CHERI CSRs that Zcherilegacy adds. (Maybe there is no difference, but you cannot tell)

andresag01:
> There is no mention of DDC, MTDC, STDC.

These are all CLEN CSRs, so they would would trap when CHERI is disabled.

> if mseccfg.CRE is 0, does this mean any accesses to the XLEN address field of mtvec, mepc, mscratch cause illegal instruction exceptions?

They would not trap. Recall that we are trying to maximize compatibility with non-CHERI RISC-V, so the regular XLEN-sized CSRs must remain accessible as usual

jamie-melling:
The mode is forced to legacy when CHERI is disabled, so you will never be able to do a CLEN operation (only XLEN) on the CSRs mentioned in 3.6, so I don't think it should refer to section 3.6 then.  It should instead just mention DDC, MTDC and STDC.

PRugg-Cap:
@jamie-melling Agreed: I don't think this was updated with the new "CLEN and XLEN CSRs alias and are mode-dependent behaviour". Therefore, any existing RISC-V CSR is safe to access, as mode is hardwired to legacy so you'll only be accessing the address, and we need to be compatible, as @andresag01 said.

The question is whether the extra capability CSRs without an existing RISC-V version can have their addresses accessed with CRE=0. It probably doesn't matter too much either way: if they've been subset (i.e. some code ran to restrict their bounds while CRE=1, then CRE was turned back to 0) then that could allow code to clear the tag by making them unrepresentable, but that goes for the ones with non-CHERI aliases as well anyway...

jrtc27:
They have to be inaccessible, otherwise that's a new side-channel that CHERI-unaware software leaves open if you run a CHERI-aware attacker on top of it

PRugg-Cap:
[...]

PR #223, opened 2024-04-29
Clarify CRE spec and note which instructions depend on it

tariqkurd-repo:
The CRE spec was ambiguous as highlighted in https://github.com/riscv/riscv-cheri/issues/218
This is an attempt to fix it and show which instructions need CRE=1 to execute


tariqkurd-repo:
The interesting change is in riscv-legacy-integration.adoc
The large number of others are adding the exception in to the relevant instruction pages

src/insns/sh4add_32bit.adoc
```diff
@@ -35,6 +35,9 @@ Increment the address field of `cs2` by `rs1` shifted left by 4 bit positions. C
 Legacy Mode Description::
 Increment `rs2` by `rs1` shifted left by 4 bit positions.
 
+Exceptions::
+include::require_cre.adoc[]
```
arichardson, line 39: Do we really need CRE for this instruction? I would assume it works fine in legacy mode?

src/insns/sh4adduw_32bit.adoc
```diff
@@ -28,6 +28,9 @@ Encoding::
 Capability Mode Description::
 Increment the address field of `cs2` by the unsigned word in `rs1` shifted left by 4 bit positions. Clear the tag if the resulting capability is unrepresentable or `cs2` is sealed.
 
+Exceptions::
+include::require_cre.adoc[]
```
arichardson, line 32: Same here.

arichardson (approved):
This looks good, but I'm not sure whether the restriction is needed for the shadd in legacy mode. I'm leaning no, but maybe it makes sense for simplicity.

tariqkurd-repo:
> This looks good, but I'm not sure whether the restriction is needed for the shadd in legacy mode. I'm leaning no, but maybe it makes sense for simplicity.

It's a choice. It's easier to define "all instructions in the cheri extensions depend on CRE".
The alternative, which is quite reasonable, is to make another mini-extension for SH4ADD.*, and then treat them the same way as SH[123]ADD.*, as these are instructions which could be used on a non_CHERI core.

I'd prefer one of those two options, rather then have them in a CHERI extension but exclude them, to keep the definition clean.

maybe a new mini-extension is the way to go.

src/insns/sh4adduw_32bit.adoc
tariqkurd-repo: fixed - it's now allowed

PR #228, opened 2024-04-30
Clarify bit-slices in edge cases of the decoded bound figure

PRugg-Cap:
@tariqkurd-repo pointed out that the table was slightly wrong in big exponent cases (negative-width bit-slices). This fixes it while hopefully not making it too much more confusing!

### DIFF
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -637,7 +637,7 @@ For the address to be valid for the current bounds encoding, the value
 in the _Upper Section_ of xref:comp_addr_bounds[xrefstyle=short] _must
 not change_ as this will change the meaning of the bounds.
 
-This gives a range of `s=2^E+MW^`, which as shown in
+This gives a range of `s=2^E+MW^`, as shown in
 xref:cap_bounds_map[xrefstyle=short].
 
 The gap between the object bounds and the bound of the representable range
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -613,13 +613,20 @@ bounds are formed of two or three sections:
 * Lower bits are set to zero
 ** This is only if there is an internal exponent (EF=0)
 
-.Composition of address bounds
-[#comp_addr_bounds,options=header,align="center"]
+.Composition of the decoded top address bound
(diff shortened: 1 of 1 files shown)


src/cap-description.adoc
```diff
@@ -613,14 +613,21 @@ bounds are formed of two or three sections:
...
 
-.Composition of address bounds
-[#comp_addr_bounds,options=header,align="center"]
+.Composition of the decoded top address bound
+[#comp_addr_bounds,options=header,align="center",cols="2,4,2,2"]
 |==============================================================================
-| Configuration  | Upper section | Middle Section | Lower section
-| EF=0           | address[MXLEN-1:E + MW] + ct | T[MW - 1:0] | {E{1'b0}}
-| EF=1, i.e. E=0 | address[MXLEN-1:MW] + ct   2+| T[MW - 1:0]
+| Configuration  | Upper section (if E + MW &#8804; MXLEN) | Middle Section | Lower section
```
tariqkurd-repo, line 619: If E+MW > MXLEN then is the top sectioin excluded completely? I think that makes sense (maybe worth clarifying in the text?)

andresag01 (approved):
Change looks good. I agree with @tariqkurd-repo that it would be good to describe when/why the "upper section" disappears.

src/cap-description.adoc
tariqkurd-repo, line 619:
```suggestion
| Configuration  | Upper section (ony present if E + MW &#8804; MXLEN) | Middle Section | Lower section
```

PRugg-Cap: (I suspect that will lead to sad wrapping issues, but may not be the end of the world). Happy to add to the caption or text underneath instead/as well?

PRugg-Cap: (To be clear, I mean wrapping of the asciidoc column header, not any kind of wrapping of capability bounds!)

PRugg-Cap: Exactly. The bottom section is E bits wide, the middle section is MW bits wide, and the top section picks up the slack, so has size MXLEN + 1 - MW - E. If MW + E > MXLEN then there's no slack for it to take up. That only happens for the very largest exponents. I'll clarify in the text.

PR #232, opened 2024-05-02
rename legacy to hybrid

tariqkurd-repo:
https://github.com/tariqkurd-repo/riscv-cheri/pull/new/rename_legacy_mode


arichardson:
I think the new extension name makes a lot of sense, just not quite sure whether we want to use it for the name of the mode as well. I have a slight preference for "Integer Mode" over "Hybrid Mode" but I'm happy with either. 

CC: @jrtc27 @swm11 @rwatson

tariqkurd-repo:
oh I see, I thought we were talking about both.
Probably better resolve this one next! I'm not keen on integer mode, as it also does FP, Vector etc. so I don't think is really covers it.
Hybrid makes more sense to me because it does the existing stuff, and some new stuff too.

tariqkurd-repo:
discussed with @PeterRugg - we're now trying out `default mode` and `capability mode`
Opinions?

andresag01:
> discussed with @PeterRugg - we're now trying out `default mode` and `capability mode` Opinions?

How about "compatibility mode"?

tariqkurd-repo:
> > discussed with @PeterRugg - we're now trying out `default mode` and `capability mode` Opinions?
> 
> How about "compatibility mode" and "capability mode"? After all, the mode is a trick to make standard RISC-V software work well on a machine with CHERI support (among other things!)

then both modes begin with C. Also this one came up in the meeting yesterday as an option IIRC

andresag01:
> > > discussed with @PeterRugg - we're now trying out `default mode` and `capability mode` Opinions?
> > 
> > 
> > How about "compatibility mode" and "capability mode"? After all, the mode is a trick to make standard RISC-V software work well on a machine with CHERI support (among other things!)
> 
> then both modes begin with C. Also this one came up in the meeting yesterday as an option IIRC

Sure, but we've never used C to refer to Capability mode because that letter is used for a few other things already.

tariqkurd-repo:
> Default isn’t necessarily the default

true. I think I should put the name of the mode in a variable!

What are we going to call it? 

* `unsafe` mode?
* `compatibility` mode 
*  `coarse grained` mode, and cap mode is `fine grained`
* `default` mode 
* `riscv` mode 
* `integer pointer` mode, and cap mode is `capability pointer` mode
* `thin pointer` mode, and cap mode is `fat pointer` mode
any other options?

tariqkurd-repo:
We need to choose a name before friday! Opinions?

andresag01:
> We need to choose a name before Friday! Opinions?

My preference is still "compatibility" and "capability" mode, even though both start with the same heavily overloaded letter!

swm11:
I'm in favour of using names that have a very clear meaning, so like "capability pointer" and "integer pointer" names for modes.

jonwoodruff:
I tend to agree with Simon as well; "capability pointer" mode and "integer pointer" mode are most clear.  Maybe we can come up with an abbreviation for them.

gameboo:
Idle thoughts: I was wondering to what extent only the extension needed a name, and we are either in/entering capability mode or not in/leaving capability mode.

Otherwise, I also like @swm11 's suggestion.

francislaus:
I think the suggestion by @swm11 is very clear. It might be a bit too long for the common usage. Would "int pointer mode" and "cap pointer" be a sensible abbreviation?

arichardson (approved):
Thanks, this sounds good to me. I wonder if we want to change the macro to use e.g. italic for the mode name?

Issue #17, opened 2024-01-23
Zcheri_pte lacks support for load-side revocation


jrtc27:
The current state of the art revocation scheme, and the only one present in CheriBSD releases, uses a load-side, which requires the ability to conditionally trap on loads based on a 1-bit generation counter, not present in the current draft.

tariqkurd-repo:
Hi Jessica - Zcheri_ptr is current WIP - and please extend it.

andresag01:
Agreed, the PTE extension is provisional and very much a work in progress. The current spec highlights this in a note and that the current proposal is missing features that exist in CHERI v9.

jonwoodruff:
Not to muddy the waters, but one could view the load-side generation barrier as an independent feature that could be useful without CHERI for other types of garbage collection.  Is there a possibility of this?  Is there a better chance of the PTE bit being allocated this way if it had support from a wider community?  (Presumably there is a [garbage collection SIG](https://github.com/riscvarchive/groups/blob/main/Managed-Runtimes/CHARTER.md)?)

PeterRugg:
Just noticed this as well. I also wonder about `cloadtags`. We really need to get stats on how much that matters with the load-side barrier, but in Cornucopia it looks like it saved about 50% for xalanc and 20% for omnet, which are the only intensive benchmarks, and I doubt there's too much change there between store-side and load-side in terms of `cloadtags` saving since the core intra-page sweeping loop is the same.

PR #234, opened 2024-05-04
PTE proposal with load-side barrier bits

andresag01:
This is a draft proposal that reworks the Zcheripte extension to include load-side revocation support in the PTE. The goal of the extension is to support the following:

- Load-side revocation
- Optimize sweep phase of capability revocation.
- Limit capability propagation at page granularity

In summary, CHERI would add 3 bits to the PTE:

* Capability Access (CA): Clear read/written tags if clear
* Capability Dirty (CD): Trap when capability with tag=1 is written to page and CD=0
* Capability Read Generation (CRG): generation of the page for revocation

Note that this is only a draft to stimulate discussion, not a finished proposal. For example, @jrtc27 previously mentioned that conflating read/write into CA is not desirable. My intuition is that some combinations are not useful, so more efficient encodings are possible. Specifically:

* The CRG bit of a page is not very useful if the page does not allow capability reads
* The CD bit of a page is not very useful if the page does not allow capability writes

However, is it always the case that the capability read/write bit of a page is set on allocation and remains unchanged throughout the lifetime of a page? If so, then how about something like this:

|Cap Read|Cap Write|CD|CRG|Notes|
|-|-|-|-|-|
|1|1|0|0|Cap read/write set means CD/CRG needed
|1|1|1|1|
|1|1|0|1|
|1|1|1|0|
|1|0|?|1|Cap write clear means CD is not useful
|1|0|?|0|
|0|?|?|?|Cap read clear means CRG is not useful. Cap read clear also implies that Cap write is probably not useful at the page granularity

If this works, then we must be able to encode 7 values which is possible in 3 bits (and we get one spare encoding!). What do you think @jrtc27 @tariqkurd-repo @arichardson?

Fixes https://github.com/riscv/riscv-cheri/issues/17


src/cheri-pte-ext.adoc
```diff
@@ -4,21 +4,32 @@
...
+{cheri_pte_ext_name} extension adds new bits to RISC-V's Page Table Entry (PTE)
+format.
 
+=== Limiting Capability Propagation
+
+TODO
+
+=== Capability Revocation
+
+TODO
```
andresag01, line 18: I will fill in these bits when there is agreement on the rest of the extension.

arichardson:
CC: @nwf

jrtc27:
If you mprotect a page it looses its capability write permission but is still capability dirty

nwf:
OK, let me give this a shot.  I'm sorry this is long, but I hope it's all accurate and provides, if nothing else, a useful software-driven perspective.

It's also worth remembering that we have done... very little experimentation, relative to the body of work on MMUs in general, when it comes to "what software wants from CHERI PTE bits".  Some of this, then, is necessarily a _best guess_ extrapolating from very few data points.

I've said it before, but to repeat: it's a bit sad we have to resort to using PTE bits for this, and I hope that in the future there are alternative extensions that build on physically-indexed metadata tables, like AMD's SEV-SNP's RMTs or Arm's RME's GPTs.  In any case, until such time, we are stuck using PTEs to reflect properties of a logical page (or even of the underlying physical frame), but hopefully future Zcheripte editions can use fewer PTE bits in composition with those other mechanisms.

Some of the design space is dependent on how willing microarchitecture is to take on data dependence in operations.  It's most useful to software if we can trigger events only on tagged loads or stores, but these are new, and possibly expensive, dependencies, keeping state speculative for longer.

Some of the design space is also dependent on how willing software is to perform expensive operations on the _set_ of page mappings for a given page.  Historically, such need was limited to things like paging, and so such operations are, generally, not particularly speedy.  The use of PTE mapping bits to approximate page state sometimes requires traversing this set.

----

OK, with all that out of the way... it's probably most useful to think about what logical states we need a logical page or an individual page mapping to be in.  As relevant here, pages can be given four states:

1. "Always cap-clean": no mapping of this page has or will ever accept a capability store.  This includes `mmap`-ed file data.

2. "Cap-dirty": this page has been the target of a capability store and has not yet been found to have no capabilities within.  There may exist mappings of this page that accept tagged stores without other side-effects.

3. "Ephemerally cap-clean": this page is _potentially permissive_ of capability stores, but there do not exist mappings of it that accept tagged stores without trapping.

4. "Cap-dirtyable": this page is _potentially permissive_ of capability stores, and there may exist mappings that accept tagged stores, possibly without trapping (but with logging; that is, where the PTW will perform an AMO on the PTE).

Correctness absent temporal safety concerns really only differentiates between 1 and 2, smashing 3 and 4 up into an amalgamated 2.  This is how the TLBE bits of CHERI-MIPS were used.

Under Cornucopia (Reloaded) revocation, the revoker must visit (at least) all cap-dirty and cap-dirtyable pages.  Transition _from_ ephemerally cap-clean is induced by traps or new mappings, and may require broadcast updates of PTEs, depending on exactly which things PTEs can represent (see below).  Re-entry _to_ ephemerally cap-clean requires broadcast down-grading all of a page's mappings, and may require retries across multiple revocation epochs.

Mappings also come in multiple flavors.  The easy cases:

1. A mapping, read-only or read-write, of an "always cap-clean" page .  If mapped read-write, untagged stores are permitted, but tagged stores either trap or clear the tag, and loads either trap if tagged or clear the tag.  No need for CD/CRG tracking.

2. A mapping of a cap-dirty page: if read-write, stores of all flavors permitted; tagged (or cap-width) loads must be mediated by CRG.

3. A mapping of a cap-dirtyable page: if read-write, tagged (or cap-width) stores either AMO the PTE or trap; tagged (or cap-width) loads must be mediated by CRG.  The AMO effectively upgrades the page from cap-dirtyable to cap-dirty, but this transition will be discovered late, possibly as late as attempted re-entry to cap-clean.

And then there is the hard case involving ephemerally cap-clean (or "recently" ephemerally cap-clean) pages.  We would, ideally, like for the revoker to be able to pay no attention to cap-clean pages, be they ephemerally or always so.  This requires that software either be able to...

a. broadcast updates to PTEs for pages transitioning from cap-clean.  That's potentially expensive.

b. configure mappings of cap-clean pages so that they _become as if_ they were the wrong CRG whenever the page -- not this mapping, but the page via another mapping -- transitions away from cap-clean.

The latter seems preferable to me, but it seems to necessitate a PTE state of "all tagged loads trap" and we almost surely want that to be a true dependence on the loaded value, so that cap-width reads of data don't cause software to have to start paying attention to the page again.

In any case, assuming it were microarchitecturally agreeable, I think the following set of _six_ states suffice for all software we've written so far:

1. (**TN**) No tagged loads or stores permitted (clearing tags OK); the mapping may be read and/or written with untagged data.
2. (**TP**_g_) Tagged stores permitted if page writable, with 1 CRG bit
3. (**TA**_g_) Tagged stores AMO (to state 2 above) or trap if page writable, with 1 CRG bit
4. (**TT**) Tagged stores and tagged loads always trap; the mapping may be read and/or written with untagged data.

If we had to remove the last one, we could (by using state 1 instead, and broadcasting to PTEs).  That's still 5 states, tho', so still 3 bits.

The distinction between **TP** and **TA** states is only meaningful if the **W** bit is asserted, but that's probably not useful for any kind of representation compression.

-----

Specifically, to the point of
> If you mprotect a page it can lose its capability write permission but still be capability dirty
Yes, but that cap-dirty-ness doesn't need to be reflected in the updated _mapping_, so long as it's tracked in the logical page state in software.

Whew.

andresag01:
@nwf: Thanks for the explanation and your proposal. Its very informative!

First of all, I am new to temporal safety using CHERI! Could you elaborate a bit on how could a page get into the "ephemeral cap-clean" state? Is this purely due to the revocation algorithm?

My guess is that a page that the revoker can ignore pages that are ephemerally cap clean altogether. However, these pages are distinct from cap-dirtyable because in the latter you may have other mappings that are cap-dirty for example -- would the revoker at any point scan all mappings of a cap-dirtyable page and, for example, conclude that they are all "cap-clean" so that page should really be ephemeral and is safe to ignore it in the next revocation cycle?

nwf:
> @nwf: Thanks for the explanation and your proposal. Its very informative!

Great to hear that that brain-dump communicated something. :)

> First of all, I am new to temporal safety using CHERI! Could you elaborate a bit on how could a page get into the "ephemeral cap-clean" state? Is this purely due to the revocation algorithm?
>
> My guess is that a page that the revoker can ignore pages that are ephemerally cap clean altogether. However, these pages are distinct from cap-dirtyable because in the latter you may have other mappings that are cap-dirty for example -- would the revoker at any point scan all mappings of a cap-dirtyable page and, for example, conclude that they are all "cap-clean" so that page should really be ephemeral and is safe to ignore it in the next revocation cycle?

That's exactly right: it's easy for the revoker to keep track of whether or not a given page was found to have a capability on it during the sweep, and, if not, that's a good reason to try downgrading it to ephemerally cap-clean (even though the application could write a capability later, we might hope that it'll be a few revocation epochs in the future).

That downgrade is a frightful wad of complexity that eventually involves looking at all aliases of a page and which we stage across multiple revocation epochs (so that we can avoid per-page TLB shoot-downs and IPIs on platforms that require those).  I have a... long-abandoned, long-winded draft of a technical report that has an exhaustively tested Murφ model of multiple TLBs and multiple page aliases and all the permitted flows through the system.  While there's a huge semantic gap between that and the implementation in CheriBSD, it was, nevertheless, quite useful for understanding.  I'll try to dig that up, stamp "DRAFT" on it in big letters, and put it up as a PDF somewhere for people to look at.

andresag01:
Thanks for the info! My understanding so far is that we there are 3 key goals for the PTEs:
* Limit capability propagation -- TN state (1)
* Revocation load barrier -- CRG states (2) & (3)
* Revocation optimizations -- dirty and ephimeral states (2), (3) & (4)

Is that correct? For the optimizations, are there any studies/figures showing what improvements (e.g. performance, pause times, etc) the dirty/ephimeral ideas provide? Is it also feasible to consider these optimizations as separate features?

jonwoodruff:
@nwf and @andresag01 
Further to this, can "Revocation Barrier", if gating both loads and stores, be used to "Limit Capability Propagation" in the case of extreme PTE poverty?

(I didn't quite parse how the 6 states behaved with respect to capability loads, but I assume the "with 1 CRG bit" meant that it traps on capability read if CRG doesn't match.)

jonwoodruff:
If CRW, the Capability Read Generation bit, is ignored when CW, Capability Write, is not set, then I think we get all of the states we want.

State 1 & 2: CRW True, CRG 1/0
> Allow capability writes, and allow capability reads only if CRG matches the CSR generation bit.

State 3: CRW False, CRG 0 (for example)
> Trap on valid capability writes, allow but strip tags on all capability reads. This should work fine for non-capability-bearing mappings such that capability-wide memcpy can be used without flipping CRG on each generation swap.

State 4: CRW False, CRG 1 (for example)
> Allow silent upgrade to CRW True, CRG = CSR generation bit. This allows hardware dirty tracking; software must detect the transition of CRW at a later point, as must be usual for dirty tracking.

nwf:
After a little bit of discussion elsewhere, here's a different 2-bit proposal.  It's not great -- I think the 3-bit proposals are nicer -- but I think this is at least workable, if a little trap-heavy in use.

# Revocation with four-state PTEs 2024/06/07

The four states proposed here are

|   State   |Cap Read          | Cap Write   |
|-----------|------------------|-------------|
|CRW, LCRG=0| GCRG=0?          | OK          | 
|CRW, LCRG=1| GCRG=1?          | OK          | 
|NoCaps     | Assumed untagged | Tag dep trap|
|CapsTrap   | Tag dep trap     | Tag dep trap|

That is, this approach jetissons CAP-DIRTYABLE / AMO-cap-dirtying PTEs in favor of the "CapsTrap" state.

## Steady states

* An always data-only page (e.g., mmap'd file) has all of its aliases parked in NoCaps, and the kernel will deliver SIGSEGV on traps.

* A cap-dirty page with all of its aliases aware of its dirtiness will have all of its alises bouncing between the two CRW states as revocation epochs elapse.

* A (logically) cap-permissive but ephemerally clean page (safely being ignored by the revoker) can have all its aliases in the CapsTrap state.

## State transitions

* The first trap on any alias of a cap-permissive but ephemerally clean page will upgrade the page's logical state to cap-dirty and the PTE to the correct CRW LCRG state.

  Subsequent traps through a CapsTrap alias of a cap-dirty page will trigger revocation (if a sweep of the containing address space is active) and then move the PTE to the appropriate CRW state.

  No broadcast update is necessary.

  We may presume that the capability store triggering the first fault has been inspected by any ongoing revocation (even if it just started; there is a somewhat amusing case in which a tagged cap is stored and triggers the trap, then revocation starts and clears the tag in the spilled register file, and so when the store retries, it will store a clear tag, but that's fine).

  Subsequent traps through CapsTrap aliases of known-dirty pages must engage the revocation machinery during sweeps because other capabilities on that page may not have been yet inspected by the revoker.

* Becoming clean remains somewhat tricky, given all the moving pieces.

  The existing revoker uses cap-dirtyable PTE states to stage its transition to cleanliness, relying on the existence of such a state -- and insistance that the PTW perform an AMO CAS to exit that state -- to ensure that at least one epoch has begun, and so at least one broadcast TLB invalidation has happened, within the address space containing an alias between that alias's PTE being in a cap-dirty (equivalently, CRW) and cap-clean but cap-load-permissive (no   direct analogy exists in the proposal here).

  Absent such a state, we probably cannot get away from needing a TLB shootdown on the transition from cap-dirty to ephemerally cap-clean.  It may suffice to have this slightly more expensive procedure:

    1. Downgrade aliases from CRW to CapsTrap in the background revoker if no caps are observed during a visit.

    2. If the background revoker finds a logically cap-dirty page with a CapsTrap alias, visit the page, and...

         * if capabilities were found and not revoked, upgrade the alias to CRW

	 * if there are no aliases, the beginning of this epoch has already ensured that this alias's TLB entry is also CapsTrap, like its PTE, and so we may safely mark the page ephemerally cap-clean.

	 * otherwise, gather the page's PTEs and, if they are all CapsTrap,

	    * issue a TLB shootdown for each alias (which, recall, might be in multiple address spaces)

	    * visit the page *again*, now that stores through stale TLB entries are not possible.  If no caps are found, mark the page ephemerally cap-clean.

       The exact expense of this approach depends on how often aliased pages become ephemerally cap-clean, but that is probably fairly rare.

## Other Caveats

* It continues to have a store-path tag dependency.  Relaxing this is somewhat difficult due to the conflicting desire to have hybrid executables (DDC) and limit capability propagation through file mappings.

  The tag dependency could be relaxed in the CapsTrap state more easily, at the cost of considering more pages during revocation, but since it probably has to exist for NoCaps, that hardly seems worthwhile.

* Its load-path dependencies are also possibly tricky.  Relaxing to cap-capable loads rather than loads of tagged capabilities again just means considering more pages cap-dirty.  (And requires that software be prepared for marking a page cap-dirty in response to a *load* through a CapsTrap alias.)

* The "assumed untagged" NoCaps behavior of NoCaps places the onus on kernel software to ensure that there are no caps on these pages (but permits hardware enforcement with clearing or trapping).

jonwoodruff:
@nwf Thank for the detailed thoughts!

Is there a justification for jettisoning the pre-CRW state for the CapsTrap state?
Maybe you could make a distinction in the OS for mapping pages with no alias into the pre-CRW state, and map aliased dirtyable pages as NoCaps, trapping on the first write and implementing the CRW upgrade in software.

This would maybe also eliminate the tag sensitivity for loads?

andresag01:
@jonwoodruff has updated the proposal as discussed last week! Thanks for taking the time to do this @jonwoodruff! https://github.com/andresag01/riscv-cheri/pull/2

jonwoodruff:
With some help, I've now performed a large-scale experiment on the feasibility of a restricted scheme using Morello.

I created [a branch of CheriBSD](https://github.com/CTSRD-CHERI/cheribsd/pull/2153/files) that does not flip individual CHERI protection bits (of which there are four), but simply assigns the bits to one of our four states.  Quite surprisingly, I got this to work correctly such that it runs all attempted software and benchmarks with revocation enabled without crashing. Sadly, tracing PTE bits on QEMU indicates that I never get the pre-CRW state, and thus am only using 3 states.  That is, NoCaps, Caps0, and Caps1.

We then booted Morello hardware with this kernel and ran spec benchmarks.  The changes were confirmed by tracing PTE cap-dirtying events, which were zero with the new kernel (and non-zero with the old one). Surprisingly, we see no overhead from running with these reduced states, with performance being consistently slightly faster than previously.  This is thought to be due to lack of optimisation of the baseline and some unknown effects that we will now be tracking down.  Nevertheless, the average overhead for temporal safety/revocation on the ref workload for Spec2006 benchmarks that would run (all int workloads but gcc and perlbench) is about 5%.  It's likely that this can be reduced significantly with optimisation such that the three-state solution would be slower than the four state solution with hardware dirtying, which this proposal would also support.  It seems more of a stretch to say that any other missing states would have a significant impact on performance.

@nwf has made a cursory review of my modifications, and believes that they should be safe.

In summary, I believe we have demonstrated that the 3-state subset of the proposal works and does not have surprising overheads with Spec2006 benchmarks.  The full 4-state solution should be very close indeed to the best we can do.

========

Random note:
Morello previously had a "always allow caps" state which was used for the kernel.  This was replaced with Caps0, with the assumption that the kernel would never flip its generation bit.  This proved to be true.

tariqkurd-repo:
I think that `menvcfg` should also get the CW, CRG bits, otherwise this scheme is different to how the other bit-fields work where `menvcfg` is a superset.

src/cheri-pte-ext.adoc
```diff
@@ -75,26 +107,28 @@ access that caused the update, and the sequence is interruptible. However, the
...
 
-The *menvcfg* register is extended to allow discovering whether the
-implementation supports the CD bit.
+NOTE: An implementation may avoid a dependency on the tag
+bit value of the capability read. This will introduce additional traps during revocation sweeps.
 
-The *menvcfg* register operates as described in cite:[riscv-priv-spec]. {cheri_base_ext_name} adds a new enable bit as shown in
-xref:menvcfgreg[xrefstyle=short] when XLEN=64.
+[#senvcfg_pte,reftext="senvcfg PTE bits"]
+=== Extending the Supervisor Environment Configuration Register (menvcfg)
```
andresag01, line 125: There is a typo here. This section is for `senvcfg` instead of `menvcfg`.

andresag01:
> I think that `menvcfg` should also get the CW, CRG bits, otherwise this scheme is different to how the other bit-fields work where `menvcfg` is a superset.

The key is that virtual memory is associated to S-mode, but not M-Mode. Perhaps the right place for the `CRG` bit is `satp` (or something like it) which is actually related to virtual memory instead of `senvcfg`.

src/cheri-pte-ext.adoc
tariqkurd-repo: yes - I'm fixing up various things - will commit that change soon.

src/cheri-pte-ext.adoc
```diff
@@ -121,14 +123,40 @@ and the virtual page's CRG bit does not equal senvcfg.CRG.
...
-=== Extending the Supervisor Environment Configuration Register (menvcfg)
+=== Extending the Machine and Supervisor Environment Configuration Registers (menvcfg/senvcfg)
 
-The *senvcfg* register operates as described in cite:[riscv-priv-spec].
+The *menvcfg* and *senvcfg* register operate as described in cite:[riscv-priv-spec].
 {cheri_pte_ext_name} adds a new Capability Read Generation (CRG) bit as shown
-in xref:senvcfgreg_pte[xrefstyle=short] when XLEN=64. The CRG bit in *senvcfg*
+in xref:menvcfgreg_pte[xrefstyle=short] and xref:senvcfgreg_pte[xrefstyle=short]
+when XLEN=64. The CRG bit in *senvcfg*
 is used as described in xref:section_extending_pte[xrefstyle=short].
```
andresag01, line 154: Does that mean that the CRG bit is mirrored in both `senvcfg` and `menvcfg`? The extension would not work well if there were multiple, differing  instances of the bit around.

tariqkurd-repo: it's the same as all other bits in Xengcfg - it's the same bit

tariqkurd-repo (approved):
it's probably worth adding something about the benefits/consequences of each of the 4 options, but I think the actual spec is ok now

tariqkurd-repo:
> it's probably worth adding something about the benefits/consequences of each of the 4 options, but I think the actual spec is ok now

Jon has pointed out that this is already covered in the doc

src/cheri-pte-ext.adoc
```diff
@@ -156,10 +156,14 @@ in xref:menvcfgreg_pte[xrefstyle=short] and xref:senvcfgreg_pte[xrefstyle=short]
 when XLEN=64. The CRG bit in *senvcfg*
 is used as described in xref:section_extending_pte[xrefstyle=short].
 
-.Machine environment configuration register (*menvcfg*) with Zcheripte, RV64 only
+WARNING: Unlike all other bits in *Xenvcfg*, CRG is in *senvcfg* and *not* in *menvcfg*. This suggests
+ this it is in the wrong place and is likely to move at some point in the future to a location
```
andresag01, line 160: Typo? It sounds a little odd.

src/cheri-pte-ext.adoc
```diff
@@ -139,10 +139,13 @@ bit value of the capability read. This will introduce additional traps during re
 |===
 |PTE.CW |PTE.CRG           |Store/AMO
 | 0     | 0                | Page fault if stored tag is set
-| 0     | 1                | Page fault if stored tag is set, or hardware CW, CRG update
+| 0     | 1                | Page fault if stored tag is set, or hardware CW and CRG update^1^
 | 1     | X                | Normal operation
 |===
 
+^1^ The choice here follows the pattern of whether to implement the _Svade_ extension
+ to take page-fault exceptions for A and D PTE bit state changes, or whether to a implement
```
andresag01, line 147: Typo: "whether to a implement"

src/cheri-pte-ext.adoc
tariqkurd-repo: what's the problem?

src/cheri-pte-ext.adoc
```diff
@@ -75,26 +109,61 @@ access that caused the update, and the sequence is interruptible. However, the
...
+[%autowidth,float="center",align="left",cols="<,<,<",options="header"]
+|===
+|PTE.CW |PTE.CRG           |Store/AMO
+| 0     | 0                | Page fault if stored tag is set
+| 0     | 1                | Page fault if stored tag is set, or hardware CW and CRG update^1^
+| 1     | X                | Normal operation
+|===
+
+^1^ The choice here follows the pattern of whether to implement the _Svade_ extension
+ to take page-fault exceptions for A and D PTE bit state changes, or whether to a implement
```
tariqkurd-repo, line 147:
```suggestion
 to take page-fault exceptions for A and D PTE bit state changes, or whether to implement
```

src/cheri-pte-ext.adoc
andresag01: "This suggests this it is in the wrong place..."

tariqkurd-repo: this is the point - we know it's wrong because it doesn't follow the pattern of Xenvcfg - but it's not worth changing now as all this stuff will change when we finally get to ARC review - so I think it's good enough for now. You can reword it if you like.

PeterRugg:
> I think that menvcfg should also get the CW, CRG bits, otherwise this scheme is different to how the other bit-fields work where menvcfg is a superset.

Looking through the PR, I don't think there is a CW bit in either `senvcfg` or `menvcfg`, and I don't think there should be. CRG only needs to be there as it's what you compare the PTE CRG bits against. It's not an enable as such, so `senvcfg` likely isn't the best fit. `satp` could make sense, but I don't think you want to insist on running an `sfence.vma` for it to take effect...

jrtc27:
(v)sstatus would make the most sense for it. There's precedent from SUM and MXR (and TVM in the full mstatus).

andresag01:
> > I think that menvcfg should also get the CW, CRG bits, otherwise this scheme is different to how the other bit-fields work where menvcfg is a superset.
> 
> Looking through the PR, I don't think there is a CW bit in either `senvcfg` or `menvcfg`, and I don't think there should be. CRG only needs to be there as it's what you compare the PTE CRG bits against. It's not an enable as such, so `senvcfg` likely isn't the best fit. `satp` could make sense, but I don't think you want to insist on running an `sfence.vma` for it to take effect...

I agree, `satp` would be great, but there isn't much space in the encoding. I also like the idea of putting it in `(v)sstatus`

andresag01:
> (v)sstatus would make the most sense for it. There's precedent from SUM and MXR (and TVM in the full mstatus).

@jrtc27 I agree that (v)sstatus would be a good place to put this flag. Does hstatus also need its own CRG bit? The RISC-V privileged spec says this:

> An OS or hypervisor running in HS-mode uses the supervisor CSRs to interact with the exception, interrupt, and address-translation subsystems

Wouldn't we expect that the OS or hypervisor running in HS-mode to also take advantage of revokation?

tariqkurd-repo:
> > (v)sstatus would make the most sense for it. There's precedent from SUM and MXR (and TVM in the full mstatus).
> 
> @jrtc27 I agree that (v)sstatus would be a good place to put this flag. Does hstatus also need its own CRG bit? The RISC-V privileged spec says this:
> 
> > An OS or hypervisor running in HS-mode uses the supervisor CSRs to interact with the exception, interrupt, and address-translation subsystems
> 
> Wouldn't we expect that the OS or hypervisor running in HS-mode to also take advantage of revokation?

Probably. I think we can safely rule out mstatus. If in doubt maybe add the bit in hstatus?

andresag01:
It needs to be in `mstatus` because it is a superset of `sstatus`. Here is the relevant fragment form the privileged spec

Screenshot from 2024-09-10 11-15-49

tariqkurd-repo:
> It needs to be in `mstatus` because it is a superset of `sstatus`. Here is the relevant fragment form the privileged spec
> 
yes ok - but I noticed that hstatus doesn't have all the mstatus/status fields.

src/cheri-pte-ext.adoc
andresag01: This text was removed when migrating CRG to xstatus.

andresag01:
I migrated the CRG bit from senvcfg to xstatus excluding hstatus. We can add it to hstatus if its necessary.

tariqkurd-repo (approved):

src/cheri-pte-ext.adoc
```diff
@@ -75,26 +109,148 @@ access that caused the update, and the sequence is interruptible. However, the
...
+
+The <<mstatusreg_pte,mstatus>>, <<sstatusreg_pte,status>> and <<vsstatusreg_pte,vsstatus>> CSRs are extended to include the new Capability Read Generation (CRG) bit as shown.
 
-[#menvcfg_pte,reftext="menvcfg PTE bits"]
-=== Extending the Machine Environment Configuration Register (menvcfg)
+mstatus.CRG is made visible in sstatus.CRG.
 
-The *menvcfg* register is extended to allow discovering whether the
-implementation supports the CD bit.
+When V=1 vstatus.CRG is in effect.
```
andresag01, line 157:
```suggestion
When V=1 vsstatus.CRG is in effect.
```

tariqkurd-repo:
any objections to merging this?

arichardson:
ping @jonwoodruff @nwf @nwf-msr: Is this ready to be merged?

arichardson (approved):
I think we can merge this now since Wes is also ok with it.

src/cheri-pte-ext.adoc
```diff
@@ -4,21 +4,63 @@
...
+management to facilitate capability revocation and control the flow of
+capabilities in memory at the page granularity. For this reason, the
+{cheri_pte_ext_name} extension adds new bits to RISC-V's Page Table Entry (PTE)
+format.
 
+=== Limiting Capability Propagation
+
+Page table enforcement can allow the operating system to limit the flow
+of capabilities between processes.
+It is highly desirable that a process should only posses capabilities that have
```
tariqkurd-repo, line 16:
```suggestion
It is highly desirable that a process should only possess capabilities that have
```

jonwoodruff:
Just found one run-on sentence.  Otherwise, it looked good to me!

--- a/src/cheri-pte-ext.adoc
+++ b/src/cheri-pte-ext.adoc
@@ -20,8 +20,8 @@ but capability pointers must not be shared across these channels
 into a foreign address space.
 An operating system might defend against this by only issuing a capability
 to the shared
-region that does not grant the load/store capability permission,
-there are circumstances where portions of general-purpose, mmapped^*^ memory become shared,
+region that does not grant the load/store capability permission.
+However, there are circumstances where portions of general-purpose, mmapped^*^ memory become shared,
 and the operating system must prevent future capability communication
 through those pages.
 This is not possible without restructuring software, as the capability for

andresag01:
> Just found one run-on sentence. Otherwise, it looked good to me!

Thanks for spotting this! I applied your change in this commit: https://github.com/riscv/riscv-cheri/pull/234/commits/48d576167dfb79f846f5ed36904f4ef1e000f415

Issue #245, opened 2024-05-09
DDC requires CRE=1 to access


tariqkurd-repo:
https://github.com/riscv/riscv-cheri/blob/main/src/riscv-hybrid-integration.adoc

This text refers to a section including tables: 

Executing CSR instructions accessing any capability wide CSR addresses ([[csr-numbers-section]](https://github.com/riscv/riscv-cheri/blob/main/src/riscv-hybrid-integration.adoc#csr-numbers-section)) cause illegal instruction exceptions

https://github.com/riscv/riscv-cheri/blob/main/src/riscv-hybrid-integration.adoc#csr-numbers-section

but this section doesn't include DDC which should be made inaccessible by CRE=0

so the text is correct, but the list in the section is incomplete.

tariqkurd-repo:
in fact what we should do is explicitly state whether CRE is required to access each CSR in the CSR description sections.

tariqkurd-repo:
closed as the MR is merged

PR #247, opened 2024-05-09
fix xrefs, and add note about debug mode

tariqkurd-repo:


tariqkurd-repo:
Fixes: https://github.com/riscv/riscv-cheri/issues/245

PR #250, opened 2024-05-15
Make DDC, MTDC, STDC require CRE

tariqkurd-repo:


arichardson (approved):
I guess implicit DDC access doesn't need it but the note makes sense.

tariqkurd-repo:
> I guess implicit DDC access doesn't need it but the note makes sense.

added, good point

PR #251, opened 2024-05-16
Fix scmode

tariqkurd-repo:
SCMODE didn't specify how to treat updating the M-bit when X=0
and I've done some associated tidying up to the M-bit section.


tariqkurd-repo:
There's something odd going on - the change to attributres.adoc is already on main so isn't a change.
It looks like it hasn't rebased against https://github.com/riscv/riscv-cheri/pull/250

arichardson:
This GitHub behaviour is very strange, could you do a `git pull --rebase` on your side and then  `push -f` to get it updated?

arichardson (approved):
This looks good, but it's probably best to rebase and update the PR before squash-merging since github seems confused about the state of the repository.

PR #265, opened 2024-05-24
Note that the M-bit is not a permission

Timmmm:
This has caused some confusion so is worth being very clear about.


Timmmm:
Thanks to @jrtc27 for pointing out this distinction. I accept all criticism for the use of "progenitor". :-D

tariqkurd-repo:
I'll close this one as I'm planning another more general clarification of the M-bit, coming soon.....

Issue #260, opened 2024-05-23
ACPERM/GCPERM - use of the M-bit is ambiguous


tariqkurd-repo:
The AP field tables include the M-bit
ACPERM doesn't AND that permission bit but the rules _do_ include clearing it if X is cleared
GCPERM doesn't return that permission

But the M-bit is encoded with the other architectural permissions, for RV32 and for RV64, in tables 3 and 4.

My proposed fix is to include the M-bit into the definition of ACPERM/GCPERM so we have a consistent definition of AP.

tariqkurd-repo:
The M-bit _isn't_ a permission, it can me toggled arbitrarily (providing the cap isn't sealed)

It can be cleared as a result of ACPERM, if X is removed, sort of. In fact if X=0 then it is no longer the M-bit, so it becomes non-existant - so it _can_ change state as a result of ACPERM.

I'll tidy up the doc to _remove_ the M-bit from the AP field. The current spec for ACPERM, GCPERM _is_ believed correct.

PR #268, opened 2024-05-28
remove the M-bit from AP

tariqkurd-repo:
The M-bit is no longer in the AP field as it's not a permission


tariqkurd-repo:
Fixes https://github.com/riscv/riscv-cheri/issues/260

PR #271, opened 2024-05-31
another attempt at removing M from AP

tariqkurd-repo:


tariqkurd-repo:
@Timmmm try this one - there were other problems remaining.

src/riscv-hybrid-integration.adoc
```diff
@@ -58,23 +58,31 @@ top is 2^MXLEN^
...
+when the <<x_perm>> is set.
+
+* When MXLEN=32, the Mode is encoded in bit 0 of quadrant 1 from the AP field
+_even though it is not a permission_ as shown in <<cap_perms_encoding32>>.
+** Only quadrant 1 represents executable capabilities, and so it's the only
+one which encodes the Mode.
+* When MXLEN=64, the Mode is encoded separately; a new <<m_bit>> field is
 added to the capability format as shown in
 xref:cap_perms_encoding64[xrefstyle=short].
+It is _only_ the <<m_bit>> for code capabilities, otherwise the field is reserved.
```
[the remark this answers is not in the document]
tariqkurd-repo: `code capabilities` is used elsewhere in the spec, and I made an update

src/cap-description.adoc
```diff
@@ -189,7 +192,7 @@ reserved values as if it were 0b00000 (no permissions). Future extensions may as
 meanings to the reserved bit patterns, in which case <<GCPERM>> is allowed to report a
 non-zero value.
 
-A {cap_rv64_perms_width}-bit vector encodes the permissions and <<m_bit>> when MXLEN=64. In
+A {cap_rv64_perms_width}-bit vector encodes the permissions MXLEN=64. In
```
andresag01, line 195: Perhaps a "when" missing here before MXLEN=64?

src/riscv-hybrid-integration.adoc
```diff
@@ -58,23 +58,31 @@ top is 2^MXLEN^
...
+* When MXLEN=32, the Mode is encoded in bit 0 of quadrant 1 from the AP field
+_even though it is not a permission_ as shown in <<cap_perms_encoding32>>.
+** Only quadrant 1 represents executable capabilities, and so it's the only
+one which encodes the Mode.
+* When MXLEN=64, the Mode is encoded separately; a new <<m_bit>> field is
 added to the capability format as shown in
 xref:cap_perms_encoding64[xrefstyle=short].
+The <<m_bit>> is only valid for code capabilities, otherwise the field is reserved.
+
+NOTE: Mode is similar to a permission, but is not considered one as it can varied
```
andresag01, line 74: > is not considered one as it can varied arbitrarily.

I think of it as orthogonal to permissions entirely. It is a trick to change the behaviour of instructions and the view of CSRs while saving instruction encoding space. I normally see it as an extension bit to the instruction encoding.

src/cap-description.adoc
andresag01, line 195: It looks like the table [here](https://github.com/riscv/riscv-cheri/pull/271/files#diff-2ac200119dffa8d1c8c860fa8f9d6614dc07cf4563a3cc1820cceb25152a12fbR212) has the M bit commented out, so we are no longer showing the M bit anywhere in any diagram or table. Am I missing something? It will be hard to understand where the M bit is encoded in the capability format for MXLEN=64

src/cap-description.adoc
```diff
@@ -135,9 +135,9 @@ Therefore, it is only possible to encode a subset of all combinations.
 .Permissions widths depending on MXLEN
 [#perms_bit_width,options=header,align="center",width="80%",cols="^13%,^27%,<60%"]
 |==============================================================================
-^| MXLEN | Permissions width      | Comment
-^| 32    | {cap_rv32_perms_width} | Encodes some combinations of 5 permission bits and _includes_ the <<m_bit>> in quadrant 1 (see <<cap_perms_encoding32>>)
-^| 64    | {cap_rv32_perms_width} | Separate bits for each permission, _excluding_ the <<m_bit>>
+^| MXLEN | AP field width         | Comment
+^| 32    | {cap_rv32_perms_width} | Encodes some combinations of 5 permission bits, including the <<m_bit>> if {cheri_default_ext_name} is supported.
```
andresag01, line 139: There is a hard-coded 5 here. I think its best to remove it or make it {cap_rv64_perms_width}

src/riscv-hybrid-integration.adoc
tariqkurd-repo: ok, updated the comment

src/cap-description.adoc
```diff
@@ -542,6 +545,7 @@ expanded base is 0 and top is 2^MXLEN^.
 | Tag      | zero  | Capability is not valid
 | SDP      | zeros | Grants no permissions
 | AP       | zeros | Grants no permissions
+| M (MXLEN=64) | zero | <<m_bit>>, encoded separately to AP for MXLEN=64
```
andresag01, line 548: I think this is not right. This chapter is about the capability encoding for zcheripurecap, so technically this field does not exist/has not been defined yet and is reserved

src/cap-description.adoc
andresag01: I initially meant that I thought these figures did not have the M bit separately because I did not see a change for that here
Screenshot from 2024-06-03 13-56-44
But the M bit is indeed there (although there is a misalignment in the bit number labels!), so we can resolve this!

src/cap-description.adoc
```diff
@@ -573,6 +577,7 @@ or 'root' capability.
                         | Grants all permissions
 | AP (MXLEN=64) | 0x1F (see xref:cap_perms_encoding64[xrefstyle=short])
                         | Grants all permissions
+| M (MXLEN=64)  | one   | <<m_bit>>, encoded separately to AP for MXLEN=64
```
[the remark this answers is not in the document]
tariqkurd-repo: we need a forward reference then, otherwise we'll have two definitions of the infinite-cap which will be too confusing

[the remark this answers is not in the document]
tariqkurd-repo: which is already there, so no problem, update coming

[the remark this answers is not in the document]
tariqkurd-repo: try again - there was already a forwards reference

PR #275, opened 2024-06-03
change utidc address to avoid conflict

tariqkurd-repo:
Fixes  https://github.com/tariqkurd-repo/riscv-cheri/pull/new/utid_address


PRugg-Cap:
Shouldn't it have upper 2 bits 0b11 to make it read-only?

tariqkurd-repo:
of course, I forgot that it was supposed to be read only - I'll reallocate

PRugg-Cap (approved):

PR #285, opened 2024-06-07
clarify malformed cap handling

tariqkurd-repo:
More clarification on malformed capability handling, following the discussion in https://github.com/riscv/riscv-cheri/pull/279

cmv, lc, sc, amoswap.c propagate malformed caps
caddi, cadd, scaddr, scbnds* de-tag
gchi, gcperm ignore
gcbase, gclen return 0

also if the permission can't be created by ACPERM then clear _all_ permissions, which affect load/store/cbo/amo dereferencing and also the target capability for JALR

### DIFF
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -45,7 +45,7 @@ include::img/cap-encoding-xlen64.edn[]
 
 Reserved bits are available for future extensions to {cheri_base_ext_name}.
 
-NOTE: Reserved bits must be 0 in valid capabilities.
+NOTE: Reserved bits must be 0 in tagged capabilities.
 
 === Components of a Capability
 
@@ -521,8 +521,9 @@ malformed    =  !EF && (malformedMSB || malformedLSB)
 NOTE: The check is for malformed _bounds_, so it does not include reserved
 bits!
 
-Capabilities with malformed bounds are always invalid anywhere in the system
-i.e. their tags are always 0.
+It is impossible for a CHERI core to generate a tagged capability with malformed
+bounds, or with any reserved bits set. If such a capability exists then it must
+have been caused by a logic or memory fault.
 
 [#section_special_caps]
 === Special Capabilities
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -525,9 +525,12 @@ It is impossible for a CHERI core to generate a tagged capability with malformed
 bounds, or with any reserved bits set. If such a capability exists then it must
 have been caused by a logic or memory fault.
 
-Capabilities with malformed bounds return both base and top bounds as zero. This
-causes all capability manipulation instructions such as <<CADDI>> to clear the tag
-on the result.
+Capabilities with malformed bounds:
+
+. Return both base and top bounds as zero, which affects instructions like <<GCBASE>>.
+. Cause certain manipulation instructions like <<CADDI>> to always clear the tag of the result.
+
+See specific instruction pages for full details of the effect of malformed capabilities.
 
 [#section_special_caps]
 === Special Capabilities
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -525,6 +525,10 @@ It is impossible for a CHERI core to generate a tagged capability with malformed
 bounds, or with any reserved bits set. If such a capability exists then it must
 have been caused by a logic or memory fault.
 
+Capabilities with malformed bounds return both base and top bounds as zero. This
+causes all capability manipulation instructions such as <<CADDI>> to clear the tag
+on the result.
+
 [#section_special_caps]
 === Special Capabilities
 
--- a/src/insns/atomic_exceptions.adoc
+++ b/src/insns/atomic_exceptions.adoc
@@ -28,7 +28,7 @@ reported in the CAUSE field of <<mtval>> or <<stval>>:
 | Seal violation        | Authority capability is sealed
 | Permission violation  | Authority capability does not grant <<r_perm>> or <<w_perm>>, or the AP field could not have been produced by <<ACPERM>>
 | Invalid address violation  | The effective address is invalid according to xref:section_invalid_addr_conv[xrefstyle=short]
-| Length violation      | At least one byte accessed is outside the authority capability bounds
+| Length violation      | At least one byte accessed is outside the authority capability bounds, which includes the case where the capability is <<section_cap_malformed,malformed>> as the bounds decode as zero
 |==============================================================================
 
 :!cap_atomic:
--- a/src/insns/atomic_exceptions.adoc
+++ b/src/insns/atomic_exceptions.adoc
@@ -28,7 +28,7 @@ reported in the CAUSE field of <<mtval>> or <<stval>>:
 | Seal violation        | Authority capability is sealed
 | Permission violation  | Authority capability does not grant <<r_perm>> or <<w_perm>>, or the AP field could not have been produced by <<ACPERM>>
 | Invalid address violation  | The effective address is invalid according to xref:section_invalid_addr_conv[xrefstyle=short]
-| Length violation      | At least one byte accessed is outside the authority capability bounds, or the capability is <<section_cap_malformed,malformed>>
+| Length violation      | At least one byte accessed is outside the authority capability bounds, or the capability has <<section_cap_malformed,malformed>> bounds
 |==============================================================================
 
 :!cap_atomic:
--- a/src/insns/atomic_exceptions.adoc
+++ b/src/insns/atomic_exceptions.adoc
@@ -21,10 +21,12 @@ When these instructions cause CHERI exceptions, _CHERI data fault_
 is reported in the TYPE field and the following codes may be
 reported in the CAUSE field of <<mtval>> or <<stval>>:
 
-[width="50%",options=header,cols="2,^1",align=center]
+<<<
+
+[%autowidth,options=header,align=center]
 |==============================================================================
 | CAUSE                 | Reason
-| Tag violation         | Authority capability tag set to 0
+| Tag violation         | Authority capability tag set to 0, or has any reserved bits set
 | Seal violation        | Authority capability is sealed
 | Permission violation  | Authority capability does not grant <<r_perm>> or <<w_perm>>, or the AP field could not have been produced by <<ACPERM>>
 | Invalid address violation  | The effective address is invalid according to xref:section_invalid_addr_conv[xrefstyle=short]
--- a/src/insns/atomic_exceptions.adoc
+++ b/src/insns/atomic_exceptions.adoc
@@ -28,7 +28,7 @@ reported in the CAUSE field of <<mtval>> or <<stval>>:
 | Seal violation        | Authority capability is sealed
 | Permission violation  | Authority capability does not grant <<r_perm>> or <<w_perm>>, or the AP field could not have been produced by <<ACPERM>>
 | Invalid address violation  | The effective address is invalid according to xref:section_invalid_addr_conv[xrefstyle=short]
-| Length violation      | At least one byte accessed is outside the authority capability bounds, which includes the case where the capability is <<section_cap_malformed,malformed>> as the bounds decode as zero
+| Length violation      | At least one byte accessed is outside the authority capability bounds, or the capability is <<section_cap_malformed,malformed>>
 |==============================================================================
 
 :!cap_atomic:
--- a/src/insns/atomic_exceptions.adoc
+++ b/src/insns/atomic_exceptions.adoc
@@ -26,7 +26,7 @@ reported in the CAUSE field of <<mtval>> or <<stval>>:
 | CAUSE                 | Reason
 | Tag violation         | Authority capability tag set to 0
 | Seal violation        | Authority capability is sealed
-| Permission violation  | Authority capability does not grant <<r_perm>> or <<w_perm>>
+| Permission violation  | Authority capability does not grant <<r_perm>> or <<w_perm>>, or the AP field could not have been produced by <<ACPERM>>
 | Invalid address violation  | The effective address is invalid according to xref:section_invalid_addr_conv[xrefstyle=short]
 | Length violation      | At least one byte accessed is outside the authority capability bounds
 |==============================================================================
--- a/src/insns/cadd_32bit.adoc
+++ b/src/insns/cadd_32bit.adoc
@@ -34,7 +34,7 @@ Encoding::
 include::wavedrom/cadd.adoc[]
 
 NOTE: <<CADD>> with `rs2=x0` is decoded as <<CMV>> instead, the key
-difference being that tagged and sealed capabilities do not have their tag
+difference being that tagged capabilities cannot their tag
 cleared by <<CMV>>.
 
 Description::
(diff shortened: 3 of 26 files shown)


src/insns/scbnds_32bit.adoc
```diff
@@ -42,6 +42,8 @@ tag is 0  or `cs1` is sealed.
 +
 `immediate = ZeroExtend(s ? uimm<<4 : uimm)`
 
+include::malformed_return_0.adoc[]
```
arichardson, line 45: Should this be "returns zero-bounds capability" or did you mean treat input as untagged?

src/insns/scbndsr_32bit.adoc
```diff
@@ -29,6 +29,8 @@ if its bounds exceed `cs1` 's bounds, `cs1` 's tag is 0 or `cs1`  is sealed.
 Exceptions::
 include::require_cre.adoc[]
 
+include::malformed_return_0.adoc[]
```
arichardson, line 32: same as above.

src/cap-description.adoc
```diff
@@ -521,8 +521,9 @@ malformed    =  !EF && (malformedMSB || malformedLSB)
 NOTE: The check is for malformed _bounds_, so it does not include reserved
 bits!
 
-Capabilities with malformed bounds are always invalid anywhere in the system
-i.e. their tags are always 0.
+It is impossible for a CHERI core to generate a tagged capability with malformed
+bounds, or with any reserved bits set. If such a capability exists then it must
+have been caused by a logic or memory fault.
```
arichardson, line 526: How about adding a note here that if the bounds of such a capability are requested they must decode as 0/0 and address manipulation instructions (CADD(I)) detag the result when such a capability is received as input.

arichardson (commented):
I believe the scbnds note is wrong? Otherwise this sounds good to me.

jamie-melling:
Will we change `SCSS` to return false if either cs1 or cs2 is malformed?  Also, I think we need to include a malformed check on cs1 for `CBLD` as then you could use that to get infinite bounds.

src/cap-description.adoc
tariqkurd-repo: ok - done - please have another look

src/insns/scbnds_32bit.adoc
tariqkurd-repo: ok - done - please have another look

src/insns/scbndsr_32bit.adoc
tariqkurd-repo: ok - done - please have another look

tariqkurd-repo:
> Will we change `SCSS` to return false if either cs1 or cs2 is malformed? Also, I think we need to include a malformed check on cs1 for `CBLD` as then you could use that to get infinite bounds.

yes - I've fixed both of these

jamie-melling:
Did we also mention that if you try to dereference a capability with malformed bounds, it would give a length exception?  I think it is implied by the sentence that malformed bounds are treated as 0, but it would be good to have it explicitly stated.

tariqkurd-repo:
> Did we also mention that if you try to dereference a capability with malformed bounds, it would give a length exception? I think it is implied by the sentence that malformed bounds are treated as 0, but it would be good to have it explicitly stated.

I've added that statement into all load/stores/atomics

src/cap-description.adoc
```diff
@@ -521,8 +521,13 @@ malformed    =  !EF && (malformedMSB || malformedLSB)
...
+
+Capabilities with malformed bounds return both base and top bounds as zero. This
+causes all capability manipulation instructions such as <<CADDI>> to clear the tag
+on the result.
```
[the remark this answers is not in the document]
tariqkurd-repo: I was wondering about that, I wasn't convinced that the two statements were equivalent.
Fix coming up.....

src/insns/atomic_exceptions.adoc
```diff
@@ -26,9 +26,9 @@ reported in the CAUSE field of <<mtval>> or <<stval>>:
 | CAUSE                 | Reason
 | Tag violation         | Authority capability tag set to 0
 | Seal violation        | Authority capability is sealed
-| Permission violation  | Authority capability does not grant <<r_perm>> or <<w_perm>>
+| Permission violation  | Authority capability does not grant <<r_perm>> or <<w_perm>>, or the AP field could not have been produced by <<ACPERM>>
 | Invalid address violation  | The effective address is invalid according to xref:section_invalid_addr_conv[xrefstyle=short]
-| Length violation      | At least one byte accessed is outside the authority capability bounds
+| Length violation      | At least one byte accessed is outside the authority capability bounds, which includes the case where the capability is <<section_cap_malformed,malformed>> as the bounds decode as zero
```
PRugg-Cap, line 31: I'd cut "as the bounds decode as zero": I think it just confuses things, though in this case it is accurate: no accesses are in bounds for [0, 0)

src/insns/malformed_clear_tag.adoc
```diff
@@ -0,0 +1,2 @@
+NOTE: This instruction sets `cd.tag=0` if `cs1` 's bounds are <<section_cap_malformed,malformed>>, as malformed bounds
+cause both the base and top bounds to be zero which is not legal in a tagged capability.
```
PRugg-Cap, line 2: The base and top bounds can be zero in a tagged capability.

src/insns/malformed_no_check.adoc
```diff
@@ -0,0 +1,3 @@
+NOTE: This instruction does not perform a <<section_cap_malformed,malformed>>
+bounds check, and so can propagate a <<section_cap_malformed,malformed>>
+tagged capability.
```
PRugg-Cap, line 3: If we're specifying this for every instruction, is there a reason bounds are special. Why not also say the same about reserved bits and permission bits?

src/insns/scbnds_32bit.adoc
PRugg-Cap: There's something odd here: I think we want to say the bounds check fails. Otherwise a malformed cap should allow `SCBNDS` to address zero and length zero.

src/insns/scss_32bit.adoc
```diff
@@ -25,8 +25,8 @@ Description::
 `rd` is set to 1 if the tag of capabilities `cs1` and `cs2` are equal and the
 bounds and permissions of `cs2` are a subset of those of `cs1`.
 
-Capabilities with <<section_cap_malformed,malformed bounds>> are considered to
-have a base and top of 0.
+NOTE: If either `cs1` 's or `cs2` 's bounds are <<section_cap_malformed,malformed>>
+then the base and top bounds decode as zero, which causes this instruction to return zero.
```
PRugg-Cap, line 29: Again, not quite the same thing. If the bounds just decoded as zero, then we should be happy to say that a malformed cap is a subset of another malformed cap.

src/cap-description.adoc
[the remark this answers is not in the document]
tariqkurd-repo: fixed

src/insns/malformed_clear_tag.adoc
tariqkurd-repo: fixed

src/insns/scss_32bit.adoc
tariqkurd-repo: true

src/insns/malformed_no_check.adoc
tariqkurd-repo: good point

src/insns/atomic_exceptions.adoc
tariqkurd-repo: done

src/insns/atomic_exceptions.adoc
```diff
@@ -26,9 +26,9 @@ reported in the CAUSE field of <<mtval>> or <<stval>>:
 | CAUSE                 | Reason
 | Tag violation         | Authority capability tag set to 0
 | Seal violation        | Authority capability is sealed
-| Permission violation  | Authority capability does not grant <<r_perm>> or <<w_perm>>
+| Permission violation  | Authority capability does not grant <<r_perm>> or <<w_perm>>, or the AP field could not have been produced by <<ACPERM>>
 | Invalid address violation  | The effective address is invalid according to xref:section_invalid_addr_conv[xrefstyle=short]
-| Length violation      | At least one byte accessed is outside the authority capability bounds
+| Length violation      | At least one byte accessed is outside the authority capability bounds, or the capability is <<section_cap_malformed,malformed>>
```
PRugg-Cap, line 31: "has malformed bounds"?

src/insns/jalr_32bit.adoc
```diff
@@ -51,7 +51,7 @@ reported in the CAUSE field of <<mtval>> or <<stval>>:
 | CAUSE                 | pass:attributes,quotes[{cheri_int_mode_name}] | pass:attributes,quotes[{cheri_cap_mode_name}] | Reason
 | Tag violation         |      | ✔     | `cs1` has tag set to 0
 | Seal violation        |      | ✔     | `cs1` is sealed and the immediate is not 0
-| Permission violation  |      | ✔     | `cs1` does not grant <<x_perm>>
+| Permission violation  |      | ✔     | `cs1` does not grant <<x_perm>>, or the AP field could not have been produced by <<ACPERM>>
 | Invalid address violation  | ✔    | ✔     | The target address is invalid according to xref:section_invalid_addr_conv[xrefstyle=short]
 | Length violation      | ✔    | ✔     | Minimum length instruction is not within the target capability's bounds.
```
jamie-melling, line 56: Malformed check for capability pointer mode?

tariqkurd-repo: that's included in the length violation case - but I've added it anyway

src/insns/scss_32bit.adoc
```diff
@@ -25,8 +25,8 @@ Description::
 `rd` is set to 1 if the tag of capabilities `cs1` and `cs2` are equal and the
 bounds and permissions of `cs2` are a subset of those of `cs1`.
 
-Capabilities with <<section_cap_malformed,malformed bounds>> are considered to
-have a base and top of 0.
+NOTE: If either `cs1` 's or `cs2` 's bounds are <<section_cap_malformed,malformed>>
+then instruction returns zero.
```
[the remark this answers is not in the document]
tariqkurd-repo: added

src/insns/atomic_exceptions.adoc
tariqkurd-repo: done

src/insns/scss_32bit.adoc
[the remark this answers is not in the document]
tariqkurd-repo: I'm leaning to having any reserved bits set to cause detagging, so take a tag violation

src/insns/malformed_no_check.adoc
```diff
@@ -0,0 +1,2 @@
+NOTE: This instruction can propagate tagged capabilities which are <<section_cap_malformed,malformed>>,
```
PRugg-Cap, line 1: "are malformed" -> "have malformed bounds"

jrtc27:
Why are we putting so much effort into specifying what happens if the impossible happens? If you observe memory corruption, all bets are off anyway, your system is no longer working. Specify that it should be impossible to create a malformed cap; if you then observe one due to an implementation or vendor extension bug, it’s in violation of the spec, and if you observe one due to a random uncorrected bitflip then your system has failed. I am concerned that by putting all kinds of specifications on what happens when the impossible happens we’re complicating implementations by adding logic to deal with cases that are, in the formal specification, unreachable. As an example, RISC-V has all the WARL fields in CSRs, and doesn’t go on to say what happens if a bit flip turns the field into an illegal value. It just assumes that won’t happen, and leaves it completely unspecified if it does.

PRugg-Cap:
@jrtc27 (See discussion [here](https://github.com/riscv/riscv-cheri/pull/279#issuecomment-2152619663)). I agree we could probably have left it unspecified, though a bitflip or reading uninitialised memory seem significantly more likely than a CSR bit flip. I don't think anywhere else is there an invariant the spec tries to maintain over all of memory.

> cases that are, in the formal specification, unreachable.

This does have to be addressed in the formal specification, at least as an assert. One practical issue others (e.g. Dapeng) have run into is that it's difficult to prove equivalence if an implementation differs from the spec on this. Since the implementation does have to do *something*, we may as well standardise what that something is. We could go less strict though.

tariqkurd-repo:
> Why are we putting so much effort into specifying what happens if the impossible happens? If you observe memory corruption, all bets are off anyway, your system is no longer working. Specify that it should be impossible to create a malformed cap; if you then observe one due to an implementation or vendor extension bug, it’s in violation of the spec, and if you observe one due to a random uncorrected bitflip then your system has failed. I am concerned that by putting all kinds of specifications on what happens when the impossible happens we’re complicating implementations by adding logic to deal with cases that are, in the formal specification, unreachable. As an example, RISC-V has all the WARL fields in CSRs, and doesn’t go on to say what happens if a bit flip turns the field into an illegal value. It just assumes that won’t happen, and leaves it completely unspecified if it does.

there's a big difference between the RISC-V spec which has a lot of unspecified behaviour, and this spec. What we *could* do is say that malformed bounds etc. cause implementation defined results in the open source spec, and then all these extra details go into a company internal spec doc, but we've chosen to specify the cases in the open source.

it's also very useful - because a faulty SoC could return a tag with incorrect data in a weird error case - and we're tieing down what would actually happen in that case.

it's also true that alpha particle strikes causing bit flips, or voltage dropout when the vector unit gets going and draws a lot of current are more common than you might think, which is why all the server chips have RAS, ECC etc. features. When these errors do occur, it means we've defined what he behaviour should be.

of course if an error causes a tagged cap to have _different_ bounds then we still wouldn't know - at that point we'd need to start adding ECC to the buses to detect it, which is definitely out of scope, but yuou might that that for RV64 you could use some of the reserved fields for ECC checking the bounds and address. That could be a future extension.

tariqkurd-repo:
> @jrtc27 (See discussion [here](https://github.com/riscv/riscv-cheri/pull/279#issuecomment-2152619663)). I agree we could probably have left it unspecified, though a bitflip or reading uninitialised memory seem significantly more likely than a CSR bit flip. I don't think anywhere else is there an invariant the spec tries to maintain over all of memory.
> 
> > cases that are, in the formal specification, unreachable.
> 
> This does have to be addressed in the formal specification, at least as an assert. One practical issue others (e.g. Dapeng) have run into is that it's difficult to prove equivalence if an implementation differs from the spec on this. Since the implementation does have to do _something_, we may as well standardise what that something is. We could go less strict though.

this is the other useful aspect - for formal environments it's very hard to constrain the input to be valid tagged caps only. With this spec we can simply have _any_ input and a deterministic output.

PRugg-Cap:
@jrtc27 The other sneaky advantage of this approach is that we maximise the chance of being forwards compatible (security-wise) with another extended CHERI implementation on the same SoC.

src/insns/cadd_32bit.adoc
```diff
@@ -34,7 +34,7 @@ Encoding::
 include::wavedrom/cadd.adoc[]
 
 NOTE: <<CADD>> with `rs2=x0` is decoded as <<CMV>> instead, the key
-difference being that tagged and sealed capabilities do not have their tag
+difference being that tagged capabilities cannot their tag
```
arichardson, line 37:
```suggestion
difference being that tagged capabilities cannot have their tag
```

arichardson (approved):

Issue #290, opened 2024-06-18
RV32 Bounds checking when exponent = 0


buxtonpaul:
It appears that the specification is not quite correct in the handling of MXLEN==32 bounds when exponent is zero.
MW = 10
EW = 5
Consider a case 

length= 0x7
base = 0x800089d0

Breaking this down to pesbt bits
B = 0x1d0
B + length = 0x1d7
T = T8,T[7:0] = 1, 0xd7
E = 0

Reconstructing T those :-

B=0x1d0
LCOut = T[7:0] < B[7:0] ? 1:0 = 0
L_MSB = T8 = 1

Reconstituting top two bits 
T[MW-1:MW-2] = B[MW-1:MW-2] +LCout + MSB
= T[9:8] = B[9:8] + 1
= T[9:8] = 1 + 1
therefore becomes  = 0x2d7 which is not correct!

I think this is due to the fact that in MXLEN==32 only 1 bit of T is lost
I believe the correct behaviour should be that 
LCOut = T[8:0] < B[8:0] ? 1:0 = 0
T[MW-1] = B[MW-1] +LCout 
Note that T8 is just passed through as bit 8 of T.

If this sounds reasonable I can try and reformulate it as a MR to update the spec.

Paul

PRugg-Cap:
Nice catch.

> I think this is due to the fact that in MXLEN==32 only 1 bit of T is lost

Agreed.

> I believe the correct behaviour should be that
> LCOut = T[8:0] < B[8:0] ? 1:0 = 0
> T[MW-1] = B[MW-1] +LCout

I need to give this some more thought.

PRugg-Cap:
I think the issue is more fundamental. The name `T8` got picked up, but I really think it's meant to be `L8`, i.e. the MSB of the length that is normally implied to be 1 when we have an internal exponent. I was looking through the paper, but I can't actually find any reference to using a bit this way, so I'm losing track of where the idea came from so that I can check what the original intention was.

In any case, the current decoding specification makes a lot more sense when interpreting the bit that way.

We *could* encode the top bit of `T`, and amend things as you suggest. `T8` is just `L8 ^ B8 ^ LCOut`, and then I think it's equivalent to just do the subtraction on the 9 bits and carry into the 10th. I think `L8` is much clearer though!

In your example, the `L8` of that cap would be 0, so it all works as expected.

To summarise, I think there's no actual bug here, but `T8` should definitely get renamed to `L8`.

[...]

PR #291, opened 2024-06-20
Replace references to T8 with L8 in the bounds encoding, and provide …

buxtonpaul:
The field T8 is named incorrectly and should be L8 to indicate the field contains bit 8 of the length. Which is used to compute then top field for the capability. 

Resolves issue #(290) https://github.com/riscv/riscv-cheri/issues/290


src/cap-description.adoc
```diff
@@ -456,15 +456,18 @@ alignment boundaries.
 
 The EF bit selects between two cases:
 
-1. EF = 1: The exponent is 0 for regions less than 2^MW-2^ bytes long
+1. EF = 1: The exponent is 0 for regions less than 2^MW-2^ bytes long. L~8~ is 
+used to encode the MSB of the lenght and is added to B along with T[MW-3:0] to
```
tariqkurd-repo, line 460:
```suggestion
used to encode the MSB of the length and is added to B along with T[MW-3:0] to
```

buxtonpaul:
I have re-pushed to the branch fixing both the length typo as well as any line ending issues(I hope)

buxtonpaul:
It was an amend commit which replaced the original

andresag01 (approved):

PRugg-Cap:
Yep, looks like the right fix

PR #299, opened 2024-07-01
fix issues 295, 297

tariqkurd-repo:
This does include a subtle spec change for both CBLD and SCSS

### DIFF
--- a/src/insns/cbld_32bit.adoc
+++ b/src/insns/cbld_32bit.adoc
@@ -25,6 +25,7 @@ Copy `cs2` to `cd` and set `cd.tag` to 1 if
 
 . `cs1.tag` is set, and
 . `cs1` 's bounds are not <<section_cap_malformed,malformed>>, and all reserved fields are zero, and
+. `cs1` 's permissions could have been legally produced by <<ACPERM>>, and
 . `cs1` is not sealed, and
 . `cs2` 's permissions and bounds are equal to or a subset of `cs1` 's, and
 . `cs2` 's bounds are not <<section_cap_malformed,malformed>>, and all reserved fields are zero, and
--- a/src/insns/scss_32bit.adoc
+++ b/src/insns/scss_32bit.adoc
@@ -25,8 +25,13 @@ Description::
 `rd` is set to 1 if the tag of capabilities `cs1` and `cs2` are equal and the
 bounds and permissions of `cs2` are a subset of those of `cs1`.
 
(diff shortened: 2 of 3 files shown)


arichardson (approved):
LGTM, but I'd like @jrtc27 to take a look too.

src/insns/cbld_32bit.adoc
```diff
@@ -25,6 +25,7 @@ Copy `cs2` to `cd` and set `cd.tag` to 1 if
 
 . `cs1.tag` is set, and
 . `cs1` 's bounds are not <<section_cap_malformed,malformed>>, and all reserved fields are zero, and
+. `cs1` 's permissions could have been legally produced by <<ACPERM>>, and
```
jrtc27, line 28: Both of these are implied by the tag being set

arichardson: That's true, but I think it still makes sense to write it here explicitly.

tariqkurd-repo: they are - but we're including cases where the memory system has faults in and so can't be relied upon just like in https://github.com/riscv/riscv-cheri/pull/285

so if someone does a rowhammer style attack for example, resulting in a cap which is obviously faulty, then detect it when possible and clear the tag

if the implementer isn't interested in such cases then they don't have to implement the logic

jrtc27: If rowhammer works all bets are off, your hardware has been broken and nothing is known any more. It is outside the scope of the architecture. Otherwise you have to add specific hardware on every single CHERI implementation to deal with cases that are architecturally impossible, which adds area. If you are an implementation that cares about memory corruption, use ECC.

arichardson: I agree with rowhammer being out of scope. I still think it's useful to define a specific outcome for all of the 129/65 input bits that are used by the instruction even if those inputs can never occur in a correct execution. Alternatively, we could change this to explicitly say the outcome is undefined. However, I think for cases that don't add much/any HW complexity having a well defined outcome is nicer.

tariqkurd-repo: yes - this was the main purpose - so that we can 100% specify what happens for all inputs - just like in [the other PR](https://github.com/riscv/riscv-cheri/pull/285) - it's extremely useful for verification.

tariqkurd-repo: merging as it's in line with the previous PR and is very useful for verification

Issue #300, opened 2024-07-01
The operation of C.JALR seems to be wrong


francislaus:
The description of `C.JALR` links to `JALR`. In the `JALR`  the PCC's offset is incremented by 4 and then written to the `cd` register. However, I think for `C.JALR`, it seems necessary to increment the PCC by 2 because the instruction is only 16 bits long.

PR #304, opened 2024-07-01
remove pcc + 4 for the next PC, as it's not always correct

tariqkurd-repo:


tariqkurd-repo:
as @francislaus pointed out, sometimes the use of <<pcc>> + 4 for the next instruction was incorrect, so I've removed all references as they were redundant anyway

tariqkurd-repo:
fixes https://github.com/riscv/riscv-cheri/issues/300

arichardson (approved):

francislaus:
Thanks for doing that change @tariqkurd-repo. I think the description of `C.JAL` does not always not refer to `JAL`. Can we change that?

francislaus:
Ah, I see @arichardson just merged

francislaus:
I was about to suggest "See <<JAL>> for execution of the expanded instruction as shown above." for the descriptions. Can we fit that in somehow?

tariqkurd-repo:
image

Both lines could be the same - the cap mode one couls also refer to the JAL definition.

Issue #303, opened 2024-07-01
M bit should be inverted


jrtc27:
Currently, we have M=1 for capability mode, but M=0 either means integer mode, if supported, or capability mode, if integer mode is not. Moreover, the infinite capability is defined to have M=0, so you have to SCMODE it if you want to derive things from it that should remain in capability mode. It would be much more natural to have M=0 mean capability mode, as the thing that you are likely using if interacting with capabilities, and then the bit only needs to be encoded to represent a non-zero state when you want to use legacy mode, making things unambiguous. This also means the infinite capability would be capability mode. Note that null is unaffected since it is non-X and thus M does not exist there.

tariqkurd-repo:
why is the infinite-cap in integer mode?

I support this change - I think it makes the definition clearer if ti doesn't cause problems.

arichardson:
I guess this means we have to define the PCC reset capability to be != infinite-cap since it has to be integer mode if the core supports hybrid? That's fine by me but probably needs calling out explicitly?

tariqkurd-repo:
For 100% compatibility then it must boot into integer mode, yes.
The mode is orthogonal to the other features of the infinite-cap so I think we can make it work..... I'll give it a go

PR #305, opened 2024-07-02
swap sense of M-bit, so that 0 is cap mode

tariqkurd-repo:
it makes more sense for zero to mean cap mode, as it's the same as not having hybrid mode

### DIFF
--- a/src/attributes.adoc
+++ b/src/attributes.adoc
@@ -67,6 +67,8 @@ endif::[]
 :TAG_RESET_CSR: The tag of the CSR must be reset to zero. The reset values of the metadata and address fields are UNSPECIFIED.
 :REQUIRE_CRE_CSR: Access to this CSR is illegal if CRE for the current mode is zero (see <<section_cheri_disable>>).
 
+:CAP_MODE: 0
+:INT_MODE: 1
 
 ///////////////////////////////////////////////////////////////////////////////
 // Cap definitions
--- a/src/attributes.adoc
+++ b/src/attributes.adoc
@@ -67,8 +67,8 @@ endif::[]
 :TAG_RESET_CSR: The tag of the CSR must be reset to zero. The reset values of the metadata and address fields are UNSPECIFIED.
 :REQUIRE_CRE_CSR: Access to this CSR is illegal if CRE for the current mode is zero (see <<section_cheri_disable>>).
 
-:CAP_MODE: 0
-:INT_MODE: 1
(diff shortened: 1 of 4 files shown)


tariqkurd-repo:
fixes https://github.com/riscv/riscv-cheri/issues/303

src/attributes.adoc
```diff
@@ -67,6 +67,8 @@ endif::[]
 :TAG_RESET_CSR: The tag of the CSR must be reset to zero. The reset values of the metadata and address fields are UNSPECIFIED.
 :REQUIRE_CRE_CSR: Access to this CSR is illegal if CRE for the current mode is zero (see <<section_cheri_disable>>).
 
+:CAP_MODE: 0
+:INT_MODE: 1
```
arichardson, line 71:
```suggestion
:CAP_MODE_VALUE: 0
:INT_MODE_VALUE: 1
```

Maybe something like this to indicate the macro holds the integer M bit value for the mode?

src/cap-description.adoc
```diff
@@ -570,16 +570,14 @@ expanded base is 0 and top is 2^MXLEN^.
 | Reserved | zeros | All reserved fields
 |==============================================================================
 
-If {cheri_default_ext_name} is supported:
-
-* For MXLEN=32, the <<m_bit>> is set to zero (pass:attributes,quotes[{cheri_int_mode_name}]) in the AP field
-* For MXLEN=64, the <<m_bit>> is set to zero (pass:attributes,quotes[{cheri_int_mode_name}]) in a separate M field
+NOTE: <<null-cap>> is not a code capability, and so the CHERI execution mode is not encoded.
```
arichardson, line 573:
```suggestion
NOTE: For MXLEN=64, the <<m_bit>> is encoded outside of AP and is set to zero for <<null-cap>>.
```

I'd just say we encode M as zero? Alternatively, we could add a `M` row to the table and just add a note there `(not present for MXLEN=32)`: e.g.
```
| M        | zero  | No meaning since non-executable (MXLEN=64 only)
```

src/attributes.adoc
tariqkurd-repo: I've done a find-and-replace on all the uses, so will commit that instead

src/cap-description.adoc
tariqkurd-repo: ok done - will commit shortly

arichardson (approved):
One minor suggestion otherwise LGTM.

Issue #312, opened 2024-07-04
Prefetch instruction sections does not mention the various malformed checks


jamie-melling:
It was missed when doing https://github.com/riscv/riscv-cheri/pull/285

PR #315, opened 2024-07-05
clarify when prefetches don't prefetch

tariqkurd-repo:
the malformed cases were missing, and also the cases should be symmetrical between it mode / cap mode which wasn't clear before
fixes https://github.com/riscv/riscv-cheri/issues/312

### DIFF
--- a/src/insns/malformed_shadd_clear_tag.adoc
+++ b/src/insns/malformed_shadd_clear_tag.adoc
@@ -1,0 +1,2 @@
+NOTE: This instruction sets `cd.tag=0` if `cs2` 's bounds are
+<<section_cap_malformed,malformed>>, or if any of the reserved fields are set.
--- a/src/insns/prefetch.i.adoc
+++ b/src/insns/prefetch.i.adoc
@@ -44,13 +44,16 @@ likely to be accessed by an instruction fetch in the near future. The encoding
 is only valid if imm[4:0]=0. The authorising capability for this operation is
 <<ddc>>.
 
-In either mode, instruction does not perform a memory access
+In either mode, PREFETCH.I does not perform a memory access
 if one or more of the following conditions of the authorising capability are met:
+
 * The tag is not set
 * The sealed bit is set
 * No bytes of the cache line requested is in bounds
 * The <<x_perm>> is not set
-* Any reserved bits are set, the permissions could not have been produced by <<ACPERM>> or the bounds are <<section_cap_malformed,malformed>>.
+* Any reserved bits are set
+* The permissions could not have been produced by <<ACPERM>>
+* The bounds are <<section_cap_malformed,malformed>>.
 
 Prerequisites for pass:attributes,quotes[{cheri_cap_mode_name}]::
 Zicbop, {cheri_base_ext_name}
--- a/src/insns/prefetch.i.adoc
+++ b/src/insns/prefetch.i.adoc
@@ -44,16 +44,9 @@ likely to be accessed by an instruction fetch in the near future. The encoding
 is only valid if imm[4:0]=0. The authorising capability for this operation is
 <<ddc>>.
 
-In either mode, PREFETCH.I does not perform a memory access
-if one or more of the following conditions of the authorising capability are met:
-
-* The tag is not set
-* The sealed bit is set
-* No bytes of the cache line requested is in bounds
-* The <<x_perm>> is not set
-* Any reserved bits are set
-* The permissions could not have been produced by <<ACPERM>>
-* The bounds are <<section_cap_malformed,malformed>>
+:prefetch_insn: PREFETCH.I
+:prefetch_perm: <<x_perm>>
+include::prefetch_cap_checks.adoc[]
 
 Prerequisites for pass:attributes,quotes[{cheri_cap_mode_name}]::
 Zicbop, {cheri_base_ext_name}
--- a/src/insns/prefetch.i.adoc
+++ b/src/insns/prefetch.i.adoc
@@ -34,12 +34,7 @@ likely to be accessed by an instruction fetch in the near future. The encoding
 is only valid if imm[4:0]=0. The authorising capability for this operation is
 `cs1`. This instruction does not throw any exceptions. However, following
 <<CHERI_SPEC>>, this instruction does not perform a prefetch if it is
-not authorized by `cs1`. This instruction does not perform a memory access
-if one or more of the following conditions of `cs1` are met:
-* The tag is not set
-* The sealed bit is set
-* No bytes of the cache line requested is in bounds
-* The <<x_perm>> is not set
+not authorized by `cs1`.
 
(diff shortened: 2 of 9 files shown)


src/insns/prefetch.i.adoc
```diff
@@ -49,6 +44,13 @@ likely to be accessed by an instruction fetch in the near future. The encoding
 is only valid if imm[4:0]=0. The authorising capability for this operation is
 <<ddc>>.
 
+In either mode, instruction does not perform a memory access
```
andresag01, line 47: Looks like there is a 'the' missing before 'instruction', or even better if you replace 'instruction' with 'prefetch.i'

src/insns/prefetch.i.adoc
```diff
@@ -49,6 +44,13 @@ likely to be accessed by an instruction fetch in the near future. The encoding
...
+if one or more of the following conditions of the authorising capability are met:
+* The tag is not set
+* The sealed bit is set
+* No bytes of the cache line requested is in bounds
+* The <<x_perm>> is not set
+* Any reserved bits are set, the permissions could not have been produced by <<ACPERM>> or the bounds are <<section_cap_malformed,malformed>>.
```
andresag01, line 53: These would read better if they were separate bullet points

src/insns/prefetch.i.adoc
```diff
@@ -49,6 +44,13 @@ likely to be accessed by an instruction fetch in the near future. The encoding
 is only valid if imm[4:0]=0. The authorising capability for this operation is
 <<ddc>>.
 
+In either mode, instruction does not perform a memory access
+if one or more of the following conditions of the authorising capability are met:
+* The tag is not set
+* The sealed bit is set
```
andresag01, line 50: I would change this for bullet point with "The capability is sealed"

src/insns/prefetch.r.adoc
```diff
@@ -49,6 +44,13 @@ likely to be accessed by a data read (i.e. load) in the near future. The
 encoding is only valid if imm[4:0]=0. The authorising capability for this
 operation is <<ddc>>.
 
+In either mode, instruction does not perform a memory access
+if one or more of the following conditions of the authorising capability are met:
+* The tag is not set
+* The sealed bit is set
+* No bytes of the cache line requested is in bounds
+* The <<r_perm>> is not set
+* Any reserved bits are set, the permissions could not have been produced by <<ACPERM>> or the bounds are <<section_cap_malformed,malformed>>.
```
andresag01, line 53: Please see my other comments.

src/insns/prefetch.w.adoc
```diff
@@ -49,6 +44,14 @@ likely to be accessed by a data write (i.e. store) in the near future.  The
 encoding is only valid if imm[4:0]=0. The authorising capability for this
 operation is <<ddc>>.
 
+In either mode, instruction does not perform a memory access
+if one or more of the following conditions of the authorising capability are met:
+* The tag is not set
+* The sealed bit is set
+* No bytes of the cache line requested is in bounds
+* The <<w_perm>> is not set
+* Any reserved bits are set, the permissions could not have been produced by <<ACPERM>> or the bounds are <<section_cap_malformed,malformed>>.
```
andresag01, line 53: Please see my other comments.

andresag01 (commented):
Looks good. Just minor suggestions

jamie-melling:
Alexandra/Evangelos has also pointed out that SHADD do not have the same malformed/reserved conditions as CADD

tariqkurd-repo:
> Alexandra/Evangelos has also pointed out that SHADD do not have the same malformed/reserved conditions as CADD

that should have another issue filed please

andresag01:
@tariqkurd-repo : I expanded the information to make it clearer instead of using cross-references.

andresag01 (approved):

src/insns/prefetch.w.adoc
```diff
@@ -49,6 +44,17 @@ likely to be accessed by a data write (i.e. store) in the near future.  The
 encoding is only valid if imm[4:0]=0. The authorising capability for this
 operation is <<ddc>>.
 
+In either mode, PREFETCH.W does not perform a memory access
+if one or more of the following conditions of the authorising capability are met:
+
+* The tag is not set
```
arichardson, line 50: Not necessary to fix this before merging, but maybe worth factoring this out into a helper file since the only difference is the permission.

src/insns/malformed_shadd_clear_tag.adoc
```diff
@@ -0,0 +1,2 @@
+NOTE: This instruction sets `cd.tag=0` if `cs2` 's bounds are
```
arichardson, line 1: Should this be a common file to be included from cadd as well, just to keep all these similar instructions in sync.

src/insns/sh4add_32bit.adoc
```diff
@@ -30,10 +30,16 @@ Encoding::
 ....
 
 pass:attributes,quotes[{cheri_cap_mode_name}] Description::
-Increment the address field of `cs2` by `rs1` shifted left by 4 bit positions. Clear the tag if the resulting capability is unrepresentable or `cs2` is sealed.
+Increment the address field of `cs2` by `rs1` shifted left by 4 bit positions
+and write the result to `cd`. The tag bit of the output capability is 0 if
+`cs2` did not have its tag set to 1, the incremented address is outside `cs2`
+'s <<section_cap_representable_check>> or `cs2` is sealed.
+
+include::malformed_shadd_clear_tag.adoc[]
```
arichardson, line 38: Maybe using something like this that can also be reused by cadd?:
```suggestion
and write the result to `cd`.

include::cap_add_representable_check.adoc[]
```

With `cap_add_representable_check.adoc` being:
```
The tag bit of the output capability is 0 if
`cs2` did not have its tag set to 1, the incremented address is outside `cs2`
's <<section_cap_representable_check>> or `cs2` is sealed.

NOTE: This instruction sets `cd.tag=0` if `cs2` 's bounds are
<<section_cap_malformed,malformed>>, or if any of the reserved fields are set.

```

arichardson (approved):
Change looks good just some suggestions for mantainability of the spec.

src/insns/malformed_shadd_clear_tag.adoc
andresag01: There is a small difference because in `cadd` the capability is in `cs1` rather than `cs2`. I guess we can use a variable to change the text slightly just before/after the file is included. I can give it a try

src/insns/prefetch.w.adoc
andresag01: Done

src/insns/sh4add_32bit.adoc
andresag01: The operand is different between these instructions (cadd is cs1 and shxxx is cs2), so the include didn't work very well because adoc did not expand code ticks ` in macros. I will leave it like this

PR #332, opened 2024-07-23
Make CRE not a WARL field

### LINKED ISSUE #331: Does senvcfg.CRE retain its value?
The spec says `senvcfg.CRE` is read-only zero in various dynamic situations (e.g. if `menvcfg.CRE` is zero). However it doesn't say if the underlying value is retained and just *reads* as zero, or if the values is actually mutated to be zero.

I.e. what does this yield?

```
menvcfg.CRE = 1
senvcfg.CRE = 1
menvcfg.CRE = 0
menvcfg.CRE = 1
print(senvcfg.CRE)
```

Timmmm:
Change the CRE bits so they are always writable, but the effective CRE depends on multiple values. This is significantly simpler to specify, avoids an entire cross-CSR WARL field (WARL fields are a pain), and removes some undefined behaviour (do the bits retain their values when they are read-only zero).

Fixes #331

Note, I deleted the text about SXLEN/UXLEN. Those didn't seem fully specified tbh, but I'll need to add them back somehow.

### DIFF
--- a/src/attributes.adoc
+++ b/src/attributes.adoc
@@ -65,7 +65,7 @@ endif::[]
 :non-csrrw-and:  <<CSRRWI>>, <<CSRRS>>, <<CSRRSI>>, <<CSRRC>> and <<CSRRCI>>
 
 :TAG_RESET_CSR: The tag of the CSR must be reset to zero. The reset values of the metadata and address fields are UNSPECIFIED.
-:REQUIRE_CRE_CSR: Access to this CSR is illegal if the <<section_cheri_disable,effective CRE>> for the current privilege is zero.
+:REQUIRE_CRE_CSR: Access to this CSR is illegal if <<section_cheri_disable,CHERI register access is disabled>> for the current privilege.
 
 :CAP_MODE_VALUE: 0
 :INT_MODE_VALUE: 1
--- a/src/attributes.adoc
+++ b/src/attributes.adoc
@@ -65,7 +65,7 @@ endif::[]
 :non-csrrw-and:  <<CSRRWI>>, <<CSRRS>>, <<CSRRSI>>, <<CSRRC>> and <<CSRRCI>>
 
 :TAG_RESET_CSR: The tag of the CSR must be reset to zero. The reset values of the metadata and address fields are UNSPECIFIED.
-:REQUIRE_CRE_CSR: Access to this CSR is illegal if CRE for the current mode is zero (see <<section_cheri_disable>>).
+:REQUIRE_CRE_CSR: Access to this CSR is illegal if the effective CRE for the current privilege is zero (see <<section_cheri_disable>>).
 
 :CAP_MODE_VALUE: 0
 :INT_MODE_VALUE: 1
--- a/src/attributes.adoc
+++ b/src/attributes.adoc
@@ -65,7 +65,7 @@ endif::[]
 :non-csrrw-and:  <<CSRRWI>>, <<CSRRS>>, <<CSRRSI>>, <<CSRRC>> and <<CSRRCI>>
 
 :TAG_RESET_CSR: The tag of the CSR must be reset to zero. The reset values of the metadata and address fields are UNSPECIFIED.
-:REQUIRE_CRE_CSR: Access to this CSR is illegal if the effective CRE for the current privilege is zero (see <<section_cheri_disable>>).
+:REQUIRE_CRE_CSR: Access to this CSR is illegal if the <<section_cheri_disable,effective CRE>> for the current privilege is zero.
 
 :CAP_MODE_VALUE: 0
 :INT_MODE_VALUE: 1
--- a/src/insns/require_cre.adoc
+++ b/src/insns/require_cre.adoc
@@ -1,1 +1,1 @@
-This instruction is illegal if the <<section_cheri_disable,CHERI register access is disabled>> for the current privilege is.
+This instruction is illegal if the <<section_cheri_disable,CHERI register access is disabled>> for the current privilege.
--- a/src/insns/require_cre.adoc
+++ b/src/insns/require_cre.adoc
@@ -1,1 +1,1 @@
-This instruction is illegal if the effective CRE for the current privilege is zero (see <<section_cheri_disable>>).
+This instruction is illegal if the <<section_cheri_disable,effective CRE>> for the current privilege is zero.
--- a/src/insns/require_cre.adoc
+++ b/src/insns/require_cre.adoc
@@ -1,1 +1,1 @@
-This instruction is illegal if CRE for the current privilege mode is zero (see <<section_cheri_disable>>).
+This instruction is illegal if the effective CRE for the current privilege is zero (see <<section_cheri_disable>>).
--- a/src/insns/require_cre.adoc
+++ b/src/insns/require_cre.adoc
@@ -1,1 +1,1 @@
-This instruction is illegal if the <<section_cheri_disable,effective CRE>> for the current privilege is zero.
+This instruction is illegal if the <<section_cheri_disable,CHERI register access is disabled>> for the current privilege is.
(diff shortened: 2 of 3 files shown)


src/attributes.adoc
```diff
@@ -65,7 +65,7 @@ endif::[]
 :non-csrrw-and:  <<CSRRWI>>, <<CSRRS>>, <<CSRRSI>>, <<CSRRC>> and <<CSRRCI>>
 
 :TAG_RESET_CSR: The tag of the CSR must be reset to zero. The reset values of the metadata and address fields are UNSPECIFIED.
-:REQUIRE_CRE_CSR: Access to this CSR is illegal if CRE for the current mode is zero (see <<section_cheri_disable>>).
+:REQUIRE_CRE_CSR: Access to this CSR is illegal if the effective CRE for the current privilege is zero (see <<section_cheri_disable>>).
```
arichardson, line 68: Unrelated to this change but it's a bit odd we use a macro for csrs and include for instructions.

src/insns/require_cre.adoc
```diff
@@ -1 +1 @@
-This instruction is illegal if CRE for the current privilege mode is zero (see <<section_cheri_disable>>).
+This instruction is illegal if the effective CRE for the current privilege is zero (see <<section_cheri_disable>>).
```
arichardson, line 1:
```suggestion
This instruction is illegal if the <<section_cheri_disable,effective CRE>> for the current privilege is zero.
```
Also make this text an xref and drop the `see`?

src/attributes.adoc
arichardson, line 68:
```suggestion
:REQUIRE_CRE_CSR: Access to this CSR is illegal if the <<section_cheri_disable,effective CRE>> for the current privilege is zero.
```
same as below?

src/riscv-hybrid-integration.adoc
```diff
@@ -452,9 +437,9 @@ xref:menvcfgmodereg[xrefstyle=short].
 include::img/menvcfgmodereg.edn[]
 
 The CHERI Register Enable (CRE) bit controls whether less privileged levels can
-perform explicit accesses to CHERI registers.  When CRE=1, CHERI registers can
-be read and written by less privileged levels.  When CRE=0, CHERI registers are
-disabled in less privileged levels as described in
+perform explicit accesses to CHERI registers. When CRE=1 and <<mseccfg>>.CRE=1,
+CHERI registers can be read and written by less privileged levels. When CRE=0,
+CHERI registers are disabled in less privileged levels as described in
```
arichardson, line 442:
```suggestion
perform explicit accesses to CHERI registers. perform explicit accesses to CHERI registers. When the CRE bit is set to 1 and <<mseccfg>>.CRE=1,
CHERI registers can be read and written by less privileged levels. When set to 0,
CHERI registers are disabled in less privileged levels as described in
```

I briefly got confused here whether we are talking about the effective CRE or the CRE bit here, maybe worth being explicit?

src/riscv-hybrid-integration.adoc
```diff
@@ -486,15 +471,9 @@ xref:senvcfgreg[xrefstyle=short].
...
-explicit accesses to CHERI registers.  When CRE=1, CHERI registers can be read
-and written by U-mode.  When CRE=0, CHERI registers are disabled in U-mode as
-described.
-
-* <<senvcfg>>.CRE is read-only-zero if:
-** <<mstatus>>.MBE is not the reset value OR
-* UXLEN<MXLEN OR
-* <<mseccfg>>.CRE==0 OR
-* <<menvcfg>>.CRE==0
+explicit accesses to CHERI registers.  When CRE=1 and <<menvcfg>>.CRE=1 and
```
arichardson, line 474: Same comment as above.

arichardson (approved):
As noted in #332, I'm fine with this change but I'd wait for others to comment before merging. 

A few small comments inline.

buxtonpaul:
Agree this makes implementation simpler (at least in Qemu), and addresses the undefined behaviour regarding if bits are cleared. But there still isn't a way for S/U mode code to know what the effective CRE is as it is dependent on registers they cannot read. Unless they try an instruction and catch the resulting exception.

src/attributes.adoc
tariqkurd-repo: I've typically used includes for multi-line and variables for single line, but there's no hard rule there

src/riscv-hybrid-integration.adoc
```diff
@@ -452,9 +437,9 @@ xref:menvcfgmodereg[xrefstyle=short].
 include::img/menvcfgmodereg.edn[]
 
 The CHERI Register Enable (CRE) bit controls whether less privileged levels can
-perform explicit accesses to CHERI registers.  When CRE=1, CHERI registers can
-be read and written by less privileged levels.  When CRE=0, CHERI registers are
-disabled in less privileged levels as described in
+perform explicit accesses to CHERI registers. perform explicit accesses to CHERI registers. When the CRE bit is set to 1 and <<mseccfg>>.CRE=1,
```
tariqkurd-repo, line 440:
```suggestion
perform explicit accesses to CHERI registers. When the CRE bit is set to 1 and <<mseccfg>>.CRE=1,
```

buxtonpaul:
> > Agree this makes implementation simpler (at least in Qemu), and addresses the undefined behaviour regarding if bits are cleared. But there still isn't a way for S/U mode code to know what the effective CRE is as it is dependent on registers they cannot read. Unless they try an instruction and catch the resulting exception.
> 
> That’s what the device tree or ACPI whatever is for.
That makes sense, I guess there aren't many use cases for dynamically changing the CRE for a running system.

Timmmm:
I've updated it so now the text talks about "CHERI register access" being enabled/disabled rather than the `CRE` bits or effective CRE, because the former takes account of `mstatus.MBE` and `XLEN<MXLEN` which can also disable CHERI register access. Hopefully this makes sense!

arichardson (approved):

src/insns/require_cre.adoc
```diff
@@ -1 +1 @@
-This instruction is illegal if CRE for the current privilege mode is zero (see <<section_cheri_disable>>).
+This instruction is illegal if the <<section_cheri_disable,CHERI register access is disabled>> for the current privilege is.
```
andresag01, line 1:
```suggestion
This instruction is illegal if the <<section_cheri_disable,CHERI register access is disabled>> for the current privilege.
```

src/riscv-hybrid-integration.adoc
```diff
@@ -79,12 +79,10 @@ In both encodings:
...
-CRE bits in <<mseccfg>>, <<menvcfg>>, and <<senvcfg>> as follows:
+The current CHERI execution mode is given by the <<m_bit>> of the <<pcc>> and the <<section_cheri_disable,CHERI register access settings>> as follows:
 
-* The Mode is pass:attributes,quotes[{cheri_cap_mode_name}] when the <<m_bit>> of the <<pcc>> is {CAP_MODE_VALUE}, *and* the effective
-CRE=1 for the current privilege level
-* The Mode is pass:attributes,quotes[{cheri_int_mode_name}] when the effective CRE=0 for the current privilege level *or* the <<m_bit>> of the <<pcc>> is {INT_MODE_VALUE}
+* The Mode is pass:attributes,quotes[{cheri_cap_mode_name}] when the <<m_bit>> of the <<pcc>> is {CAP_MODE_VALUE}, *and* <<section_cheri_disable,CHERI register access is enabled>> for the current privilege.
+* Otherwise the Mode is pass:attributes,quotes[{cheri_int_mode_name}].
 
 When the <<m_bit>> can be set follows the rules defined by <<ACPERM>>.
```
andresag01, line 87:
```suggestion
When the <<m_bit>> can be set, the rules defined by <<ACPERM>> must be followed.
```

andresag01 (changes requested):

andresag01 (approved):

PR #337, opened 2024-08-01
Introduce GCTYPE to obtain the capability type

arichardson:
While the base CHERI specification only has unsealed capabilities and one type of sentry capability, other extensions (e.g. CherIoT) that will build upon the base CHERI-RISC-V standard have additional types of capabilities. Defining the specification in terms of object types instead of just having unsealed and sentry is one important step towards allowing CherIoT to be a RV32E extension on top of the base standard.

See https://lists.riscv.org/g/tech-tg-cheri/message/6


arichardson:
@DavidChisnall @nwf I haven't yet updated the semantics of sealing but hopefully you agree this is a correct first step.

src/cap-description.adoc
```diff
@@ -258,25 +258,20 @@ references if that SDP bit is set because it "owns" that object.
 |==============================================================================
 
 [#section_cap_sealed]
-==== Sealed (S) Bit
+==== Capability Type (CT) Bit
```
arichardson, line 261: I used capability type instead of object type here since I think the former is more consistent with the rest of the specification.

jrtc27: I agree in using something other than "o(bject )type", which is a misnomer in the post-CInvoke world.

src/insns/wavedrom/gctype.adoc
```diff
@@ -0,0 +1,12 @@
+
+[wavedrom, ,svg]
+....
+{reg: [
+  {bits: 7,  name: 'opcode',  attr: ['7', 'OP=0110011'], type: 8},
+  {bits: 5,  name: 'rd',      attr: ['5', 'dest'], type: 2},
+  {bits: 3,  name: 'funct3',  attr: ['3', 'GCTYPE=000'], type: 8},
+  {bits: 5,  name: 'cs1',     attr: ['5', 'src'], type: 4},
+  {bits: 5,  name: 'funct5',  attr: ['5', 'GCTYPE=00010'], type: 3},
```
arichardson, line 9: Looking at https://github.com/tariqkurd-repo/riscv-opcodes/blob/tariqkurd-repo/master/unratified/rv64_cheri, the getter with value 2 is not allocated right now, so I used that one.

the CSV files still need updating but I was going to wait until we have agreement on this PR.

src/insns/gctype_32bit.adoc
```diff
@@ -0,0 +1,47 @@
...
+
+Encoding::
+include::wavedrom/gctype.adoc[]
+
+Description::
+Decode the architectural capability type from `cs1` and write the result to `rd`.
+It is not required that the input capability `cs1`  has its tag set to 1.
+
+NOTE: While the architectural capability type maps directly to the value of the `CT` capability bit in {cheri_base_ext_name}, future extensions may define an alternate mapping.
+Therefore, software should always use <<GCTYPE>> to obtain the capability type rather than directly reading the high bits of the capability using <<GCHI>>.
```
nwf-msr, line 26: Thanks for calling that out! :+1:

nwf-msr (approved):
LGTM

tariqkurd-repo (approved):

src/insns/wavedrom/gctype.adoc
arichardson: Updated CSV now.

PR #350, opened 2024-08-28
Add GCMODE instruction

### LINKED ISSUE #302: No GCMODE
We have SCMODE but no way to extract it back out. For RV64 one could slice it out of the high bits, but this gets rather ugly on RV32.

andresag01:
Add GCMODE instruction that decodes the mode from an arbitrary capability in an *x* register and writes the value to another *x* register.

GCMODE allows inspecting a capability's CHERI execution mode without the need for software to "manually" decode the capability's metadata using `gchi`.

Fixes #302.

### DIFF
--- a/src/insns/gcmode_32bit.adoc
+++ b/src/insns/gcmode_32bit.adoc
@@ -1,0 +1,30 @@
+<<<
+
+[#GCMODE,reftext="GCMODE"]
+==== GCMODE
+
+Synopsis::
+Capability get CHERI execution mode
+
+Mnemonic::
+`gcmode rd, cs1`
+
+Encoding::
+include::wavedrom/gcmode.adoc[]
+
+Description::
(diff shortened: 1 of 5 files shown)


arichardson (approved):

src/insns/gcmode_32bit.adoc
```diff
@@ -0,0 +1,30 @@
...
+`gcmode rd, cs1`
+
+Encoding::
+include::wavedrom/gcmode.adoc[]
+
+Description::
+Decode the architectural capability CHERI execution mode from `cs1` and write
+the result to `rd`. It is not required that the input capability `cs1`  has its
+tag set to 1.
+
```
tariqkurd-repo, line 19: it's worth mentioning that 0-cap mode, 1-int mode - i.e. clearly state what it returns

andresag01: Fixed

PR #351, opened 2024-08-30
Permute permissions bitmap for acperm/gcperm

### LINKED ISSUE #325: Permute ACPERMS integer operand's bits
As discussed on the mailing list ([here](https://lists.riscv.org/g/tech-tg-cheri/message/6), [here](https://lists.riscv.org/g/tech-tg-cheri/message/11), and [here](https://lists.riscv.org/g/tech-tg-cheri/message/16)) and as part of https://github.com/riscv/riscv-cheri/issues/294 and https://github.com/riscv/riscv-cheri/issues/324, we should permute the current bit pattern used by ACPERM (https://github.com/riscv/riscv-cheri/blob/main/src/img/acperm_bit_field.edn) to leave the bottom-most bits available for extensions.

In RISC-V, an arbitrary value for the least significant four bits of an otherwise all-1s XLEN-length word, `n` in `0xF...Fn`, can be synthesized with a single compressed instruction.  That is, these bits are somewhat more precious than the others.

In more detail, presently, ACPERM's bit pattern's least significant bits are C, R, W, and X.  Experience (between big-CHERI and CHERIoT) suggests that C, R, and X are relatively rarely modified, while W and non-base-specification permissions (load-global, load-mutable) are more often changed as part of exchanging data and structures across trust boundaries.

Therefore, it would be nice for C, R, and X to be moved up to higher bits in the ACPERM bit pattern, with their prior locations left reserved for further extensions (such as the as-yet hypothetical Zcheriot).

andresag01:
This a first attempt at permuting the permissions bitmap used for ACPERM/GCPERM. The main goals are:

* Put 3 reserved bits along with W and LM within bits [4:0]
    * These can be changed using the immediate from `c.andi`
* Put software-defined permissions (SDP) within bits [10:0] (even in the RV64 case!)
    * These can be changed using RVI instructions with 12-bit sign-extended immediates (e.g. `andi`)
* The remainder architecturally-defined permissions (X, ASR and R) above bit 12 because these are rarely changed

**NOTE:** I have not checked the allocator and kernel yet!

This is how the new bitmap looks like:

Screenshot from 2024-08-30 16-22-02

Fixes #325


arichardson:
The allocation of the lower bits makes sense to me, just not quite sure where to place the software perms.
I wonder if we should leave some space before the software permissions and maybe have them be bits 10-13 - this would allow the two RV32 bits to be checked with a single andi while leaving some space for other permissions below. If we want all RV64 software perms in the "zero-extended" andi range we would have to use bits 7-10 which only leaves one bit reserved.

I'd like to hear opinions from @nwf @LawrenceEsswood and @jrtc27 before merging this PR.

jrtc27:
Why make low bits reserved rather than high bits?

nwf-msr:
FWIW:

* I'd be happy with reserving `2:4` rather than `0:2`, if allocating from the LSB upwards is more traditional.

* The most significant use of software defined permissions proper I'm aware of is CheriBSD, which defines two:
   * https://github.com/CTSRD-CHERI/cheribsd/blob/b2ad856aac657e26a9117ea6662289575e6eb514/sys/cheri/cherireg.h#L56
     `CHERI_PERM_SYCALL` is applicable only to eXecutable caps, used to constrain which pieces of code are allowed to make system calls as part of compartmentalization.
   * https://github.com/CTSRD-CHERI/cheribsd/blob/b2ad856aac657e26a9117ea6662289575e6eb514/sys/cheri/cherireg.h#L66
     `CHERI_PERM_SW_VMEM` is applicable to all memory capabilities and conveys the rights to manage the address space (with `mmap` and friends).
     
   Only the latter of those is toggled with anything approaching regularity.

* In CHERIoT, we defined a user permission bit but it's a little strange -- it's over on the sealing root, not on the data or execute roots.  We had intended it for "spicy integers" kinds of use cases -- like file descriptors -- but AFAIK we've never actually written any non-self-test code that uses it.  (And I think we think that sealed caps are the better answer, but I don't pretend to speak for anyone else on the project.)

Anyway, all of that to say, I think `ACPERMS` is justified in _splitting_ its software defined permission field, bringing one or two of them within reach of `ANDI`'s 12-bit immediate, and leaving the rest further afield.

andresag01:
> I wonder if we should leave some space before the software permissions and maybe have them be bits 10-13 - this would allow the two RV32 bits to be checked with a single andi while leaving some space for other permissions below. If we want all RV64 software perms in the "zero-extended" andi range we would have to use bits 7-10 which only leaves one bit reserved.

Yes, the trouble is that bit 11 is the sign bit, so putting the SDPs in 10-13 means that only 1 bit is actually easily accessible with the 12 bit immediates. But I take your point that we are not leaving many reserved bits in the 12 bit immediate. Perhaps we could split the SDP in half for RV64? Not very tidy though!

andresag01:
> Why make low bits reserved rather than high bits? (As in, why 0:2 rather than 2:4?)

No specific reason. Happy to swap them!

andresag01:
> Anyway, all of that to say, I think `ACPERMS` is justified in _splitting_ its software defined permission field, bringing one or two of them within reach of `ANDI`'s 12-bit immediate, and leaving the rest further afield.

Yes! Seems like I posted my message 3 minutes after yours! Happy to do this as well.

arichardson:
How much do we care about GCPERM results being testable using andi vs lui+and? I'd imagine the more performance critical uses would be masking off permissions.

But if we care about micro optimizing checking X+permit_syscall, then X would also need to move down

andresag01:
Here is another option for the bit field:
* Shifted C, LM and W down to bit 0
* Split the SDP permissions for CLEN=64 such that 2 bits are within bits [11:0] and 2 bits are further up

Screenshot from 2024-09-02 11-13-29

LawrenceEsswood:
I am slightly concerned that if the goal is to have the most useful permissions clear-able by `c.li` + `acperm` that we only have 5 bits that can be usefully changed in the `c.li`, as the the 6'th must be set to correctly not clear all other permissions after sign extention. I reckon C is probably the bit that does not belong. Does anyone have data/intuition on whether C can be the least favoured bit for this group?

I would also add: if we just revised `acperm` as an and-not (as has been suggested a few times) then we could:
clear in [0..4] with a  `c.li`+`acperm`
clear in [12..16] with a `c.lui` + `acperm`
clear in [12..30] with a `lui` + `acperm`
clear in [0..10] with a `li` + `acperm`

This not only puts more bits in compressed instruction reach, but also allows addressing the the 19/20 bits (depending on how we feel about sign extension into the top 32 on RV64) with a single uncompressed lui, rather than just 11 bits with li.  With such a revision I think we would still want anything that needs very frequent access in [0..4], followed by [12..30] (which is easily big enough to put all currently proposed permissions).

Also, regardless of the direction we take here, could we try to not split at reserved groups so much? Currently, all permissions are 1 bit and so their grouping does not matter much. If N-bit permissions are introduced, they will want to have N contiguous bits available and splitting the reserved regions makes this harder.

andresag01:
The last one was not very popular so here is another one that:
* Has the same bit field format for CAP=64 and CAP=128
* Leaves 3 reserved bits in [4:0] for CHERIoT
* Moves all SDPs within [10:0]
Screenshot from 2024-09-05 17-12-32

LawrenceEsswood:
Looks pretty good to me. I would say the only downside now is that there are bits in SDP that are in 11-bit range on RV32 but do not actually exist on RV32, and there are still two breaks in reserved bits.

You could always slide SDP and C up, such that SDP straddles the boundary (leaving just two bits of it in 11-bit immediate range), and even R,X,ASR down to meet them. This would mean only two reserved regions (including those where CheriOT wants them), and no bits in the 11-bit zone on RV32 that do not apply.

tariqkurd-repo (approved):

andresag01:
> You could always slide SDP and C up, such that SDP straddles the boundary (leaving just two bits of it in 11-bit immediate range), and even R,X,ASR down to meet them. This would mean only two reserved regions (including those where CheriOT wants them), and no bits in the 11-bit zone on RV32 that do not apply.

Assuming that we would like to keep mostly the same format for RV32 and RV64, then this would still leave a reserved gap between SDP and R/X/ASR because SDPLEN=2 for RV32 and SDPLEN=4 for RV64. Also, the idea is to leave enough space before R/X/ASR to extend the SDP if someone needed it. I personally think we should leave it like this because also anything else are really tiny optimizations.

@LawrenceEsswood @arichardson : Should we just merge it as it currently stands in this PR?

tariqkurd-repo:
are you ok to merge this @arichardson ?

PR #354, opened 2024-09-06
Allow XLEN writes to new CSRs in Zcherihybrid

### LINKED ISSUE #352: Clarify writing non-extended CLEN CSRs in integer pointer mode
In section 5.4.4 CSR Instructions, there is this section:

> Accessing a capability CSR other than an extended CSR in Integer Pointer Mode results in an illegal instruction exception. These CSRs are listed in Table 25.

Where Table 25 lists: dddc, mtdc, stdc, vstdc, ddc

Is it correct for these CSRs to raise an exception? This seems inconsistent with accessing other CLEN CSRs in integer pointer mode. If it is correct, what is the motivation behind it?

andresag01:
Remove outdated requirement preventing XLEN writes to new CHERI CSRs added in Zcherihybrid: dddc, ddc, mtdc, stdc, vstdc. The XLEN write rules for these CSRs are the same rules that already exist for regular, extended RISC-V CSRs like mscratch.

Fixes #352


jrtc27:
I mean if anything it would make more sense to have these always be CLEN-wide accesses.

andresag01:
> I mean if anything it would make more sense to have these always be CLEN-wide accesses.

This is the alternative i.e. to make them read/write CLEN regardless of the CHERI execution mode.

I went for the other option which is to make them consistent with the behaviour of extended CSR which read/write XLEN when the CHERI execution mode is Integer Pointer Mode. @jrtc27: Do you see any issues why this would not work?

jrtc27:
I mean it works but if you want to do anything useful with them you then need to be in capability mode, which means you'd have to allocate another CSR encoding for the hypothetical future Zcherihybridthathasusefulmemops that lets you do most capability things from integer mode. I don't really see the downside to making it always be CLEN, other than that it's a special case.

jrtc27:
I realise you can switch modes. My point is that it seems stupid to introduce something new that requires you to switch mode to use it in a useful manner. Mode switching is to remain compatible with the existing ISA definition, but these are not existing CSRs.

jrtc27:
Also, what do you even name the integer versions of these?

arichardson:
It is a bit inconsistent that the GPR operand type for the csrrw changes based on the target register, it would be more consistent to just to XLEN writes even if that is not useful.
RISC-V has a lot of encodings allocated to non-useful things (zero register base for loads), so this is not different.

The advantage of making it always CLEN means we don't need a name for the XLEN alias and you can use it without a modeswitch from legacy mode for sandboxing.

I am fine with either solution, I wonder what is easier for hardware implementations?

Timmmm:
> Also, what do you even name the integer versions of these?

Presumably they can have the same name since the "new" names (`mtvecc` etc.) are now just aliases of the old ones (`mtvec`). IMO it would be worth removing the new names altogether but that's another discussion.

In the Sail implementation we currently allow XLEN writes to `ddc` etc. It's more regular and therefore simpler code. We could change it to disallow them fairly easily though but it would be extra code, and then you have more types of CSRs:

* XLEN only
* XLEN and CLEN
* CLEN only

It would be nice to only have `XLEN` and `XLEN and CLEN`, though I get your point that it's kind of a useless feature.

tariqkurd-repo:
I don't think we need new names to allow XLEN writes.
I can see both sides of the argument - having XLEN writes is compeltely regular and having CLEN writes is more useful.

How often do we really think that this will be a problem? Why not update CLEN CSRs from cap mode? `c.modesw` was added for exactly that as Andres pointed out.

So unless something large comes to light I'm in favour of teh current proposal of XLEN writes - to keep the regularity.

tariqkurd-repo (approved):
let's keep regularity and go with the current proposal

arichardson (approved):
I am happy with this change as is. @jrtc27 are you strongly opposed?

jrtc27:
I'm not strongly opposed, but I do think it's a waste of CSR encodings in integer mode. It's also really weird to name the capabilities as capabilities yet interact with them as integers; there is no other case of that in the ISA, as doing so with the GPRs uses the X names.

tariqkurd-repo:
Yes it's a waste of encodings but you wouldn't want them to be allocated to
other things which would be really confusing, so maybe think of it as
reserving them for CHERI?

jrtc27:
Yet you could have them reserved for CHERI and actually useful for CHERI. Given there are CSRs which are always integers in both modes, I don't understand what's wrong with having ones that are always capabilities in both modes.

tariqkurd-repo:
That's a fair point

arichardson:
Ultimately I think this is something the ARC may end up changing again, so it might not be too important which approach we use.

Since you can do useful things with always CLEN without changing modes (e.g. read ddc, scbnds, write dcc) it might be slightly better to go for that approach. But again I don't really mind which approach we take here.

Timmmm:
> Changed such that, on new CSRs, CSRRW always does CLEN access and everything else (CSRRWI, CSRRS, CSRRC, etc) does CLEN read and XLEN write. 

I re-read the Sail and actually I think I was wrong before and that is what we implemented already. Oops. Here's a snippet just in case I still can't read:

```
function set_mtvecc(value : Capability) -> unit = {
  mtvecc = legalize_tvecc(mtvecc, convertInvalidAddr(value));
}

function set_mtvec(value : xlenbits) -> unit = {
  let mtv = legalize_tvec(Mk_Mtvec(mtvecc.address), convertInvalidAddr(value));
  mtvecc = setCapAddrChecked(mtvecc, mtv.bits);
  // TODO: Clear tag if `value + 4 * MAX_MCAUSE` is not representable.
}

function set_ddc(value : Capability) -> unit = {
  ddc = convertInvalidAddrCap(value);
}

// You can see the lack of `dd` name here! -ed
function set_ddc_xlen(value : xlenbits) -> unit = {
  ddc = setCapAddrChecked(ddc, convertInvalidAddr(value));
}

...

// Return true if this is a CLEN-width CSR.
function is_clen_csr(csr : csreg) -> bool = match csr {
  0x305 => true, // mtvecc
...
  0x416 => true, // ddc
...

  _ => false,
}

// Write a capability CSR (is_clen_csr(csr) is true) with a Capability.
function write_clen_csr(csr : csreg, value : Capability) -> unit = {
  let stripTag = boundsMalformed(value) | permsMalformed(value) | not(capReservedValid(value));
  let value = { value with tag = value.tag & not(stripTag) };
  let res : Capability = match csr {
    0x305 => {set_mtvecc(value); mtvecc},
...
    0x416 => {set_ddc(value); ddc},
...
    _ => internal_error(__FILE__, __LINE__, "write_clen_csr for XLEN CSR; this should not be possible"),
  };
}

// Write a capability CSR (is_clen_csr(csr) is true) with xlenbits.
function write_clen_csr_with_xlen(csr : csreg, value : xlenbits) -> unit = {
  let res : Capability = match csr {
    0x305 => {set_mtvec(value); mtvecc},
...
    0x416 => {set_ddc_xlen(value); ddc},
...
    _ => internal_error(__FILE__, __LINE__, "write_clen_csr_with_xlen for XLEN CSR; this should not be possible"),
  };
}

// Read a capability CSR (is_clen_csr(csr) is true) returning the entire capability.
function read_clen_csr(csr : csreg) -> Capability = match csr {
...
  0x305 => mtvecc,
...
  0x416 => ddc,
...
  _ => internal_error(__FILE__, __LINE__, "read_clen_csr for XLEN CSR; this should not be possible"),
}

// execute(CSR) body was moved into separate functions so it can be overridden for CHERI -ed

// Handle CSR ops for CHERI. This special cases CLEN-width CSRs and delegates
// all other CSRs to the default non-CHERI handler.
function handle_csr_op(csr, rs1, rd, is_imm, op, isWrite, isRead) =
  if is_clen_csr(csr) then {
    // Reads of CLEN registers are always CLEN.
    let csr_val = if isRead then read_clen_csr(csr) else zero_reg;
    if isWrite then {
      // The only way to write metadata is with `csrrw` in CapPtrMode mode.
      // Note `op == CSRRW` means `csrrw` or` csrrwi` which is why why also
      // check `not(is_imm)`.
      if effective_cheri_mode() == CapPtrMode & op == CSRRW & not(is_imm) then {
        write_clen_csr(csr, C(rs1));
      } else {
        let rs1_val : xlenbits = if is_imm then zero_extend(rs1) else X(rs1);
        let new_val : xlenbits = match op {
          CSRRW => rs1_val,
          CSRRS => csr_val.address | rs1_val,
          CSRRC => csr_val.address & ~(rs1_val)
        };
        write_clen_csr_with_xlen(csr, new_val);
      };
    };
    if effective_cheri_mode() == CapPtrMode then {
      C(rd) = csr_val;
    } else {
      X(rd) = csr_val.address;
    };
    RETIRE_SUCCESS
    // Original code for existing CSRs.
  } else default_handle_csr_op(csr, rs1, rd, is_imm, op, isWrite, isRead)
```

jrtc27:
> > Changed such that, on new CSRs, CSRRW always does CLEN access and everything else (CSRRWI, CSRRS, CSRRC, etc) does CLEN read and XLEN write.
> 
> I re-read the Sail and actually I think I was wrong before and that is what we implemented already. Oops. Here's a snippet just in case I still can't read:
> 
> ```
> function set_mtvecc(value : Capability) -> unit = {
>   mtvecc = legalize_tvecc(mtvecc, convertInvalidAddr(value));
> }
> 
> function set_mtvec(value : xlenbits) -> unit = {
>   let mtv = legalize_tvec(Mk_Mtvec(mtvecc.address), convertInvalidAddr(value));
>   mtvecc = setCapAddrChecked(mtvecc, mtv.bits);
>   // TODO: Clear tag if `value + 4 * MAX_MCAUSE` is not representable.
> }
> 
> function set_ddc(value : Capability) -> unit = {
>   ddc = convertInvalidAddrCap(value);
> }
> 
> // You can see the lack of `dd` name here! -ed
> function set_ddc_xlen(value : xlenbits) -> unit = {
>   ddc = setCapAddrChecked(ddc, convertInvalidAddr(value));
> }
> 
> ...
> 
> // Return true if this is a CLEN-width CSR.
> function is_clen_csr(csr : csreg) -> bool = match csr {
>   0x305 => true, // mtvecc
> ...
>   0x416 => true, // ddc
> ...
> 
>   _ => false,
> }
> 
> // Write a capability CSR (is_clen_csr(csr) is true) with a Capability.
> function write_clen_csr(csr : csreg, value : Capability) -> unit = {
>   let stripTag = boundsMalformed(value) | permsMalformed(value) | not(capReservedValid(value));
>   let value = { value with tag = value.tag & not(stripTag) };
>   let res : Capability = match csr {
>     0x305 => {set_mtvecc(value); mtvecc},
> ...
>     0x416 => {set_ddc(value); ddc},
> ...
>     _ => internal_error(__FILE__, __LINE__, "write_clen_csr for XLEN CSR; this should not be possible"),
>   };
> }
> 
> // Write a capability CSR (is_clen_csr(csr) is true) with xlenbits.
> function write_clen_csr_with_xlen(csr : csreg, value : xlenbits) -> unit = {
>   let res : Capability = match csr {
>     0x305 => {set_mtvec(value); mtvecc},
> ...
>     0x416 => {set_ddc_xlen(value); ddc},
> ...
>     _ => internal_error(__FILE__, __LINE__, "write_clen_csr_with_xlen for XLEN CSR; this should not be possible"),
>   };
> }
> 
> // Read a capability CSR (is_clen_csr(csr) is true) returning the entire capability.
> function read_clen_csr(csr : csreg) -> Capability = match csr {
> ...
>   0x305 => mtvecc,
> ...
>   0x416 => ddc,
> ...
>   _ => internal_error(__FILE__, __LINE__, "read_clen_csr for XLEN CSR; this should not be possible"),
> }
> 
> // execute(CSR) body was moved into separate functions so it can be overridden for CHERI -ed
> 
> // Handle CSR ops for CHERI. This special cases CLEN-width CSRs and delegates
> // all other CSRs to the default non-CHERI handler.
> function handle_csr_op(csr, rs1, rd, is_imm, op, isWrite, isRead) =
>   if is_clen_csr(csr) then {
>     // Reads of CLEN registers are always CLEN.
>     let csr_val = if isRead then read_clen_csr(csr) else zero_reg;
>     if isWrite then {
>       // The only way to write metadata is with `csrrw` in CapPtrMode mode.
>       // Note `op == CSRRW` means `csrrw` or` csrrwi` which is why why also
>       // check `not(is_imm)`.
>       if effective_cheri_mode() == CapPtrMode & op == CSRRW & not(is_imm) then {
>         write_clen_csr(csr, C(rs1));
>       } else {
>         let rs1_val : xlenbits = if is_imm then zero_extend(rs1) else X(rs1);
>         let new_val : xlenbits = match op {
>           CSRRW => rs1_val,
>           CSRRS => csr_val.address | rs1_val,
>           CSRRC => csr_val.address & ~(rs1_val)
>         };
>         write_clen_csr_with_xlen(csr, new_val);
>       };
>     };
>     if effective_cheri_mode() == CapPtrMode then {
>       C(rd) = csr_val;
>     } else {
>       X(rd) = csr_val.address;
>     };
>     RETIRE_SUCCESS
>     // Original code for existing CSRs.
>   } else default_handle_csr_op(csr, rs1, rd, is_imm, op, isWrite, isRead)
> ```

That's making it mode-dependent like the others?

jrtc27:
> @jrtc27 : Changed such that, on new CSRs, CSRRW always does CLEN access and everything else (CSRRWI, CSRRS, CSRRC, etc) does CLEN read and XLEN write. Does this match your expectation?

Yes, that does. Thanks.

src/riscv-hybrid-integration.adoc
```diff
@@ -242,32 +244,33 @@ field.
 to MXLEN bits according to cite:[riscv-priv-spec] _(3.1.6.2. Base ISA Control in
 mstatus Register)_ and are then written to the destination *x* register.
 
-When <<CSRRW>> is executed on an extended CSR in {cheri_cap_mode_name}:
+When <<CSRRW>> is executed on an extended CSR in {cheri_cap_mode_name}, or
+on a new CSR regardless of the CHERI execution mode:
```
jrtc27, line 248:
```suggestion
on a new capability CSR regardless of the CHERI execution mode:
```

Timmmm:
> That's making it mode-dependent like the others?

Ah you're right.

PR #355, opened 2024-09-07
Introduce capability levels

arichardson:
This adds the local/global mechanism of CHERIv9, augmented with the LoadGlobal feature of CHERIoT. The specification is written in such a way that a future extension could extend the bits used for local/global to more than one.


arichardson:
CC: @nwf @davidchisnall

davidchisnall:
I *think* this is sufficiently expressive for what we need. We currently have orthogonal global and store local, and we use the following combinations:

 - Global, not StoreLocal: heap, globals.
 - Local, not StoreLocal: heap and globals that are temporarily delegated.
 - Local and StoreLocal: Stacks / register save areas.

We don’t use StoreLocal + Global because that would violate stack isolation. This means that, in our encoding, Global is an orthogonal permission but StoreLocal is a dependent permission that can exist only if you have both W and c permissions and so we can squeeze it into a corner of the encoding space.

I believe that his scheme could be represented in this proposal.

That said, I don’t think that this belongs in the core specification. We have shown that our use of local/global works well in the context of an RTOS but there is, as yet, very little demonstration that either this or the generalised multi-level variant is the right thing on larger systems. For example, when we worked through the levels we’d want in CheriBSD, we came up with a bunch of different concepts that would be nice to express:

 - Kernel vs userspace.
 - Compartment-local vs shared.
 - Thread-local (including stack) vs shared.
 - Hypervisor-owned vs guest-owned.
 - Stack, TLS, heap, global, with ordering between them.

A few people have used local-global in different ways in Morello, but this still feels like something that’s too close to research to be in the base spec. Putting it in a separate optional extension and aligning the permissions with CHERIoT so that we can appear as a two-level instantiation of that extension would be fine.

src/cap-description.adoc
```diff
@@ -81,6 +81,19 @@ The byte-address of a memory location is encoded as MXLEN integer value.
 ^| 64      ^| {cap_rv64_addr_width}
 |==============================================================================
 
+[#section_cap_level,reftext="capability level"]
+==== Capability Level (CL)
+Capabilities can be assigned a level, which in conjunction with the <<sl_perm>> and <<ll_perm>> allows enforcing invariants on capability propagation.
+For example, this can be used to ensure that a callee cannot save a copy of the passed in argument or to avoid sharing of compartment-local data between compartments.
```
andresag01, line 87: By "save", do you mean write to memory?

src/cap-description.adoc
```diff
@@ -124,6 +138,37 @@ If a capability grants <<r_perm>> and <<c_perm>>, but no <<lm_perm>>, then a cap
 The rules specified by <<ACPERM>> are followed when <<w_perm>> and <<lm_perm>> are removed, so additional permissions may also be removed.
 Clearing a capability's <<lm_perm>> and <<w_perm>> allows sharing a read-only version of a data structure (e.g. a tree or linked list) without making a copy.
 
+[#sl_perm,reftext="SL-permission"]
+Store Level Permission (SL):: This permission allows limiting the propagation of capabilities using the following logic: capabilities with a <<section_cap_level>> less than the inverse of <<sl_perm>> will be stored with the tag cleared.
+With one bit for SL and CL each, this permission functions as follows:
```
tariqkurd-repo, line 143:
```suggestion
With one bit for SL and <<section_cap_level,CL>> each, this permission functions as follows:
```

src/cap-description.adoc
```diff
@@ -81,6 +81,19 @@ The byte-address of a memory location is encoded as MXLEN integer value.
 ^| 64      ^| {cap_rv64_addr_width}
 |==============================================================================
 
+[#section_cap_level,reftext="capability level"]
+==== Capability Level (CL)
+Capabilities can be assigned a level, which in conjunction with the <<sl_perm>> and <<ll_perm>> allows enforcing invariants on capability propagation.
+For example, this can be used to ensure that a callee cannot save a copy of the passed in argument or to avoid sharing of compartment-local data between compartments.
+
+The Capability Level has two values: when set to 1 the capability is _global_ (in general allowing it to be stored using any authorizing capability), and when set to 0 the capability is _local_, and can only be stored by authorizing capabilities with the <<sl_perm>> set.
+Furthermore, the <<ll_perm>> can be used to restrict loading of _global_ capabilities -- causing the loaded value to be _local_ instead.
```
andresag01, line 90: > causing the loaded value to be _local_ instead

Is this similar to the LoadMutable permission in that the hardware automatically changes the CL of the capability read from memory to be 0 (i.e. local)?

src/cap-description.adoc
```diff
@@ -124,6 +138,37 @@ If a capability grants <<r_perm>> and <<c_perm>>, but no <<lm_perm>>, then a cap
...
+
+Thus, if LG and GL are cleared before delegating a capability then it, and any capability loaded via it (including via indirection),
+may be stored only via capabilities with PERMIT_STORE_LOCAL_CAPABILITY. This
+limits the ability of the delegee to retain capabilities to a delegated data structure or
+part thereof, making it easier to later revoke access to the delegated data structure. In
+the case of loaded sealed capabilities, GL, but not LG, is cleared, making this an exception to the immutability of sealed capabilities. Thus, a sealed capability reached
+via an authority lacking LG may be stored only through a PERMIT_STORE_LOCAL_-
+CAPABILITY authority, as may its unsealed form, but the further authority borne by
+the sealed capability once unsealed is unaltered. This differs from the behavior of
+LM on sealed capabilities. Untagged capabilities are unaffected.
```
andresag01, line 170: Was this paragraph taken from a paper or CHERIv9? It uses a lot of terms not defined in this spec like LG, GL, PERMIT_..., delegating, etc.

Could we reword this if we'd like to keep the general idea?

src/cap-description.adoc
```diff
@@ -124,6 +138,37 @@ If a capability grants <<r_perm>> and <<c_perm>>, but no <<lm_perm>>, then a cap
 The rules specified by <<ACPERM>> are followed when <<w_perm>> and <<lm_perm>> are removed, so additional permissions may also be removed.
 Clearing a capability's <<lm_perm>> and <<w_perm>> allows sharing a read-only version of a data structure (e.g. a tree or linked list) without making a copy.
 
+[#sl_perm,reftext="SL-permission"]
+Store Level Permission (SL):: This permission allows limiting the propagation of capabilities using the following logic: capabilities with a <<section_cap_level>> less than the inverse of <<sl_perm>> will be stored with the tag cleared.
+With one bit for SL and CL each, this permission functions as follows:
+// This awkward wording above allows a multi-bit scheme:
```
andresag01, line 144: Would it help to describe the permission as only allowing writes of other capabilities whose level is greater or equal than our own? (Or at least that how I understood it!)

andresag01 (commented):
Thanks for the proposal! A few questions after an initial read through.

src/img/cap-encoding-xlen32.edn
```diff
@@ -6,11 +6,12 @@
...
 (def right-margin 100)
 (def boxes-per-row 32)
-(draw-column-headers {:height 50 :font-size 26 :labels (reverse ["0" "1" "2" "" "" "" "" "" "" "9" "10" "11" "12" "" "" "" "" "17" "18" "19" "20" "21" "" "" "24" "25" "" "" "" "29" "30" "31"])})
+(draw-column-headers {:height 50 :font-size 26 :labels (reverse ["0" "1" "2" "" "" "" "" "" "" "9" "10" "11" "12" "" "" "" "" "17" "18" "19" "20" "21" "" "23" "24" "25" "" "" "" "29" "30" "31"])})
 
 (draw-box "SDP"         {:span 2})
 (draw-box "AP, M"       {:span 5})
-(draw-box "Reserved"    {:span 4})
+(draw-box "CL"          {:span 1})
+(draw-box "Reserved"    {:span 3})
```
andresag01, line 14: Could you reduce the font size for this image? Some of the labels are overflowing the fields.

src/img/cap-encoding-xlen64.edn
```diff
@@ -6,13 +6,14 @@
...
+(draw-column-headers {:height 50 :font-size 26 :labels (reverse ["0" "2" "3" "" "" "" "13" "14" "16" "17" "" "" "25" "26" "27" "28" "" "42" "43" "44" "" "" "" "" "51" "52" "53" "56" "57" "" "" "63"])})
 
 (draw-box "Reserved"    {:span 4})
 (draw-box "SDP"         {:span 2})
 (draw-box "M"           {:span 1})
-(draw-box "AP"          {:span 4})
-(draw-box "Reserved"    {:span 6})
+(draw-box "AP"          {:span 6})
+(draw-box "CL"          {:span 1})
+(draw-box "Reserved"    {:span 3})
```
andresag01, line 16: Could you reduce the font size for this image? Some of the labels are overflowing the fields.

src/cap-description.adoc
```diff
@@ -81,6 +81,19 @@ The byte-address of a memory location is encoded as MXLEN integer value.
 ^| 64      ^| {cap_rv64_addr_width}
 |==============================================================================
 
+[#section_cap_level,reftext="capability level"]
+==== Capability Level (CL)
```
andresag01, line 85: What instruction could be used to read/write a capability's level?

tariqkurd-repo:
I really don't follow the details of how this feature works, I guess it doesn't help that I'm not familiar with the feature in either CHERI v9 or CHERIoT.

I think we need a better description.

src/cap-description.adoc
arichardson: Yes, should probably make that explicit!

src/cap-description.adoc
arichardson: yes exactly, it behaves like LoadMutable.

src/cap-description.adoc
arichardson: Yeah this was taken from cheriot - I need to update this.

nwf:
There are several intertwined mechanisms here; let me try to go through each in turn, specifically in the context of CHERIoT.

At its most basic, CHERIoT uses two of the permission bits discussed here to differentiate stack memory from globals and heap, and from that builds a few useful properties.  Specifically, the RTOS bootstrap ensures that...

- All pointers to globals and the heap arena are "global" (bear a set `GL` bit) and not store-local (bear a clear `SL` bit).
- All thread initial stack pointers are marked "local" (bear a clear GL bit) and store-local (bear a set `SL` bit).

First and foremost, this architectural feature propagates up to software as the property that pointers derived from stack pointers are confined to the stack (cannot be stored into globals or onto the heap).  The policy asymmetric and the reverse is permitted: pointers to globals or into the heap can be stored to the stacks.

Beyond that static partitioning, it is possible for software to clear the `GL` bit, changing a "global" capability to a "local" one, such that storing it to memory now requires a `SL`-bearing authorizing capability.  This provides software with a "shallow no-capture" mechanism: compartment A can take a "global" capability to a heap object and down-convert it to "local" before handing it to compartment B (as part of a cross-compartment call); by the time compartment B returns, the only locations that may have come to hold copies of that passed capability are pointed to by `SL`-bearing authorities, usually just the stack.  Zeroizing the stack between the caller's high water mark (lowest used address) and the callee's high water mark is sufficient to eliminate most of these automatically. [0]

The addition of the "load global" (`LG`) permission can transmute "shallow no-capture" to "deep no-capture".  By clearing both `GL` and `LG` on a passed pointer, compartment A can ensure that no part of the graph structure transitively referenced by that pointer can be captured by compartment B.  [0] [1]

[0]: One does need to be a little careful about passing `SL`-bearing pointers as arguments in cross-compartment calls; it is possible for the callee to "capture" local capabilities into such spaces, so the caller must take defensive countermeasures.  Usually, though, the stack (automatically managed by the runtime) is the only `SL`-bearing capability passed across compartment boundaries.  See https://github.com/CHERIoT-Platform/cheriot-rtos/blob/5763f0fe7d08e8799af97722c0169db3789e62fc/sdk/core/switcher/entry.S#L780-L832 for the gory details of cross-compartment return handling.

[1]: There is some subtlety around the interaction of sealed capabilities and `LG`.  When a non-`LG`-bearing authority is used to load an unsealed capability, the result will have both `GL` and `LG` cleared.  However, if the load is of a sealed capability, `GL` will be cleared but `LG` will be preserved; that is, `GL` is "outside" the seal while `LG` is "inside".  Tangentially, both `SD` (what the draft calls `W`) and `LM` are "inside" the seal and so will not be cleared when loaded through a non-`LM`-bearing authority.  See https://github.com/CHERIoT-Platform/cheriot-sail/issues/14 and https://github.com/CHERIoT-Platform/cheriot-sail/pull/44 for the full details.

Whew.  I hope that's useful?

LawrenceEsswood:
@davidchisnall The intent is that only the Global & !StoreLocal & !LoadGlobal from your "desirable" table would be lost (specifically because clearing LG in this encoding requires clearing G). The rest should be possible.

@arichardson I am happy with what is currently here, but I still think there is a missing permission that could be specified with the others here that I feel is a logical completion of these control flow bits, and by design costs no extra encoding bits here (As "E" controls all deep properties in this encoding).

`LM` is a deep version of `W`.
`LL` is a deep version of `Level`
??? is a deep version of `SL`

Calling it `LSL` (LoadStoreLocal); it is the bit to allow "Deeply non-capturING". As opposed to `LL` (`LG` as it was previously known) which makes something "Deeply non-capturable".

CHERIoT, simply by construction, does not allow any (unsealed) capabilities on their Heap/Global areas that has the SL bit (@nwf / @davidchisnall you may wish to fact check that statement). Thus, they enforce the transitive property that nothing loaded by these non-`SL` capabilities locations has `SL`. `LSL` would, for this use case, not make any difference because the property it enforces is true by construction.

However, I maintain the ability to quickly make a structure non-capturing (without needing assumptions about its construction) is useful. It allows delegating structures that are otherwise capturing in a way that means we don't have to search them after the delegation. The stack and register saves are very flat structures, and their life-cycles easy to reason about, but in the wider scope of all the different uses we imagine for Local/Global doing things by construction will be harder. We may not wish to mutate structures before declaring we are sure they cannot capture local things. LSL allows flipping a bit and knowing that the required property that a structure cannot be used to captured is true.

@nwf on your [0], I raised this before, but CheriOT's "Zeroizing the stack between the caller's high water mark" being sufficient. This really only works if the callee does not get pointers past that water mark that can SL. Which it does, typically via arguments. Say you wanted to guarantee that a callee does not put pointers to its own stack frame (or another to some non-capture object) within the callers frame.  (without also searching the entirety of the callers frame). how do you currently do that? If you strip SL from such an argument that points to the callers stack, it may still be used to load a capability with SL, which in turn could stash a local thing. Would LSL not be a really useful knob to have here?

Like multiple levels, I don't think anything specified here prevents LSL being introduced at a later time, but I do think it is a logical completion. All the other permissions to do with movement of capabilities can have deep properties, why not SL?

And, because no review can be made without bike-shedding: L for Load and Level is getting confusing. I liked E.

andresag01:
> And, because no review can be made without bike-shedding: L for Load and Level is getting confusing. I liked E.

We also use E for _Exponent_ in the description of the bounds encoding.

Its also interesting that we use Load for LoadMutable and Store for StoreLocal, but then we have Read and Write permissions...

src/cap-description.adoc
arichardson: Added that it can be read with GCPERM.

src/cap-description.adoc
arichardson: It allows storing capabilities with `CLl >= ~SL`. I added a new table for multi-bit. Happy to change this wording to be clearer, but can't think of anything easy to parse.

arichardson:
Split out into a separate extension and allocated RV32 encoding - still needs some cleanup but hopefully better than the previous version.

src/insns/acperm_32bit.adoc
```diff
@@ -50,8 +50,10 @@ The common rules are:
 .. Clear <<m_bit>> unless <<x_perm>> is set
 . <<lm_perm>> cannot be set without <<c_perm>> being set
 .. Clear <<lm_perm>> unless <<c_perm>> is set.
-. <<sl_perm>> cannot be set without <<c_perm>> being set
-.. Clear <<sl_perm>> unless <<c_perm>> is set.
+. <<sl_perm>> cannot be set without <<c_perm>> and <<w_perm>> being set
```
LawrenceEsswood, line 53: Can we weaken this one to just "without <c_perm>"? otherwise we prohibit a future "Elevate Store Local"

src/level-ext.adoc
```diff
@@ -0,0 +1,119 @@
+[#section_ext_cheri_levels]
+== "{cheri_levels_ext_name}" Extension for Capability Levels
+
+{cheri_levels_ext_name} is an optional extension to {cheri_base_ext_name} that adds support for associating a level with capabilities and limiting flow of capabilities to specific memory region subsets.
+This extension allows assigning a level to capabilities, which in conjunction with two new permissions allows enforcing invariants on capability propagation.
+For example, this can be used to ensure that a callee cannot store a copy of the passed in argument to memory or to avoid sharing of compartment-local data between compartments.
+The number of supported levels is configurable, but this specification currently only requires supporting two levels (_local_ and _global_).
+
+=== Capability format changes
+The {cheri_levels_ext_name} adds a new `LVLBITS`-bit field to the <<section_cap_encoding,capability encoding>>, the _<<section_cap_level>>_.
```
tariqkurd-repo, line 10:
```suggestion
{cheri_levels_ext_name} adds a new `LVLBITS`-bit field to the <<section_cap_encoding,capability encoding>>, the _<<section_cap_level>>_.
```

src/level-ext.adoc
```diff
@@ -0,0 +1,119 @@
...
+| 2-3   10+| reserved
+| 4   |   | ✔ |   |   |   |   |   |   | N/A | Data WO
+| 5   | ✔ | ✔ |   |   |   |   |   |   | N/A | Data RW
+| 6-7   10+| reserved
+11+| *Quadrant 1: Executable capabilities*
+11+| bit[0] - <<m_bit>> ({CAP_MODE_VALUE}-{cheri_cap_mode_name}, {INT_MODE_VALUE}-{cheri_int_mode_name})
+|Bits[4:3]| R | W | C | LM | EL | SL  | X | ASR | Mode^1^ |
+| 0-1   | ✔ | ✔ | ✔ | ✔  | ✔  | ∞   | ✔ |  ✔  | Mode^1^  | Execute + ASR (see <<infinite-cap>>)
+| 2-3   | ✔ |   | ✔ | ✔  | ✔  | ∞   |   |     | Mode^1^  | Execute + Data & Cap RO
+| 4-5   | ✔ | ✔ | ✔ | ✔  | ✔  | ∞   | ✔ |     | Mode^1^  | Execute + Data & Cap RW
```
andresag01, line 67: I guess by infinity you mean the maximum level i.e. the most global?

src/img/acperm_bit_field.edn
```diff
@@ -15,7 +15,9 @@
 (draw-box "Reserved" {:span 6})
 (draw-box "SDP"      {:span 4})
 (draw-box "C"        {:span 1})
-(draw-box "Reserved" {:span 3})
+(draw-box "CL"       {:span 1})
+(draw-box "SL"       {:span 1})
+(draw-box "EL"       {:span 1})
```
andresag01, line 20: The indexes and width of these fields need to be adjusted because the figure is not printing correctly.

src/level-ext.adoc
arichardson: Correct, I could also just use "Any" here?

arichardson:
Updated, hopefully this addresses the outstanding comments. I plan to create a new pull request for a "load-store-local" that can be used for a deep variant of store-local, but since this has not been validated in software yet I did not feel like including it in the base spec for now.

src/level-ext.adoc
```diff
@@ -0,0 +1,124 @@
...
+
+NOTE: The <<cap_encoding_xlen64,`MXLEN=64` capability encoding>> diagram shows the layout for `LVLBITS=1`
+
+- For `MXLEN=32` the capability's <<cap_encoding_xlen32,AP>> field is able to encoding these permissions without increasing in size (provided that LVLBITS&#8804;2).
+
+[#section_cap_level,reftext="capability level"]
+==== Capability Level (CL)
+
+With `LVLBITS=1`, the _Capability Level_ can hold two values: when set to 1 the capability is _global_ (in general allowing it to be stored using any authorizing capability), and when set to 0 the capability is _local_, and can only be stored by authorizing capabilities with the <<sl_perm>> set.
+Furthermore, the <<el_perm>> can be used to restrict loading of _global_ capabilities -- causing the loaded value to be _local_ instead.
```
andresag01, line 30:
```suggestion
Furthermore, the <<el_perm>> can be used to restrict loading of _global_ capabilities -- causing the hardware to automatically set the level of loaded capabilities to _local_ instead.
```

src/level-ext.adoc
```diff
@@ -0,0 +1,124 @@
...
+[#section_ext_cheri_multiple_levels]
+=== Extending {cheri_levels_ext_name} to more than two levels
+When `LVLBITS>1`, the behaviour of <<ACPERM>> can no longer use masking to adjust the <<section_cap_level>> or <<sl_perm>>, but instead must perform an integer minimum operation on those `LVLBITS`-wide fields.
+The <<section_cap_level,CL>> field of the resulting capability is set to `min(rs2[CL], cs1[CL])`  (equivalent to `rs2[CL] & cs1[CL]` for `LVLBITS=1`).
+Similarly, <<sl_perm>> is set to `min(rs2[SL], cs1[SL])` (equivalent to `rs2[SL] & cs1[SL]` for `LVLBITS=1`).
+
+When storing capabilities, the <<sl_perm>> checks need to perform a `LVLBITS`-wide integer comparison instead of just testing a single bit.
+Considering for an example `LVLBITS=2`:
+
+[grid=rows,cols="2,3,6"]
```
andresag01, line 113:
```suggestion
[options=header,grid=rows,cols="2,3,6"]
```

src/insns/store_tag_perms.adoc
```diff
@@ -0,0 +1,5 @@
+Tag of the written capability value::
+
+The capability written to memory has the tag set to 0 if the tag of `cs2` is 0 or if the authorizing capability (<<ddc>> or `cs1`) does not grant <<c_perm>>.
++
+The stored tag is also set to zero if the authorizing capability has does not have <<sl_perm>> set but the stored data has a <<section_cap_level>> of 0.
```
andresag01, line 5:
```suggestion
The stored tag is also set to zero if the authorizing capability does not have <<sl_perm>> set but the stored data has a <<section_cap_level>> of 0.
```

andresag01 (commented):
Thanks for the changes! I just made some wording suggestions and typo/formatting fixes.

src/level-ext.adoc
```diff
@@ -0,0 +1,124 @@
...
+== "{cheri_levels_ext_name}" Extension for Capability Levels
+
+{cheri_levels_ext_name} is an extension to {cheri_base_ext_name} that adds support for associating a level with capabilities and limiting the flow of capabilities to specific memory region subsets.
+This extension allows assigning a level to capabilities, which in conjunction with two new permissions allows enforcing invariants on capability propagation.
+For example, this can be used to ensure that a callee can only write a copy of the passed-in argument capability to specific locations in memory (e.g. the callee's stack frame).
+It can also be used to avoid sharing of compartment-local data (such as pointers to a stack object) between compartments.
+
+NOTE: This specification only defines the architectural mechanics of this feature, for further information on how this can be used by software please refer to cite:[cheri-v9-spec].
+
+The number of supported capability levels is implementation-defined, but this specification currently only requires supporting two levels (_local_ and _global_).
```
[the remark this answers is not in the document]
arichardson: I don't think this is true - I tried to be generic everywhere but may have missed something.

andresag01 (approved):

src/level-ext.adoc
andresag01: I am ok with the infinity symbol

src/cap-description.adoc
```diff
@@ -124,6 +138,37 @@ If a capability grants <<r_perm>> and <<c_perm>>, but no <<lm_perm>>, then a cap
...
+// 00 -> allow storing as low as level ~00 = 11 -> only level 3 is ok
+- If this permission (as well as <<c_perm>> and <<w_perm>>) is set then capabilities may be stored via this capability regardless of their associated <<section_cap_level>>.
+- If the permission is not set, then any capability with a <<section_cap_level>> of zero (i.e. _local_), will be stored with the tag cleared.
+
+NOTE: Future extensions may widen this permission and <<section_cap_level>> to more than one bit.
+
+[#ll_perm,reftext="LL-permission"]
+Load Level Permission (LL):: If this permission is not set then any tagged capabilities loaded via this capability will have the <<ll_perm>> cleared and the <<section_cap_level>> restricted to the <<section_cap_level>> of the authorizing capability.
+
+Thus, if LG and GL are cleared before delegating a capability then it, and any capability loaded via it (including via indirection),
```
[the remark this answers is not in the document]
andresag01: This was reworked already. As far as I can tell, all terms are properly defined now.

tariqkurd-repo (approved):

src/level-ext.adoc
```diff
@@ -0,0 +1,156 @@
...
+.{cheri_levels_ext_name} `LVLBITS=1` summary table for loaded capabilities
+[#cap_level_load_summary,width="100%",options=header,align="center",cols="1,1,1,1,1,5"]
+|==============================================================================
+4+|Auth cap field | Data cap field |Notes
+|*R*|*C*|*EL*|*CL*| Tag            |
+|1  | 1 | 1  | X  | X              |Load data capability unmodified
+|1  | 1 | 0  | 1  | X              |Load data capability unmodified
+|1  | 1 | 0  | 0  | 1              |Load data capability with CL=0
+|1  | 1 | 0  | 0  | 0              |Load data capability unmodified
+|==============================================================================
```
andresag01, line 132: I think this table should be like this instead:
| CL | EL | | CL | EL |
|-|-|-|-|-|
| 0 | 0 | | 0 | 0 |
| 0 | 1 | | Unchanged | Unchanged |
| 1 | 0 | | Unchanged | 0 |
| 1 | 1 | | Unchanged | Unchanged |

src/level-ext.adoc
```diff
@@ -0,0 +1,156 @@
...
+
+.{cheri_levels_ext_name} `LVLBITS=1` summary table for stored capabilities
+[#cap_level_store_summary,width="100%",options=header,align="center",cols="1,1,1,1,5"]
+|==============================================================================
+3+|Auth cap field | Data cap field | Notes
+|*W*|*C*|*SL*|*CL*|
+|1  | 1 | 1  | X  | Store data capability unmodified
+|1  | 1 | 0  | 1  | Store data capability unmodified
+|1  | 1 | 0  | 0  | Store data capability with tag cleared
+|==============================================================================
```
andresag01, line 119: I would make this table like this instead:

|SL|CL| ~SL | Tag|
|-|-|-|-|
| 0 | 0 | 1 | 0 |
| 0 | 1 | 1 | Unchanged
| 1 | 0 | 0 | Unchanged
| 1 | 1 | 0 | Unchanged

arichardson: I've updated the table now and kept the ~SL as part of a comment instead which is easier to read IMO.

src/level-ext.adoc
arichardson: Fixed and cleaned up table.

PR #357, opened 2024-09-09
Allow alternative capability encodings

### LINKED ISSUE #324: Allowing using alternative capability encodings
As discussed in this week's CHERI TG meeting, it is sometimes desirable to use alternative capability encodings, particularly in RV32. However, the CHERI RISC-V specification currently mandates capability encodings for RV32 and RV64 in the base extension Zcheripurecap. A proposal to get around this restriction is to split out the capability encoding (i.e. most of Chapter 2) into a separate CHERI RISC-V extension and include a paragraph indicating that the encoding described in this specification could be replaced with an alternative encoding.

There is precedent for a similar arrangement in the ratified RISC-V privileged specification which has the following text as a note in the introduction:

> We briefly note that the entire privileged-level design described in this document could be replaced with an entirely different privileged-level design without changing the unprivileged ISA, and possibly without even changing the ABI...... For simplicity of expression, the text is written as if this was the only possible privileged architecture.

The CHERI RISC-V spec would require that alternative encodings at the very least support the minimal set of features that underpin the base CHERI instructions. For example, the alternative encoding would need to support all the base architectural permissions and sentries, and could add new permissions.

To enable this flexible arrangement, we would need to review the existing CHERI RISC-V instructions and features to ensure that nothing is defined in such a way that it is directly dependent on the capability encoding. For example, the current description of ACPERM springs to mind because it restricts the allowed combinations of set permissions depending on what can be encoded in the AP field. We probably need to move these restrictions into the new capability encoding extension. Another example is probably the CHERI execution mode which is alluded to in Chapter 2 although this is not part of Zcheripurecap.

andresag01:
Allow implementations supporting CHERI RISC-V to use alternative capability encodings that provide most key features, such as bounds and permissions, but may change, for example, the granularity of bounds or offer additional features. Also, allow the `sentry` instruction to become an illegal when using alternative capability encodings that do not provide ambient sealing permission.

The paragraph allowing alternative encodings is largely inspired from the paragraph in the RISC-V Privileged Specification that allows alternative privileged ISA designs.

Fix #324

### DIFF
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -27,7 +27,7 @@ initially ported to CHERIv9 while providing spatial safety, temporal safety and
 compartmentalization support alongside a good measure of compatibility with
 RISC-V software that is not aware of CHERI. Alternative capability encoding
 specifications must provide key primitives, such as permissions and bounds,
-from in this specification while using a different encoding that, for example,
+from this specification while using a different encoding that, for example,
 changes the granularity of bounds or adds new features. For simplicity of
 expression, the text is written as if this was the only possible capability
 encoding possible for CHERI RISC-V.
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -19,6 +19,19 @@ current draft and expected to be part of privileged architecture 1.13.
 MXLEN without including the tag bit. The value of CLEN is always calculated
 based on MXLEN regardless of the effective XLEN value.
 
+NOTE: We briefly note that the capability encoding described in this section
+could be replaced with an entirely different design without changing how CHERI
+integrates with the RISC-V ISA, and possibly without even changing the ABI. In
+particular, this capability encoding specification was designed to run software
+initially ported to CHERIv9 while providing spatial safety, temporal safety and
+compartmentalization support alongside a good measure of compatibility with
+RISC-V software that is not aware of CHERI. Alternative capability encoding
+specifications must provide key primitives, such as permissions and bounds,
+from in this specification while using a different encoding that, for example,
+changes the granularity of bounds or adds new features. For simplicity of
+expression, the text is written as if this was the only possible capability
+encoding possible for CHERI RISC-V.
+
 [#section_cap_encoding]
 === Capability Encoding
 
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -21,7 +21,7 @@ based on MXLEN regardless of the effective XLEN value.
 
 NOTE: We briefly note that the capability encoding described in this section
 could be replaced with an entirely different design without changing how CHERI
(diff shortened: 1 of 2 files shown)


src/cap-description.adoc
```diff
@@ -19,6 +19,19 @@ current draft and expected to be part of privileged architecture 1.13.
...
 
+NOTE: We briefly note that the capability encoding described in this section
+could be replaced with an entirely different design without changing how CHERI
+integrates with the RISC-V ISA, and possibly without even changing the ABI. In
+particular, this capability encoding specification was designed to run software
+initially ported to CHERIv9 while providing spatial safety, temporal safety and
+compartmentalization support alongside a good measure of compatibility with
+RISC-V software that is not aware of CHERI. Alternative capability encoding
+specifications must provide key primitives, such as permissions and bounds,
+from in this specification while using a different encoding that, for example,
```
tariqkurd-repo, line 30:
```suggestion
from this specification while using a different encoding that, for example,
```

tariqkurd-repo (approved):

src/cap-description.adoc
```diff
@@ -19,6 +19,19 @@ current draft and expected to be part of privileged architecture 1.13.
 MXLEN without including the tag bit. The value of CLEN is always calculated
 based on MXLEN regardless of the effective XLEN value.
 
+NOTE: We briefly note that the capability encoding described in this section
+could be replaced with an entirely different design without changing how CHERI
+integrates with the RISC-V ISA, and possibly without even changing the ABI. In
```
jrtc27, line 24: It absolutely is part of the ABI

andresag01: So you'd like to remove this part entirely?
> and possibly without even changing the ABI.

jrtc27: Yes

andresag01: Done

src/insns/sentry_32bit.adoc
```diff
@@ -22,6 +22,11 @@ Description::
 Capability `cd` is written with the capability in `cs1` with its type bit set to 1.
 Attempting to seal an already sealed capability will lead to the tag of `cd` being set to 0.
 
+NOTE: The SENTRY instruction may give rise to an illegal instruction fault when
```
arichardson, line 25: Maybe we could change this to "when the implementation does not support capability type 1 (unrestricted sentry)", since you could still imagine an implementation that has seal/unseal permissions but allows unrestricted sentry (e.g. CHERIv9).

arichardson (commented):
Other than my one comment this LGTM.

src/insns/sentry_32bit.adoc
jrtc27: Quite

andresag01: Done

Issue #333, opened 2024-07-24
Should the Zstid CSRs be guarded by CRE


tomaird:
The description for `CRE` says:

> Executing CSR instructions accessing any CSR added by Zcherihybrid (see Table 25) causes an illegal instruction exception

But this doesn't include the CSRs added by Zstid. Should they be included as well?

From @PRugg-Cap :

> We need to make sure sneaky user-mode software can't stash values that are hidden from legacy OSs. That means either TID needs to be guarded on CRE, or more likely (but more annoyingly) we need to have _another_ enable bit for the TIDs.

tariqkurd-repo:
they definitely need an enable bit - so sharing CRE seems most sensible.
opinions?

jrtc27:
If it’s a separate extension it needs its own enable. Even if it’s dependent on CHERI you could write software for CHERI without xtid, which will then be broken if you add xtid without its own enable.

PRugg-Cap:
I was worried you might say that...

arichardson:
I also think there this should not be gated by CRE, but instead using a state-enable bit. The capability extended versions should definitely be guarded by CRE though.

tariqkurd-repo:
yeah - a state enable seems likely to be required

PR #358, opened 2024-09-09
Control access to Zstid CSRs

andresag01:
Add a bit in xstateen0 CSRs to control access to new CSRs used by the Zstid extension.

Fixes https://github.com/riscv/riscv-cheri/issues/333

### DIFF
--- a/src/tid-ext.adoc
+++ b/src/tid-ext.adoc
@@ -1,4 +1,4 @@
-== "Zstid Extension for Thread Identification
+== "{tid_ext_name}" Extension for Thread Identification
 
 {tid_ext_name} is an optional extension to the RISC-V base ISA.
 Implementations that support {cheri_base_ext_name} and {tid_ext_name}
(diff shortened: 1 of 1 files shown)


tariqkurd-repo:
if you allocate bit 28 then there'll be a lot more time before it gets overwritten, but they do seem to be allocating from the top so I guess it's ok

arichardson (approved):

tomaird:
I don't think it's clear enough from this change explicitly which bit controls which CSR in which privilege mode. It seems to suggest that if `mstateen0.TID=0` this means that only M-level access can access `stid`, and if `sstateen0.TID=0` then U-level access cannot access `utid`. And if the `TID` bits are 1, then the read/write-permissions described in the `Zstid` CSR tables apply. Is that correct? This could be clearer.

Also does this change now mean `Zstid` is dependent on the `Sxstateen` extensions being enabled too? Or is possible to implement `Zstid` without `Sxstateen`? This could also be clearer.

PR #365, opened 2024-09-13
Integrating vector and CHERI for RISC-V

### LINKED ISSUE #359: Add section on vector support
The vector spec is simple "check every byte of every load/store access"

This needs adding to the doc.

The concept of adding tags into vector registers would be a future feature.

andresag01:
This PR has the following changes:
* Remove all outdated notes regarding CHERI and vector
* Remove pre-existing and incomplete vector instruction listings that were `ifdef` out
* Remove adoc defines to conditionally build the CHERI specification with/without vector
* Add a section describing how the RISC-V vector extension integrates with Zcheripurecap and Zcherihybrid:
    * Vector registers cannot hold capabilities with tag=1
    * Vector memory accesses are authorized by either `ddc` or a capability in a **c** register depending on the CHERI mode 
    * Vector has all the usual memory access exceptions, but only _body_ and _active_ elements may give rise to out-of-bounds exceptions i.e. elements that are masked off, before vstart or after vl do not trap

Fixes #359


src/vector-integration.adoc
```diff
@@ -1,38 +1,27 @@
...
-vector elements are divided into three disjoing subsets: _prestart_, _body_
-and _tail_. Only _body_ elements that are _active_, that is where the current
-mask is enabled, are considered accessed and may give rise to faults if their
-memory location is outside the capability's bounds.
-* For loads, the read permission must be set in AP
-* For stores, the write permission must be set in AP
+According to the vector extension cite:[riscv-v-spec] only _active_ elements are
+accessed or updated in memory. Therefore only _active_ elements are checked to be
+within capability bounds. If a vector load or store has no _active_ elements then
+no CHERI fault will be taken.
```
andresag01, line 24: So if VL=0 for example, and we are doing a load with an authorising cap with R-permission=0 then we do not get a trap?

src/vector-integration.adoc
```diff
@@ -1,38 +1,27 @@
...
+  Until that time, vector load and store instructions must not be used to implement generic
+  memory copying in software, such as the `memcpy()` standard C library function,
+  because the vector registers do not hold capabilities, so the tags of any
+  copied capabilities will be set to 0 in the destination memory.
 
-NOTE: Vector load and store instructions must not be used to implement generic
-memory copying in software, such as the `memcpy()` standard C library function,
-because the vector registers do not hold capabilities, so the tags of any
-copied capabilities will be set to 0 in the destination memory.
+Vector loads and stores all following the behaviour as described in
```
andresag01, line 14: Typo? Perhaps "load and stores all following the behaviour ..." -> "loads and stores behave ..."

src/vector-integration.adoc
```diff
@@ -1,38 +1,27 @@
...
+xref:section_int_load_store_insns[xrefstyle=short], as all are treated as integer data.
```
andresag01, line 15: Minor nitpicking: I avoided the expression "integer data" in the original version because technically vector registers can hold various formats of integer and floating-point data -- I recommend avoiding it here too!

src/vector-integration.adoc
tariqkurd-repo: correct - because there is no load - which follows the vector spec rules

src/vector-integration.adoc
tariqkurd-repo:
```suggestion
Vector loads and stores all follow the behaviour as described in
```

src/vector-integration.adoc
tariqkurd-repo: AFAIK vector loads don't do things like nan boxing on loaded FP data - I think they all load/store data as if it is integer data - so I think this is correct.... they certainly all follow the rules in that xref so I think this is ok as it is.

tariqkurd-repo: also, if there's a move between V and X registers then it is always treated as integer data, which this sentence is also supposed to cover. Can you think of any counter-examples?

tariqkurd-repo (approved):

src/vector-integration.adoc
arichardson: I'm not sure the "as all are integer data" makes this sentence any clearer, could just drop it?

arichardson (approved):

src/vector-integration.adoc
jrtc27: I think that should be stated more clearly; currently it talks about bounds only, then states a more general thing that doesn’t inherently follow. Ideally in a way that encourages people to do the other checks once and feed that into each iteration of bounds checking.

src/vector-integration.adoc
tariqkurd-repo:
```suggestion
xref:section_int_load_store_insns[xrefstyle=short].
```

src/vector-integration.adoc
tariqkurd-repo: I've simplified the text - it's correct now - it was a bit weak before

Issue #366, opened 2024-09-13
Rename length violation to bounds violation


jrtc27:
The name is our fault, but it's not the best, as accesses before the base also fault, not just ones beyond base + length. Bounds violation makes it clear that it's for anything outside the bounds, and is the term we actually use in practice.

arichardson:
I've looked at every use of `length` and I blieve #367 fixes all of them.

PR #367, opened 2024-09-13
Use "bounds" violation instead of "length"

arichardson:
This error also applies to access before the start of the capability, so length is not an accurate term, and most existing software already refers to it as a bounds violation.

Fixes: https://github.com/riscv/riscv-cheri/issues/366

### DIFF
--- a/src/insns/condbr_common.adoc
+++ b/src/insns/condbr_common.adoc
@@ -1,5 +1,5 @@
 Exceptions::
 When the target address is not within the <<pcc>>'s bounds, and the branch is taken,
 a _CHERI jump or
-branch fault_ is reported in the TYPE field and Length Violation is reported in
+branch fault_ is reported in the TYPE field and Bounds violation is reported in
 the CAUSE field of <<mtval>> or <<stval>>:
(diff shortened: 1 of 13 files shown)


src/insns/condbr_common.adoc
```diff
@@ -1,5 +1,5 @@
 Exceptions::
 When the target address is not within the <<pcc>>'s bounds, and the branch is taken,
 a _CHERI jump or
-branch fault_ is reported in the TYPE field and Length Violation is reported in
```
jrtc27, line 4: I assume the capital V was previously rogue and there are no other instances of that?

jrtc27 (approved):
That was quick! Haven't checked if you've missed any but the ones that you have here all look good.

tariqkurd-repo (approved):

andresag01 (approved):

src/insns/condbr_common.adoc
andresag01: It was a typo. I found and fixed another instance of this in the mret/sret instruction page.

tariqkurd-repo: seems to be fixed

PR #368, opened 2024-09-16
pointer masking support

tariqkurd-repo:
Add statement on pointer masking

### DIFF
--- a/src/ptrmask-integration.adoc
+++ b/src/ptrmask-integration.adoc
@@ -16,5 +16,4 @@ extension before the bounds check.
 NOTE: This scheme doesn't seem very useful, but the problem is the dynamic configuration
  of pointer masking which can dynamically update meaning of the address within the capability,
  and so the full address field must be used to calculate bounds.
- There is future work required to determine a more useful way of applying pointer masking in
- {cheri_cap_mode_name}.
+ There is future work required to determine a more useful way of applying pointer masking to capabilities.
--- a/src/ptrmask-integration.adoc
+++ b/src/ptrmask-integration.adoc
@@ -1,7 +1,7 @@
 [#section_pointer_masking_integration]
 == Integrating {cheri_base_ext_name} and {cheri_default_ext_name} with Pointer Masking
 
-The pointer masking extensions SSnpm, Sspm and Supm are compatible with {cheri_default_ext_name}.
+The pointer masking extensions Smmpm, Smnpm, SSnpm, Sspm and Supm are compatible with {cheri_default_ext_name}.
 
 In {cheri_int_mode_name} all data addresses are integer pointers, which
 are interpreted as being XLEN-wide, and may be subject to pointer masking.
--- a/src/ptrmask-integration.adoc
+++ b/src/ptrmask-integration.adoc
@@ -12,4 +12,5 @@ In {cheri_cap_mode_name} all pointers are capabilities, and the bounds
 encoding relies on XLEN-wide addresses, and therefore the definitions of the bounds
 will need to be updated to be usable for pointer masking.
 
-Therefore supporting pointer masking in {cheri_cap_mode_name} is future work.
+Therefore supporting pointer masking in {cheri_cap_mode_name} is future work, and so
+Xenfcfg.PMM is currently ignored in {cheri_cap_mode_name}.
--- a/src/ptrmask-integration.adoc
+++ b/src/ptrmask-integration.adoc
@@ -13,4 +13,3 @@ encoding relies on XLEN-wide addresses, and therefore the definitions
 will need to be updated to be usable for pointer masking.
 
 Therefore supporting pointer masking in {cheri_cap_mode_name} is future work.
-
--- a/src/ptrmask-integration.adoc
+++ b/src/ptrmask-integration.adoc
@@ -9,7 +9,7 @@ All data accesses are checked against <<ddc>> which is unaffected by
 pointer masking. Therefore no capability bounds encoding is affected.
 
 In {cheri_cap_mode_name} all pointers are capabilities, and the bounds
-encoding relies on XLEN-wide addresses, and therefore the definitions
+encoding relies on XLEN-wide addresses, and therefore the definitions of the bounds
 will need to be updated to be usable for pointer masking.
 
 Therefore supporting pointer masking in {cheri_cap_mode_name} is future work.
--- a/src/ptrmask-integration.adoc
+++ b/src/ptrmask-integration.adoc
@@ -15,5 +15,5 @@ extension before the bounds check.
 
 NOTE: This scheme doesn't seem very useful, but the problem is the dynamic configuration
  of pointer masking which can arbitrarily update the meaning of the address within the capability,
- and so the full address field must be used to calculate bounds.
+ so the full address field must be used to calculate bounds.
  There is future work required to determine a more useful way of applying pointer masking to capabilities.
--- a/src/ptrmask-integration.adoc
+++ b/src/ptrmask-integration.adoc
@@ -14,6 +14,6 @@ for bounds calculation. The dereferenced address has the masked bits replaced by
 extension before the bounds check.
 
 NOTE: This scheme doesn't seem very useful, but the problem is the dynamic configuration
- of pointer masking which can dynamically update meaning of the address within the capability,
+ of pointer masking which can arbitrarily update the meaning of the address within the capability,
  and so the full address field must be used to calculate bounds.
  There is future work required to determine a more useful way of applying pointer masking to capabilities.
--- a/src/ptrmask-integration.adoc
+++ b/src/ptrmask-integration.adoc
@@ -8,9 +8,9 @@ are interpreted as being XLEN-wide, and may be subject to pointer masking.
 All data accesses are checked against <<ddc>> which is unaffected by
 pointer masking. Therefore no capability bounds encoding is affected.
 
-In {cheri_cap_mode_name} all pointers are capabilities, and the bounds
-encoding relies on XLEN-wide addresses, and therefore the definitions of the bounds
-will need to be updated to be usable for pointer masking.
-
-Therefore supporting pointer masking in {cheri_cap_mode_name} is future work, and so
-Xenfcfg.PMM is currently ignored in {cheri_cap_mode_name}.
+In {cheri_cap_mode_name} all pointers are capabilities. If pointer masking is enabled
+then the masked bits are replaced with sign extension of the highest unmasked bit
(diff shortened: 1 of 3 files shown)


src/ptrmask-integration.adoc
```diff
@@ -0,0 +1,15 @@
+[#section_pointer_masking_integration]
+== Integrating {cheri_base_ext_name} and {cheri_default_ext_name} with Pointer Masking
+
+The pointer masking extensions SSnpm, Sspm and Supm are compatible with {cheri_default_ext_name}.
```
andresag01, line 4: According to the RISC-V Pointer Masking spec, the existing Pointer Masking extensions are:
* Smmpm
* Smnpm
* Ssnpm

Could we align this text with the contents of the RISC-V spec?

src/ptrmask-integration.adoc
```diff
@@ -0,0 +1,15 @@
...
+
+In {cheri_int_mode_name} all data addresses are integer pointers, which
+are interpreted as being XLEN-wide, and may be subject to pointer masking.
+All data accesses are checked against <<ddc>> which is unaffected by
+pointer masking. Therefore no capability bounds encoding is affected.
+
+In {cheri_cap_mode_name} all pointers are capabilities, and the bounds
+encoding relies on XLEN-wide addresses, and therefore the definitions
+will need to be updated to be usable for pointer masking.
```
andresag01, line 13: Which definitions are you referring to?

andresag01 (commented):
Is this going to work efficiently in hybrid code? i.e. Code that is partially purecap but has some parts still running in Integer Pointer Mode. My understanding is that pointer masking is enabled or disabled in a privilege level using CSRs that can only be read/written from higher privileges.

For example, `senvcfg.PMM` controls pointer masking for U/VU-mode. Therefore, if a program wants to jump from a piece of code in Capability Pointer Mode to a piece of code in Integer Pointer mode, would it have to also do some sort of supervisor call to enable pointer masking on the way in and disable pointer masking on return back to Capability Pointer Mode?

src/ptrmask-integration.adoc
tariqkurd-repo: I took the names from https://github.com/riscv/riscv-profiles/blob/main/src/rva23-profile.adoc

tariqkurd-repo:
```suggestion
The pointer masking extensions Smmpm, Smnpm, SSnpm, Sspm and Supm are compatible with {cheri_default_ext_name}.
```

src/ptrmask-integration.adoc
```diff
@@ -0,0 +1,15 @@
...
+
+The pointer masking extensions Smmpm, Smnpm, SSnpm, Sspm and Supm are compatible with {cheri_default_ext_name}.
+
+In {cheri_int_mode_name} all data addresses are integer pointers, which
+are interpreted as being XLEN-wide, and may be subject to pointer masking.
+All data accesses are checked against <<ddc>> which is unaffected by
+pointer masking. Therefore no capability bounds encoding is affected.
+
+In {cheri_cap_mode_name} all pointers are capabilities, and the bounds
+encoding relies on XLEN-wide addresses, and therefore the definitions
```
tariqkurd-repo, line 12:
```suggestion
encoding relies on XLEN-wide addresses, and therefore the definitions of the bounds
```

src/ptrmask-integration.adoc
```diff
@@ -0,0 +1,15 @@
...
+In {cheri_int_mode_name} all data addresses are integer pointers, which
+are interpreted as being XLEN-wide, and may be subject to pointer masking.
+All data accesses are checked against <<ddc>> which is unaffected by
+pointer masking. Therefore no capability bounds encoding is affected.
+
+In {cheri_cap_mode_name} all pointers are capabilities, and the bounds
+encoding relies on XLEN-wide addresses, and therefore the definitions of the bounds
+will need to be updated to be usable for pointer masking.
+
+Therefore supporting pointer masking in {cheri_cap_mode_name} is future work.
```
tariqkurd-repo, line 15:
```suggestion
Therefore supporting pointer masking in {cheri_cap_mode_name} is future work, and so
Xenfcfg.PMM is currently ignored in {cheri_cap_mode_name}.
```

src/ptrmask-integration.adoc
```diff
@@ -0,0 +1,16 @@
...
+In {cheri_int_mode_name} all data addresses are integer pointers, which
+are interpreted as being XLEN-wide, and may be subject to pointer masking.
+All data accesses are checked against <<ddc>> which is unaffected by
+pointer masking. Therefore no capability bounds encoding is affected.
+
+In {cheri_cap_mode_name} all pointers are capabilities, and the bounds
+encoding relies on XLEN-wide addresses, and therefore the definitions of the bounds
+will need to be updated to be usable for pointer masking.
+
+Therefore supporting pointer masking in {cheri_cap_mode_name} is future work, and so
```
arichardson, line 15: I'm not sure this is ideal since it means you also have to mask the PMM bit whenever you do a modesw in hybrid code? I think it would be more consistent to always mask off the bits and just feed those into the bounds check, and rephrase this as "pointer masking is not applied to capability decoding"? 

Setting/clearing the high bits could create an unrepresentable capability (which is why Morello sort of ignored them) and you can't use them for metadata unless you have an almighty capability. As long as the bounds are still decoded relative to the address stored in the capability rather than the masked address, masking can't change the meaning of capabilities. And since it just changes inaccessible virtual addresses I don't think masking prior to the bounds check is a problem? It's not particularly useful but at least it's more consistent than forcing PMM off?

tariqkurd-repo: ok - I've changed it - please have another look

src/ptrmask-integration.adoc
```diff
@@ -0,0 +1,16 @@
...
+== Integrating {cheri_base_ext_name} and {cheri_default_ext_name} with Pointer Masking
+
+The pointer masking extensions Smmpm, Smnpm, SSnpm, Sspm and Supm are compatible with {cheri_default_ext_name}.
+
+In {cheri_int_mode_name} all data addresses are integer pointers, which
+are interpreted as being XLEN-wide, and may be subject to pointer masking.
+All data accesses are checked against <<ddc>> which is unaffected by
+pointer masking. Therefore no capability bounds encoding is affected.
+
+In {cheri_cap_mode_name} all pointers are capabilities. If pointer masking is enabled
```
LawrenceEsswood, line 11: This is the way that Morello did it, but they faced some unintended consequences as a result. Because of how sign extension would change bounds decode, modifying a capability address (completely in bounds) can result in dramatically different bounds. This was even a bug for a while as it beat the fast representability check. They ended up having to special case changing the 56th bit of address, if the capability had a specific range of E's, to always untag when modified even though the capability was never taken out of bounds.

I think we should probably learn from this lesson that whilst the address should eventually be sign extended for the purpose of generating a vaddr for the next stage of memory fetch, sign extension should likely not be a part of bounds calculation / bounds check so that in-bounds capabilities are always representable.

I would suggest for the purpose of bounds decode / bounds check high bits just be masked (not extended)

tariqkurd-repo:
the proposal should be updated to

- use the address including the masked bits for the bounds calculation
- apply the mask before the actual dereference

and then do something better later for capability mode

src/ptrmask-integration.adoc
```diff
@@ -0,0 +1,20 @@
...
+then the entire address field, including any bits representing the pointer mask, are used
+for bounds calculation. The dereferenced address has the masked bits replaced by sign
```
arichardson, line 13:
```suggestion
For instructions using capabilities (e.g. loads/stores in {cheri_cap_mode_name}), the final access address is subject to pointer masking, but the computed bounds are not. The entire address field, including any bits representing the pointer mask, are used
for bounds calculation. When pointer masking is enabled, the dereferenced address has the masked bits replaced by sign
```

maybe something like this?

src/ptrmask-integration.adoc
```diff
@@ -0,0 +1,20 @@
...
+extension before the bounds check.
+
+NOTE: This scheme doesn't seem very useful, but the problem is the dynamic configuration
+ of pointer masking which can dynamically update meaning of the address within the capability,
+ and so the full address field must be used to calculate bounds.
+ There is future work required to determine a more useful way of applying pointer masking in
+ {cheri_cap_mode_name}.
```
arichardson, line 20:
```suggestion
 There is future work required to determine a more useful way of applying pointer masking to capabilities.
```

arichardson (commented):
Nit: it's not really the decoding mode that is impacted by pointer masking but rather how the instruction interprets the operands. I've made few suggestions to make it clear that we are referring to the type of operand not the mode we are executing in.

src/ptrmask-integration.adoc
```diff
@@ -0,0 +1,20 @@
...
+All data accesses are checked against <<ddc>> which is unaffected by
+pointer masking. Therefore no capability bounds encoding is affected.
+
+In {cheri_cap_mode_name} all pointers are capabilities. If pointer masking is enabled
+then the entire address field, including any bits representing the pointer mask, are used
+for bounds calculation. The dereferenced address has the masked bits replaced by sign
+extension before the bounds check.
+
+NOTE: This scheme doesn't seem very useful, but the problem is the dynamic configuration
+ of pointer masking which can dynamically update meaning of the address within the capability,
```
andresag01, line 17:
```suggestion
 of pointer masking which can arbitrarily update the meaning of the address within the capability,
```

src/ptrmask-integration.adoc
```diff
@@ -0,0 +1,20 @@
...
+ and so the full address field must be used to calculate bounds.
```
andresag01, line 18:
```suggestion
 so the full address field must be used to calculate bounds.
```

andresag01:
Made a few minor suggestions. Also rebased to resolve conflicts in `introduction.md`

arichardson (commented):
I still think we should not conflate the CHERI mode with the behaviour of the pointer extension - it only really depends on the type of the operand. Otherwise LGTM.

jrtc27:
> I still think we should not conflate the CHERI mode with the behaviour of the pointer extension - it only really depends on the type of the operand

Agreed. This matters in particular for a Zcherihybrid.

andresag01 (approved):

src/ptrmask-integration.adoc
arichardson: Agreed we need to fix this for a proper integration, but for now just doing the bare minimum makes the most sense.

arichardson (approved):

PR #369, opened 2024-09-17
Add extension maturity table

tariqkurd-repo:
Clarify which extensions are mature, and which are still subject to investigation

### DIFF
--- a/src/introduction.adoc
+++ b/src/introduction.adoc
@@ -90,7 +90,7 @@ RISC-V.
 It also adds a mode bit in the encoding of capabilities to
 allow changing the current CHERI execution mode using indirect jump
 instructions.
-{sh4add_ext_name}::   Addition of <<SH4ADD>> and <<SH4ADD_UW>> as the natural datatype is 16-bytes.
+{sh4add_ext_name}::   Addition of <<SH4ADD>> and <<SH4ADD_UW>> for RV64 only, as CHERI capabilities are 16 bytes when XLEN=64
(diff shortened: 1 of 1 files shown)


arichardson (approved):

src/introduction.adoc
```diff
@@ -90,13 +90,28 @@ RISC-V.
 It also adds a mode bit in the encoding of capabilities to
 allow changing the current CHERI execution mode using indirect jump
 instructions.
+{sh4add_ext_name}::   Addition of <<SH4ADD>> and <<SH4ADD_UW>> as the natural datatype is 16-bytes.
```
arichardson, line 93:
```suggestion
{sh4add_ext_name}::   Addition of <<SH4ADD>> and <<SH4ADD_UW>> as CHERI capabilities are 16 bytes when XLEN=64.
```
Maybe this is clearer?

src/introduction.adoc
```diff
@@ -90,13 +90,28 @@ RISC-V.
 It also adds a mode bit in the encoding of capabilities to
 allow changing the current CHERI execution mode using indirect jump
 instructions.
```
andresag01, line 92: We should delete, or at least modify, this sentence which is old. We also support changing the mode with modesw which is more efficient than the indirect jumps

src/introduction.adoc
tariqkurd-repo:
```suggestion
{sh4add_ext_name}::   Addition of <<SH4ADD>> and <<SH4ADD_UW>> for RV64 only, as CHERI capabilities are 16 bytes when XLEN=64
```

src/introduction.adoc
tariqkurd-repo: I've removed that text

Issue #322, opened 2024-07-10
Integer pointer mode in debug mode


andresag01:
The current CHERI RISC-V spec goes to great length to provide a "compatibility" mode with CHERI-less RISC-V when executing code in M, S and U mode provided that Zcherihybrid is implemented. However, the machine always enters debug in capability pointer mode. Should there be an option to enter debug mode in "integer pointer mode", perhaps something like a CRE?

NOTE: I am only pointing out that a CHERI machine supporting Zcherihybrid would need a CHERI-aware debugger even though, for whatever reason, it is only ever running non-CHERI aware code compiled by a non-CHERI aware compiler.

andresag01:
@tariqkurd-repo @jrtc27 : Do you have a view on this?

tariqkurd-repo:
The debugger only needs to issue an SCMODE before running existing debug code, but it's a reasonable point that something does have to change. I think it's normal to have to customise debugger software for specific cores anyway, just look at how many debug options there are in the RISC-V spec. Also I think that a CHERI core is likely to be delivered with a customed debugger, which in this case will include the update to switch modes, or more likely to have full CHERI support available in the debugger.

Your proposal is simple enough - but I'd hold off until the need for it is demonstrated.

I'm pretty sure that @jrtc27 will have an opinion.

tariqkurd-repo:
this has been raised as a real problem, as entering capability mode in debug mode by default prevents a legacy debugger from functioning.

jrtc27:
I don't see why that matters. If the debug spec ever bumps its version again then you'll need a debugger that understands that. It already specifies 3 versions, 2 of which are widely used, and I doubt any debugger out there supports RV128 which would similarly be broken the moment you ship 128-bit hardware even if it also supports 64-bit.

tariqkurd-repo:
I agree - and I also know that debuggers are very expensive - and getting them customised makes them even more expensive - so having a "risc-v compliant" debug interface is really important, whatever that means with the various options in the debug spec.

tariqkurd-repo:
https://github.com/riscv/riscv-cheri/pull/370

PR #370, opened 2024-09-17
controlling the cheri mode to enter when entering debug mode

tariqkurd-repo:
fixes https://github.com/riscv/riscv-cheri/issues/322

### DIFF
--- a/src/debug-integration.adoc
+++ b/src/debug-integration.adoc
@@ -178,17 +178,17 @@ include::img/dscratch1creg.edn[]
 
 The <<dinfc>> register is a CLEN-bit plus tag bit CSR only accessible in debug mode.
 
-The reset value is the <<infinite-cap>> capability with the <<m_bit>> set to {CAP_MODE_VALUE},
-regardless of whether {cheri_default_ext_name} (see xref:section-cheri-execution-mode[xrefstyle=short])
-is implemented:
+The reset value is the <<infinite-cap>> capability.
 
-The current CHERI execution mode can be determined in debug mode by reading <<dinfc>>, as the
-<<m_bit>> is updated by the hardware. The value of the <<m_bit>> when the core is not in debug
-mode is either the reset value or the value of the <<m_bit>> in debug mode when the core last
-resumed execution.
+If {cheri_default_ext_name} is implemented:
 
-<<dinfc>> is read/write but with no writeable fields, and so writes are
-ignored.
+* The <<m_bit>> is reset to be {cheri_default_ext_name}.
+* The debugger can set the <<m_bit>> to {cheri_base_mode_ext} by reading the CSR, using <<SCMODE>> and then writing the CSR.
+** This only needs doing once after resetting the core.
+* The <<m_bit>> is used on debug mode entry to determine which CHERI execution mode to enter.
+
+The <<m_bit>> is the only writeable field in <<dinfc>>.
+Therefore if {cheri_default_ext_name} is not implemented then it is read-write with no writeable fields.
 
 NOTE: A future version of this specification may add writeable fields to allow creation
  of other capabilities, if, for example, a future extension requires multiple formats for
--- a/src/debug-integration.adoc
+++ b/src/debug-integration.adoc
@@ -182,8 +182,9 @@ The reset value is the <<infinite-cap>> capability.
 
 If {cheri_default_ext_name} is implemented:
 
-* The <<m_bit>> is reset to be {INT_MODE_VALUE}.
-* The debugger can set the <<m_bit>> to {cheri_base_mode_ext} by reading the CSR, using <<SCMODE>> and then writing the CSR.
+* The <<m_bit>> is reset to be {cheri_default_ext_name} ({INT_MODE_VALUE}).
+* The debugger can set the <<m_bit>> to {cheri_base_mode_ext} ({CAP_MODE_VALUE}) by executing <<MODESW>> from the program buffer
+** if <<MODESW>> is not supported in debug mode then the same can be done by reading the CSR, using <<SCMODE>> and then writing the CSR.
 ** This only needs doing once after resetting the core.
 * The <<m_bit>> is used on debug mode entry to determine which CHERI execution mode to enter.
 
--- a/src/debug-integration.adoc
+++ b/src/debug-integration.adoc
@@ -5,9 +5,6 @@ This section describes changes to integrate the Sdext ISA and
 {cheri_base_ext_name}. It must be implemented to make external debug compatible
 with {cheri_base_ext_name}. Modifications to Sdext are kept to a minimum.
 
-WARNING: This section is preliminary as no-one has yet built debug support
- for CHERI-RISC-V so change is likely.
-
 The following features, which are optional in Sdext, must be implemented for
 use with {cheri_base_ext_name}:
 
--- a/src/debug-integration.adoc
+++ b/src/debug-integration.adoc
@@ -182,6 +182,8 @@ The reset value is the <<infinite-cap>> capability with the <<m_bit>> set to {CA
 regardless of whether {cheri_default_ext_name} (see xref:section-cheri-execution-mode[xrefstyle=short])
 is implemented:
 
+The current CHERI execution mode can be determined in debug mode by reading <<dinfc>>.
+
 <<dinfc>> is read/write but with no writeable fields, and so writes are
 ignored.
 
--- a/src/debug-integration.adoc
+++ b/src/debug-integration.adoc
@@ -182,7 +182,10 @@ The reset value is the <<infinite-cap>> capability with the <<m_bit>> set to {CA
 regardless of whether {cheri_default_ext_name} (see xref:section-cheri-execution-mode[xrefstyle=short])
 is implemented:
 
-The current CHERI execution mode can be determined in debug mode by reading <<dinfc>>.
+The current CHERI execution mode can be determined in debug mode by reading <<dinfc>>, as the
+<<m_bit>> is updated by the hardware. The value of the <<m_bit>> when the core is not in debug
+mode is either the reset value or the value of the <<m_bit>> in debug mode when the core last
+resumed execution.
 
 <<dinfc>> is read/write but with no writeable fields, and so writes are
 ignored.
--- a/src/debug-integration.adoc
+++ b/src/debug-integration.adoc
@@ -182,7 +182,7 @@ The reset value is the <<infinite-cap>> capability.
 
 If {cheri_default_ext_name} is implemented:
 
-* The <<m_bit>> is reset to be {cheri_default_ext_name}.
+* The <<m_bit>> is reset to be {INT_MODE_VALUE}.
 * The debugger can set the <<m_bit>> to {cheri_base_mode_ext} by reading the CSR, using <<SCMODE>> and then writing the CSR.
 ** This only needs doing once after resetting the core.
 * The <<m_bit>> is used on debug mode entry to determine which CHERI execution mode to enter.
--- a/src/insns/modesw_common.adoc
+++ b/src/insns/modesw_common.adoc
@@ -5,10 +5,13 @@ Toggle the hart's current CHERI execution mode in <<pcc>>.
 * If the current mode in <<pcc>> is {cheri_int_mode_name} ({INT_MODE_VALUE}), then the <<m_bit>> in <<pcc>> is set to {cheri_cap_mode_name} ({CAP_MODE_VALUE}).
 * If the current mode is {cheri_cap_mode_name} ({CAP_MODE_VALUE}), then the <<m_bit>> in <<pcc>> is set to {cheri_int_mode_name} ({INT_MODE_VALUE}).
 
-NOTE: The effective CHERI exection mode is give by the value of some CSRs and
+NOTE: The effective CHERI execution mode is give by the value of some CSRs and
 the <<pcc>>'s <<m_bit>>, so executing <<MODESW>> does not necessarily change
 the machine's current mode. The current, effective CHERI execution mode can be
 observed as described in xref:m_bit_observe[xrefstyle=short].
 
-NOTE: Implementations may optionally support executing <<C_MODESW>> from the
-program buffer while in debug mode.
+NOTE: Implementations may optionally support executing <<MODESW>> from the
+program buffer while in debug mode. If supported them the <<m_bit>> in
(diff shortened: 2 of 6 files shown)


src/riscv-hybrid-integration.adoc
```diff
@@ -279,9 +279,20 @@ shown in xref:default-csrnames-added[xrefstyle=short].
 
 {cheri_default_ext_name} optionally allows <<MODESW>> to execute in debug mode.
 
-When entering debug mode, the core always enters {cheri_cap_mode_name}.
+When entering debug mode, whether the core enters {cheri_int_mode_name} or
+{cheri_cap_mode_name} is UNSPECIFIED.
```
andresag01, line 283: Couldn't we just say that the CHERI execution mode in debug is defined, for example, through a CRE bit or through the mode bit in `dinfc`?

src/riscv-hybrid-integration.adoc
```diff
@@ -279,9 +279,20 @@ shown in xref:default-csrnames-added[xrefstyle=short].
...
 Implementations may optionally support switching CHERI execution mode by
-executing the <<MODESW>> from the program buffer.
+executing <<MODESW>> from the program buffer.
+
+The current mode can be read from <<dinfc>>.
```
andresag01, line 287: Note that we currently have not described what happens when reading `dinfc` from Integer Pointer Mode -- I just noticed this!

src/riscv-hybrid-integration.adoc
andresag01: I often find that unspecified features are tricky to handle and give rise to problems in the long term. We do not need to add extra state. We could simply:
* Define that the CHERI execution mode is given by the mode bit in `dinfc`
* Clearly state that CSR accesses to dinfc are CLEN wide regardless of CHERI execution mode

If we like the unspecified part, we could indicate that the initial value of `dinfc.mode` is implementation-defined and that implementations may make `dinfc` read-only meaning that the CHERI mode in debug cannot be changed.

I guess the only awkward bit with this proposal is what happens when the implementation does assign pcs to the program buffer and/or it is possible to execute MODESW in debug mode. Perhaps `dinfc` just reflects the current mode?

andresag01: Its minor, but the specification says this in reference to CLEN access for new CHERI CSRs:
> Also, Zcherihybrid adds the new capability CSRs listed in Table 28.

Table 28 does not list dinfc.

>dinfc.M reports the mode already

I don't think it does. We removed that feature in this commit which is already merged into `main`:

https://github.com/riscv/riscv-cheri/commit/74f05ec206e5152abb904e9b20460bd34856a642#diff-ea4440c312373a6c9e166a66f393ba1c7025400a239982f21123dbf15d06fbff

tariqkurd-repo: ok so the table needs fixing
the reset value of dinfc isn't a concern - what we need is for the mode to be updated on the entry into debug mode which is currently missing

tariqkurd-repo: the current spec is fine as dinfc is a new CLEN CSR
image

tariqkurd-repo: done

src/riscv-hybrid-integration.adoc
tariqkurd-repo: we do - see above

src/riscv-hybrid-integration.adoc
```diff
@@ -279,9 +279,20 @@ shown in xref:default-csrnames-added[xrefstyle=short].
 
 {cheri_default_ext_name} optionally allows <<MODESW>> to execute in debug mode.
 
-When entering debug mode, the core always enters {cheri_cap_mode_name}.
+When entering debug mode, whether the core enters {cheri_int_mode_name} or
```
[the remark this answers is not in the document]
tariqkurd-repo: which CRE bit are you thinking of though?

I was thinking that a specific core will specify that it will always enter a specific mode on debug mode entry, and this will be matched against the debugger (integer mode if using a legacy debugger). 

Then the core can have an input pin driven by a fuse which controls which mode it is in when entering debug mode, which can be set to cap mode when it is known that a CHERI-aware debugger is in use.

That way legacy debuggers and CHERI aware debuggers can work, and the mode can be selectable by a pin or some other mechanism. the piont is that we don't specify any of that and leave it to the core vendor.

Alternatively we need to specify a control which says which mode to enter, but which one? Are you thinking of menvcfg.CRE? Or add another bit? That is another solution.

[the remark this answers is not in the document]
arichardson: I was thinking of using `mseccfg.CRE` as the bit that determines which mode the debugger enters into. The other option could be to use dinfc.mode and have that be intmode on reset - then we could allow the debugger to write an infinite cap with a different mode to specify the mode on entry. A bit ugly allowing a write there but it avoids allocating yet another register.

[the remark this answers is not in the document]
tariqkurd-repo: this causes a problem though - you might have a core running CHERI code and need to connect it to a non-CHERI aware debugger - the non-CHERI aware debugger will be of limited use, but will be able to probe memory

the point of making the mode UNSPECIFIED is to allow legacy debuggers to be used even if the core is running CHERI code, as an essential step on the way to CHERI adoption.

so I think this PR is correct as it stands, and we shouldn't be enforcing extra restrictions on debugger usage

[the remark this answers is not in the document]
jrtc27: That's not a use case to care much about. If you think of CHERI as a new base ISA it just doesn't really make sense to do.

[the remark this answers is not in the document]
jrtc27: And there will be plenty of time to teach OpenOCD about capability mode prior to ratification.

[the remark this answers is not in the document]
jrtc27: I really don't want to see the CHERI debugging experience made worse, nor put unnecessary burden on implementations (e.g. needing a larger program buffer to be supported so mode switching can be done), in the name of compatibility.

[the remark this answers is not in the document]
tariqkurd-repo: This is a real use case that we have right now, and others will have it to when they move to CHERI cores.
My proposal of giving people the choice doesn't slow down or prevent openOCD being updated.
The recommendation is to be in capablity mode once debugger support is available.

tariqkurd-repo:
we're now thinking about allocating an unused bit in DCSR which will specify which mode to enter on debug mode entry

carlrobshaw (approved):

src/debug-integration.adoc
```diff
@@ -178,12 +178,17 @@ include::img/dscratch1creg.edn[]
...
-The reset value is the <<infinite-cap>> capability with the <<m_bit>> set to {CAP_MODE_VALUE},
-regardless of whether {cheri_default_ext_name} (see xref:section-cheri-execution-mode[xrefstyle=short])
-is implemented:
+The reset value is the <<infinite-cap>> capability.
 
-<<dinfc>> is read/write but with no writeable fields, and so writes are
-ignored.
+If {cheri_default_ext_name} is implemented:
+
+* The <<m_bit>> is reset to be {cheri_default_ext_name}.
```
arichardson, line 185:
```suggestion
* The <<m_bit>> is reset to be {INT_MODE_VALUE}.
```
I think that is the right macro?

src/debug-integration.adoc
```diff
@@ -178,12 +178,17 @@ include::img/dscratch1creg.edn[]
...
+* The debugger can set the <<m_bit>> to {cheri_base_mode_ext} by reading the CSR, using <<SCMODE>> and then writing the CSR.
```
arichardson, line 186:
```suggestion
* The debugger can set the <<m_bit>> to {CAP_MODE_VALUE} by reading the CSR, using <<SCMODE>> and then writing the CSR.
```

src/riscv-hybrid-integration.adoc
```diff
@@ -279,9 +279,20 @@ shown in xref:default-csrnames-added[xrefstyle=short].
...
+{cheri_cap_mode_name} is UNSPECIFIED.
 Implementations may optionally support switching CHERI execution mode by
-executing the <<MODESW>> from the program buffer.
+executing <<MODESW>> from the program buffer.
```
arichardson, line 285: I think one thing we should clarify here is whether this also updates dinfc.MODE and therefore whether it only affects the current program buffer execution or also future entries to debug mode. I think having it also set dinfc.mode is cleaner and allows a debugger to just execute a single modesw in the program buffer to have all future debug entries be in capmode.

arichardson (approved):
One minor questions otherwise this looks good to me.

I think it is important that we can attach both an updated one (e.g. OpenOCD) as well as an unmodified debugger to the core since it is common to use more than one debugger for a given core depending on who is debugging/what is being debugged.

andresag01:
@tariqkurd-repo: Could we remove this text in the spec? This change is based on feedback from porting debuggers to CHERI RISC-V:

> This section is preliminary as no-one has yet built debug support for CHERI-RISC-V so
change is likely.

src/debug-integration.adoc
tariqkurd-repo: done, but in a slightly different way

src/riscv-hybrid-integration.adoc
tariqkurd-repo: done

tariqkurd-repo:
> @tariqkurd-repo: Could we remove this text in the spec? This change is based on feedback from porting debuggers to CHERI RISC-V:
> 
> > This section is preliminary as no-one has yet built debug support for CHERI-RISC-V so
> > change is likely.

done

src/riscv-hybrid-integration.adoc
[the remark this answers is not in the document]
tariqkurd-repo: spec update, will request re-review

src/riscv-hybrid-integration.adoc
andresag01, line 283: This text with the UNSPECIFIED is out-of-data. It is specified now.

andresag01: The text at https://github.com/riscv/riscv-cheri/pull/370/files#diff-b366fcf81637594697f776b0c57bf43a576436df8b1fde7ed1ae593771025d94R92 is out of date. The debug mode is given by dinfc and the code sequence must us gcmode instead of gctag.

src/insns/modesw_common.adoc
```diff
@@ -5,10 +5,13 @@ Toggle the hart's current CHERI execution mode in <<pcc>>.
...
 the machine's current mode. The current, effective CHERI execution mode can be
 observed as described in xref:m_bit_observe[xrefstyle=short].
 
-NOTE: Implementations may optionally support executing <<C_MODESW>> from the
-program buffer while in debug mode.
+NOTE: Implementations may optionally support executing <<MODESW>> from the
+program buffer while in debug mode. If supported them the <<m_bit>> in
+<<dinfc>> is toggled and used to control which mode to enter next time debug
+mode is entered. The CHERI execution mode is only controlled by the <<m_bit>>
+of <<dinfc>> in debug mode.
```
andresag01, line 17: I am concerned that MODESW is allowed when the PCC is unspecified because:
1. We already have a way easily change the mode by read-modify-write dinfc using csrr, scmode, csrw
2. Changes the instruction's behaviour just for this privilege mode

Also, note that the debug spec also allows having the PCC defined properly, so that the program buffer is part of the memory address space allowing jumps, branches, auipc, etc to be executed from there. What determines the CHERI mode in this case? If its the PCC then do we need to update dinfc whenever PCC is written with (e.g.) JALR/MODESW turning the M bit across these two registers into some WARL-like field?

It would be best to disallow MODESW while in debug mode altogether and simply have the mode there be determined by dinfc in all cases.

andresag01 (changes requested):

src/riscv-hybrid-integration.adoc
tariqkurd-repo: fixed.

andresag01:
I broadly agree with this change. However, it would be best if we did not have MODESW writing `dinfc` when executing in debug mode. First, its awkward to change the behavior of this instruction based on the privileged mode. And second, we already have another well-defined mechanism to change the m-bit in `dinfc` using csrr, scmode and csrw which we can fully execute from the program buffer given that CHERI RISC-V requires the program buffer to have at least 4 entries.

@arichardson : What do you think?

arichardson:
> I broadly agree with this change. However, it would be best if we did not have MODESW writing `dinfc` when executing in debug mode. First, its awkward to change the behavior of this instruction based on the privileged mode. And second, we already have another well-defined mechanism to change the m-bit in `dinfc` using csrr, scmode and csrw which we can fully execute from the program buffer given that CHERI RISC-V requires the program buffer to have at least 4 entries.
> 
> @arichardson : What do you think?

I did not realize that it was possible to have pcc in debug mode, I though it always used a program buffer without a real pc. Your explanation makes sense and I think not having it affect dinfc.mode is fine.

Alternatively, for implementations that have a PCC in debug mode and support modesw, you could copy PCC.mode into dindc.mode at the end of debug mode execution, but I'm not convinced that's needed. All we really need is a one time selection on debugger startup and for that the CSR write seems fine.

arichardson (approved):

arichardson:
> > I broadly agree with this change. However, it would be best if we did not have MODESW writing `dinfc` when executing in debug mode. First, its awkward to change the behavior of this instruction based on the privileged mode. And second, we already have another well-defined mechanism to change the m-bit in `dinfc` using csrr, scmode and csrw which we can fully execute from the program buffer given that CHERI RISC-V requires the program buffer to have at least 4 entries.
> > @arichardson : What do you think?
> 
> I did not realize that it was possible to have pcc in debug mode, I though it always used a program buffer without a real pc. Your explanation makes sense and I think not having it affect dinfc.mode is fine.
> 
> Alternatively, for implementations that have a PCC in debug mode and support modesw, you could copy PCC.mode into dindc.mode at the end of debug mode execution, but I'm not convinced that's needed. All we really need is a one time selection on debugger startup and for that the CSR write seems fine.

Alternatively, we could say for implementations that have a real pcc in debug mode, the m bit if dinfc is a view of the PCC mode bit?

tariqkurd-repo:
If there is a PCC then yes, it will be another view of the same bit. This
isn't currently stated.

PR #373, opened 2024-09-19
add extra information into exception reporting

tariqkurd-repo:
We need to have the exception faulting address when a CHERI data memory access faults.
Not having so is causing problems, so this adds two fields:

mtval.CAP_IDX - which cap caused the fault
mtval2 - XLEN wide address which faulted (already a standard CSR)
stval2 - as mtval2

### DIFF
--- a/src/hypervisor-integration.adoc
+++ b/src/hypervisor-integration.adoc
@@ -169,7 +169,7 @@ include::img/vstdcreg.edn[]
 
 The <<vstval>> register is an SXLEN-bit read-write register.
 
-When a fault is taken into VS-mode <<vstval>> is updated as for <<stval>> for all CHERI memory access traps.
+When a fault is taken into VS-mode <<vstval>> is updated as for <<stval>> for all CHERI exceptions.
 
 .Virtual supervisor trap value register
 [#vstval-format]
@@ -183,7 +183,7 @@ The <<vstval2>> register is a VSXLEN-bit read-write register.
 
 The CSR address is 0x24b.
 
-When a fault is taken into VS-mode <<vstval2>> is updated as for <<mtval2>> for all CHERI memory access traps.
+When a fault is taken into VS-mode <<vstval2>> is updated as for <<mtval2>> for all CHERI exceptions.
 It is set to zero in all other cases.
 
 NOTE: This is not a standard RISC-V CSR, but <<mtval2>> is. 0x24b is the regular location for the CSR.


src/riscv-integration.adoc
```diff
@@ -995,9 +1026,25 @@ include::img/stvalreg.edn[]
 
 TYPE is a CHERI-specific fault type that caused the exception while CAUSE
 is the cause of the fault. The possible CHERI types and causes are encoded as
-shown in xref:mtval-cheri-type[xrefstyle=short] and
+shown in xref:mtval-cheri-type[xrefstyle=short], xref:mtval-cheri-cap-idx[xrefstyle=short] and
 xref:mtval-cheri-causes[xrefstyle=short] respectively.
 
+[#stval2,reftext="stval2"]
+==== Supervisor Trap Value Register 2 (stval2)
```
andresag01, line 1033: I guess we will need a vstval2 for hypervisor

andresag01: Would also be good to add a diagram for it just like with the other registers

src/riscv-integration.adoc
```diff
@@ -791,6 +805,23 @@ CHERI violations have the following order in priority:
...
 . Bounds violation (_Lowest_)
 
+[#mtval2,reftext="mtval2"]
+==== Machine Trap Value Register 2 (mtval2)
+
+The <<mtval2>> register is an MXLEN-bit read-write register, which is added as part of the
+Hypervisor extension. {cheri_base_ext_name} also requires the implementation of this CSR.
+
+When a CHERI fault is taken on any data memory access into into M-mode, <<mtval2>> is written
+with the MXLEN-bit address which caused the fault. This can be used together with the
```
andresag01, line 815: Why address is that? For example, if I load 4 bytes from address 0x10, but only bytes at 0x10 and 0x11 are in bounds while 0x12 and 0x13 are out-of-bounds, then what address do I get in mtval2? Is it 0x10 or 0x12 (or something else!)?

src/riscv-integration.adoc
```diff
@@ -791,6 +805,23 @@ CHERI violations have the following order in priority:
...
+CAP_IDX field of <<mtval>> to diagnose the authorising capability which caused the
+exception.
+
+<<mtval2>> is set to zero for all PCC faults, and follows the standard rules for non-CHERI exceptions.
```
andresag01, line 819: By PCC faults, do you mean faults for checks where the authorising capability is PCC?

src/riscv-integration.adoc
tariqkurd-repo: the load/store address, which would normally be in mtval. the lowest byte of the access

src/riscv-integration.adoc
tariqkurd-repo: `PCC fault` is already in the doc

src/riscv-integration.adoc
tariqkurd-repo: fixed - it follows the standard Xtval rules

src/riscv-integration.adoc
```diff
@@ -791,6 +805,24 @@ CHERI violations have the following order in priority:
...
 . Bounds violation (_Lowest_)
 
+[#mtval2,reftext="mtval2"]
+==== Machine Trap Value Register 2 (mtval2)
+
+The <<mtval2>> register is an MXLEN-bit read-write register, which is added as part of the
+Hypervisor extension. {cheri_base_ext_name} also requires the implementation of this CSR.
+
+When a CHERI fault is taken on any data memory access into into M-mode, <<mtval2>> is written
+with the MXLEN-bit address which caused the fault. This follows the existing rules for reporting
```
arichardson, line 815: The priv spec calls this register "Machine bad guest physical address.", so we probably want to clarify that we are writing the virtual address here?

arichardson: hmm in the hypervisor spec chapter it is listed as "Machine Second Trap Value (mtval2) Register" which makes a lot more sense. 

I might send a PR to the ISA manual to change that in `2.2. CSR Listing`.

tariqkurd-repo: for the purpose of this spec it's the "effective address", as reported in Xtval if there's (e.g.) a page fault

arichardson (approved):
Missing the vstval2reg.edn file otherwise LGTM.

jrtc27:
Why do you need the address, and why do you need CAP_IDX? That was always meant to be a crutch that went away in the standard version.

src/riscv-integration.adoc
```diff
@@ -1032,17 +1034,37 @@ The fields are identical to <<mtval>> for CHERI exceptions.
...
+When a fault is taken into VS-mode <<vstval2>> is updated as for <<mtval2>> for all CHERI memory access traps.
+It is set to zero in all other cases.
+
+NOTE: This is not a standard RISC-V CSR, but <<mtval2>> is. 0x24b is the regular location for the CSR.
+
+<<vstval2>> holds the same set of values that <<stval2>> can hold.
+
+.Virtual supervisor trap value register 2
+[#vstval2-format]
+include::img/vstval2reg.edn[]
```
carlrobshaw, line 1066: Maybe I missed it. Where is img/vstval2reg.edn ?

src/riscv-integration.adoc
```diff
@@ -1032,17 +1034,37 @@ The fields are identical to <<mtval>> for CHERI exceptions.
...
 NOTE: This is not a standard RISC-V CSR, but <<mtval2>> is. 0x14b is the regular location for the CSR.
 
+<<stval2>> holds the same set of values that <<mtval2>> can hold.
+
 .Supervisor trap value register 2
 [#stval2-format]
 include::img/stval2reg.edn[]
 
+[#vstval2,reftext="vstval2"]
+==== Virtual Supervisor Trap Value Register 2 (vstval2)
```
carlrobshaw, line 1051: Should this go with the hypervisor extension support rather than the base CHERI support CSRs?

src/img/stval2reg.edn
```diff
@@ -6,9 +6,9 @@
...
 (def right-margin 100)
 (def boxes-per-row 32)
-(draw-column-headers {:height 20 :font-size 18 :labels (reverse ["0" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "MXLEN-1"])})
+(draw-column-headers {:height 20 :font-size 18 :labels (reverse ["0" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "SXLEN-1"])})
 
 (draw-box "Addr"   {:span 32})
 
-(draw-box "MXLEN" {:span 32 :borders {}})
+(draw-box "SXLEN" {:span 32 :borders {}})
 ----
```
carlrobshaw, line 14: This commit should really be separate as it's a fix to stval2 and not part of adding vstval2

src/riscv-integration.adoc
tariqkurd-repo: I missed it - it's committed now - hopefully the build will work as a result

src/img/stval2reg.edn
tariqkurd-repo: yeah fair point

src/riscv-integration.adoc
tariqkurd-repo: done

src/riscv-integration.adoc
tariqkurd-repo: I've added the word "effective" to clarify

tariqkurd-repo:
> Why do you need the address, and why do you need CAP_IDX? That was always meant to be a crutch that went away in the standard version.

So David asked me to post the case where I would have wanted the fault address of a CHERI fault to be available:
On a (normal) page fault the linux kernel has this check:
Well the linux kernel does this on a page fault in kernel mode:

        if (!user_mode(regs) && addr < TASK_SIZE && unlikely(!(regs->status & SR_SUM))) {
                if (fixup_exception(regs))
                        return;

                die_kernel_fault("access to user memory without uaccess routines", addr, regs);
        }
Ignoring the SR_SUM check as it is irrelevant to the discussion what this does on a (normal page fault is):

Check if the code that triggered the fault is executing in kernel mode (!user_mode(regs))
Check that the address that causes the fault is a user address (addr < TASK_SIZE)
Here addr is the fault address provided by hardware on a page fault.
The fixup_exception() will then check if the instruction pointer is in one of the user_copy routines and if this is true will abort the user copy operation with an error that will eventually be reported to the user as an EFAULT return value from the attempted syscall.
Now if we do a copy_to_user in CHERI, we want the user to provide a capability for the target of the copy and the kernel will use that capability as it would use an address.
If the user provided an invalid or unsuitable capability the result should be the same as for a page fault (at least that's what I think).
I.e. we don't want the user process to abort with a CHERI fault. Instead we want to terminate the syscall with an error and let user mode handle it.
The most prominent example where this showed up is a user mode program (seen in various syscall test suites) passing in a NULL pointer to read simply to test the EFAULT behaviour but of course the case of a real invalid capability must be handled as well.
Now, if we want to do the same check for a CHERI fault we would also need to know if we are executing in kernel mode and we would need to check the fault address against the user space address limit. The latter is what I want the fault address for.

jrtc27:
We do not need the faulting address in order to disambiguate the various cases in CheriBSD. I am perplexed that there is the belief Linux does.

jrtc27:
If you take a capability fault and PC is within copyin/out (copy_to/from_user) then you return an error from the routines. If you take a capability fault and PC is not within copyin/out then that's a kernel error (whether because a kernel access was junk or because it failed to use copyin/out) and deserves a panic.

tariqkurd-repo:
another issue is whether we want to carry CAP_IDX down the pipe, our feeling is that just reporting the address in Xtval2 is enough for the reported case

djmsip:
I haven't looked at this particular case in any detail, but my understanding is that having the address allows the code to basically use the same path as the MMU fault. Without the address we have to omit the check the code is doing on the MMU fault case in the CHERI case. As to why the check is done, that is a good question which we are looking into, but all the arches do it.

There are a few other points on this as well:

- Most of the people who have been working on this have complained that the faulting address is not available easily
- It does help with debugging in my opinion.
- This will be the only memory access trap that does *not* report the faulting address as far as I know.
- All the information is already available as I understand it, so it is not hard to implement
- I'm concerned that there may be some other operating systems or use case that may find this useful - I just don't know. Can we say we will *never* need the faulting address?

Of course, we can always find out what the address is by looking at the instruction and operands (and potentially the boundaries of the instruction as well) , but given the machine already knows all this, why not make the information available? Better a crutch now than legless later. Feel free to shoot me down:-)

tariqkurd-repo:
> I haven't looked at this particular case in any detail, but my understanding is that having the address allows the code to basically use the same path as the MMU fault. Without the address we have to omit the check the code is doing on the MMU fault case in the CHERI case. As to why the check is done, that is a good question which we are looking into, but all the arches do it.
> 
> There are a few other points on this as well:
> 
> * Most of the people who have been working on this have complained that the faulting address is not available easily
> * It does help with debugging in my opinion.
> * This will be the only memory access trap that does _not_ report the faulting address as far as I know.
> * All the information is already available as I understand it, so it is not hard to implement
> * I'm concerned that there may be some other operating systems or use case that may find this useful - I just don't know. Can we say we will _never_ need the faulting address?
> 
> Of course, we can always find out what the address is by looking at the instruction and operands (and potentially the boundaries of the instruction as well) , but given the machine already knows all this, why not make the information available? Better a crutch now than legless later. Feel free to shoot me down:-)

The address is already available, as if we take (e.g.) a page fault we'll report it. It's hard to justify not reporting it in this case, as many people report wanting to know it.

jrtc27:
The reason information is communicated to trap handlers for page faults is so that you can do all the usual CoW, swapping, etc tricks that rely on knowing what page is being accessed. Yes, we _can_ provide the implementation, but we can provide lots of other things too. There is a concern in my head that, if you provide the information, it implies to people that they should be using it to _recover_ from the trap and resume execution, and without the full capability what are you going to do with it beyond say "yes that's an address"? The only thing you can usefully do is change what errors you report, but you still need to report an error, and it's not clear to me that it's useful to differentiate cases, especially since you can't trust anything about the address other than that it was the address the program attempted to use, and you don't have the capability that was used as the authority.

tariqkurd-repo:
> The reason information is communicated to trap handlers for page faults is so that you can do all the usual CoW, swapping, etc tricks that rely on knowing what page is being accessed. Yes, we _can_ provide the implementation, but we can provide lots of other things too. There is a concern in my head that, if you provide the information, it implies to people that they should be using it to _recover_ from the trap and resume execution, and without the full capability what are you going to do with it beyond say "yes that's an address"? The only thing you can usefully do is change what errors you report, but you still need to report an error, and it's not clear to me that it's useful to differentiate cases, especially since you can't trust anything about the address other than that it was the address the program attempted to use, and you don't have the capability that was used as the authority.

The viable alternative is to make all the new CSRs cap width and include the metadata with the tag cleared as we don't want to add rep checks. If the rep check would have failed on the access (particularly for a vector indexed load store where the index is very large) is the metadata useful? I suspect that it is useful in almost all cases.

Then the information is more complete.

Again - the address is already reported in many cases, so is trivial to add. Adding the metadata is also ok.

>but we can provide lots of other things too. 

no-one has asked for anything else AFAIK, and this one has come up repeatedly in software development

djmsip:
Ok, that is a different argument, about intent, which I hadn't thought of I confess. Personally, I don't think we can stop people doing dumb things, after all they could do that by decoding the instruction. I think it is useful to have the address (metadata would useful as well I think) for ease of debugging if nothing else. In this case, IIUC it means the changes for CHERI are a bit less as we can common up more code, though I need to look at the code in depth. I don't like the fact that the program has tried to access something and we don't immediately know what, even though the machine does.

chrehrhardt:
Someone else asking for this (on arm morello). Turns out they have it:
https://community.arm.com/support-forums/f/morello-forum/55922/decoding-morello-instructions-to-determine-the-faulting-memory-address-on-a-sigprot
Morello Linux handles CHERI traps with a SIGSEGV and if this is done something should be filled into siginfo_t's ->si_addr field

src/riscv-integration.adoc
```diff
@@ -791,6 +805,26 @@ CHERI violations have the following order in priority:
...
 
+[#mtval2,reftext="mtval2"]
+==== Machine Trap Value Register 2 (mtval2)
+
+The <<mtval2>> register is an MXLEN-bit read-write register, which is added as part of the
+Hypervisor extension. {cheri_base_ext_name} also requires the implementation of this CSR.
+
+When a CHERI fault is taken on any data memory access into into M-mode, <<mtval2>> is written
+with the MXLEN-bit effective address which caused the fault. This follows the existing rules for reporting
+load/store addresses to <<mtval>> from [riscv-priv-spec].
```
andresag01, line 816: Fix broken citation.
```suggestion
load/store addresses to <<mtval>> from cite:[riscv-priv-spec].
```

src/riscv-integration.adoc
```diff
@@ -993,10 +1027,25 @@ xref:stval-format[xrefstyle=short] to assist software in handling the trap.
...
+The fields are identical to <<mtval>> for CHERI exceptions.
+
+[#stval2,reftext="stval2"]
+==== Supervisor Trap Value Register 2 (stval2)
+
+The <<stval2>> register is an SXLEN-bit read-write register.
+
+The CSR address is 0x24b.
+
+When a fault is taken into S-mode <<stval2>> is updated as for <<mtval2>> for all CHERI memory access traps.
```
andresag01, line 1039: The mtval2 case says this:
> mtval2 is set to zero for all PCC faults

Which is a little different to saying that its only updated for CHERI memory access faults. It would be good to be a little more precise in the intention for both of these registers (I was trying to hint at this in an earlier comments although I did not do it very well!)

src/hypervisor-integration.adoc
```diff
@@ -164,6 +164,36 @@ present when the implementation supports {cheri_default_ext_name}.
 .Virtual supervisor trap default capability register
 include::img/vstdcreg.edn[]
 
+[#vstval,reftext="vstval"]
+==== Virtual Supervisor Trap Value Register (vstval)
```
andresag01, line 168:
```suggestion
=== Virtual Supervisor Trap Value Register (vstval)
```

src/hypervisor-integration.adoc
```diff
@@ -164,6 +164,36 @@ present when the implementation supports {cheri_default_ext_name}.
...
+
+When a fault is taken into VS-mode <<vstval>> is updated as for <<stval>> for all CHERI memory access traps.
+
+.Virtual supervisor trap value register
+[#vstval-format]
+include::img/vstvalreg.edn[]
+
+
+[#vstval2,reftext="vstval2"]
+==== Virtual Supervisor Trap Value Register 2 (vstval2)
```
andresag01, line 180:
```suggestion
=== Virtual Supervisor Trap Value Register 2 (vstval2)
```

src/hypervisor-integration.adoc
andresag01, line 168: Do we need this section at all? The H spec clearly states that these registers are analogous to the S-mode versions, and nothing is changing between stval vs vstval, so I did not add them initially

src/hypervisor-integration.adoc
```diff
@@ -164,6 +164,36 @@ present when the implementation supports {cheri_default_ext_name}.
...
+
+The <<vstval2>> register is an SXLEN-bit read-write register.
```
andresag01, line 182:
```suggestion
The <<vstval2>> register is a VSXLEN-bit read-write register.
```

src/riscv-integration.adoc
```diff
@@ -791,6 +805,26 @@ CHERI violations have the following order in priority:
...
+==== Machine Trap Value Register 2 (mtval2)
+
+The <<mtval2>> register is an MXLEN-bit read-write register, which is added as part of the
+Hypervisor extension. {cheri_base_ext_name} also requires the implementation of this CSR.
+
+When a CHERI fault is taken on any data memory access into into M-mode, <<mtval2>> is written
+with the MXLEN-bit effective address which caused the fault. This follows the existing rules for reporting
+load/store addresses to <<mtval>> from [riscv-priv-spec].
+//This reported address can be used together with the CAP_IDX field of <<mtval>> to diagnose the
+//authorising capability which caused the exception.
```
andresag01, line 818: Could we remove the comment before merging?

andresag01 (approved):
Happy with the changes. It would be good to remove the comments and addressing some of the minor typos/comments before merging though

arichardson (approved):
I think reporting the address so you can put it in the SEGV addr field without decoding the instruction makes a lot of sense. 

Capidx is just nice to have information that I guess could be dropped and we'd force GDB to decode the instruction instead.

 @bsdjhb what do you think?

src/hypervisor-integration.adoc
tariqkurd-repo: well - if we have one we should probably have all of them.
In practice it's enough to say that stval is updated and vstval is updated accordingly which would shorten the doc

src/riscv-integration.adoc
```diff
@@ -771,6 +771,20 @@ xref:mtval-cheri-causes[xrefstyle=short] respectively.
...
+//| 0                         | Faulting cap was c0
+//| 1                         | Faulting cap was c1
+//| ...                       |
+//| 30                        | Faulting cap was c30
+//| 31                        | Faulting cap was c31
+//| 32                        | Faulting cap was <<ddc>>
+//| 33                        | Faulting cap was <<pcc>>
+//| 34-63                     | Reserved
+//|==============================================================================
+
```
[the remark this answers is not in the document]
tariqkurd-repo: feel free

src/riscv-integration.adoc
tariqkurd-repo: feel free

src/riscv-integration.adoc
```diff
@@ -791,6 +805,26 @@ CHERI violations have the following order in priority:
...
+The <<mtval2>> register is an MXLEN-bit read-write register, which is added as part of the
+Hypervisor extension. {cheri_base_ext_name} also requires the implementation of this CSR.
+
+When a CHERI fault is taken on any data memory access into into M-mode, <<mtval2>> is written
+with the MXLEN-bit effective address which caused the fault. This follows the existing rules for reporting
+load/store addresses to <<mtval>> from cite:[riscv-priv-spec].
+//This reported address can be used together with the CAP_IDX field of <<mtval>> to diagnose the
+//authorising capability which caused the exception.
+
+<<mtval2>> is set to zero for all PCC faults, and follows the standard rules for non-CHERI exceptions.
```
tariqkurd-repo, line 820:
```suggestion
<<mtval2>> is set to zero for all other CHERI exceptions, and follows the standard rules for non-CHERI exceptions.
```

src/riscv-integration.adoc
tariqkurd-repo, line 1023:
```suggestion
When a fault is taken into S-mode <<stval2>> is updated as for <<mtval2>> for all CHERI exceptions.
```

tariqkurd-repo: clarified

src/riscv-integration.adoc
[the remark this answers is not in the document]
tariqkurd-repo: done

bsdjhb:
As a general comment, yes, I do think it would be useful to have the faulting address.  I have need of it in a CheriBSD PR where I convert CHERI faults for hybrid processes into SIGSEGV instead of SIGPROT.  The current PR works on Morello but not on CHERI RISC-V due to not having the faulting address to hand.

CAPIDX is not really needed IMO.  Morello doesn't provide it, and any adaptation of CHERI to x86 can never provide it (not all memory operands in x86 are in registers).  In general for a debugger, a user can just look at the faulting instruction disassembly to work out which register is involved.  More typically a user will be looking at the source line that faulted and printing variables anyway.  If you do not need a separate error code field in `Xtval` (e.g. if you are using separate `Xcause` codes for tag vs length faults like Morello), then I would suggest just storing the effective faulting VA in `Xtval` similar to MMU faults.

bsdjhb:
For reference, the CheriBSD PR I mentioned: https://github.com/CTSRD-CHERI/cheribsd/pull/2145

jrtc27:
Hm, the SIGSEGV points are fair. In that case, I would suggest putting the address in xtval and the error code in xtval2 to avoid special casing in both hardware and software.

tariqkurd-repo:
It would be more regular to have the address in Xtval, certainly

djmsip:
I think we may as well move it now, it does make it more regular and I suspect it would get shifted in review anyway.

PR #381, opened 2024-09-26
add note about chossing when to check the tag on a load

tariqkurd-repo:


src/cheri-pte-ext.adoc
```diff
@@ -141,11 +141,18 @@ bit value of the capability read. This will introduce additional traps during re
...
 | 0     | 0      | Page fault if stored tag is set
-| 0     | 1      | Page fault if stored tag is set, or hardware CW and CRG update^1^
+| 0     | 1      | Page fault if stored tag is set, or hardware CW and CRG update^2^
 | 1     | X      | Normal operation
 |===
 
-^1^ The choice here follows the pattern of whether to implement the _Svade_ extension
+^1^ The choice here is whether to take a data dependent exceptions on loads or atomic operations.
+ If the exception is taken then the software will sweep the page and update the generation
+ (CRG). If the exception is taken without checking the tag, then the sweep will happen sooner.
```
andresag01, line 150: I think that not checking the tag in a data-dependent way for loads is likely to cause more (unnecessary) traps which might degrade performance a little. It might also mean that pages are swept sooner (?), but I don't know if that is true, and you can probably write your sweeper to avoid that situation.

src/cheri-pte-ext.adoc
```diff
@@ -141,11 +141,18 @@ bit value of the capability read. This will introduce additional traps during re
...
+ The choice on whether to check the tag or not can be static or dynamic. It is also permitted to
```
andresag01, line 151: Do you mean that the data-dependent check can be turned on/off at run-time, for example, by writing a CSR?

tariqkurd-repo:
Basically the choice is implementation defined, and can vary dynamically
according to that micro architecture

src/cheri-pte-ext.adoc
tariqkurd-repo: it can vary as the implementor sees fit

src/cheri-pte-ext.adoc
tariqkurd-repo: text removed

src/cheri-pte-ext.adoc
```diff
@@ -141,11 +141,14 @@ bit value of the capability read. This will introduce additional traps during re
...
 |PTE.CW |PTE.CRG |Store/AMO
 | 0     | 0      | Page fault if stored tag is set
-| 0     | 1      | Page fault if stored tag is set, or hardware CW and CRG update^1^
+| 0     | 1      | Page fault if stored tag is set, or hardware CW and CRG update^2^
 | 1     | X      | Normal operation
 |===
 
-^1^ The choice here follows the pattern of whether to implement the _Svade_ extension
+^1^ The choice here is whether to take data dependent exceptions on loads or atomic operations.
+ The implementation is free to define when (if ever) it checks the tag.
```
arichardson, line 149:
```suggestion
 It is legal for the implementation to fault even if the tag is set since this behaviour is only an optimization for software.
 This means it is also legal to only check the tag in certain conditions and conservatively fault otherwise.
```

Maybe something slightly more wordy to make it explicit that all of `always check/never check/only check for atomics/etc.` is a valid implemenetation choice.

arichardson (approved):
Suggested a slighly more explicit wording. Feel free to use this or ignore.

Issue #372, opened 2024-09-18
Compatibility with Zilsd


arichardson:
It looks like Zilsd (https://github.com/riscv/riscv-zilsd/blob/v0.10/zilsd.adoc) wants to repurpose the DWORD load/store opcodes for RV32 to be load-store pair. Should we use the RV64 encoding of loads/stores for RV32 to avoid this conflict?

This would solve the non-compressed conflict, however for compressed instructions they are are using the floating-point load-stores and forcing capability loads to use the RV64 format would prevent push-pop instructions from being supported. Not sure how best to approach this problem.

arichardson:
https://lists.riscv.org/g/tech-arch-review/message/259 suggests this extension looks likely to be ratified as-is?

tariqkurd-repo:
The confusing 16-bit encoding mapping is why I put this table in the doc:

<img width="643" alt="image" src="https://github.com/user-attachments/assets/9eac1dbd-b84a-4e8c-b2df-816eff02e37c">

So - I would say - that no change is needed for compressed encodings - as we already support both Zcmp (push/pop) and Zcmt (table jump) in cap mode.

For the 32-bit encodings - then we have quite a few which change depending on XLEN - so I think I'd be in favour of unifying all of them. Apart from anything else it will simplify the document - so my proposal is to move all load/store/AMO cap encodings to the RV64 ones.

arichardson:
Simplifying the encodings to be consistent across XLEN for 32-bit encodings sounds good to me!

PR #385, opened 2024-09-27
unify 32-bit encodings for load/store/amo to use RV64 versions

tariqkurd-repo:
fixes https://github.com/riscv/riscv-cheri/issues/372

due to the conflict with Zilsd meaning we lost access to LC/SC in RV32, it seems wise to move all to the RV64 encodings which have no current conflicts.

all is subject to change on ARC review.


andresag01 (approved):

src/insns/load_16bit_fp_sp.adoc
```diff
@@ -27,12 +27,6 @@ Standard floating point load instructions, authorised by the capability in <<ddc
 NOTE: These instructions are available in RV32 {cheri_int_mode_name} only.
 In {cheri_cap_mode_name} they are remapped to <<C.LC>>/<<C.LCSP>>.
 
-NOTE: In {cheri_int_mode_name}, these instructions may be remapped to
```
arichardson, line 30: This is still true? But deleting it seems fine.

arichardson (approved):

src/insns/load_16bit_fp_sp.adoc
tariqkurd-repo: it's true but as Zilsd support is now fully in I thought I'd remove it

tariqkurd-repo:
I checked with Jessica on slack and she's ok with this

PR #387, opened 2024-10-01
optionally simplify lm_perm

tariqkurd-repo:
LM permission adds 4 cases to RV32:

output R
output RC
output original permissions
output no permissions (reserved cases only)

This note allows the implementation to propagate reserved permissions through ACPERM for improved timing in the design.
The reserved permissions cannot be used for dereferencing, and so this does not cause a security problem.

It's just an implementation defined note - no spec change (it's also ok to clear the permissions).

### DIFF
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -137,8 +137,7 @@ If a capability grants <<r_perm>> and <<c_perm>>, but no <<lm_perm>>, then a cap
 The rules specified by <<ACPERM>> are followed when <<w_perm>> and <<lm_perm>> are removed, so additional permissions may also be removed.
 Clearing a capability's <<lm_perm>> and <<w_perm>> allows sharing a read-only version of a data structure (e.g. a tree or linked list) without making a copy.
 
-NOTE: Implementations are allowed to propagate, but not alter, reserved permissions when performing the <<lm_perm>> check on a tagged loaded capability, which means a different rule to <<ACPERM>>.
- This simplifies the number of cases handled by the implementation as it's not required to remove _all_ permissions.
+NOTE: Implementations are allowed to retain invalid capabilities loaded from memory instead of reducing them to _no permissions_ by using <<ACPERM>> behaviour.
 
 [#asr_perm,reftext="ASR-permission"]
 Access System Registers Permission (ASR):: Allow read and write access to all
--- a/src/cap-description.adoc
+++ b/src/cap-description.adoc
@@ -137,7 +137,7 @@ If a capability grants <<r_perm>> and <<c_perm>>, but no <<lm_perm>>, then a cap
 The rules specified by <<ACPERM>> are followed when <<w_perm>> and <<lm_perm>> are removed, so additional permissions may also be removed.
 Clearing a capability's <<lm_perm>> and <<w_perm>> allows sharing a read-only version of a data structure (e.g. a tree or linked list) without making a copy.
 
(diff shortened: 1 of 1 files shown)


src/cap-description.adoc
```diff
@@ -137,6 +137,9 @@ If a capability grants <<r_perm>> and <<c_perm>>, but no <<lm_perm>>, then a cap
 The rules specified by <<ACPERM>> are followed when <<w_perm>> and <<lm_perm>> are removed, so additional permissions may also be removed.
 Clearing a capability's <<lm_perm>> and <<w_perm>> allows sharing a read-only version of a data structure (e.g. a tree or linked list) without making a copy.
 
+NOTE: Implementations are allowed to propagate, but not alter, reserved permissions when performing <<lm_perm>> check on a loaded capability.
+ This simplifies the number of cases handled by the implementation as it's a choice whether to remove _all_ permissions.
```
andresag01, line 141: Isn't this already covered as part of this statement?
> loaded capabilities that are sealed or untagged do not have their permissions changed

I would expect a capability that has "reserved" or "invalid" permissions would be untagged because of the way CHERI RISC-V is defined. I guess the strange case where this could happen is if there is some fault/bit-flip and the capability ended up tagged with reserved permissions, which would be very awkward, but ideally it would be detected as an error.

tariqkurd-repo: it's exactly the case you mentioned - what happens if we get an illegal cap -tagged with a reserved perm value.
certainly these should be untagged in an ideal world, but we need to handle the unexpected cases.

andresag01: Ok, that case is a little concerning. What damage could a capability that is tagged, but invalid in some way, cause? For example, would a LC succeed if it was provided a capability with "reserved" permissions but somehow has R and C set?

At the very least, it would be good to amend the note here to clearly state the case we are concerned about.

tariqkurd-repo: optional divergence from ACPERM for tagged caps with reserved permissions, this is a don't care state and so it's important to have the option to simplify the hardware

arichardson: I was a bit confused by this wording, but if it's just about invalid capabilities loaded from memory due to an error, maybe this could be simplified to something like "unlike  explicit <<ACPERM>>, implemementations are permitted to retain invalid combinations loaded from memory instead of canonicalizing to _no permissions_"?

tariqkurd-repo: ok - try again - I had a look and `canonicalizing` isn't currently in the spec so I avoided using it.

src/cap-description.adoc
```diff
@@ -137,6 +137,8 @@ If a capability grants <<r_perm>> and <<c_perm>>, but no <<lm_perm>>, then a cap
 The rules specified by <<ACPERM>> are followed when <<w_perm>> and <<lm_perm>> are removed, so additional permissions may also be removed.
 Clearing a capability's <<lm_perm>> and <<w_perm>> allows sharing a read-only version of a data structure (e.g. a tree or linked list) without making a copy.
 
+NOTE: Implementations are allowed to retain invalid capabilities loaded from memory instead of reducing them to _no permissions_ by using <<ACPERM>> behaviour.
```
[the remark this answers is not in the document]
arichardson: Oh and I guess we should use en_US here as well.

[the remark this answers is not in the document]
tariqkurd-repo: yeah probably, but that will need to be global in the doc, I think there's a mixture throughout it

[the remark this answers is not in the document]
tariqkurd-repo: I guess that can be another PR

andresag01 (approved):

PR #406, opened 2024-10-04
split MODESW into MODESW.CAP and MODESW.INT

tariqkurd-repo:
MODESW is difficult to implement in higher performance implementations as it causes a lot of uncertainty in pre-decode and decode about which mode is currently being entered. It also causes a long sequential path through a wide-issue decode packet, compared with entering a specific mode which makes it much easier to determine the mode for each instruction.

Therefore it is split into two which directly enter the named mode MODESW.CAP for cap mode and MODESW.INT for int mode, so that instructions can be decoded much more easily.

It was only a single encoding before to try and reduced the 16-bit encoding space for C.MODESW. This is guaranteed not to be accepted without showing significant code-size reduction or performance improvement as part of the standardisation process.

Only using 32-bit encodings means we can add a second encoding, and so be specific about which mode to enter.


arichardson (approved):
Since there are no arguments encoding space should not matter. I'm happy with this change since I don't think switching without knowing the target mode is ever useful.

@jrtc27 @jonwoodruff what do you think?

andresag01 (approved):

PeterRugg:
Chatting around the office, @jonwoodruff, @gameboo and I are happy.

Issue #416, opened 2024-10-10
dfinc updating not very clear


tariqkurd-repo:
dinfc is written to to control the cheri mode to enter when entering debug mode, and also reports the current cheri mode when in debug mode.

this table needs updating:

![image](https://github.com/user-attachments/assets/5efe12e9-1bcd-4f20-85b1-1285bba0005e)

but also, what do we do about writing caps which don't have X permission, or, for RV32 have invalid permission bits to dfinc.
or are untagged? what should the rules be?

One option is to ignore all writes and only allow MODESW.* to write to it.
This seems the simplest approach.

This is the part which needs clarification:

<img width="638" alt="image" src="https://github.com/user-attachments/assets/f5a00550-a02f-46ab-89a5-ea55321b861e">

tariqkurd-repo:
PR merged

arichardson:
If you include `Fixes: #N` or `Closes: #N` in the commit message when merging, it will autoclose the linked issue.

PR #417, opened 2024-10-10
clarify writes to dinfc

tariqkurd-repo:
https://github.com/riscv/riscv-cheri/issues/416


src/debug-integration.adoc
```diff
@@ -185,8 +185,12 @@ If {cheri_default_ext_name} is implemented:
 ** This only needs doing once after resetting the core.
 * The <<m_bit>> is used on debug mode entry to determine which CHERI execution mode to enter.
 
-The <<m_bit>> is the only writeable field in <<dinfc>>.
-Therefore if {cheri_default_ext_name} is not implemented then it is read-write with no writeable fields.
+The <<m_bit>> is the only writeable field in <<dinfc>>:
+
+* It is only updated on a <<CSRRW>> write _if_ the capability being written has a valid <<m_bit>>.
```
arichardson, line 190:
```suggestion
* It is only updated on a <<CSRRW>> write _if_ the capability being written is an <<infinite-cap>> capability. 
```
Should we add this constraint (which implies valid permissions encoding with M) or do we just ignore all other bits?

src/debug-integration.adoc
```diff
@@ -185,8 +185,12 @@ If {cheri_default_ext_name} is implemented:
...
+* If any other CSR instruction is used, or if the written <<m_bit>> is not valid then <<dinfc>> is not updated.
```
arichardson, line 191:
```suggestion
* If any other CSR instruction is used, or if the written capability is not <<infinite-cap>> then <<dinfc>> is not updated.
```

arichardson (commented):
Not sure if we need to restrict it to infinite caps, but might make more sense? 

Alternatively I'd suggest using `permissions encoding includes the <<m_bit>>` instead of `has a valid <<m_bit>>`.

src/debug-integration.adoc
tariqkurd-repo: we now have a new approach, I'll put a comment below

src/debug-integration.adoc
tariqkurd-repo: we now have a new approach, I'll put a comment below

tariqkurd-repo:
We've had an internal review on this, and what it would really take to be able to write debug software.

The previous spec didn't specify whether MODESW.* updated the execution of the next instruction, or delayed the update to the next entry into debug mode

We spent some time discussing how to update dinfc, as checking for a full infinite-cap is quite expensive in hardware.
Our conclusion was to not update dinfc, and *require* MODESW.* in debug mode

The point is that not requiring MODESW.* means that the debugger won't be able to rely on them. It will have to try and execute one, see if it gets an error and then fall back to dong the CSR RMW sequence to update dinfc. Or just not both with them.

So we've gone for the simple approach of requiring them, having them update M in dinfc, and making that observable by reading dinfc.

arichardson (approved):
I like this simplification

src/riscv-hybrid-integration.adoc
```diff
@@ -278,18 +278,16 @@ All CSR instructions cause CHERI exceptions if the <<pcc>> does not grant
...
 {cheri_cap_mode_name} is controlled by the <<m_bit>> in <<dinfc>>.
-Implementations may optionally support switching CHERI execution mode by
-executing <<MODESW_CAP>> or <<MODESW_INT>> from the program buffer.
 
 The current mode can be read from <<dinfc>>.
 
 ifdef::cheri_v9_annotations[]
 NOTE: *CHERI v9 Note:* The mode change instructions <<MODESW_CAP>> and <<MODESW_INT>> are new
-and the requirement to optionally support it in debug mode is also new.
+and the requirement to support it in debug mode is also new.
```
tariqkurd-repo, line 290:
```suggestion
and the requirement to support them in debug mode is also new.
```

Issue #402, opened 2024-10-02
SCBNDSI redundancy?


nwf:
`SCBNDSI` has a 5-bit immediate, `uimm`, and a gated (by `s`) shift-by-4 operation, in
https://github.com/riscv/riscv-cheri/blob/1c43ce2fe5688057d6108a5f901574f2dac0acd0/src/insns/scbnds_32bit.adoc?plain=1#L43 which means that there are two ways to request bound lengths of `0` (unshifed or shifted 0) and `16` (unshifed 16, shifted 1).  Perhaps "shifted 0" and "shifted 1" could be put to better use (as requests for `512` and `528`, say)?

arichardson:
Avoiding redundant encodings here would be nice and I agree that having shifted zero be `512` would make a lot of sense here. Not sure how much that increases decoder complexity, but if we are already special casing, maybe we could also have shifted 1 be something more useful like `1024` or `4096`?

davidchisnall:
I'd like to see some distribution of object sizes before thinking about adding that.  And it will probably be different on 32- and 64-bit systems.  It might be worth marking those as explicitly invalid encodings that are reserved for future expansion.

arichardson:
> I'd like to see some distribution of object sizes before thinking about adding that. And it will probably be different on 32- and 64-bit systems. It might be worth marking those as explicitly invalid encodings that are reserved for future expansion.

That is a good argument - I believe @jonwoodruff did some analysis here.
For now I've opened a PR to reserve these encodings.

jonwoodruff:
These are data samples from Spec2006, averaging distributions from each of 10 benchmarks.

This first set is for flat "bits of precision". It starts at 1 bit because we assume signed values.  12 reaches 100% because we're just measuring the used sizes in RISC-V code generation from CHERI ISA-v9, which only has a 12-bit immediate.

 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
incoffseti | 0.14% | 41.46% | 47.54% | 53.55% | 62.63% | 69.90% | 72.67% | 76.13% | 87.26% | 97.36% | 98.02% | 100.00%

[...]

PR #418, opened 2024-10-10
Reserve redundant SCBNDSI encodings

arichardson:
These encodings could be used for larger values such as 512 in the future.

See https://github.com/riscv/riscv-cheri/issues/402

CC: @nwf @davidchisnall


tariqkurd-repo (approved):

nwf (approved):
Seems OK to me

Issue #426, opened 2024-10-17
ACPERM RV32 order of rules inconsistency


jamie-melling:
Depending on the order in which you carry out the RV32 ACPERM rules you can get different results.  This does assume that the rules are cascaded, which is not explicitly stated.

Contrived example:

The MXLEN=32 additional rules are:

2. Clear C-permission and X-permission if R-permission is not set.
3. Clear W-permission if C-permission is set, but LM-permission is not set.

Example 1

| R | C | LM | W |  |
|--------|--------|--------|--------|--------|
| 0 | 1 | 0 | 1 | After & with rs2 |
| 0 | 0 | 0 | 1 | After RV32 Rule 2 - Clear C-perm |
| 0 | 0 | 0 | 1 | After RV32 Rule 3 - Does not clear W-perm |

Example 2

| R | C | LM | W |  |
|--------|--------|--------|--------|--------|
| 0 | 1 | 0 | 1 | After & with rs2 |
| 0 | 1 | 0 | 0 | After RV32 Rule 3 - Clear W-perm |
| 0 | 0 | 0 | 0 | After RV32 Rule 2 - Clear C-perm  |

To solve this, the spec needs to be explicit about if and when the rules are cascaded and that in such a case, the ordering of the rules should be enforced.

tariqkurd-repo:
interesting, I see no reason _not_ to cascade the rules

tomaird:
Applying them all exactly in order doesn't quite work either
Currently this rule (for all architectures):

> 3. Clear M-bit unless X-permission is set

is above this rule (for RV32 only):

> 4. Clear X-permission if X-permission and R-permission are set, and C-permission and W-permission are both clear.

But if we apply 3 before 4, then we can end up with X-permission cleared but M-bit set, which isn't valid.

I think there needs to be a deeper thought about the dependencies between these rules

PR #427, opened 2024-10-17
apply ACPERM rules in order

tariqkurd-repo:
fixes https://github.com/riscv/riscv-cheri/issues/426
fixes https://github.com/riscv/riscv-cheri/issues/428

### DIFF
--- a/src/insns/acperm_32bit.adoc
+++ b/src/insns/acperm_32bit.adoc
@@ -45,11 +45,11 @@ The rules from <<acperm_rules>> must be followed When removing permissions.
 | Rule | Permission   | Only valid if
 |  1   | <<c_perm>>   | <<r_perm>> or <<w_perm>>
 3+| See <<acperm_rules_rv32>> for rules 2-5 (RV32 only)
-|  6   | <<sl_perm>>  | <<w_perm>> and <<c_perm>>
-|  7   | <<el_perm>>  | <<c_perm>> and <<r_perm>>
-|  8   | <<lm_perm>>  | <<c_perm>> and <<r_perm>>
-|  9   | <<asr_perm>> | <<x_perm>>
-| 10   | <<m_bit>>    | <<x_perm>>
+|  7   | <<sl_perm>>  | <<w_perm>> and <<c_perm>>
+|  8   | <<el_perm>>  | <<c_perm>> and <<r_perm>>
+|  9   | <<lm_perm>>  | <<c_perm>> and <<r_perm>>
+| 10   | <<asr_perm>> | <<x_perm>>
+| 11   | <<m_bit>>    | <<x_perm>>
 3+| See <<acperm_rules_rv32>> for rule 12 (RV32 only)
 |===
 
@@ -70,8 +70,9 @@ These rules are specifically to meet the encoding requirements for <<cap_perms_e
 |  2   | <<c_perm>>   | <<r_perm>>
 |  3   | <<x_perm>>   | <<r_perm>>
 |  4   | <<w_perm>>   | not(<<c_perm>>) or (<<c_perm>> and not(<<lm_perm>>))
-|  5   | <<x_perm>>   | <<w_perm>> or (<<c_perm>> and <<lm_perm>> and <<el_perm>>)
-| 11   | <<asr_perm>> | All other permissions are set.
+|  5   | <<x_perm>>   | <<w_perm>> or <<c_perm>>
+|  6   | <<x_perm>>   | <<c_perm>> == <<lm_perm>> == <<el_perm>>
+| 12   | <<asr_perm>> | All other permissions are set.
 |===
 
 In all cases invalid permissions from <<acperm_rules_rv32>> are cleared.
--- a/src/insns/acperm_32bit.adoc
+++ b/src/insns/acperm_32bit.adoc
@@ -40,42 +40,28 @@ The rules from <<acperm_rules>> must be followed when removing permissions.
 
 [#acperm_rules]
 .ACPERM common rules
-[%autowidth,float="center",align="center",cols="2,3,3",options="header"]
+[%autowidth,float="center",align="center",cols="3,3,3",options="header"]
 |===
-| Rule | Permission   | Only valid if
-|  1   | <<c_perm>>   | <<r_perm>> or <<w_perm>>
-3+| See <<acperm_rules_rv32>> for rules 2-5 (RV32 only)
-|  7   | <<sl_perm>>  | <<w_perm>> and <<c_perm>>
-|  8   | <<el_perm>>  | <<c_perm>> and <<r_perm>>
-|  9   | <<lm_perm>>  | <<c_perm>> and <<r_perm>>
-| 10   | <<asr_perm>> | <<x_perm>>
-| 11   | <<m_bit>>    | <<x_perm>>
-3+| See <<acperm_rules_rv32>> for rule 12 (RV32 only)
+| Rule            | Permission   | Only valid if
+|  1  (RV32 only) | <<asr_perm>> | All other permissions are set.
+|  2              | <<c_perm>>   | <<r_perm>> or <<w_perm>>
+|  3  (RV32 only) | <<c_perm>>   | <<r_perm>>
+|  4  (RV32 only) | <<x_perm>>   | <<r_perm>>
+|  5  (RV32 only) | <<w_perm>>   | not(<<c_perm>>) or (<<c_perm>> and not(<<lm_perm>>))
+|  6  (RV32 only) | <<x_perm>>   | <<w_perm>> or <<c_perm>>
+|  7  (RV32 only) | <<x_perm>>   | <<c_perm>> == <<lm_perm>> == <<el_perm>>
+|  8              | <<sl_perm>>  | <<w_perm>> and <<c_perm>>
+|  9              | <<el_perm>>  | <<c_perm>> and <<r_perm>>
+| 10              | <<lm_perm>>  | <<c_perm>> and <<r_perm>>
+| 11              | <<asr_perm>> | <<x_perm>>
+| 12              | <<m_bit>>    | <<x_perm>>
 |===
 
 The behaviour of currently illegal combinations from <<acperm_rules>> is to clear the permission if invalid (or in the case of <<sl_perm>> set it to 0 (_local_)).
 
 * For RV64 all such combinations may be redefined by future extensions.
-* For RV32 none of these combinations are possible.
-
-RV32 needs additional rules as some combinations which are legal on RV64 are not possible on RV32.
-These rules are specifically to meet the encoding requirements for <<cap_perms_encoding32>>, or
-<<cap_perms_encoding_levels32>> if <<cheri_levels_ext,{cheri_levels_ext_name}>> is implemented.
-
-[#acperm_rules_rv32]
-.ACPERM RV32 additional rules
-[%autowidth,float="center",align="center",cols="2,2,3",options="header"]
-|===
-| Rule | Permission   | Only valid if
-|  2   | <<c_perm>>   | <<r_perm>>
-|  3   | <<x_perm>>   | <<r_perm>>
-|  4   | <<w_perm>>   | not(<<c_perm>>) or (<<c_perm>> and not(<<lm_perm>>))
-|  5   | <<x_perm>>   | <<w_perm>> or <<c_perm>>
-|  6   | <<x_perm>>   | <<c_perm>> == <<lm_perm>> == <<el_perm>>
-| 12   | <<asr_perm>> | All other permissions are set.
-|===
-
-In all cases invalid permissions from <<acperm_rules_rv32>> are cleared.
+* For the RV32 only rules none of these combinations are possible to meet the encoding requirements for <<cap_perms_encoding32>>, or
+<<cap_perms_encoding_levels32>> if <<cheri_levels_ext,{cheri_levels_ext_name}>> is implemented. In all cases invalid permissions are cleared.
 
 .Capability permissions bit field
 [#acperm_bit_field]
--- a/src/insns/acperm_32bit.adoc
+++ b/src/insns/acperm_32bit.adoc
@@ -30,7 +30,6 @@ format shown in xref:acperm_bit_field[xrefstyle=short].
 . Calculate the bitwise AND of the bit field with the mask `rs2`.
 . If the AP and <<m_bit>> field in `cs1` could not have been produced by <<ACPERM>> then
 clear all AP permissions, and the <<m_bit>> to {CAP_MODE_VALUE}. Skip the next step.
-.. If <<cheri_levels_ext,{cheri_levels_ext_name}>> is implemented then also set <<section_cap_level,CL>> to 0 (_local_).
 . Clear AP permissions as required to meet the rules below.
 . Encode the AP permissions for RV32 according to <<cap_perms_encoding32>>.
 . Copy `cs1` to `cd`, and update the AP and SDP fields with the newly calculated versions.
--- a/src/insns/acperm_32bit.adoc
+++ b/src/insns/acperm_32bit.adoc
@@ -47,12 +47,12 @@ The rules from <<acperm_rules>> must be followed when removing permissions.
 |  2              | <<c_perm>>   | <<r_perm>> or <<w_perm>>
 |  3  (RV32 only) | <<c_perm>>   | <<r_perm>>
 |  4  (RV32 only) | <<x_perm>>   | <<r_perm>>
-|  5  (RV32 only) | <<w_perm>>   | not(<<c_perm>>) or (<<c_perm>> and not(<<lm_perm>>))
+|  5  (RV32 only) | <<w_perm>>   | not(<<c_perm>>) or <<lm_perm>>
 |  6  (RV32 only) | <<x_perm>>   | <<w_perm>> or <<c_perm>>
-|  7  (RV32 only) | <<x_perm>>   | <<c_perm>> == <<lm_perm>> == <<el_perm>>
-|  8              | <<sl_perm>>  | <<w_perm>> and <<c_perm>>
-|  9              | <<el_perm>>  | <<c_perm>> and <<r_perm>>
-| 10              | <<lm_perm>>  | <<c_perm>> and <<r_perm>>
+|  7              | <<el_perm>>  | <<c_perm>> and <<r_perm>>
+|  8              | <<lm_perm>>  | <<c_perm>> and <<r_perm>>
+|  9  (RV32 only) | <<x_perm>>   | <<c_perm>> == <<lm_perm>> == <<el_perm>>
+| 10              | <<sl_perm>>  | <<w_perm>> and <<c_perm>>
 | 11              | <<asr_perm>> | <<x_perm>>
 | 12              | <<m_bit>>    | <<x_perm>>
 |===
--- a/src/insns/acperm_32bit.adoc
+++ b/src/insns/acperm_32bit.adoc
@@ -50,7 +50,7 @@ The rules from <<acperm_rules>> must be followed when removing permissions.
 |  6  (RV32 only) | <<x_perm>>   | <<w_perm>> or <<c_perm>>
 |  7              | <<el_perm>>  | <<c_perm>> and <<r_perm>>
 |  8              | <<lm_perm>>  | <<c_perm>> and <<r_perm>>
-|  9  (RV32 only) | <<x_perm>>   | <<c_perm>> == <<lm_perm>> == <<el_perm>>
+|  9  (RV32 only) | <<x_perm>>   | (<<c_perm>> == <<lm_perm>> == <<el_perm>>) and (<<sl_perm>> == ∞)
 | 10              | <<sl_perm>>  | <<c_perm>>
 | 11              | <<asr_perm>> | <<x_perm>>
 | 12              | <<m_bit>>    | <<x_perm>>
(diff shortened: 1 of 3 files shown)


src/insns/acperm_32bit.adoc
```diff
@@ -40,6 +40,8 @@ These cases are defined to return useful minimal sets of permissions, which may
 +
 NOTE: Future extensions may allow more combinations of permissions, especially for MXLEN=64.
 
+All rules are applied in the order shown on this page.
+
```
arichardson, line 44: I guess the main problem is that the RV32 rules need to be inserted at the right location. Maybe we should change this to a list that includes all of them in order and prefix the RV32 ones with `RV32-only:...`

tariqkurd-repo: ok - we've reworked the rules - please have a look

src/insns/acperm_32bit.adoc
```diff
@@ -29,41 +29,53 @@ ACPERM performs the following operations:
...
-. If removing <<lm_perm>> results in unrepresentable permissions, then set <<r_perm>> and <<c_perm>> only.
+
+The rules from <<acperm_rules>> must be followed When removing permissions.
+
+[#acperm_rules]
+.ACPERM common rules
+[%autowidth,float="center",align="center",cols="2,3,3",options="header"]
+|===
+| Rule | Permission   | Only valid if
+|  1   | <<c_perm>>   | <<r_perm>> or <<w_perm>>
```
arichardson, line 46: Maybe this is simpler if we merge the tables and just do something like this:
```suggestion
|  1             | <<c_perm>>   | <<r_perm>> or <<w_perm>>
|  2 (RV32-only) | <<c_perm>>   | <<r_perm>>
```
Could also use a footnote instead of the parens?

src/insns/acperm_32bit.adoc
```diff
@@ -29,41 +29,53 @@ ACPERM performs the following operations:
...
+.ACPERM RV32 additional rules
+[%autowidth,float="center",align="center",cols="2,2,3",options="header"]
+|===
+| Rule | Permission   | Only valid if
+|  2   | <<c_perm>>   | <<r_perm>>
+|  3   | <<x_perm>>   | <<r_perm>>
+|  4   | <<w_perm>>   | not(<<c_perm>>) or (<<c_perm>> and not(<<lm_perm>>))
+|  5   | <<x_perm>>   | <<w_perm>> or <<c_perm>>
+|  6   | <<x_perm>>   | <<c_perm>> == <<lm_perm>> == <<el_perm>>
+| 12   | <<asr_perm>> | All other permissions are set.
```
arichardson, line 75: This could also be the first step, right? If we have all perms initially then the check is a no-op, and if anything is missing we also clear ASR.

arichardson (commented):
I think this is correct, but I'd like @LawrenceEsswood to have a look as well.

src/insns/acperm_32bit.adoc
tariqkurd-repo: yes I think that's fair comment - if anything missing after ANDing then clear ASR

src/insns/acperm_32bit.adoc
tariqkurd-repo: reformatted into one table

src/insns/acperm_32bit.adoc
```diff
@@ -29,41 +29,39 @@ ACPERM performs the following operations:
...
+[#acperm_rules]
+.ACPERM common rules
+[%autowidth,float="center",align="center",cols="3,3,3",options="header"]
+|===
+| Rule            | Permission   | Only valid if
+|  1  (RV32 only) | <<asr_perm>> | All other permissions are set.
+|  2              | <<c_perm>>   | <<r_perm>> or <<w_perm>>
+|  3  (RV32 only) | <<c_perm>>   | <<r_perm>>
+|  4  (RV32 only) | <<x_perm>>   | <<r_perm>>
+|  5  (RV32 only) | <<w_perm>>   | not(<<c_perm>>) or (<<c_perm>> and not(<<lm_perm>>))
```
LawrenceEsswood, line 50: `not(<<c_perm>>) or (<<c_perm>> and not(<<lm_perm>>))`

Simplifies to

`not(<<c_perm>>) or not(<<lm_perm>>)`

The negation (condition to clear) for that is `<<c_perm>> and <<lm_perm>>` which does not correctly match what it was before (clear if `<<c_perm>> and NOT <<lm_perm>>`).

This row should be:

|  5  (RV32 only) | <<w_perm>>   | not(<<c_perm>>) or <<lm_perm>>))

src/insns/acperm_32bit.adoc
```diff
@@ -29,41 +29,39 @@ ACPERM performs the following operations:
...
+|  6  (RV32 only) | <<x_perm>>   | <<w_perm>> or <<c_perm>>
+|  7  (RV32 only) | <<x_perm>>   | <<c_perm>> == <<lm_perm>> == <<el_perm>>
```
LawrenceEsswood, line 52: Disagree with the order here.

if I have a `Execute + Data & Cap RW` and strip C, I expect to end up with `Execute + Data RW`. This would drop me out of X completely.

Rules 9 and 10 should apply first.

src/insns/acperm_32bit.adoc
```diff
@@ -29,41 +29,39 @@ ACPERM performs the following operations:
...
+|  8              | <<sl_perm>>  | <<w_perm>> and <<c_perm>>
```
LawrenceEsswood, line 53: Still very much of the opinion that it was correct above and this was only conditioned on C, not W,  so that a read-only cap could still have a store-level for the purpose of future extensions where we limit the store level transtively.

LawrenceEsswood (changes requested):

src/insns/acperm_32bit.adoc
tariqkurd-repo: ok fixed

src/insns/acperm_32bit.adoc
tariqkurd-repo: understood - fixed

src/insns/acperm_32bit.adoc
```diff
@@ -47,12 +47,12 @@ The rules from <<acperm_rules>> must be followed when removing permissions.
...
+|  5  (RV32 only) | <<w_perm>>   | not(<<c_perm>>) or <<lm_perm>>
 |  6  (RV32 only) | <<x_perm>>   | <<w_perm>> or <<c_perm>>
-|  7  (RV32 only) | <<x_perm>>   | <<c_perm>> == <<lm_perm>> == <<el_perm>>
-|  8              | <<sl_perm>>  | <<w_perm>> and <<c_perm>>
-|  9              | <<el_perm>>  | <<c_perm>> and <<r_perm>>
-| 10              | <<lm_perm>>  | <<c_perm>> and <<r_perm>>
+|  7              | <<el_perm>>  | <<c_perm>> and <<r_perm>>
+|  8              | <<lm_perm>>  | <<c_perm>> and <<r_perm>>
+|  9  (RV32 only) | <<x_perm>>   | <<c_perm>> == <<lm_perm>> == <<el_perm>>
+| 10              | <<sl_perm>>  | <<w_perm>> and <<c_perm>>
```
LawrenceEsswood: `10              | <<sl_perm>>  | <<w_perm>> and <<c_perm>>`

Should be

`10              | <<sl_perm>>  | <<c_perm>>` ?

src/insns/acperm_32bit.adoc
LawrenceEsswood: I meant the text before change "as above" (just because thats where it appeared on my screen...)


It was

```
. <<sl_perm>> cannot be set without <<c_perm>> being set
.. Zero <<sl_perm>> unless <<c_perm>> is set.
```

Which had no condition on W. This is not to protect a transition, but to ensure if we introduce a future implict ANDPERM with SL on load, then even capabilities without W can theoretically have an SL so that such an AND does what we want.

LawrenceEsswood (approved):

tariqkurd-repo:
@tomaird - I can't add you as a reviewer but can you check this one?

arichardson (approved):
This looks good to me now - but I'll wait a bit for other to comment on this before merging.

src/insns/acperm_32bit.adoc
```diff
@@ -29,41 +29,38 @@ ACPERM performs the following operations:
...
+|  9  (RV32 only) | <<x_perm>>   | <<c_perm>> == <<lm_perm>> == <<el_perm>>
+| 10              | <<sl_perm>>  | <<c_perm>>
+| 11              | <<asr_perm>> | <<x_perm>>
+| 12              | <<m_bit>>    | <<x_perm>>
+|===
+
+The behaviour of currently illegal combinations from <<acperm_rules>> is to clear the permission if invalid (or in the case of <<sl_perm>> set it to 0 (_local_)).
+
+* For RV64 all such combinations may be redefined by future extensions.
+* For the RV32 only rules none of these combinations are possible to meet the encoding requirements for <<cap_perms_encoding32>>, or
```
tariqkurd-repo, line 62:
```suggestion
* The RV32 only rules are added because they remove combinations which do not meet the encoding requirements for <<cap_perms_encoding32>>, or
```

src/insns/acperm_32bit.adoc
```diff
@@ -29,41 +29,38 @@ ACPERM performs the following operations:
...
+<<cap_perms_encoding_levels32>> if <<cheri_levels_ext,{cheri_levels_ext_name}>> is implemented. In all cases invalid permissions are cleared.
```
tariqkurd-repo, line 63:
```suggestion
<<cap_perms_encoding_levels32>> if <<cheri_levels_ext,{cheri_levels_ext_name}>> is implemented.
```

tomaird:
If we start with the following perms set: `R W C X LM ASR EL, mode=1 (int)` (i.e. all bits set except SL), e.g from doing an `acperm` with cs1 having infinite perms and rs2 being `0xfffffff7`

Then apply the new rules:
```
1 - clear ASR because SL is clear
2 - leave C set
3 - leave C set
4 - leave X set
5 - leave W set (beacuse LM set)
6 - leave X set
7 - leave EL set
8 - leave LM set
9 - leave X set
10 - SL was already clear
11 - ASR was cleared by (1)
12 - leave M bit set (int mode)
```

So we only cleared ASR.

This leaves us with `R W C X LM EL, mode=1 (int)` which doesn't have an encoding in RV32

So I think rule 9 possible needs extending to something like: 
```
 9  (RV32 only) | <<x_perm>>   | <<c_perm>> == <<lm_perm>> == <<el_perm>> and <<sl_perm>> == ∞
```
so we clear X and end up in quadrant 3 for this case.

arichardson:
> If we start with the following perms set: `R W C X LM ASR EL, mode=1 (int)` (i.e. all bits set except SL), e.g from doing an `acperm` with cs1 having infinite perms and rs2 being `0xfffffff7`
> 
> Then apply the new rules:
> 
> ```
> 1 - clear ASR because SL is clear
> 2 - leave C set
> 3 - leave C set
> 4 - leave X set
> 5 - leave W set (beacuse LM set)
> 6 - leave X set
> 7 - leave EL set
> 8 - leave LM set
> 9 - leave X set
> 10 - SL was already clear
> 11 - ASR was cleared by (1)
> 12 - leave M bit set (int mode)
> ```
> 
> So we only cleared ASR.
> 
> This leaves us with `R W C X LM EL, mode=1 (int)` which doesn't have an encoding in RV32
> 
> So I think we possibly need a new rule to clear X in this case so we end up in quadrant 3, or remove C+LM+EL to stay in quadrant 1 as `Execute + Data RW`

Good catch - I'd keep X in this case so go for your second suggestion.

tariqkurd-repo:
One thing to consider is what happens if SL is cleared but is in a N/A state, we could just ignore it

src/insns/acperm_32bit.adoc
```diff
@@ -29,41 +29,38 @@ ACPERM performs the following operations:
...
 . Clear AP permissions as required to meet the rules below.
-. Encode the AP permissions for MXLEN=32 according to <<cap_perms_encoding32>>.
+. Encode the AP permissions for RV32 according to <<cap_perms_encoding32>>.
 . Copy `cs1` to `cd`, and update the AP and SDP fields with the newly calculated versions.
 . Set `cd.tag=0` if `cs1` is sealed or if any reserved fields of `cs1` are set.
-+
-Some combinations of permissions cannot be encoded for MXLEN=32, and are not useful when MXLEN=64.
-These cases are defined to return useful minimal sets of permissions, which may be no permissions.
-+
-NOTE: Future extensions may allow more combinations of permissions, especially for MXLEN=64.
```
nwf, line 41: Please keep something along these lines in the specification so that CHERIoT's permission semilattice could be a valid extension of Zcheripurecap.

arichardson: Agreed, I'd suggest keeping this note.

src/level-ext.adoc
```diff
@@ -74,14 +74,14 @@ endif::[]
...
+| 2-3   | ✔ |   | ✔ | ✔  | ✔  | ∞^1^| ✔ |     | Mode^1^  | Execute + Data & Cap RO
 | 4-5   | ✔ | ✔ | ✔ | ✔  | ✔  | ∞   | ✔ |     | Mode^1^  | Execute + Data & Cap RW
-| 6-7   | ✔ | ✔ |   |    |    | N/A | ✔ |     | Mode^1^  | Execute + Data RW
+| 6-7   | ✔ | ✔ |   |    |    | ∞^1^| ✔ |     | Mode^1^  | Execute + Data RW
 11+| *Quadrant 2: Restricted capability data read/write*
 11+| bit[2] = write, bit[1:0] = store level. R and C implicitly granted, LM dependent on W permission.
 |Bits[4:3]| R | W | C | LM | EL | SL    | X | ASR | Mode^1^ |
 | 0-2   10+| reserved
-| 3       | ✔ |   | ✔ |    |    | N/A   |   |     | N/A | Data & Cap R0 (without <<lm_perm>>)
+| 3       | ✔ |   | ✔ |    |    | ∞^1^  |   |     | N/A | Data & Cap R0 (without <<lm_perm>>)
```
arichardson, line 84:
```suggestion
| 3       | ✔ |   | ✔ |    |    | 0^1^  |   |     | N/A | Data & Cap R0 (without <<lm_perm>>)
```

src/level-ext.adoc
```diff
@@ -91,13 +91,15 @@ endif::[]
 11+| _Reserved bits for future extensions must be 1 so they are implicitly granted_
 |Bits[4:3]| R | W | C | LM | EL | SL    | X | ASR | Mode^1^ |
 | 0-2   10+| reserved
-| 3       | ✔ |   | ✔ | ✔  | ✔  | N/A   |   |     | N/A | Data & Cap R0
+| 3       | ✔ |   | ✔ | ✔  | ✔  | ∞^1^  |   |     | N/A | Data & Cap R0
```
arichardson, line 94:
```suggestion
| 3       | ✔ |   | ✔ | ✔  | ✔  | 0^1^  |   |     | N/A | Data & Cap R0
```

src/level-ext.adoc
```diff
@@ -91,13 +91,15 @@ endif::[]
...
 | 4       | ✔ | ✔ | ✔ | ✔  | ✔  | _(3)_ |   |     | N/A | Reserved for `LVLBITS=2`
 | 5       | ✔ | ✔ | ✔ | ✔  | ✔  | _(2)_ |   |     | N/A | Reserved for `LVLBITS=2`
 | 6       | ✔ | ✔ | ✔ | ✔  | ✔  | 1     |   |     | N/A | Data & Cap RW (with store _local_)
 | 7       | ✔ | ✔ | ✔ | ✔  | ✔  | 0     |   |     | N/A | Data & Cap RW (no store _local_)
 |==============================================================================
 
+^1^ SL isn't applicable in these cases, but is specified as ∞ to help the definition of <<ACPERM>>
```
arichardson, line 101:
```suggestion
^1^ SL isn't applicable in these cases, but reports this value in <<GCPERM>> to simplify the rules followed by <<ACPERM>>
```

Maybe something like this so we can report 0 for the Q2&Q3 cases.

src/level-ext.adoc
arichardson: IMO this should be zero for consistency: just report the negation of bit 0 (and bit 0-1 for multiple level bits).

src/level-ext.adoc
arichardson: IMO this should be zero for consistency: just report the negation of bit 0 (and bit 0-1 for multiple level bits).

src/insns/acperm_32bit.adoc
```diff
@@ -29,41 +29,38 @@ ACPERM performs the following operations:
...
-NOTE: The combination of <<x_perm>> clear and <<m_bit>> set is reserved for future extensions.
-
-The MXLEN=32 additional rules are:
-
-. Clear <<asr_perm>> unless _all_ other permissions are set
-. Clear <<c_perm>> and <<x_perm>> if <<r_perm>> is not set
-. Clear <<w_perm>> if <<c_perm>> _is_ set, but <<lm_perm>> _is not_ set.
-. Clear <<x_perm>> if <<x_perm>> and <<r_perm>> are set, and <<c_perm>> and <<w_perm>> are both clear.
-. If removing <<lm_perm>> results in unrepresentable permissions, then set <<r_perm>> and <<c_perm>> only.
+
```
arichardson, line 37:
```suggestion
+
Some combinations of permissions cannot be encoded for MXLEN=32, and are not useful when MXLEN=64.
These cases are defined to return useful minimal sets of permissions, which may be no permissions.
+
NOTE: Future extensions may allow more combinations of permissions, especially for MXLEN=64.
```

arichardson: Let's keep this note as suggested.

arichardson (approved):
Minor comment otherwise LGTM

src/level-ext.adoc
LawrenceEsswood: I don't think it hurts to have 0 here (if as Alex says it adds consistency). Given this cap does not have LM, it will never load anything to which SL (even transitively) would ever apply.

src/level-ext.adoc
LawrenceEsswood: This one can load a W cap, and so its SL would be important if we add a transitive SL. In an earlier sketch we had all combinations of SL and W=0 in this quadrant for this reason.

I think we decided it was fine to reserve 3/4 of these now (and just have the one, 3 or 0). Whichever one we have selected here with the bit pattern is what we should report. I think that is (as Alex suggests) 0, to match the W=1 case. We can add 1/2/3 if required. This is slightly unlike Quad 1, where I expect to have infinity regardless of how many levels there are.

src/level-ext.adoc
```diff
@@ -74,14 +74,14 @@ endif::[]
 11+| bit[0] - <<m_bit>> ({CAP_MODE_VALUE}-{cheri_cap_mode_name}, {INT_MODE_VALUE}-{cheri_int_mode_name})
 |Bits[4:3]| R | W | C | LM | EL | SL  | X | ASR | Mode^1^ |
 | 0-1   | ✔ | ✔ | ✔ | ✔  | ✔  | ∞   | ✔ |  ✔  | Mode^1^  | Execute + ASR (see <<infinite-cap>>)
-| 2-3   | ✔ |   | ✔ | ✔  | ✔  | ∞   | ✔ |     | Mode^1^  | Execute + Data & Cap RO
+| 2-3   | ✔ |   | ✔ | ✔  | ✔  | ∞^1^| ✔ |     | Mode^1^  | Execute + Data & Cap RO
 | 4-5   | ✔ | ✔ | ✔ | ✔  | ✔  | ∞   | ✔ |     | Mode^1^  | Execute + Data & Cap RW
-| 6-7   | ✔ | ✔ |   |    |    | N/A | ✔ |     | Mode^1^  | Execute + Data RW
+| 6-7   | ✔ | ✔ |   |    |    | ∞^1^| ✔ |     | Mode^1^  | Execute + Data RW
```
LawrenceEsswood, line 79: This could be 0 or ∞, because it truly doesn't matter as we will never load a writable cap via this (bearing future transitive SL in mind). 0 might be slightly friendlier. One could imagine a structure made deeply not SL, with some function pointers in it. Those should be sealed, which is why I don't really think this will ever be bothersome. But if there were any unsealed ones, and they had SL reduced, I think it would be more expected that they lose C, not X, as a result (possibly, they really should be sealed, so who knows what is more useful there).

Given that it is probably not important, and the rule "lose X if SL not inf" is easier than something like: "lose X if SL if inf != W", I don't have strong feelings here. Maybe somebody else has thoughts here?

LawrenceEsswood:
I think I agree with Alex on Quads 3/4. The encoding was meant to align certain bits with SL (possibly inverted).

src/level-ext.adoc
arichardson: Yes I agree if we end up adding the transitive property, then 2 should report SL=1, and for n bit we add 1 with SL=2 and zero with SL=3.

src/level-ext.adoc
tariqkurd-repo: I have changed it. And we've also changed the order of rules 9 and 10, and update (former) rule 9 to allow for SL=0 in this case.

Previously it was testing for SL (former rule 9) and then clearning SL (former rule 10) which seems the wrong way around

With the latest commit if you start with infinite permissions, just remove C then you stay in quadrant 1 and end on 6-7, and so retain the X permission.

Previously the X was lost which seemed a shame, as it could have been kept.

arichardson (approved):

PR #433, opened 2024-10-24
Fix minor typo in PTE section

Timmmm:
what -> that, and I think "known to not contain" is a bit easier to parse than "known not to contain".


jrtc27:
Splitting infinitives can be a contentious subject. Some might also prefer "which" to "that", at least in British English, though in this instance it seems American English prescribes "that".

arichardson (approved):
IMO the what->that makes the sentence easier to parse. No opinion on the `not to` -> `to not`.

Issue #432, opened 2024-10-22
EL permission rules clarification


tomaird:
The summary table describes that `CL` and `EL` are only cleared if `CL=0`  on the auth cap (amongst other conditions):

![image](https://github.com/user-attachments/assets/fea583c6-a134-4e26-ba1c-d5aaf9908c80)

But all the descriptions of `EL` don't mention that `CL` must be zero on the auth cap, but do mention all the other conditions. e.g. in `LC`:

![image](https://github.com/user-attachments/assets/dd6bf43a-6753-4855-a8d0-06546036d76c)

I'm assuming the table is correct and `auth_cap.CL=0` is required to clear `CL` and `EL` in the data cap, but it's a bit confusing to leave this out of these text descriptions.

Note, this is not a problem for the SL-perm rules which describe the conditions in full.

jamie-melling:
To me, it seems like changing the `CL` is more like a demotion than removing a permission.  If your auth cap is Global, then you can load Local or Global capabilities, but if your capability is Local, then can only load Local capabilities using it - hence setting `CL=0` when `auth_cap.CL ==0`. 
"restricting capability level to the level of the authorizing capability" is more future proof wording if the capability levels were ever increased.

PR #436, opened 2024-10-29
local/global clarification

tariqkurd-repo:
Fix https://github.com/riscv/riscv-cheri/issues/432

### DIFF
--- a/src/insns/atomic_exceptions.adoc
+++ b/src/insns/atomic_exceptions.adoc
@@ -8,7 +8,7 @@ If <<c_perm>> is not granted then store the memory tag as zero, and load `cd.tag
 +
 If the authorizing capability does not grant <<lm_perm>>, and the tag of `cd` is 1 and `cd` is not sealed, then an implicit <<ACPERM>> clearing <<w_perm>> and <<lm_perm>> is performed to obtain the intermediate permissions on `cd` (see <<LC>>).
 +
-If the authorizing capability does not grant <<el_perm>>, and the tag of `cd` is 1, then an implicit <<ACPERM>> clearing <<el_perm>> and restricting <<section_cap_level>> to the level of the authorizing capability is performed to obtain the final permissions on `cd` (see <<LC>>).
+If the authorizing capability does not grant <<el_perm>>, the authorizing <<section_cap_level>> is _local_ and the tag of `cd` is 1, then an implicit <<ACPERM>> clearing <<el_perm>> and restricting the <<section_cap_level>> to the level of the authorizing capability is performed to obtain the final permissions on `cd` (see <<LC>>).
 +
 endif::[]
 ifndef::cap_atomic[]
--- a/src/insns/atomic_exceptions.adoc
+++ b/src/insns/atomic_exceptions.adoc
@@ -8,7 +8,7 @@ If <<c_perm>> is not granted then store the memory tag as zero, and load `cd.tag
 +
 If the authorizing capability does not grant <<lm_perm>>, and the tag of `cd` is 1 and `cd` is not sealed, then an implicit <<ACPERM>> clearing <<w_perm>> and <<lm_perm>> is performed to obtain the intermediate permissions on `cd` (see <<LC>>).
 +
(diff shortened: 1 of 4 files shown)


src/insns/atomic_exceptions.adoc
```diff
@@ -8,7 +8,7 @@ If <<c_perm>> is not granted then store the memory tag as zero, and load `cd.tag
 +
 If the authorizing capability does not grant <<lm_perm>>, and the tag of `cd` is 1 and `cd` is not sealed, then an implicit <<ACPERM>> clearing <<w_perm>> and <<lm_perm>> is performed to obtain the intermediate permissions on `cd` (see <<LC>>).
 +
-If the authorizing capability does not grant <<el_perm>>, and the tag of `cd` is 1, then an implicit <<ACPERM>> clearing <<el_perm>> and restricting <<section_cap_level>> to the level of the authorizing capability is performed to obtain the final permissions on `cd` (see <<LC>>).
+If the authorizing capability does not grant <<el_perm>>, the authorizing <<section_cap_level>> is _local_ and the tag of `cd` is 1, then an implicit <<ACPERM>> clearing <<el_perm>> and restricting the <<section_cap_level>> to the level of the authorizing capability is performed to obtain the final permissions on `cd` (see <<LC>>).
```
arichardson, line 11: I omitted `the authorizing <<section_cap_level>>` initially, since the <<el_perm>> should always be cleared and limiting CL to the authorizing cap's CL does nothing if the auth cap has a CL of _global_.

CC: @LawrenceEsswood

src/level-ext.adoc
```diff
@@ -108,6 +108,8 @@ This avoids the need for a dedicated instruction and allows reducing the level a
 [#section_cap_level_summary]
 === Capability level summary table
 
+NOTE: A capability with <<section_cap_level,CL>>=1 is *global* and with <<section_cap_level,CL>>=0 is *local*.
```
arichardson, line 111: We use _local_ and _global_ in other places.
```suggestion
NOTE: A capability with <<section_cap_level,CL>>=1 is _global_ and with <<section_cap_level,CL>>=0 is _local_.
```

arichardson (commented):
I think the added condition on the load is not correct, but otherwise the changes look good.

tariqkurd-repo:
> I think the added condition on the load is not correct, but otherwise the changes look good.

which condition?

davidchisnall:
@nwf

src/insns/atomic_exceptions.adoc
LawrenceEsswood: Correct. This ACPERM should happen just for the purpose of clearing EL even if loading by a global capability.

src/level-ext.adoc
```diff
@@ -74,7 +74,7 @@ endif::[]
 11+| bit[0] - <<m_bit>> ({CAP_MODE_VALUE}-{cheri_cap_mode_name}, {INT_MODE_VALUE}-{cheri_int_mode_name})
 |Bits[4:3]| R | W | C | LM | EL | SL  | X | ASR | Mode^1^ |
 | 0-1   | ✔ | ✔ | ✔ | ✔  | ✔  | ∞   | ✔ |  ✔  | Mode^1^  | Execute + ASR (see <<infinite-cap>>)
-| 2-3   | ✔ |   | ✔ | ✔  | ✔  | ∞   | ✔ |     | Mode^1^  | Execute + Data & Cap RO
```
LawrenceEsswood, line 77: This was meant to be inf.

Although it doesn't seem to matter what store-level a capability is if it cannot store, if we later want to introduce a  transitive store level policy this would be used to limit any capabilities loaded.

Getperm has to report something anyway, min and max being sensible defaults, and max is useful here.

src/insns/atomic_exceptions.adoc
tariqkurd-repo: ok fixed

src/level-ext.adoc
tariqkurd-repo: ok, fair enough

arichardson:
> > I think the added condition on the load is not correct, but otherwise the changes look good.
> 
> which condition?

The one you just removed (`the authorizing <<section_cap_level>> is _local_`). Looks good now.

arichardson (approved):

LawrenceEsswood (approved):

PR #439, opened 2024-10-30
Clarify interaction of levels with sealed capabilities

arichardson:
Sealed capabilities should only have their `CL` changed, but not the `EL` permission.

See also https://github.com/CHERIoT-Platform/cheriot-sail/issues/14

### DIFF
--- a/src/level-ext.adoc
+++ b/src/level-ext.adoc
@@ -122,19 +122,15 @@ NOTE: A capability with <<section_cap_level,CL>>=1 is _global_ and with <<sectio
 
 NOTE: if W=0 or C=0 then SL is irrelevant
 
-.{cheri_levels_ext_name} `LVLBITS=1` summary table for loaded capabilities
-[#cap_level_load_summary,width="100%",options=header,align=center,cols="1,1,1,1,1,6"]
+.{cheri_levels_ext_name} additional rules for loading capabilities
+[#cap_level_load_summary,width="100%",options=header,align=center,cols="1,1,1,1,1,1,6"]
 |==============================================================================
-   4+|Auth cap field             | Data cap field |
-    h|*R*   h|*C* h|*EL* h|*CL* h| Tag           h|Action
-.5+.^|1 .5+.^| 1   | 1    | X    | X              |Load data capability unmodified
-              .4+.^| 0    | 1    | 1              |Load data capability with EL=0
-                          | 1    | 0              |Load data capability unmodified
-                     .2+.^| 0    | 1              |Load data capability with CL=0,EL=0
-                                 | 0              |Load data capability unmodified
+   4+|Auth cap field                  2+| Data cap field |
+    h|*R*   h|*C* h|*EL*      h|*CL*   h| Tag h| Sealed h|Action
+.2+.^|1 .2+.^| 1 .2+.^| 0 .2+.^| X .2+.^| 1    | Yes     |Load data capability with `CL=min(auth.CL, data.CL)`, EL unchanged
+                                               | No      |Load data capability with `EL=0, CL=min(auth.CL, data.CL)`
+   6+| All other cases                                   |Load data capability with EL, CL unmodified
 |==============================================================================
-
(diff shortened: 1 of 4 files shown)


src/level-ext.adoc
```diff
@@ -123,15 +123,15 @@ NOTE: A capability with <<section_cap_level,CL>>=1 is _global_ and with <<sectio
...
-.5+.^|1 .5+.^| 1   | 1    | X    | X              |Load data capability unmodified
-              .4+.^| 0    | 1    | 1              |Load data capability with EL=0
-                          | 1    | 0              |Load data capability unmodified
-                     .2+.^| 0    | 1              |Load data capability with CL=0,EL=0
-                                 | 0              |Load data capability unmodified
+   4+|Auth cap field             2+| Data cap field |
+    h|*R*   h|*C* h|*EL* h|*CL*   h| Tag h| Sealed h|Action
+.5+.^|1 .5+.^| 1   | 1 .3+.^| X    | X    | X  .2+.^|Load data capability unmodified
+              .4+.^| 0             | 0    | N/A
+                              .3+.^| 1    | Yes     |Load data capability with `CL=min(auth.CL, data.CL)`, EL unchanged
```
arichardson, line 132:
```suggestion
                              .3+.^| 1    | Yes     |Load data capability with `CL=min(auth.CL, data.CL)`, EL unchanged
```

I could use AND here instead, but `min` makes it portable to LVLBITS>1

LawrenceEsswood:
I agree with the CheriOT decisions here and that only level is outside of sealing,  this LGTM.

nwf:
It's worth noting that we're not _completely_ happy with the CHERIoT solution, since it means that it's possible to strip `GL`(obal) off a sealed sentry and so deprive the unsealed form, and all its progeny, of `GL`.

In practice, this isn't a huge problem _for us_, because we
1. generally don't derive long-lived capabilities to sub-objects from _data-typed_ sentries, and
2. are quite restrictive in the kinds of _executable-typed_ sentries we pass around:
   - intra-compartment return sentries
   - library function entry points (forward sentries)
   - forward and reverse sentries for the switcher

Flipping bits on intra-compartment anything is just going to hurt yourself.  Libraries are much like your own code, and jumping in with a non-`GL` sentry is just going to mean anything derived from its PCC is also non-`GL`, but the library is unlikely to store those beyond the stack anyway (of course, the caller might, but again: don't hurt yourself).  The switcher does not derive sub-objects from PCC.

But, even though it seems to work out OK for us, it's still not _ideal_, but we have yet to come up with a better solution.  It _might_ be a problem for different compartmentalization models that try to use sentries more directly.

tariqkurd-repo:
The table is very hard to read formatted like that, I've made a local modification so it looks like this:

image

so now I can start to unpick the differences. For example:

EL=0 CL=1 Tag=1 Sealed=Yes -> reduce CL, EL unchanged (was unmodified)
EL=0 CL=1 Tag=1 Sealed=No -> EL=0, CL unchanged (was unmodified)

but also the table isn't fully specified, for example

EL=0 CL=0 Tag=0 Sealed=X -> what happens here? It should be unmodified.

I think there must be a clearer way of specifying this and covering all of the cases.

davidchisnall:
> It's worth noting that we're not _completely_ happy with the CHERIoT solution, since it means that it's possible to strip `GL`(obal) off a sealed sentry and so deprive the unsealed form, and all its progeny, of `GL`.

I think we are happy with that behaviour (we use sealing for software-defined capabilities and it *is* useful to be able to provide a no-capture guarantee to delegated capabilities.  Software that needs to capture can explicitly check for G and error, or can embed a G capability to the object inside the outer sealed object).  The thing that we're not happy about (and have an issue to track) is that we can strip G by round tripping through memory but we can't do it via candperm.  We aim to fix that before 1.0.

src/level-ext.adoc
tariqkurd-repo: ok

tariqkurd-repo (approved):
I added the note about CL changing evenon a sealed cap

Issue #428, opened 2024-10-21
Zcherilevels RV32 stripping EL clarification


jamie-melling:
If you have a quadrant 1 capability with perms[2:0] equal to either 0-1 or 4-5 (with w-perm) and remove just `EL`, there are two possible outcomes, neither of which break monotonicity. Either you stay in quadrant 1 and strip `C` and `LM`  to get an `R`, `W` and `X` capability.  The other option is to create a quadrant 2 capability where you strip `LM`, `W` and `X` to get an `R` and `C` capability.  The rules laid out in ACPERM do not cover which mutation is desired.

One other note, small spec bug in ACPERM rules.
![image](https://github.com/user-attachments/assets/d5148926-c0d5-4912-ac53-959ae9e062bd)

tariqkurd-repo:
It looks like these rules need a thorough review.
Here: https://github.com/riscv/riscv-cheri/pull/421 we default to RC permission where there is a choice, so maybe we shoudl do the same here.

tomaird:
Typo fix here: #429

jamie-melling:
One other thing worth double-checking, in ACPERM, there is

> If the AP and M-bit field in cs1 could not have been produced by ACPERM then clear all AP permissions and the M-bit, and skip the next step

and further down on the ACPERM section

> Even though being included here CL is not considered an architectural permission.

So does the ACPERM instruction set the `CL` field to 0 when `cs1` could not have been produced by ACPERM?

It is implied that GCPERM will return 0's if the same check fails for that instruction

> If the AP field cannot be produced by ACPERM then all architectural permission bits in rd are set to 0.

PR #440, opened 2024-11-05
Fix RV32 Zcherilevels ACPERM

tariqkurd-repo:
This version has been verified by a python script which checks all possible combinations of removing permissions from maximum permissions

additional fix for https://github.com/riscv/riscv-cheri/issues/428

### DIFF
--- a/src/insns/acperm_32bit.adoc
+++ b/src/insns/acperm_32bit.adoc
@@ -52,9 +52,9 @@ The rules from <<acperm_rules>> must be followed when removing permissions.
 |  4  (RV32 only) | <<x_perm>>   | <<r_perm>>
 |  5  (RV32 only) | <<w_perm>>   | not(<<c_perm>>) or <<lm_perm>>
 |  6  (RV32 only) | <<x_perm>>   | <<w_perm>> or <<c_perm>>
-|  7              | <<el_perm>>  | <<c_perm>> and <<r_perm>>
-|  8              | <<lm_perm>>  | <<c_perm>> and <<r_perm>>
-|  9              | <<sl_perm>>  | <<c_perm>>
+|  7              | <<el_perm>>  | <<c_perm>> and <<r_perm>> and <<lm_perm>>
+|  8              | <<lm_perm>>  | <<c_perm>> and <<r_perm>> and (<<w_perm>> or <<el_perm>>)
+|  9              | <<sl_perm>>  | (<<c_perm>> and <<lm_perm>> and (<<x_perm>> or <<w_perm>>)
 | 10  (RV32 only) | <<x_perm>>   | (<<c_perm>> and <<lm_perm>> and <<el_perm>> and (<<sl_perm>> == ∞)) or +
                                    (not(<<c_perm>> and not(<<lm_perm>>) and not(<<el_perm>>) and (<<sl_perm>>==0)))^1^
 | 11              | <<asr_perm>> | <<x_perm>>
--- a/src/insns/acperm_32bit.adoc
+++ b/src/insns/acperm_32bit.adoc
@@ -52,13 +52,16 @@ The rules from <<acperm_rules>> must be followed when removing permissions.
 |  4  (RV32 only) | <<x_perm>>   | <<r_perm>>
 |  5  (RV32 only) | <<w_perm>>   | not(<<c_perm>>) or <<lm_perm>>
 |  6  (RV32 only) | <<x_perm>>   | <<w_perm>> or <<c_perm>>
-|  7              | <<el_perm>>  | <<c_perm>> and <<r_perm>> and <<lm_perm>>
-|  8              | <<lm_perm>>  | <<c_perm>> and <<r_perm>> and (<<w_perm>> or <<el_perm>>)
-|  9              | <<sl_perm>>  | (<<c_perm>> and <<lm_perm>> and (<<x_perm>> or <<w_perm>>)
-| 10  (RV32 only) | <<x_perm>>   | (<<c_perm>> and <<lm_perm>> and <<el_perm>> and (<<sl_perm>> == ∞)) or +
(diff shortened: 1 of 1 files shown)


src/insns/acperm_32bit.adoc
```diff
@@ -52,9 +52,9 @@ The rules from <<acperm_rules>> must be followed when removing permissions.
 |  4  (RV32 only) | <<x_perm>>   | <<r_perm>>
 |  5  (RV32 only) | <<w_perm>>   | not(<<c_perm>>) or <<lm_perm>>
 |  6  (RV32 only) | <<x_perm>>   | <<w_perm>> or <<c_perm>>
-|  7              | <<el_perm>>  | <<c_perm>> and <<r_perm>>
-|  8              | <<lm_perm>>  | <<c_perm>> and <<r_perm>>
-|  9              | <<sl_perm>>  | <<c_perm>>
+|  7              | <<el_perm>>  | <<c_perm>> and <<r_perm>> and <<lm_perm>>
+|  8              | <<lm_perm>>  | <<c_perm>> and <<r_perm>> and (<<w_perm>> or <<el_perm>>)
+|  9              | <<sl_perm>>  | (<<c_perm>> and <<lm_perm>> and (<<x_perm>> or <<w_perm>>)
```
arichardson, line 57: Do we really need to explicitly list LM here? I guess it's implied since SL depends on W, and W implies LM. Not opposed to keeping it since I imagine the checks will be optimized out but wondering if the text can be simplified.

tariqkurd-repo:
The modifications should have been for RV32 only, so this needs fixing

src/insns/acperm_32bit.adoc
```diff
@@ -52,9 +52,9 @@ The rules from <<acperm_rules>> must be followed when removing permissions.
 |  4  (RV32 only) | <<x_perm>>   | <<r_perm>>
 |  5  (RV32 only) | <<w_perm>>   | not(<<c_perm>>) or <<lm_perm>>
 |  6  (RV32 only) | <<x_perm>>   | <<w_perm>> or <<c_perm>>
-|  7              | <<el_perm>>  | <<c_perm>> and <<r_perm>>
-|  8              | <<lm_perm>>  | <<c_perm>> and <<r_perm>>
-|  9              | <<sl_perm>>  | <<c_perm>>
+|  7              | <<el_perm>>  | <<c_perm>> and <<r_perm>> and <<lm_perm>>
```
arichardson, line 55: This is basically `(quandrant == 3) || (quadrant == 1 && C)` which I guess lines up with these constraints. It's a bit weird to list LM as requirement for the unrelated EL permission, so this should be RV32 only.
```suggestion
|  7  (RV32 only) | <<el_perm>>  | <<c_perm>> and <<r_perm>> and <<lm_perm>>
```

src/insns/acperm_32bit.adoc
```diff
@@ -52,9 +52,9 @@ The rules from <<acperm_rules>> must be followed when removing permissions.
...
+|  8              | <<lm_perm>>  | <<c_perm>> and <<r_perm>> and (<<w_perm>> or <<el_perm>>)
```
arichardson, line 56: The second part of the constraint should be RV32-only - for RV64 we should be able to have LM without EL.

src/insns/acperm_32bit.adoc
arichardson:
```suggestion
|  9  (RV32 only) | <<sl_perm>>  | (<<c_perm>> and <<lm_perm>> and (<<x_perm>> or <<w_perm>>)
```

arichardson (commented):
I think these constraints should be scoped to RV32 but it looks sensible to me. The logic should be folded to something simpler anyway so I don't think it's a huge problem that these clauses are getting somewhat complex.

src/insns/acperm_32bit.adoc
tariqkurd-repo: yes - RV32 only was missing - will push a new commit soon

src/insns/acperm_32bit.adoc
tariqkurd-repo: same comment as above

src/insns/acperm_32bit.adoc
tariqkurd-repo: same comment as above

src/insns/acperm_32bit.adoc
jamie-melling: If you remove `LM` from the SL term in rule 9 there are 10 routes to get an invalid encoding, but they all end with the same permissions. 
| R | W | C | LM | SL | X | Rule |
|--------|--------|--------|--------|--------|--------|--------|
| 1 | 1 | 1 | 0 | 1 | 1 | After & |
| 1 | 0 | 1 | 0 | 1 | 1 | Rule 5 - clear W |
| 1 | 0 | 1 | 0 | 1 | 1 | Rule 9 - but no LM inclusion (no change) |
| 1 | 0 | 1 | 0 | 1 | 0 | Rule 10 - clear X |

This leaves us with `R, C, SL` which doesn't map to any table entry.  If you have `LM` as part of the rule, then you remove `SL` end at `R, C`.

jamie-melling: Rule 9 without `LM` does not clear `SL` because `X` is still HIGH, but `SL` needs to be aware of `X` because Quadrant 1, row 2/3 has  `W == 0` and `C, LM, EL == 1` and `SL == inf` while Quadrant 3, row 3 has identical values for `W, C, LM, EL` however `SL=0`.  So the only differentiator between these two cases is that in Quadrant 1 has `X`

arichardson (approved):

src/insns/acperm_32bit.adoc
tariqkurd-repo: closing as Alex has approved the PR

PR #442, opened 2024-11-06
clarify that quad 2 and 3 entries 4 and 5 are reserved when LVLBITS=2

tariqkurd-repo:
also note that Zcherilevels with LVLBITS=1 is in the prototype phase, I think LVLBITS=2 will need more analysis to prove that the spec is correct (at least for the ACPERM rules).


arichardson (approved):
Wording is less ambigious now so LGTM.

Issue #441, opened 2024-11-06
Ambiguous wording for sstatus.CRG 


tomaird:
The section in `Zcheripte` describing adding the `CRG` to `xstatus` CSRs is ambiguous on the writability of the new bits, specifically for `sstatus`:

> mstatus.CRG is made visible in sstatus.CRG.

This could imply that `sstatus.CRG` is a read-only mirror of `mstatus.CRG`, rather than being writable.

It would be better if the writability of these bits was explicit

PR #443, opened 2024-11-06
Xstatus.CRG clarification

tariqkurd-repo:
https://github.com/riscv/riscv-cheri/issues/441

When looking at the text I noticed that it had typos and xrefs missing.

This PR doesn't solve the actual problem yet.

What's the relationship between mstatus.CFG and sstatus.CFG? @jonwoodruff ? @jrtc27 ?

The text is very vague, should it be that the same bit is in mstatus and sstatus like in mip/sip.ssip? for example.....

### DIFF
--- a/src/cheri-pte-ext.adoc
+++ b/src/cheri-pte-ext.adoc
@@ -157,11 +157,11 @@ bit value of the capability read. This will introduce additional traps during re
 [#xstatus_pte]
 === Extending the Machine (mstatus), Supervisor (sstatus) and Virtual Supervisor (vsstatus) Status Registers
 
-The <<mstatusreg_pte,mstatus>>, <<sstatusreg_pte,status>> and <<vsstatusreg_pte,vsstatus>> CSRs are extended to include the new Capability Read Generation (CRG) bit as shown.
+The <<mstatusreg_pte,mstatus>>, <<sstatusreg_pte,sstatus>> and <<vsstatusreg_pte,vsstatus>> CSRs are extended to include the new Capability Read Generation (CRG) bit as shown.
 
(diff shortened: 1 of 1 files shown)


jonwoodruff:
This bit was written by @andresag01 , so I'm not confident of any considerations he might have had, but, if I understand what this is asking:
The generation bit is tied to the current page table.  If a new page table root is swapped in on any exception level change, then CRG should change as well.  If the page table root must be swapped manually, then CRG should also be swapped manually.

Sorry for the lazy response!

jonwoodruff:
This bit was written by @andresag01 , so I'm not confident of any considerations he might have had, but, if I understand what this is asking:
The generation bit is tied to the current page table. If a new page table root is swapped in on any exception level change, then CRG should change as well. If the page table root must be swapped manually, then CRG should also be swapped manually.

(Assuming that "CFG" in this issue is meant to be "CRG" in the original issue...)

tariqkurd-repo:
> This bit was written by @andresag01 , so I'm not confident of any considerations he might have had, but, if I understand what this is asking: The generation bit is tied to the current page table. If a new page table root is swapped in on any exception level change, then CRG should change as well. If the page table root must be swapped manually, then CRG should also be swapped manually.
> 
> (Assuming that "CFG" in this issue is meant to be "CRG" in the original issue...)

yes - CFG was a typo, sorry! It should have been CRG.
All the exception handling and PTE management is in S-mode.
Personally I can't see any use for mstatus.CRG, would M-mode software ever want control over this?
I don't know the answer but maybe it's safer to leave it in, following the clarification in the second commit.

tariqkurd-repo:
We've discussed this internally and can't think of any use for mstatus.CRG, so I've removed it.

PR #447, opened 2024-11-12
vm requires pte scheme

tariqkurd-repo:
there's no provision in the spec for VM in capability mode without the PTE extension, so the path of least resistance is to require it.

fixes https://github.com/riscv/riscv-cheri/issues/448

### DIFF
--- a/src/cheri-pte-ext.adoc
+++ b/src/cheri-pte-ext.adoc
@@ -10,7 +10,9 @@ capabilities in memory at the page granularity. For this reason, the
 {cheri_pte_ext_name} extension adds new bits to RISC-V's Page Table Entry (PTE)
 format.
 
-Implementing any virtual memory translation scheme (_Sv39_, _Sv48_ or _Sv57_) requires {cheri_pte_ext_name} to be implemented.
+Implementing any virtual memory translation scheme (_Sv39_, _Sv48_ or _Sv57_) and {cheri_base_ext_name} requires {cheri_pte_ext_name} to be implemented.
+
+NOTE: There is no explicit mechanism for enabling or disabling {cheri_pte_ext_name}. A VM-enabled legacy (non-CHERI) OS running in {cheri_int_mode_name} will not load or store capabilities, and so the default state of CW=0 causing loaded capabilities to have their tags cleared, and stored capabilities with their tags set to cause a page fault, won't occur. A CHERI-aware OS running a VM-enabled OS is required to support {cheri_pte_ext_name}, and the minimum level of support is to set CW to 1 in all PTEs and leave <<sstatusreg_pte,sstatus>>.CRG and CRG in all PTEs set to 0, which will allow capabilities with their tags set to be loaded and stored successfully.
 
 === Limiting Capability Propagation
 
--- a/src/cheri-pte-ext.adoc
+++ b/src/cheri-pte-ext.adoc
@@ -10,7 +10,7 @@ capabilities in memory at the page granularity. For this reason, the
 {cheri_pte_ext_name} extension adds new bits to RISC-V's Page Table Entry (PTE)
 format.
 
-{cheri_pte_ext_name} requires at least one virtual memory translation scheme (_Sv39_, _Sv48_ or _Sv57_) to be implemented.
+Implementing any virtual memory translation scheme (_Sv39_, _Sv48_ or _Sv57_) requires {cheri_pte_ext_name} to be implemented.
 
 === Limiting Capability Propagation
 


src/cheri-pte-ext.adoc
```diff
@@ -10,7 +10,9 @@ capabilities in memory at the page granularity. For this reason, the
 {cheri_pte_ext_name} extension adds new bits to RISC-V's Page Table Entry (PTE)
 format.
 
-{cheri_pte_ext_name} requires at least one virtual memory translation scheme (_Sv39_, _Sv48_ or _Sv57_) to be implemented.
+Implementing any virtual memory translation scheme (_Sv39_, _Sv48_ or _Sv57_) and {cheri_base_ext_name} requires {cheri_pte_ext_name} to be implemented.
+
+NOTE: There is no explicit mechanism for enabling or disabling {cheri_pte_ext_name}. A VM-enabled legacy (non-CHERI) OS running in {cheri_int_mode_name} will not load or store capabilities, and so the default state of CW=0 causing loaded capabilities to have their tags cleared, and stored capabilities with their tags set to cause a page fault, won't occur. A CHERI-aware OS running a VM-enabled OS is required to support {cheri_pte_ext_name}, and the minimum level of support is to set CW to 1 in all PTEs and leave <<sstatusreg_pte,sstatus>>.CRG and CRG in all PTEs set to 0, which will allow capabilities with their tags set to be loaded and stored successfully.
```
jrtc27, line 15: This is dangerous advice, any OS that followed it would let you store capabilities to things like shared memory (allowing capabilities to pass between address spaces) or file-backed pages (same problem as shared memory, but also it just doesn't make sense to do, and if the page is swapped out and back the tags will disappear).

tariqkurd-repo:
Ok, please suggest an improvement to the text!

src/cheri-pte-ext.adoc
```diff
@@ -10,7 +10,9 @@ capabilities in memory at the page granularity. For this reason, the
 {cheri_pte_ext_name} extension adds new bits to RISC-V's Page Table Entry (PTE)
 format.
 
-{cheri_pte_ext_name} requires at least one virtual memory translation scheme (_Sv39_, _Sv48_ or _Sv57_) to be implemented.
+Implementing any virtual memory translation scheme (_Sv39_, _Sv48_ or _Sv57_) and {cheri_base_ext_name} requires {cheri_pte_ext_name} to be implemented.
```
arichardson, line 13: For sv32 we can't support this extension, so I don't think VM implies PTE extension. Not supporting the extension should be fine since it just allows capability loads/stores anywhere. I suggest reverting this line.

src/cheri-pte-ext.adoc
arichardson: How about just adding "pages intended to store capabilities (i.e. anonymous mappings)" :
```suggestion
NOTE: There is no explicit mechanism for enabling or disabling {cheri_pte_ext_name}. A VM-enabled legacy (non-CHERI) OS running in {cheri_int_mode_name} will not load or store capabilities, and so the default state of CW=0 causing loaded capabilities to have their tags cleared, and stored capabilities with their tags set to cause a page fault, won't occur.
A CHERI-aware OS running a VM-enabled OS is required to support {cheri_pte_ext_name}, and the minimum level of support is to set CW to 1 in all PTEs intended for storing capabilities (i.e. anonymous mappings) and leave <<sstatusreg_pte,sstatus>>.CRG and CRG in all PTEs set to 0, which will allow capabilities with their tags set to be loaded and stored successfully.
```

src/cheri-pte-ext.adoc
tariqkurd-repo: Sorry - I clearly merged this far too soon - the problem is what the software sees if ZcheriPTE isn't implemented, because PTE.CW will be 0 and sstatus.CRG will be 0.
if PTE.CW=0 then cap load/stores either tag clear or except, so they aren't allowed anywhere.
For RV32 you're right - the PTE bits dont exist - so this is all a non-argument.

But for RV64 - how can the software know whether to look at the CW and CRG bits? If a CHERI core supports Zcheripte and the software doesn't know about it - isn't it a problem? I think that PTE.CW will always be zero and so cap load/store won't work.

So it seemed cleaner to require it, then there's no ambiguity - but certainly - only for RV64.

src/cheri-pte-ext.adoc
tariqkurd-repo: ok - I'll make a new PR - let's discuss there

PR #462, opened 2024-12-05
Deleted incorrect statement about ASR needed for all M and S mode CSRs

francislaus:
The `mtidc` and `stidc` registers do not require `ASR` permission when read. This makes the deleted statements wrong.

### DIFF
--- a/src/riscv-integration.adoc
+++ b/src/riscv-integration.adoc
@@ -831,7 +831,7 @@ CHERI violations have the following order in priority:
 === Supervisor-Level CSRs
 
 {cheri_base_ext_name} extends some of the existing RISC-V CSRs to be able to
-hold capabilities or with other new functions.
+hold capabilities or with other new functions. <<asr-perm>> in the <<pcc>> is typically required for access.
 
 [#stvec,reftext="stvec"]
 ==== Supervisor Trap Vector Base Address Register (stvec)
--- a/src/riscv-integration.adoc
+++ b/src/riscv-integration.adoc
@@ -425,7 +425,7 @@ include::generated/csr_renamed_purecap_mode_u_table_body.adoc[]
 === Machine-Level CSRs
 
 {cheri_base_ext_name} extends some M-mode CSRs to hold capabilities or
-otherwise add new functions.
+otherwise add new functions. <<asr-perm>> in the <<pcc>> is typically required for access.
 
 [#mstatus,reftext="mstatus"]
 ==== Machine Status Registers (mstatus and mstatush)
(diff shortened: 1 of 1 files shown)


francislaus:
Thinking about it, it might be better to amend the statements as follows:

 `<<pcc>> must grant <<asr_perm>> to access M-mode CSRs regardless of the RISC-V privilege mode with exception of the <<mtidc>> not requiring <<pcc>> to grant <<asr_perm>> for read accesses.`

francislaus:
Also, this statement is not true any longer for unprivileged CSRs anymore: https://github.com/riscv/riscv-cheri/blob/12c5565defc6a3f363cf9578504e7377d463563e/src/riscv-integration.adoc?plain=1#L1054

I propose: `Only writing the <<utidc>> register requires <<pcc> to grant <<asr_perm>>, all other unprivileged CSRs do not require <<pcc>> to grant <<asr_perm>> to access unprivileged CSRs.`
`

tariqkurd-repo (approved):

tariqkurd-repo:
I think the definiton of ASR-perm is correct - so maybe the PR is ok as it is?

src/riscv-integration.adoc
```diff
@@ -425,8 +425,7 @@ include::generated/csr_renamed_purecap_mode_u_table_body.adoc[]
 === Machine-Level CSRs
 
 {cheri_base_ext_name} extends some M-mode CSRs to hold capabilities or
-otherwise add new functions. <<pcc>> must grant <<asr_perm>> to access M-mode
-CSRs regardless of the RISC-V privilege mode.
+otherwise add new functions.
```
tariqkurd-repo, line 428:
```suggestion
otherwise add new functions. <<asr-perm>> in the <<pcc>> is typically required for access.
```

src/riscv-integration.adoc
```diff
@@ -1051,7 +1049,8 @@ include::img/stval2reg.edn[]
 
 === Unprivileged CSRs
 
-Unlike machine and supervisor level CSRs, {cheri_base_ext_name} does not require
+In {cheri_base_ext_name} only writing the <<utidc>> register requires
+<<pcc>> to grant <<asr_perm>>, all other unprivileged CSRs do not require
```
arichardson, line 1053:
```suggestion
```suggestion
In {cheri_base_ext_name} the only register that requires <<asr_perm>> is <<utidc>> (for updates but not for reads), and all other unprivileged CSRs do not require
```
How about this wording?

arichardson (approved):

src/riscv-integration.adoc
tariqkurd-repo: seems to have been updated accordingly

PR #464, opened 2024-12-07
Extended Zcherihybrid to reflect that access is not allowed to CHERI instructions

francislaus:
The text in most cases only considered CHERI registers, but `CRE` also prevents executing CHERI instructions.

### DIFF
--- a/src/riscv-hybrid-integration.adoc
+++ b/src/riscv-hybrid-integration.adoc
@@ -324,7 +324,7 @@ it is not possible to disable CHERI checks completely.
 endif::[]
 
 {cheri_default_ext_name} includes functions to disable explicit access to CHERI
-registers.  The following occurs when executing code in a privilege mode that
+registers and instructions.  The following occurs when executing code in a privilege mode that
 has CHERI register access disabled:
 
 * The CHERI instructions in xref:section_cap_instructions[xrefstyle=short] and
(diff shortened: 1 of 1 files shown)


src/riscv-hybrid-integration.adoc
```diff
@@ -324,7 +324,7 @@ it is not possible to disable CHERI checks completely.
 endif::[]
 
 {cheri_default_ext_name} includes functions to disable explicit access to CHERI
-registers.  The following occurs when executing code in a privilege mode that
+registers and instructions.  The following occurs when executing code in a privilege mode that
```
arichardson, line 327: I thought this was implicit because any CHERI instruction that uses a CHERI register is not possible, but happy to add this clarification.

francislaus: @arichardson I agree, it is implicit. For some CSRs, the text explicitly says "registers and instructions" (https://github.com/riscv/riscv-cheri/blob/main/src/riscv-hybrid-integration.adoc?plain=1#L446-L448) whereas for others the text only refers to registers. We might want to have that consistent. I am fine with either way.

If we go with only referring to registers, we might want to make it clear somewhere that this implies that code cannot execute CHERI instructions.

tariqkurd-repo (approved):

src/riscv-hybrid-integration.adoc
tariqkurd-repo: pushing to this branch resulted in a new PR (no idea why) but here it is: https://github.com/riscv/riscv-cheri/pull/474

PR #469, opened 2024-12-10
Clarifications to PTE specification

Timmmm:
Some minor clarification & formatting improvements in the PTE spec. I can break these into separate PRs if you like.

### DIFF
--- a/src/cheri-pte-ext.adoc
+++ b/src/cheri-pte-ext.adoc
@@ -129,9 +129,9 @@ regards to the ongoing capability revocation cycle. Two schemes are permitted:
 
 * A load page fault exception is raised when a capability load or AMO instruction is executed
 and the virtual page's CRG bit does not equal <<sstatusreg_pte,sstatus>>.CRG.
-* A load page fault exception is raised when a capability load or AMO instruction is executed,
-the capability read from memory has tag set
-and the virtual page's CRG bit does not equal <<sstatusreg_pte,sstatus>>.CRG.
+* A load page fault exception is raised when a capability load or AMO instruction is executed
+and the virtual page's CRG bit does not equal <<sstatusreg_pte,sstatus>>.CRG
+and the capability read from memory has its tag set.
 
 NOTE: An implementation may avoid a dependency on the tag
 bit value of the capability read. This will introduce additional traps during revocation sweeps.
--- a/src/cheri-pte-ext.adoc
+++ b/src/cheri-pte-ext.adoc
@@ -134,10 +134,10 @@ When CW is set, the CRG bit indicates the current generation of the virtual memo
 regards to the ongoing capability revocation cycle. Two schemes are permitted:
 
 * A load page fault exception is raised when a capability load or AMO instruction is executed
-and the virtual page's CRG bit does not equal <<sstatusreg_pte,sstatus>>.CRG.
+with <<c_perm>> granted and the virtual page's CRG bit does not equal <<sstatusreg_pte,sstatus>>.CRG.
 * A load page fault exception is raised when a capability load or AMO instruction is executed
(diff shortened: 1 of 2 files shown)


tariqkurd-repo (approved):
We haven't previously specified when happens on non-leaf PTEs, and there's no reason to allocate these bits there so it's better if we define them to be reserved, so I think this is sensible.

The other edits are all no-change, just improving the formatting and explicitly pointing at that the tag clear on SC is before the PTE check.

tariqkurd-repo:
> We haven't previously specified when happens on non-leaf PTEs, and there's no reason to allocate these bits there so it's better if we define them to be reserved, so I think this is sensible.
> 
> The other edits are all no-change, just improving the formatting and explicitly pointing at that the tag clear on SC is before the PTE check.

@Timmmm has pointed out that it is just a clarification as the text only talks about leaf PTEs

src/cheri-pte-ext.adoc
```diff
@@ -129,9 +135,9 @@ regards to the ongoing capability revocation cycle. Two schemes are permitted:
 
 * A load page fault exception is raised when a capability load or AMO instruction is executed
 and the virtual page's CRG bit does not equal <<sstatusreg_pte,sstatus>>.CRG.
-* A load page fault exception is raised when a capability load or AMO instruction is executed,
-the capability read from memory has tag set
-and the virtual page's CRG bit does not equal <<sstatusreg_pte,sstatus>>.CRG.
+* A load page fault exception is raised when a capability load or AMO instruction is executed
+and the virtual page's CRG bit does not equal <<sstatusreg_pte,sstatus>>.CRG
+and the capability read from memory has its tag set.
```
arichardson, line 140: Do we need to define here what happens if for loads via capabilities without <<c_perm>>? I imagine there is no need to trap if we aren't actually reading caps, but I do wonder if it could be a problem pulling tagged data into the cache. I don't think so since the next load with c_perm would perform this check again and then trap?

arichardson (approved):
Clarifications on store looks good to me.

The non-leaf note was previously implicit, spelling it out makes sense.

src/cheri-pte-ext.adoc
tariqkurd-repo: I think we don't need to trap if there's no C-permission, and have clarified that in my latest commit.
@Timmmm thinks we do. @jonwoodruff / @jrtc27 can you add your opinion?

Timmmm: Sorry I didn't mean that we do need to trap, I just meant your commit changed the behaviour whereas the rest of this PR is just clarifications, so I wasn't sure if it was intentional. Maybe take it out of this PR and make a new one?

tariqkurd-repo: this isn't a change - it's the intention - I guess it's not what the SAIL does then?

jamie-melling: The spec already has this note although it doesn't mention Zcherilevels can also strip tags, so maybe this note should be adjusted
image

Timmmm: Ah good spot. I think this change makes sense then. I am currently implementing Zcheripte in Sail.

jonwoodruff: I'm trying to catch up here, and also found the note @jamie-melling mentioned.

As for adding further clarifications, if necessary, c_perm would be required to get a trap in both implementation options for CRG load checking (not just the second one).  That is, if a capability load or AMO instruction is executed, the authorizing capability has c_perm, and pte.CRG != csr.CRG.

On balance, I prefer the simple existing statement to adding more complexity to the conditions, but maybe people will find it confusing to have incomplete text!  Maybe pseudocode is the answer?

tariqkurd-repo: Note updated

arichardson (approved):

PR #475, opened 2024-12-11
Capitalised MODE everywhere in the csv

francislaus:


arichardson:
This is still pulled from a Google sheet (which we should probably change and use the repo as the source of truth instead), so that needs updating. @tariqkurd-repo .

arichardson (approved):

tariqkurd-repo:
the CSV technically is the master, it's just more convenient to update it by updating the spreadsheet and downloading. A point for discussion.

Issue #470, opened 2024-12-10
CHERI PTE exceptions might need their own cause code


PeterRugg:
The Zcheripte spec implies that faults caused by CHERI PTE exceptions use the existing RISC-V load and store/AMO page fault cause codes. The ISAv9 spec has separate cause codes for these: 0x1a and 0x1b.

Are we sure that we can do without separate cause codes?

chrehrhardt:
I think it would be useful to know if a fault is capability related (either through an additional exception code or a flag in say in xtval2). If you get e.g. a write fault you always have to make the page writable. However, a OS may not want to deal with capability faults if it does not have to.

arichardson:
The normal capability faults triggered when checking bounds and permissions of a register already use a separate code to ensure the kernel can handle them correctly, and I believe we also need to so the same for PTE based faults since otherwise the revocation logic ends up being mixed in with the normal page fault handling. While I think it should be possible to make it work with extra software complexity I'd prefer having separate codes. Does anyone recall how Morello handled this?

@nwf, @jrtc27 and @jonwoodruff: could you comment on this?

nwf:
I think that agrees with my preferences: we probably could come to the conclusion in software that the problem _might_ be, say, capability load generation related by looking at the PTE, but I'd rather the hardware just told us.

PR #480, opened 2024-12-12
add more exception types to distinguish CHERI page fault types

tariqkurd-repo:
Fixes https://github.com/riscv/riscv-cheri/issues/470

### DIFF
--- a/src/riscv-integration.adoc
+++ b/src/riscv-integration.adoc
@@ -800,14 +800,14 @@ is the cause of the fault. The possible CHERI types and causes are encoded as
 shown in xref:mtval2-cheri-type[xrefstyle=short] and
 xref:mtval2-cheri-causes[xrefstyle=short] respectively.
 
-.Encoding of TYPE field for CHERI exceptions
+.Encoding of TYPE field for CHERI Faults
 [#mtval2-cheri-type,width=65%,float="center",align="center",options=header,cols="30%,70%"]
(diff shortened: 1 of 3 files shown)


src/riscv-integration.adoc
```diff
@@ -801,18 +801,23 @@ xref:mtval2-cheri-causes[xrefstyle=short] respectively.
...
 |==============================================================================
 
+^1^ if {cheri_pte_ext_name} is implemented
+
 .Encoding of CAUSE field
 [#mtval2-cheri-causes,width=55%,float="center",align="center",options=header]
 |==============================================================================
 | CHERI Cause Code              | Description
-| {cheri_excep_cause_tag}       | Tag violation
+| {cheri_excep_cause_tag}       | Tag violation, or TYPE is a <<cheri_pte_ext,PTE>> fault
```
jrtc27, line 820: Oh no, you don't want to lump this under CHERI faults. It's still a page fault, just with extra information.

tariqkurd-repo: this just means that the field isn't populated for the PTE faults

tariqkurd-repo: the TYPE field is all you need, and CAUSE=0 if it's a PTE fault, so it's redundant

tariqkurd-repo: reworked

PR #483, opened 2024-12-12
Use stval/stval2 for HS-mode exceptions, not htval/htval2

### LINKED ISSUE #482: Use of htval(2) for CHERI faults is odd
HS-mode should still use stval and stval2 for CHERI exceptions that come from HS-mode. The point of htval is for communicating the guest physical address when VS-mode takes a G-stage page fault to HS-mode, but when you're taking exceptions from HS-mode itself it's still S-mode and you use stval(2). It's not HS-mode's stval equivalent.

jrtc27:
htval exists solely to communiate extra information to HS-mode about VS-mode faults to HS-mode, currently just the guest physical address for G-stage page faults. CHERI exceptions can and must (to align with other kinds of faults, including HS-mode's own faults) use stval/stval2 when trapping from HS-mode.

Fixes #482


chrehrhardt:
The last sentence of the commit message should probably read "...when trapping _to_ HS-mode". Otherwise LGTM.

jrtc27:
> The last sentence of the commit message should probably read "...when trapping _to_ HS-mode". Otherwise LGTM.

The text I deleted talked of exceptions "taken in HS-mode", which to me means "from HS-mode". But I agree that, on rereading, it probably _meant_ to say "to HS-mode", because otherwise M-mode could be handling them and then using htval(2) makes even less sense.

Issue #484, opened 2024-12-13
Priority of CHERI page faults


Timmmm:
It didn't matter before #480 because CHERI page faults were indistinguishable from standard page faults, but now we need to specify their priorities. There are a couple of sources of page faults that you *could* do before or after the CHERI checks:

* Standard WRXU bit checks
* Page fault if A/D bits need updating but hardware updates aren't supported

From a Sail point of view I think the easiest thing is if CHERI checks are the lowest priority; i.e. the above checks are done first.

Also there's a question of what happens if hardware A/D bit updates are supported but hardware CRG/CW bit updates are not, or vice versa. It would simplify the Sail implementation (and spec!) if they always had to be the same - either all hardware updates or none at all. Would that be acceptable?

tariqkurd-repo:
>we need to specify their priorities

The global rule is that CHERI checks come first - and anything existing in RISC-V comes after. 
Is there any reason why this wouldn't work here?

(the only counter-example is a PC breakpoint - which is higher than CHERI checks on the PC - otherwise it's impossible to debug CHERI code)

>Also there's a question of what happens if hardware A/D bit updates are supported but hardware CRG/CW bit updates are not, or vice versa. It would simplify the Sail implementation (and spec!) if they always had to be the same - either all hardware updates or none at all. Would that be acceptable?

I've already tied them together in section 5.4 (@buxtonpaul asked about this before)

tariqkurd-repo:
> Ah ok, no I think that should be fine. Is that general rule written down anywhere?

I think the correct solution here is to add the PTE page faults into the exception check information on the instruction pages, and add the statement that the PTE page faults come first.

PR to follow

PR #485, opened 2024-12-13
clarify PTE fault exception priority, and add into instruction pages

tariqkurd-repo:
Fixes https://github.com/riscv/riscv-cheri/issues/484

### DIFF
--- a/src/cheri-pte-ext.adoc
+++ b/src/cheri-pte-ext.adoc
@@ -100,7 +100,7 @@ If the CW bit is clear then:
 
 * When a capability load or AMO instruction is executed, the implementation
  clears the tag bit of the capability read from the virtual page.
-* When CRG is clear, the "no capability state", a store page fault exception
+* When CRG is clear, the "no capability state", a store/AMO page fault exception
  is raised when a capability store or AMO instruction is executed and the tag bit
  of the capability being written is set.
 * When CRG is set, the "pre-CW state", two schemes are permitted (also see <<section_hardware_pte_updates>>):
@@ -166,6 +166,13 @@ and the capability read from memory optionally has its tag set^1^.
  Taking a trap when the tag is not set will introduce additional traps during revocation sweeps.
 
 ^2^ See <<section_hardware_pte_updates>>.
+
+The exceptions added by {cheri_pte_ext_name}:
+
+* reuse the load page fault and store/AMO page fault exception cause values,
+and so the cause of the exception can be determined by software by checking bit zero of <<mtval2>>,
+<<stval2>> etc.
+* are all prioritized _above_ standard page fault exceptions.
 
 [#section_hardware_pte_updates]
 === Enabling Software or Hardware PTE updates
--- a/src/hypervisor-integration.adoc
+++ b/src/hypervisor-integration.adoc
@@ -75,7 +75,7 @@ part of {cheri_base_ext_name} when the hypervisor extension is supported. Its
 CSR address is 0x64b.
 
 <<htval2>> is updated following the same rules as <<mtval2>> for CHERI exceptions,
-load page fault and store page fault exceptions which are taken in HS-mode.
+load page fault and store/AMO page fault exceptions which are taken in HS-mode.
 
 The fields are identical to <<mtval2>> for CHERI exceptions.
 
@@ -196,7 +196,7 @@ include::img/vstdcreg.edn[]
 The <<vstval>> register is a VSXLEN-bit read-write register.
 
 <<vstval>> is updated following the same rules as <<mtval>> for CHERI exceptions,
-load page fault and store page faults which are taken in VS-mode.
+load page fault and store/AMO page faults which are taken in VS-mode.
 
 .Virtual supervisor trap value register
 [#vstval-format]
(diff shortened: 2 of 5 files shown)


Timmmm (approved):
Thanks!

src/cheri-pte-ext.adoc
```diff
@@ -167,6 +167,13 @@ and the capability read from memory optionally has its tag set^1^.
 
 ^2^ See <<section_hardware_pte_updates>>.
 
+The exceptions added by {cheri_pte_ext_name}:
+
+* reuse the load page fault and store/AMO page fault exception cause values,
+and so the cause of the exception can be determined by software by checking bit zero of <<mtval2>>,
+<<stval2>> etc.
+* are all prioritized _above_ standard page fault exceptions.
```
Timmmm, line 175: Wait actually does this include the page faults that check the tag on the loaded capability?

> A load page fault exception is raised when a capability load or AMO instruction is executed with [C-permission](https://riscv.github.io/riscv-cheri/#c_perm) granted and the virtual page’s CRG bit does not equal [sstatus](https://riscv.github.io/riscv-cheri/#sstatusreg_pte).CRG **and the capability read from memory optionally has its tag set**.

That must happen after the RWXU checks I think? Should it happen before the A/D page fault too?

In fact is it possible for the A/D bits to be updated and *then* you get a page fault from loading a tagged capability?

tariqkurd-repo: ok - fixed - please have another look

Timmmm: Yeah I agree with your comment. I'm not sure the spec update makes that clear though because it still says

> * are all prioritized _above_ standard page fault exceptions.

Maybe something like:

> * are all prioritized _above_ standard page fault exceptions (except possibly <<for loads,zcheripte_load_page_priority>>)

And

> A load page fault exception is raised when a capability load or AMO instruction is executed with C-permission granted and the virtual page’s CRG bit does not equal sstatus.CRG and the capability read from memory optionally has its tag set. **[#zcheripte_load_page_priority]#If this scheme is used the page fault is lower priority than standard page faults.#**

tariqkurd-repo: I've just used xref's to the table to avoid repeating the text, have another look....

src/cheri-pte-ext.adoc
```diff
@@ -164,9 +164,16 @@ and the capability read from memory optionally has its tag set^1^.
...
  Taking a trap when the tag is not set will introduce additional traps during revocation sweeps.
+ Checking the loaded tag affects the exception priority, see <<exception-priority>>.
 
 ^2^ See <<section_hardware_pte_updates>>.
 
+The exceptions added by {cheri_pte_ext_name} reuse the load page fault and store/AMO page fault exception cause values,
+and so the cause of the exception can be determined by software by checking the value in <<mtval2>>,
+<<stval2>> etc.
+
+TThe behavior when multiple page fault types are detected at once is shown in <<mtval2-page-fault>>.
```
Timmmm, line 175: TThe

Timmmm (commented):
Yeah I think this is unambiguous, and can be implemented easily in Sail, and I can't see any reason why it would be hard to implement in designs. Just one minor typo.

src/cheri-pte-ext.adoc
tariqkurd-repo, line 175:
```suggestion
The behavior when multiple page fault types are detected at once is shown in <<mtval2-page-fault>>.
```

Timmmm (approved):

src/hypervisor-integration.adoc
```diff
@@ -196,7 +196,7 @@ include::img/vstdcreg.edn[]
 The <<vstval>> register is a VSXLEN-bit read-write register.
 
 <<vstval>> is updated following the same rules as <<mtval>> for CHERI exceptions,
-load page fault and store page faults which are taken in VS-mode.
+load page fault and store/AMO page faults which are taken in VS-mode.
```
tariqkurd-repo, line 199:
```suggestion
load page fault and store page faults which are taken in VS-mode.
```

src/riscv-integration.adoc
```diff
@@ -830,13 +837,24 @@ CHERI violations have the following order in priority:
...
+[#mtval2-page-fault,width=70%,float="center",align="center",cols="2,1",options=header]
+|==============================================================================
+| Fault                                                   | Value
+| RISC-V page fault                                       | 0
+| CHERI <<cheri_pte_ext,PTE>> fault                       | 1
+| RISC-V page fault and CHERI <<cheri_pte_ext,PTE>> fault | 2
+|==============================================================================
 
-If a store page fault is caused by a CHERI <<cheri_pte_ext,PTE>> fault, then set bit zero of <<mtval2>> to 1 and all other bits to 0, otherwise set all bits to zero.
+NOTE: Reporting both allows the software the choice about which action to take first, for example a write to a
```
jrtc27, line 854: This confuses me. Load-side revocation will never do any sweeping on stores, so value 2 isn't useful, it will be treated the same as 1. As for loads, yes, sweeping occurs there, but there's nothing else really to do if both faults occur (for non-tag-dependent implementations) over and above just a CHERI PTE fault, and if your implementation is tag-dependent then the RISC-V page fault and CHERI PTE fault are strictly serial so it's impossible to have both at once.

PR #487, opened 2024-12-17
clarify element zero exceptions for FoF loads

tariqkurd-repo:


arichardson (approved):
This makes more sense to me

Issue #494, opened 2024-12-19
PCC bounds checking depends on fetched instruction


tomaird:
One of the checks for PCC is that:

> All bytes of the instruction must be in bounds

This check requires to know the size of the instruction (16 bit or 32 bit), which means the instruction must have been successfully fetched.

However the exception priority list has PCC checks _above_ instruction access/page faults. So if the PCC address is at the upper end of its bounds such that a 16 bit instruction would be in bounds but a 32 bit instruction wouldn't, and during instruction address translation there is an access fault so no instruction will be fetched, it will be impossible to know if all bytes of the instruction will be in bounds. 

The spec isn't clear on how to handle this case. Should it assume a minimum sized instruction?

jrtc27:
Our spec was that you fetch in 16-bit units, doing bounds then normal translation + access fault checks for each in turn. If you had a non-C core then you could instead just fetch the whole 32-bit instruction in one go.

tariqkurd-repo:
So the PCC check here needs qualifying to say that if no valid instruction is availabe, then a minimum instruction is checked for PCC bounds.

![image](https://github.com/user-attachments/assets/65b3f444-d22d-4df5-a8f8-2645839f148e)

PR #495, opened 2024-12-19
clarify PCC fault vs instruction access fault

tariqkurd-repo:
fixes https://github.com/riscv/riscv-cheri/issues/494
@tomaird 

the diff is a bit confused as a renumbered the notes below the table
the only real addition is the new note number 1

### DIFF
--- a/src/riscv-integration.adoc
+++ b/src/riscv-integration.adoc
@@ -710,7 +710,7 @@ _Reserved_
 |Priority |Exc.Code |Description
 |_Highest_ |3 |Instruction address breakpoint
 | .>|*{cheri_excep_mcause}* .<|*Prior to instruction address translation:* +
-*CHERI fault due to PCC checks (tag, execute permission, invalid address and bounds)*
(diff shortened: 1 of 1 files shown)


src/riscv-integration.adoc
```diff
@@ -736,19 +736,22 @@ Load/store/AMO address breakpoint
...
 Load/store/AMO access fault
 | .>|4,6 .<|If not higher priority: +
 .>|_Lowest_ .>|13 .<|If not higher priority: +
-CHERI load PTE fault^3^
+CHERI load PTE fault^4^
 |===
 
-^1^ The higher priority CHERI <<cheri_pte_ext,PTE>> page fault covers capability loads or atomics where the loaded tag _is not_ checked, and all capability stores and atomics where the stored tag is set.
+^1^ PCC bounds are intended to be checked against all the bytes of fetched instructions.
+ In the case of variable length instruction encoding, and that the fetch has failed to return any data, then only a minimum length instruction is checked against the PCC bounds.
```
jrtc27, line 748: This sounds a lot like “fetch the whole instruction and only then check bounds” which can be dangerous wrt side channels

PR #497, opened 2024-12-20
Remove support for downloading CSVs from Google Docs

### LINKED ISSUE #496: Move CSVs from Google Docs into repo
We probably shouldn't be downloading unversioned CSVs from Google Docs at this point...

I would suggest the simplest option for now is to vendor them in the repo and manually update them when necessary. Or just edit them in place with OpenOffice or whatever. You lose the colouring but everything else should work.

Timmmm:
The Google version hasn't been kept up to date with the version in the repo, and it's a bit weird to still have this anyway.

Fixes #496

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,11 +21,6 @@ DATE    ?= $(shell date +%Y-%m-%d)
 VERSION ?= v0.9.2
 REVMARK ?= Draft
 
-# URLs for downloaded CSV files
-URL_BASE = https://docs.google.com/spreadsheets/d/1nyxKamsYZaeTyG8qP-JX4_oQwcuQ_4nZ_Ihm3RK0NEY/gviz/tq?tqx=out:csv
-URL_ISA  = $(URL_BASE)&gid=0
-URL_CSR  = $(URL_BASE)&gid=1927549494
-
 # Directories and files
 BUILD_DIR   = build
 SRC_DIR     = src
@@ -130,7 +125,6 @@ pdf: $(PDF_RESULT)
 html: $(HTML_RESULT)
 all: pdf html
 generate: $(GEN_SRC)
-download: $(CSVS)
 
 $(BUILD_DIR):
 	@echo "  DIR $@"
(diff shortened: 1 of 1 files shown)


tariqkurd-repo:
There is one bug fix in the google sheet which hasn't made it to an MR yet - @francislaus are you planning to do that?
I'm not planning to remove the google sheet as maintaining the CSV without it will be a real pain, we'll need a flow for doing that (possibly to/fromyaml) first.

tariqkurd-repo (changes requested):
Not until a replacement flow is in place to allow easy maintainance.

Timmmm:
You can edit the CSV directly using a CSV editor no? I don't see any features except for colours that are used in the Google Doc that aren't preserved in the CSV.

Here's [one I like in VSCode](https://marketplace.visualstudio.com/items?itemName=janisdd.vscode-edit-csv):

image

It's write format doesn't *quite* match what Google Docs outputs (Google outputs `,"",` for empty fields), but they seem to be on their own there. LibreOffice and Gnumeric and that extension just do `,,` so I'd suggest just switching to that.

arichardson:
I also think it's time to change the source of truth to be the repository. I can also edit the CSV as a table in Intellij:

image

We could convert it to YAML to make it easier to edit, but I'm not sure that's really necessary?

tariqkurd-repo:
> We could convert it to YAML to make it easier to edit, but I'm not sure that's really necessary?

the point of yaml is to make the differences easier to spot, but as the spreadsheet is fairly small it's probably not so important

also the spreadsheet as a lot of repeated entries, which are lost when converting to CSV, so it's possible that text will diverge (e.g. a lot of CSRs have the same `Update the CSR using <<SCADDR>>.` text

Additionally. which is also important, a huge advantage of the spreadsheet is filtering and sorting columns which I can't obviously see how to do in EditCSV. The information is quite stable so maybe it's not so important not but these spreadsheet features have made a lot of difference when developing the spec.

tariqkurd-repo (approved):

PR #501, opened 2025-01-07
Ensure GCPERM result is forwards-compatible with software

arichardson:
When new extensions add permissions that are subsets of current permissions
(e.g. Zcherilevels refines loads/stores), then a system that does not
support these permissions should still report ones in GCPERM to ensure that
code that inspects permissions knows that the load/store behaves as if all
of these permission bits were set.
We reserve the low 24 bits as ones (to be used as refinements of the current
permissions) and the upper 8 (40 for RV64) bits as zeros (to be used for
new permissions that aren't subsets (e.g. seal/unseal/compartment ID).

Partially addresses https://github.com/riscv/riscv-cheri/issues/502


arichardson:
We could also report a hardcoded 1 here since not implementing the levels extension ensures that all loads/stores behave as if the permissions were always granted. But I think being able to easily detect whether the levels ext is implemented probably makes more sense?

src/insns/gcperm_32bit.adoc
```diff
@@ -33,6 +33,9 @@ permission bits in `rd` are set to 0.
 [#gcperm_bit_field]
 include::../img/acperm_bit_field.edn[]
 
+NOTE: When {cheri_levels_ext_name} is not implemented, the `CL`, `SL`, `EL` and EL fields always report 0.
```
jrtc27, line 36: EL and EL (with the latter copy also not in backticks)?

src/insns/gcperm_32bit.adoc
```diff
@@ -33,6 +33,9 @@ permission bits in `rd` are set to 0.
...
+Therefore, the presence of the {cheri_levels_ext_name} can be detected by checking the <<GCPERM>> result on the <<infinite-cap>>.
```
jrtc27, line 37: I mean, the RISC-V way is to get that information externally, for better or (generally) worse.

src/insns/gcperm_32bit.adoc
arichardson: thanks, fixed.

src/insns/gcperm_32bit.adoc
jrtc27: Reserved bits should be 0, and without levels the bits are reserved. When an extension adds meaning to those bits, the existing bit pattern should be unchanged in its meaning. I'm concerned the latter isn't true if the natural default is 1.

arichardson: I agree that reserved bits should be zero. The problem is that if we add new permissions, those will generally be '1' for permissive behaviour and '0' for restricted, so the natural default is 1 even though cores that don't implement this extension will be reporting zero and doing the permissive behaviour. I am not sure if this can really be fixed.

jrtc27: Either you invert them or you have RES1 bits. https://github.com/riscv/riscv-cheri/issues/502 is the short dump of thoughts I have on the matter, and that I've brought up many times over the years wanting people to think about this carefully with worked examples, to no avail.

tariqkurd-repo: ok - so RES1 seems like the way to go

arichardson: For the in-memory format it seems like we would have to use inverted bit meaning? Otherwise if we use RES1 we end up having those permissions set to 0 for the null cap which sounds like it would be a problem?

arichardson:
If we go for RES1 on GCPERM/ACPERM, does that mean we need to have all unused bits hardwired to one?

That would break patterns like `gcperm(acperm(cap, REQUESTED_PERMS)) == REQUESTED_PERMS` to check if you got exactly what you asked for.

arichardson:
Updated to reserve bits 0-23 as 1 and 24-xlen-1 as zero in the GCPERM mask. Also added a note on how new permissions need to be allocated.

arichardson:
Added note on ACPERM behaviour based on offline feedback from @LawrenceEsswood. This matches what was previously discussed in the CHERI meetings, so will merge this now. We can always adjust the number of reserved one bits during review.

PR #506, opened 2025-01-13
Fix observing CHERI execution mode code

James-Williams:
The example for observing the CHERI execution mode for privilege modes other than debug incorrectly uses `gctag` instead of `gcmode`.


jrtc27 (changes requested):
Existing code is correct, if confusing. New code is not.

PR #507, opened 2025-01-15
specify M-bit behaviour when there's no integer mode

tariqkurd-repo:
The spec doesn't currently specify what to do about encodings when Zcherihybrid isn't implemented.
The proposal is:

* for RV32 - setting M in quadrant 1 gives invalid permissions
* for RV64 - M is simply a reserved bit

which seems a bit inconsistent, but it's because in RV32 it's part of the permissions encodings, whereas in RV64 it's just a separate bit, and so reserved.

### DIFF
--- a/src/insns/acperm_32bit.adoc
+++ b/src/insns/acperm_32bit.adoc
@@ -61,12 +61,14 @@ The rules from <<acperm_rules>> must be followed when removing permissions.
 | 13  (RV32 only) | <<x_perm>>   | (<<c_perm>> and <<lm_perm>> and <<el_perm>> and (<<sl_perm>> == ∞)) or +
                                    (not(<<c_perm>> and not(<<lm_perm>>) and not(<<el_perm>>) and (<<sl_perm>>==0)))^1^
 | 14              | <<asr_perm>> | <<x_perm>>
-| 15^2^           | <<m_bit>>    | <<x_perm>>
+| 15^2^ (RV32 only) | <<m_bit>>  | <<x_perm>> and {cheri_default_ext_name} is implemented
 |===
 
 ^1^ All the listed permissions in the set are either minimum or maximum. +
-^2^ The <<m_bit>> only exists if {cheri_default_ext_name} is implemented.
- Otherwise it is reserved and this rule is not relevant.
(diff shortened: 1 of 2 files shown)


jrtc27:
It's not inconsistent, for RV32 it's a reserved encoding?

tariqkurd-repo:
The question is whether it's `reserved` causing a tag violation on
load/store dereference, or `invalid permissions` causing a permission
violation.

That's the only distinction I'm aware of between the two cases.

We just need to choose one.

jrtc27:
> The question is whether it's `reserved` causing a tag violation on load/store dereference, or `invalid permissions` causing a permission violation. That's the only distinction I'm aware of between the two cases. We just need to choose one.

How would it ever have a tag?

tariqkurd-repo:
Only if it's from memory corruption, the same as the other reserved bit +
tag set cases, and invalid permission cases

arichardson:
If it's easier for consistency we could move the M bit into AP for RV64 as well? And then just treat it as a reserved permission?

jrtc27:
I don't like thinking of it as a permission any more than necessary, it'll confuse people...

arichardson:
> I don't like thinking of it as a permission any more than necessary, it'll confuse people...

That's a good point since it's not restricted to monotonic behaviour. Probably better to keep it conceptually separate for RV64.

jrtc27:
I'm still concerned that for RV32 this is allocating the bit / AP encodings for M even when not implemented, when instead they should all be _reserved_ encodings that can be allocated for a different purpose by another extension. This is particularly important given the limited encoding space available on RV32; it is quite likely that a capability mode-only RV32 implementation would have a use for those encodings.

tariqkurd-repo:
That's fair comment, it's restricting the usage to permissions.

jrtc27:
> That's fair comment, it's restricting the usage to permissions.

Not just permissions, it's enforcing it as being M.

tariqkurd-repo:
This is fair comment - the way that the bit is encoded means that it can only represent something which is used for code-capabilities in future - so some other attribute we havne't though of yet. It effectively means that bit 0 of quadrant 1 for RV32 only is up for grabs.

Issue #293, opened 2024-06-21
Representability check description doesn't capture address-wrapping edge cases


PRugg-Cap:
The description of the representability check is slightly imprecise in cases where the base or top want to wrap the address space. A correction is described in the bounds section: https://github.com/riscv/riscv-cheri/blob/8f587528c7d1a0005be68c38d955408bb22485f2/src/cap-description.adoc?plain=1#L501-L502

but not in the representable region section, where it is also relevant:
https://github.com/riscv/riscv-cheri/blob/8f587528c7d1a0005be68c38d955408bb22485f2/src/cap-description.adoc?plain=1#L649-L653 

In addition, we need to clarify what happens if the base wraps.

I'll need to think about how to describe this precisely, but just wanted to record the issue so it doesn't get missed. I don't think there's a bug, but somebody trying to implement it just from this description would certainly get confused!

tariqkurd-repo:
PR for this one? @PRugg-Cap ?

PRugg-Cap:
Not yet, sorry, I'll try tomorrow!

tariqkurd-repo:
fixed by the referenced MR

Issue #362, opened 2024-09-12
Include SAIL snippets into the specification


andresag01:
This includes the snippets into the instruction listings as well as the capability description chapter.

francislaus:
Can we (and should we) pick up a dependency to the Sail model for that? Would it be possible to automatically pull in the relevant lines from the Sail model? @Timmmm @PeterRugg

wmat:
There is for PDFs.  See https://github.com/Alasdair/asciidoctor-sail

I have a branch of riscv-isa-manual called sail-inclusion here:
https://github.com/riscv/riscv-isa-manual/tree/sail-inclusion with an
example of using Alasdair's extension to pull SAIL into the doc.  If you
look at my commit here:
https://github.com/riscv/riscv-isa-manual/commit/abd45a5a56154bcabbd07daaf41d91a1945729ce

you can see the changes I had to make for the example.  But please see
Alasdair's documentation in his repository for more information here:
https://github.com/Alasdair/asciidoctor-sail/blob/master/doc/built/sail_to_asciidoc.pdf

Timmmm:
I did set this up when the spec was in Codasip but we removed it when it was moved to GitHub since the Sail code wasn't open at the time. I'll see if I can resurrect it.

The big question is how do you get the JSON file. It's very big so I don't think you just want to add it to Git. But adding Sail compilation to the build process is also not great (although it's a lot better since we made the Sail binary release). In Codasip we had a cache on a web server which worked pretty well. How do people feel about using GitHub releases for that?

So every push (or maybe tag) in the Sail repo would generate the JSON and add it to a GitHub release, and then this repo would just point to the URL and download it.

We'll have to answer this question for riscv-isa-manual too and I think this repo could serve as a good demonstrator.

wmat:
For my testing purposes on riscv-isa-manual I've been generating the index at the top of tree of the sail repository and adding it directly into the repo and referencing it with an AsciiDoc attribute in the document header (riscv-unprivileged.adoc) with `:sail-doc: src/sail-index/riscv_RV64.json`.

Timmmm:
Yeah the issue is it's 4.5 MB (3.1 MB minified). Feels a bit much for a file that we'd continually update? Though I guess it probably does diff pretty well at least.

Timmmm:
[...]

PR #512, opened 2025-01-30
clarify top bound and how it feeds into rep range check

tariqkurd-repo:
@PeterRugg some small clarifications in the top bound and representable range check


src/cap-description.adoc
```diff
@@ -448,8 +448,8 @@ T[MW - 1:MW - 2] = B[MW - 1:MW - 2] + LCout + LMSB
 Decoding the bounds:
 
 ```
-top:    t = { a[MXLEN - 1:E + MW] + ct, T[MW - 1:0]    , {E{1'b0}} }
-base:   b = { a[MXLEN - 1:E + MW] + cb, B[MW - 1:0]    , {E{1'b0}} }
+top:    t = { 1'b0, a[MXLEN - 1:E + MW] + ct, T[MW - 1:0], {E{1'b0}} }
```
andresag01, line 451: Hi Tariq! Does this mean that the `top` is now 66 bits and `top[65]==1'b0` always? Not sure about this.

This pseudo-code was always meant to be temporary and it has no formal definition, but I seem to recall that `ct` could overflow `a[MXLEN - 1:E + MW]` and that's how we would get the 65 bits to form the top bound of the infinite cap bounds. This is why I suggested [here](https://github.com/riscv/riscv-cheri/issues/362#issuecomment-2622884049) that this be replaced with SAIL snippets given that those now work although this might make the cap encoding description a bit more difficult to read.

tariqkurd-repo, line 451:
```suggestion
top:    t = { {1'b0, a[MXLEN - 1:E + MW]} + ct, T[MW - 1:0], {E{1'b0}} }
```

src/cap-description.adoc
```diff
@@ -448,8 +448,8 @@ T[MW - 1:MW - 2] = B[MW - 1:MW - 2] + LCout + LMSB
 Decoding the bounds:
 
 ```
-top:    t = { a[MXLEN - 1:E + MW] + ct, T[MW - 1:0]    , {E{1'b0}} }
-base:   b = { a[MXLEN - 1:E + MW] + cb, B[MW - 1:0]    , {E{1'b0}} }
+top:    t = { {1'b0, a[MXLEN - 1:E + MW]} + ct, T[MW - 1:0], {E{1'b0}} }
+base:   b = {       a[MXLEN - 1:E + MW] + cb, B[MW - 1:0], {E{1'b0}} }
```
tariqkurd-repo, line 452:
```suggestion
base:   b = {        a[MXLEN - 1:E + MW]  + cb, B[MW - 1:0], {E{1'b0}} }
```

src/cap-description.adoc
tariqkurd-repo: yes - the top bound is wider, but we're only talking about the rep range check here which isn't affected by any bits above MXLEN-1

PeterRugg:
Yes, nice tidy-ups, and you're right that the overflow should be ignored for the representability check. The tricky thing to get right is the bounds decoding for wrapping cases, but I think that's made clear in the bounds decoding section. Thanks for sorting this. I think this closes https://github.com/riscv/riscv-cheri/issues/293. Sorry for the delay on that, but I'm glad this has had multiple eyes now!

PR #520, opened 2025-02-05
Minor revision

marnovandermaas:
Some minor revisions that I found when scrolling through the document.


src/instructions.adoc
```diff
@@ -260,6 +272,8 @@ include::insns/prefetch.w.adoc[]
 <<<
 === "Zba" Extension for Bit Manipulation Instructions
 
+Bit manimulation instruction reference.
```
tariqkurd-repo, line 275:
```suggestion
Bit manipulation instruction reference.
```

arichardson (approved):
Thanks!

PR #521, opened 2025-02-05
detection of out of range top

tariqkurd-repo:
Suggestions about how to handle the malformed caps which have an out of range top, which are difficult to detect without expanding the bounds

@PeterRugg 

fixes https://github.com/riscv/riscv-cheri/issues/522


src/cap-description.adoc
```diff
@@ -537,6 +537,18 @@ NOTE: A capability has infinite bounds if E=CAP_MAX_E and it is not
...
+[#section_cap_malformed_top_gt_inf_top]
+===== The Top Bound is Out Of Range
+
+The top bound of the <<infinite-cap>> is defined to be 2^MXLEN^.
+
+There are cases where the decoded bounds of a capability could, in theory, exceed this value.
+
+These cases are computationally expensive to detect without expanding the bounds, therefore
+all instructions such as <<CBLD>> which expand bounds and output a capability will clear
+the output tag if the top bound is out of range.
```
jrtc27, line 549: I don't understand this point. No such capability is given to you in the initial system state, so of course CBLD must ensure that it doesn't give you back one of these, but that's by virtue of it never being a subset of an existing capability?

src/cap-description.adoc
```diff
@@ -561,6 +573,7 @@ CHERI enforces the following invariants for all valid (i.e., tagged) capabilitie
 . The bounds are not malformed.
 . No reserved bit in the capability encoding is set.
 . The permissions can be legally produced by <<ACPERM>>.
+. The <<section_cap_malformed_top_gt_inf_top,top bound must be no greater than 2^XLEN^>>.
```
jrtc27, line 576: MXLEN

src/insns/cadd_32bit.adoc
```diff
@@ -48,7 +48,8 @@ For <<CADD>>, the address is incremented by the value in
 For <<CADDI>>, the address is incremented by the immediate value
 `imm`.
 
-include::malformed_clear_tag.adoc[]
+NOTE: This instruction sets `cd.tag=0` if `cs1` 's bounds are <<section_cap_malformed,malformed>>,
```
jrtc27, line 51: Why is ADD singled out?

src/cap-description.adoc
tariqkurd-repo: this is like the malformed bounds case, and so I'd like (at least the option of) detagging if such a thing is observed on either input of CBLD, just like we do for malformed bounds, reserved bits set etc...

src/cap-description.adoc
tariqkurd-repo, line 576:
```suggestion
. The <<section_cap_malformed_top_gt_inf_top,top bound must be no greater than 2^MXLEN^>>.
```

src/insns/cadd_32bit.adoc
tariqkurd-repo: because they don't expand the bounds, so for cheaper implementation we're only including instructions which do expand the bounds

jrtc27: They still need to perform a representability check

tariqkurd-repo: this is true, but don't expand the bounds. A functional unit which does CADD/CADDI doesn't need to do anything with the bounds, and the expansion is expensive just for this check.
If the expansion has already been done then the check is cheap.

src/cap-description.adoc
andresag01: @tariqkurd-repo : Could you provide an example situation you are thinking that CBLD would produce a capability that has this property? (i.e. its top > 2^MXLEN and is tagged).

As @jrtc27 explained, CBLD will only ever output capabilities that have equal or smaller bounds than the input capability to maintain the property that software cannot gain permissions it did not have. So, if at reset the most permissive/powerful capability provided to software is the infinite capability, then CBLD should never return a capability that is tagged and has top > infinite_cap.top.

tariqkurd-repo: These are currently (mostly) undocumented malformed bounds cases, and there are many of them.
They can only be caused by fault injection or bugs e.g. the tag cache attaching the wrong tag, and so should be filtered out.
They are too expensive to detect directly from the bounds, but trivial if the bounds are expanded.
Strictly speaking nothing needs to change in the hardware as they are impossible cases, just like reserved bits set or bad permissions.
The objective is to get formal verification to work allowing _any_ bit pattern against the design, and this is one of the awkard cases, so we need to choose an approach.

Issue #523, opened 2025-02-05
Minor formatting issue in Zcheripte


Timmmm:
![Image](https://github.com/user-attachments/assets/f69e0e7f-b52c-4ee1-82f6-52914e51c1dc)

The final two bullets should be indented (or maybe numbered).

PR #524, opened 2025-02-05
Fix indentation of sub-items

arichardson:
Move the note after the two sub-items to ensure asciidoc renders the document correctly.

Fixes: https://github.com/riscv/riscv-cheri/issues/523


arichardson:
CC: @Timmmm

Timmmm:
LGTM, I can't really remember why it was in that location in the first place. Would it make more sense on line 109?

arichardson:
> LGTM, I can't really remember why it was in that location in the first place. Would it make more sense on line 109?

Yes I think moving it up there makes sense. Will move it and then merge. Thanks for taking a look.

andresag01 (approved):

tariqkurd-repo (approved):

PR #525, opened 2025-02-05
Changes to pointer masking

LawrenceEsswood:
Change interaction with pointer masking to suggest that it should affect both bounds decode and checks.

### DIFF
--- a/src/ptrmask-integration.adoc
+++ b/src/ptrmask-integration.adoc
@@ -4,6 +4,10 @@
 The pointer masking extensions Smmpm, Smnpm, SSnpm, Sspm and Supm are compatible with both {cheri_base_ext_name} and {cheri_default_ext_name}.
 Whenever pointer masking is enabled, all bounds decoding and bounds checks are affected.
 
+NOTE: The suggestion in this section is based on the pointer masking approach from Morello but with changes
+to sign extension and to address the dynamic nature of bit masking.
+Further evaluation on RISCV, especially for {cheri_cap_mode_name}, is still required.
+
 When bounds are <<section_cap_encoding,encoded or decoded>>, a masked but _not_ sign extended address is used.
 Changing how many bits are masked can therefore change the interpretation of the bounds of a capability,
 both for the purpose of implicit accesses via bounds checks and any instructions that report the bounds of a capability.
@@ -14,6 +18,7 @@ it is not invalid to have a capability that covers a longer range than could be 
 (such as one that covers the entire MXLEN address space).
 For the <<section_cap_representable_check, representable range check>>, both the original and new address
 are masked.
+Bounds setting instructions also mask the address in the same way.
 
 NOTE: Because dynamically changing the number of masked bits changes the interpretation of a capability,
 software must take the same care when sharing capabilities between address spaces with differing pointer masking
@@ -21,4 +26,4 @@ as it generally must when sharing capabilities between address spaces with diffe
 
 Any address that is checked against a capability (whether via loads/stores in {cheri_int_mode_name} or {cheri_cap_mode_name}),
 is also first subject to the same masking as bounds decode (masking without extension).
-After any CHERI operations, the final access address is still subject to sign extension as normal.
+After any CHERI operations, the final access address is still subject to as much sign extension as the masking extensions mandate.
--- a/src/ptrmask-integration.adoc
+++ b/src/ptrmask-integration.adoc
@@ -1,18 +1,24 @@
 [#section_pointer_masking_integration]
 == Integrating {cheri_base_ext_name} and {cheri_default_ext_name} with Pointer Masking
 
-The pointer masking extensions Smmpm, Smnpm, SSnpm, Sspm and Supm are compatible with {cheri_default_ext_name}.
+The pointer masking extensions Smmpm, Smnpm, SSnpm, Sspm and Supm are compatible with both {cheri_base_ext_name} and {cheri_default_ext_name}.
+Whenever pointer masking is enabled, all bounds decoding and bounds checks are affected.
 
-For instructions using integer addresses (e.g. loads/stores in {cheri_int_mode_name}), they
-are interpreted as being XLEN-wide, and may be subject to pointer masking.
-All data accesses are checked against <<ddc>> which is unaffected by
-pointer masking. Therefore no capability bounds encoding is affected.
+When bounds are <<section_cap_encoding,encoded or decoded>>, a masked but _not_ sign extended address is used.
+Changing how many bits are masked can therefore change the interpretation of the bounds of a capability,
+both for the purpose of implicit accesses via bounds checks and any instructions that report the bounds of a capability.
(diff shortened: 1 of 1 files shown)


src/ptrmask-integration.adoc
```diff
@@ -1,18 +1,24 @@
...
-The pointer masking extensions Smmpm, Smnpm, SSnpm, Sspm and Supm are compatible with {cheri_default_ext_name}.
+The pointer masking extensions Smmpm, Smnpm, SSnpm, Sspm and Supm are compatible with both {cheri_base_ext_name} and {cheri_default_ext_name}.
+Whenever pointer masking is enabled, all bounds decoding and bounds checks are affected.
 
-For instructions using integer addresses (e.g. loads/stores in {cheri_int_mode_name}), they
-are interpreted as being XLEN-wide, and may be subject to pointer masking.
-All data accesses are checked against <<ddc>> which is unaffected by
-pointer masking. Therefore no capability bounds encoding is affected.
+When bounds are <<section_cap_encoding,encoded or decoded>>, a masked but _not_ sign extended address is used.
+Changing how many bits are masked can therefore change the interpretation of the bounds of a capability,
```
jrtc27, line 8: Well, this is the problem, and why we didn't choose this scheme.

arichardson: We could argue that changing translation scheme (or page table base) already changes interpretation of all capability and therefore changing the number of masked bits is essentially the same thing. Both operations require ASR, so hopefully only an issue for privileged code.

src/ptrmask-integration.adoc
```diff
@@ -1,18 +1,24 @@
...
+software must take the same care when sharing capabilities between address spaces with differing pointer masking
+as it generally must when sharing capabilities between address spaces with different page mappings.
 
-NOTE: This scheme doesn't seem very useful, but the problem is the dynamic configuration
- of pointer masking which can arbitrarily update the meaning of the address within the capability,
- so the full address field must be used to calculate bounds.
- There is future work required to determine a more useful way of applying pointer masking to capabilities.
+Any address that is checked against a capability (whether via loads/stores in {cheri_int_mode_name} or {cheri_cap_mode_name}),
+is also first subject to the same masking as bounds decode (masking without extension).
+After any CHERI operations, the final access address is still subject to sign extension as normal.
```
arichardson, line 24: This is only true of pointer masking is on? Otherwise there is no sign extension.

src/ptrmask-integration.adoc
LawrenceEsswood: Is it actually a problem though?

Masking high bits of pointers is equivalent to just repeating entries in the top level page table, and so is just an implicit modification to an address space. Changing the address space and then trying to use the same capabilities has always been a silly thing for software to do, this is no different.

The only issue that is left is the practicality of doing this in hardware. i.e., can we squeeze an extra mux on address in, and resolve pipeline hazards. Loads/stores _already_ had hazards on changing the number of masked bits.

src/ptrmask-integration.adoc
LawrenceEsswood: Well, masking and sign extension of all no masked bits. We can frame this as either no extension, or sign extension of all masked bits (which can be none).

src/ptrmask-integration.adoc
jrtc27: Changing the page tables does not cause the actual GCFOO to change what they report. Nor did the previous text. This now does. And aside from that, doesn't this mean that some valid capabilities become illegal encodings with base < top? And on top of that, masking the address for bounds decoding doesn't do anything to deal with B and T with a large exponent (when viewing it as positive, rather than negative as it effectively is in the spec).

LawrenceEsswood: Agreed on changes to GCFOO. Not sure if that poses a problem. Software that was manually comparing bounds has to deal with masking in any case. There are admittedly new hazards.

> masking the address for bounds decoding doesn't do anything to deal with B and T with a large exponent

Agreed, and this would cause a fault, and I don't think it matters. Having a capability that has bounds covering _only_ an address range not addressable with TBI, and then trying to access it with TBI on,  is a clear case of software mismanagement of capabilities. A trap is fine, if not desirable. If found, unsealed, in a context with masked bits, all capabilities should really be allocated in either the top or bottom of the address space, or possibly the whole space or some large portion of it.

> doesn't this mean that some valid capabilities become illegal encodings with base < top.

Hmm, possibly I overlook something. I had reckoned that taking a capability covering somewhere in the middle of the address space with no mask into a space mask would change cb / ct. I didn't think it would matter _what_ bounds it would end up having (because it is a clear sign of software mismanagement, as it was with large T and B). As long as top >= base, and base wouldn't go under 0, or top over 2^XLEN.

With 

```
top:    t = { {1'b0, a[MXLEN - 1:E + MW]} + ct, T[MW - 1:0], {E{1'b0}} }
base:   b = {        a[MXLEN - 1:E + MW]  + cb, B[MW - 1:0], {E{1'b0}} }
```

And ct/cb being between +1/-1, there is no way to get base > top without underflow of base.

top would not get larger for zero-ing more bits of a.

For arguing it cannot cause base to underflow: to do that, you would need B=0. Thus, B < R, so cb could not be -1. I think its safe.

LawrenceEsswood: I have made a mistake there.

Underflow does not require B=0, it would require a[MXLEN - 1:E + MW]=0. Which can happen.

Possibly then, yes, this can cause `cb=-1` with  `a[MXLEN - 1:E + MW]=0` (only in the case the OS has done something wrong), so then either a clamp is needed, or we have to treat that as an invalid capability.

That is sad, as previously there was not a way of architecturally getting invalid bounds without memory error.

src/ptrmask-integration.adoc
LawrenceEsswood: I've been clear that only as much sign extension as the base spec requires is done here.

src/ptrmask-integration.adoc
```diff
@@ -1,18 +1,29 @@
...
-for bounds calculation. When pointer masking is enabled, the dereferenced address has the masked bits replaced by sign
-extension before the bounds check.
+When bounds are <<section_cap_encoding,encoded or decoded>>, a masked but _not_ sign extended address is used.
+Changing how many bits are masked can therefore change the interpretation of the bounds of a capability,
+both for the purpose of implicit accesses via bounds checks and any instructions that report the bounds of a capability.
+Apart from treating high address bits as 0, there are no other changes to bounds decode,
+which are still based on MXLEN, not the new effectively addressable space.
+That is, the maximum length of a capability does not change, and
+it is not invalid to have a capability that covers a longer range than could be actually be addressed with pointer masking enabled.
+(such as one that covers the entire MXLEN address space).
```
arichardson, line 18:
```suggestion
it is not invalid to have a capability that covers a longer range than could be actually be addressed with pointer masking enabled (such as one that covers the entire MXLEN address space).
```

arichardson (approved):
This matches what was discussed.

Issue #476, opened 2024-12-11
Tag clearing vs trapping for store capability violations: resolve TODO?


rmn30:
https://riscv.github.io/riscv-cheri/#_capability_load_and_store_instructions currently contains a TODO saying that storing a tagged capability via a capability that does not have the C permission may cause a trap in future rather than clearing the tag of the stored capability as it does at the moment.

This is a point of divergence for CHERIoT which currently traps in this case (but not in the case of a store-local / capability level violation). Trapping may be slightly preferable for software as it detects the error sooner, but is more difficult for hardware as it introduces a data dependent trap.

davidchisnall:
I would not want the standard to mandate trapping in the core specification because (as we learned with Morello), the data-dependent trap in the store can be painful on superscalar implementations.  At the same time, it's fairly easy to implement on in-order pipelines.

From a software perspective, the trap is useful.  We can avoid traps in `memcpy` and other forms of type-oblivious copy by propagating the MC permission from the destination to the source (if the destination lacks store-capability, remove load-capability from the source).  This lets you avoid taking traps when copying untrusted data, but otherwise storing a capability via a !MC capability is silent data corruption in the tag-clearing case and so catching it early is beneficial.

Prior to ratifying the standard, it would be nice to ensure that both forms are supported.  On big systems, the trapping mode may be useful as a per-process debugging feature for things willing to take the performance hit (and as a store barrier for concurrent garbage collection), but the tag-clearing option is a better default.  A *NIX profile should at least provide a tag-clearing mode, and optionally a trapping mode.  An embedded profile should be free to use only the trapping mode.

andresag01:
I think this TODO was added from the initial version of the spec. I agree, the tag-clearing option should be the default although this is best controlled through the profiles.

@davidchisnall: When you mentioned "optional" tag-clearing / trapping mode, did you mean behavior controlled through (e.g.) a CSR such that it can be changed at run-time or fixed by the hardware designers? I would have thought that one would generally like either tag-clearing or trapping, but not both options in a system.

davidchisnall:
[...]

PR #528, opened 2025-02-07
Write note about traps for capability stores

andresag01:
Add a note indicating that capability stores may trap in future extensions when the authorizing capability does not grant capability write permissions. Also remove outdated TODO.

Fixes https://github.com/riscv/riscv-cheri/issues/476

### DIFF
--- a/src/riscv-integration.adoc
+++ b/src/riscv-integration.adoc
@@ -224,8 +224,9 @@ capability written to memory is cleared if the authorizing capability does not
 grant permission to write capabilities (i.e. both <<w_perm>> and <<c_perm>>
 must be set in AP).
 
-WARNING: #TODO: these cases may cause exceptions in the future - we need a way
-for software to discover and/or control the behavior#
+NOTE: Future extensions to {cheri_base_ext_name} may add mechanisms that cause
+stores to raise exceptions when the authorizing capability does not grant both
+<<w_perm>> and <<c_perm>>.
 
 [#section_existing_riscv_insns]
 === Existing RISC-V Instructions


tariqkurd-repo (approved):

src/riscv-integration.adoc
```diff
@@ -224,8 +224,9 @@ capability written to memory is cleared if the authorizing capability does not
 grant permission to write capabilities (i.e. both <<w_perm>> and <<c_perm>>
 must be set in AP).
 
-WARNING: #TODO: these cases may cause exceptions in the future - we need a way
-for software to discover and/or control the behavior#
+NOTE: Future extensions to {cheri_base_ext_name} may add mechanisms that cause
+stores to raise exceptions when the authorizing capability does not grant both
+<<w_perm>> and <<c_perm>>.
```
jrtc27, line 229: I think this needs an “if the tag of the capability being stored to memory is set” somewhere?

arichardson: We could add that to be extra clear but I since this is a note underneath "what happens when you don't have "C+W" I think it's probably fine to omit this clarification.

jrtc27: I see nothing that states this any of this isn't for SC as a whole. The paragraph above just says the tag is cleared. C says "Allow reading capability data from memory if the authorizing capability also grants R-permission. Allow writing capability data to memory if the authorizing capability also grants W-permission." We don't define capability data anywhere, so a completely reasonable, if false, interpretation would be that _any_ LC/SC counts as a load/store of capability data. 3.2.1 Tag talks about valid and invalid capabilities. I would in fact argue that capability data is both valid and invalid capabilities. Similarly, we have "Missing LM-permission does not affect untagged values since this could result in surprising bit patterns when copying non-capability data. Similarly, sealed capabilities are not modified as they are not directly dereferenceable." in various places, which is assuming a definition that only encompasses valid capabilities. This is all quite loose, making implicit assumptions and, I would argue, using the wrong definition.

arichardson: I imagine a future extension could have at least two options: 1) trap for tagged data missing W|C, 2) trap for any cap store missing W|C (even if that is less useful.

I do agree that we should make sure to be explicit about anything defined in the spec but since this is about some hypothetical future thing being vague until that extension actually exists seems ok?

jrtc27: Well my point is the latter is something that should not be permitted, because it breaks memcpy

arichardson: Fair, would you be willing to fix this note?

PR #530, opened 2025-02-10
small fixes

tariqkurd-repo:


arichardson (approved):

src/riscv-integration.adoc
```diff
@@ -26,14 +26,14 @@ privileged architecture specified in the RISC-V ISA.
 === Memory
 
 A hart supporting {cheri_base_ext_name} has a single byte-addressable address
-space of 2^XLEN^ bytes for all memory accesses. Each memory region capable of
+space of 2^MXLEN^ bytes for all memory accesses. Each memory region capable of
```
andresag01, line 29: @tariqkurd-repo @arichardson : This was correct and it was taken directly from the RISC-V unprivileged specification where it says (see Page 16 Section 1.4 [here](https://drive.google.com/file/d/1uviu1nH-tScFfgrovvFCrj7Omv8tFtkp/view)):
> A RISC-V hart has a single byte-addressable address space of bytes for all memory accesses.

It is meant to be the effective XLEN of the machine which can be MXLEN or SXLEN or UXLEN depending CSRs. The idea is that the byte-addressable memory space is just like in CHERI-less RISC-V so we are fully compatible, but capability bounds remain encoded in MXLEN width regardless of the effective XLEN.

PR #531, opened 2025-02-12
Pte fault overhaul

tariqkurd-repo:
the previous definitions really weren't very clear


andresag01:
Looks like a commit from an earlier PR sneaked into this PR?

nwf:
The top commit ("CHERI PTE fault overhaul") looks OK by me, and I agree it's just clarifying the existing behaviors.

tariqkurd-repo:
> The top commit ("CHERI PTE fault overhaul") looks OK by me, and I agree it's just clarifying the existing behaviors.

yeah thanks - I was worried then when this goes to review they'd be confused by what a CHERI PTE fault actually is

arichardson (commented):
This looks good, but it would be good to remove the initial commit since I'm not sure what github will do when we merge this.

PR #532, opened 2025-02-12
Revert MXLEN to XLEN in RV memory description

andresag01:
The text was originally taken from RISC-V and ideally it should remain as in the Unprivileged RISC-V spec to ensure that the CHERI extension is backwards-compatible with RISC-V.

### DIFF
--- a/src/riscv-integration.adoc
+++ b/src/riscv-integration.adoc
@@ -26,14 +26,14 @@ privileged architecture specified in the RISC-V ISA.
 === Memory
 
 A hart supporting {cheri_base_ext_name} has a single byte-addressable address
-space of 2^MXLEN^ bytes for all memory accesses. Each memory region capable of
+space of 2^XLEN^ bytes for all memory accesses. Each memory region capable of
 holding a capability also stores a tag bit for each naturally aligned CLEN bits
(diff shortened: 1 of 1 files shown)


tariqkurd-repo:
it's not necessary as MXLEN must be the largest of all possible XLENs, so the current text is correct
the newer privilege spec 1.13 includes this limitations

andresag01:
> it's not necessary as MXLEN must be the largest of all possible XLENs, so the current text is correct the newer privilege spec 1.13 includes this limitation

Yes, the value of `misa.MXL` is read-only and set to MXLEN in the newer privileged spec, but that is only telling us the effective XLEN while in M-mode, and for the purposes of CHERI, it defines the bit width of the capability.

However, the sentence changed in this PR is _specifically_ about memory and the effective XLEN which could be read-write and <MXLEN depending on mstatus.SXL, mstatus.UXL and whether we are in S-mode or U-mode. For example, if MXLEN=64, but mstatus.SXL=32-bit and we are in S-mode, how would the machine access memory addresses above `2^32 - 1` if registers are only 32 bits?

tariqkurd-repo:
yes ok - the reference in the RISC-V ISA spec does talk about 2^XLEN so better to be consistent with that

PR #533, opened 2025-02-12
remove unnecessary line breaks

tariqkurd-repo:


arichardson (approved):
Agreed we don't need that many page breaks.

PR #537, opened 2025-02-14
Fix encoding of C.FSD instruction

tomaird:
The encoding diagram for `C.FSD` was completely wrong, it seem to have been a bad copy paste from `C.FSDSP`

### DIFF
--- a/src/insns/wavedrom/c-sp-store-css-fp-dp.adoc
+++ b/src/insns/wavedrom/c-sp-store-css-fp-dp.adoc
@@ -3,9 +3,11 @@
 [wavedrom, ,svg]
 ....
 {reg: [
-  {bits: 2, name: 'op',     type: 8, attr: ['2','C0=00']},
-  {bits: 5, name: 'rs2',    type: 4, attr: ['5','src']},
-  {bits: 6, name: 'imm',    type: 3, attr: ['6','offset[5:3|8:6]']},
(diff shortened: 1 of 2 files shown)


src/insns/wavedrom/c-sp-store-css-fp-dp.adoc
```diff
@@ -3,9 +3,11 @@
 [wavedrom, ,svg]
 ....
 {reg: [
-  {bits: 2, name: 'op',     type: 8, attr: ['2','C0=00']},
-  {bits: 5, name: 'rs2',    type: 4, attr: ['5','src']},
-  {bits: 6, name: 'imm',    type: 3, attr: ['6','offset[5:3|8:6]']},
+  {bits: 2, name: 'op',     type: 8, attr: ['2', 'C0=00']},
+  {bits: 3, name: 'rs2\'',  type: 3, attr: ['3', 'src']},
```
arichardson, line 7: Looks like there is some inconsistency with the compressed instruction here, some of them use `rs2`, some `rs2'`, and some use `rs2``

src/insns/wavedrom/c-sp-store-css-fp-dp.adoc
```diff
@@ -3,9 +3,11 @@
 [wavedrom, ,svg]
```
arichardson, line 3: Remove the "is this correct?' above?

arichardson (approved):
This matches the unpriv spec, so LGTM.

Issue #540, opened 2025-02-18
Clarify when mtvec.mode should be included in representability check


tomaird:
The description for `mtvecc` has the following: 

![Image](https://github.com/user-attachments/assets/1fbb920e-1f07-4ba5-97d5-6a990d4041ac)

From these two highlighted sections, it's a bit unclear when writing to `xtvecc` whether the representability check should be done on 

1.  `mtvecc.address` (including mode bits) - this is the address that is visible to SW
2. OR `{mtvecc.address[MXLEN-1:2], 2'b00}` - this is the address that is actually jumped to (for non-vectored traps)
3. OR both of the above

(Obviously when `MODE=vectored` we also do the representability check on `{mtvecc.address[MXLEN-1:2], 2'b00} + 4 * HICAUSE` - but that is more clear, since we're checking the maximum possible vector target is in range.)

tariqkurd-repo:
Yes - this is clearly wrong - the rep check where the MODE is set to Vectored has the LSBs set to 2'b01, but the address jumped to on exception always has the LSBs set to 2'b00.
In very weird cases, it's possible for the value in mtvecc to only be representable when the LSB is set.
Therefore we must be clear that the address range checked for representability _must_ include all possible exceptoin vector addresses.

arichardson:
We also don't specify whether you can use a sentry in direct mode. I am not sure if this is something that should be supported by unsealing on trap (@jrtc27 @nwf)?

PR #542, opened 2025-02-19
do mtvecc rep check with MODE field zeroed

tariqkurd-repo:
We must do the rep check on the value which is actually used on exception launch, otherwise it's possible to jump to a value outside the rep range by clearing the LSB of mtvecc in vectored mode

fixes https://github.com/riscv/riscv-cheri/issues/540

This effects all versions of the spec from at least version 0.9.0

### DIFF
--- a/src/riscv-integration.adoc
+++ b/src/riscv-integration.adoc
@@ -511,8 +511,9 @@ change on writing. If the MODE field is non-zero after any WARL conversion,
 then the address used for the exception vector (which has **mtvecc.address[1:0]=2'b00**)
 will not match the value read by software from the CSR.
 
-In this situation, the same capability can have more than one address.
-When updating the CSR _all possible_ values of the address must be in the
+In this situation, the same capability can effectively have more than one address.
+When updating the CSR _all possible_ values of the address visible either by CSR read,
+or which are used as the <<pcc>> on exception launch, must be in the
 <<section_cap_representable_check>>, and the tag cleared if _any_ are not.
 
 The capability is always written using <<SCADDR>> semantics to update the address field,
@@ -520,11 +521,11 @@ even when writing the full capability, and so sealed capabilities will always ha
 their tags cleared.
 
 Additionally, when MODE=Vectored the capability has its tag bit cleared if the
-capability address + 4 x HICAUSE is not within the <<section_cap_representable_check>>.
+capability address~3 + 4xHICAUSE is not within the <<section_cap_representable_check>>.
 HICAUSE is the largest interrupt cause value that the implementation can write
 to <<mcause>> or <<scause>>/<<vscause>> when an interrupt is taken.
 
-NOTE: When MODE=Vectored, it is only required that address + 4 x HICAUSE is
+NOTE: When MODE=Vectored, it is only required that the capability address + 4xHICAUSE is
 within the <<section_cap_representable_check>> instead of the capability's bounds.
 This ensures
 that software is not forced to allocate a capability granting access to more
@@ -532,7 +533,7 @@ memory for the trap-vector than necessary to handle the trap causes that
 actually occur in the system.
 
 NOTE: When MODE=Vectored, if either the capability address _or_ the capability
-address + 4 x HICAUSE are invalid then the <<section_invalid_addr_conv>> rules
+address + 4xHICAUSE are invalid then the <<section_invalid_addr_conv>> rules
 are followed which may require the tag to be cleared. In particular, if any part
 of the range is in the invalid address space then clearing the tag is strongly
 recommended.
--- a/src/riscv-integration.adoc
+++ b/src/riscv-integration.adoc
@@ -507,7 +507,7 @@ its address incremented by four times the interrupt cause number.
 
 Capabilities written to <<mtvecc>> also include writing the MODE field in
 **mtvecc.address[1:0]**, which is a WARL field, meaning that the capability address can
-change on writing. If the MODE field is non-zero after any legalization,
+be legalized. If the MODE field is non-zero after any legalization,
 then the address used for the exception vector (which has **mtvecc.address[1:0]=2'b00**)
 will not match the value read by software from the CSR.
 
--- a/src/riscv-integration.adoc
+++ b/src/riscv-integration.adoc
@@ -521,11 +521,11 @@ even when writing the full capability, and so sealed capabilities will always ha
 their tags cleared.
 
 Additionally, when MODE=Vectored the capability has its tag bit cleared if the
-capability address~3 + 4xHICAUSE is not within the <<section_cap_representable_check>>.
+`(capability address & ~3) + 4 * HICAUSE` is not within the <<section_cap_representable_check>>.
 HICAUSE is the largest interrupt cause value that the implementation can write
 to <<mcause>> or <<scause>>/<<vscause>> when an interrupt is taken.
 
-NOTE: When MODE=Vectored, it is only required that the capability address + 4xHICAUSE is
+NOTE: When MODE=Vectored, it is only required that the capability address + `4 *  HICAUSE` is
 within the <<section_cap_representable_check>> instead of the capability's bounds.
 This ensures
 that software is not forced to allocate a capability granting access to more
@@ -533,7 +533,7 @@ memory for the trap-vector than necessary to handle the trap causes that
 actually occur in the system.
 
 NOTE: When MODE=Vectored, if either the capability address _or_ the capability
-address + 4xHICAUSE are invalid then the <<section_invalid_addr_conv>> rules
+address + `4 * HICAUSE` are invalid then the <<section_invalid_addr_conv>> rules
 are followed which may require the tag to be cleared. In particular, if any part
 of the range is in the invalid address space then clearing the tag is strongly
 recommended.
--- a/src/riscv-integration.adoc
+++ b/src/riscv-integration.adoc
@@ -506,13 +506,17 @@ interrupts cause the <<pcc>> to be set to the capability with
 its address incremented by four times the interrupt cause number.
 
 Capabilities written to <<mtvecc>> also include writing the MODE field in
-**mtvecc.address[1:0]**, even though it is not actually part of the exception vector address.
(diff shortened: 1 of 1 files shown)


src/riscv-integration.adoc
```diff
@@ -507,7 +507,7 @@ its address incremented by four times the interrupt cause number.
 
 Capabilities written to <<mtvecc>> also include writing the MODE field in
 **mtvecc.address[1:0]**. As a result, a representability and sealing check is
-performed on the capability with the legalized (WARL) MODE field included in
+performed on the capability with the MODE field zeroed in
```
arichardson, line 510: Hmm I'm not sure about this one - ISAv9 Qemu does the check with the legalized mode. Then again I am not convinced this actually matters since you can use mode==1 with sentries. 

I also feel that we should be a bit more explicit here about what happens if you install a sentry in mtvecc.

tariqkurd-repo: The point here is that it's possible (in a very bizarre way) to have  a cap which is representable when the LSB of the address is set, and not representable when the LSB is clear, so jumping to MTVECC in this case means jumping to a bad capability. That's the bug I'm trying to fix here, and it implies that Qemu on CHERI v9 has this an an undetected bug.

For sentries in mtvecc - I agree the spec isn't clear - clarifying that should be another PR.

tariqkurd-repo: ok - updated - including clarifying sentry behaviour

tariqkurd-repo: V9 wouldn't have this problem - it's because we hade the A<R check use many more bits to in the extreme case even address bit [0] can actually affect the representability result, which was probably an oversight

src/riscv-integration.adoc
```diff
@@ -506,10 +506,18 @@ interrupts cause the <<pcc>> to be set to the capability with
 its address incremented by four times the interrupt cause number.
 
 Capabilities written to <<mtvecc>> also include writing the MODE field in
-**mtvecc.address[1:0]**. As a result, a representability and sealing check is
-performed on the capability with the legalized (WARL) MODE field included in
-the address. The tag of the capability written to <<mtvecc>> is cleared if
-either check fails.
+**mtvecc.address[1:0]**, which is a WARL field, meaning that the capability address can
```
arichardson, line 509: In theory it would be possible to do this legalization on read and just store the "illegal value"?

src/riscv-integration.adoc
```diff
@@ -506,10 +506,18 @@ interrupts cause the <<pcc>> to be set to the capability with
...
+change on writing. If the MODE field is non-zero after any WARL conversion,
```
arichardson, line 510:
```suggestion
change on writing. If the MODE field is non-zero after legalization,
```

src/riscv-integration.adoc
```diff
@@ -506,10 +506,18 @@ interrupts cause the <<pcc>> to be set to the capability with
...
+then the address used for the exception vector (which has **mtvecc.address[1:0]=2'b00**)
+will not match the value read by software from the CSR.
```
arichardson, line 512: I'm not sure about this - won't we always read the legalized value? Or do you mean won't match the value "written to the CSR"?

src/riscv-integration.adoc
```diff
@@ -506,10 +506,18 @@ interrupts cause the <<pcc>> to be set to the capability with
...
+
+In this situation, the same capability can have more than one address.
+When updating the CSR _all possible_ values of the address must be in the
```
arichardson, line 515: Does this _all possible_ refer to the max address with vectored mode? Or just to mode=0/1?

src/riscv-integration.adoc
```diff
@@ -506,10 +506,18 @@ interrupts cause the <<pcc>> to be set to the capability with
...
+<<section_cap_representable_check>>, and the tag cleared if _any_ are not.
+
+The capability is always written using <<SCADDR>> semantics to update the address field,
```
arichardson, line 518: I'm not sure how useful it is to support this but couldn't we install a sentry for non-vectored mode? On exception it is unsealed and in Direct mode there is no address modification so that would work.

For vectored, I agree we have to check that cap address with the low bit cleared (since it is set to 1 for vectored) is representable.

arichardson (commented):
Am I understanding correctly that this clarification is only needed for MODE=VECTORED, where the LSB=1 when writing the capability, but when you get a trap the target address could be cap.addr with LSB clear, so we have to check that both `cap.addr&~1` is representable as well as `(cap.addr&~1)+4*HICAUSE`?

For MODE=direct no change is needed, and for mode=illegal we can choose to treat it as MODE=zero=direct?

tariqkurd-repo:
Yes, but we can't assume only 2 valid values in the mode field as CLIC defines more, and other extensions may, but essentially yes. We need to do a rep check on _all_ possible values which can be returned by a CSR read or which can be used as the PCC on exception launch.

tariqkurd-repo:
> Or maybe something like
> 
> > When writing a capability to mtvecc, a representability check must be performed for _all_ possible values which can be returned by a CSR read or which can be used as the PCC on exception launch. Based on the constraints of trap vector behavior, this check can be simplified to a representability check for `mtvecc.addr&~3` and `mtvecc.addr&~3+HICAUSE*4`.

almost - but if address[1:0]=01 is in the rep range and [1:0]=00 isn't then the tag should remain set

src/riscv-integration.adoc
tariqkurd-repo:
```suggestion
change on writing. If the MODE field is non-zero after any legalization,
```

src/riscv-integration.adoc
tariqkurd-repo: yes - so I fixed the text

src/riscv-integration.adoc
tariqkurd-repo: it's possible, but that would be a spec change......

src/riscv-integration.adoc
arichardson: New wording fixes this issue thanks.

src/riscv-integration.adoc
arichardson: For MODE=DIRECT the address matches the raw value of the capability so there is only one address that can be observed? We should call out that sentries are invalidated even in this case since there is no unsealing on MRET.

src/riscv-integration.adoc
```diff
@@ -506,25 +506,34 @@ interrupts cause the <<pcc>> to be set to the capability with
...
+NOTE: When MODE=Vectored, it is only required that the capability address + 4xHICAUSE is
 within the <<section_cap_representable_check>> instead of the capability's bounds.
 This ensures
 that software is not forced to allocate a capability granting access to more
 memory for the trap-vector than necessary to handle the trap causes that
 actually occur in the system.
 
 NOTE: When MODE=Vectored, if either the capability address _or_ the capability
-address + 4 x HICAUSE are invalid then the <<section_invalid_addr_conv>> rules
+address + 4xHICAUSE are invalid then the <<section_invalid_addr_conv>> rules
```
arichardson, line 536:
```suggestion
`address + 4 * HICAUSE` are invalid then the <<section_invalid_addr_conv>> rules
```
Maybe like this?

src/riscv-integration.adoc
```diff
@@ -506,25 +506,34 @@ interrupts cause the <<pcc>> to be set to the capability with
...
+or which are used as the <<pcc>> on exception launch, must be in the
+<<section_cap_representable_check>>, and the tag cleared if _any_ are not.
+
+The capability is always written using <<SCADDR>> semantics to update the address field,
+even when writing the full capability, and so sealed capabilities will always have
+their tags cleared.
 
 Additionally, when MODE=Vectored the capability has its tag bit cleared if the
-capability address + 4 x HICAUSE is not within the <<section_cap_representable_check>>.
+capability address~3 + 4xHICAUSE is not within the <<section_cap_representable_check>>.
```
arichardson, line 524:
```suggestion
`(capability address & ~3) + 4 * HICAUSE` is not within the <<section_cap_representable_check>>.
```

src/riscv-integration.adoc
```diff
@@ -506,25 +506,34 @@ interrupts cause the <<pcc>> to be set to the capability with
...
 HICAUSE is the largest interrupt cause value that the implementation can write
 to <<mcause>> or <<scause>>/<<vscause>> when an interrupt is taken.
 
-NOTE: When MODE=Vectored, it is only required that address + 4 x HICAUSE is
+NOTE: When MODE=Vectored, it is only required that the capability address + 4xHICAUSE is
```
arichardson, line 528:
```suggestion
NOTE: When MODE=Vectored, it is only required that the capability address + `4 *  HICAUSE` is
```

arichardson (commented):
I'd like @jrtc27 to take a look too but overall this looks good.

src/riscv-integration.adoc
```diff
@@ -506,25 +506,34 @@ interrupts cause the <<pcc>> to be set to the capability with
...
-**mtvecc.address[1:0]**. As a result, a representability and sealing check is
-performed on the capability with the legalized (WARL) MODE field included in
-the address. The tag of the capability written to <<mtvecc>> is cleared if
-either check fails.
+**mtvecc.address[1:0]**, which is a WARL field, meaning that the capability address can
+be legalized. If the MODE field is non-zero after any legalization,
+then the address used for the exception vector (which has **mtvecc.address[1:0]=2'b00**)
+will not match the value read by software from the CSR.
+
+In this situation, the same capability can effectively have more than one address.
```
arichardson, line 514:
```suggestion
```
I think this sentence is confusing and is best removed. We're basically just using this capability and with the address updated.

src/riscv-integration.adoc
arichardson: https://github.com/riscv/riscv-cheri/issues/561 opened to track the sentry issue. Otherwise this has been resolved.

src/riscv-integration.adoc
arichardson: https://github.com/riscv/riscv-cheri/issues/561

arichardson (approved):

PR #544, opened 2025-02-21
Improve consistency of CSR descriptions

arichardson:
Drop the "renamed extension" wording and make the descriptions consistent.

### DIFF
--- a/src/riscv-integration.adoc
+++ b/src/riscv-integration.adoc
@@ -330,8 +330,7 @@ two tag bits in memory.
 
 {cheri_base_ext_name} requires that RISC-V CSRs intended to hold addresses,
 are now able to hold capabilities. Therefore, such registers are
-extended to CLEN+1 bits in {cheri_base_ext_name} and a `c` suffix is added such
-as <<mtvec>> is extended to <<mtvecc>>.
+extended to CLEN+1 bits in {cheri_base_ext_name} and a `c` suffix is added (e.g. <<mtvec>> is extended to <<mtvecc>>).
 
 Reading or writing any part of a CLEN-bit CSR may cause
 side effects. For example, the CSR's tag bit may be cleared if a new address
--- a/src/riscv-integration.adoc
+++ b/src/riscv-integration.adoc
@@ -329,8 +329,9 @@ two tag bits in memory.
 === Zicsr, Control and Status Register (CSR) Instructions
 
 {cheri_base_ext_name} requires that RISC-V CSRs intended to hold addresses,
-like <<mtvec>>, are now able to hold capabilities. Therefore, such registers are
-renamed and extended to CLEN+1 bits in {cheri_base_ext_name}.
+are now able to hold capabilities. Therefore, such registers are
+extended to CLEN+1 bits in {cheri_base_ext_name} and a `c` suffix is added such
+as <<mtvec>> is extended to <<mtvecc>>.
 
 Reading or writing any part of a CLEN-bit CSR may cause
 side effects. For example, the CSR's tag bit may be cleared if a new address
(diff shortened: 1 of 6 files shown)


src/riscv-integration.adoc
```diff
@@ -330,7 +330,7 @@ two tag bits in memory.
 
 {cheri_base_ext_name} requires that RISC-V CSRs intended to hold addresses,
 like <<mtvec>>, are now able to hold capabilities. Therefore, such registers are
-renamed and extended to CLEN-bit in {cheri_base_ext_name}.
+renamed and extended to CLEN+1 bits in {cheri_base_ext_name}.
```
nwf, line 333: Perhaps something like
> for brevity, this document simply says "The Xc register extends X" rather than repeatedly pointing out the simultaneous renaming.

arichardson: Thatnks that sounds good.

andresag01: You are right, the "rename" idea was from a previous iteration of the spec when we had separate addresses for the Xc CSRs. IMHO, this chapter makes the spec confusing because it still talks about "rename", so I suggest removing that altogether and just talk about extended registers.

Also, note that at this point in the spec we are defining Zcheripurecap which does not support the CHERI mode, so in fact, a CPU only implementing this extension would only ever have access to the "extended" Xc registers.

src/riscv-integration.adoc
```diff
@@ -388,6 +388,7 @@ xref:vscsrnames-renamed[xrefstyle=short] and
 xref:ucsrnames-renamed[xrefstyle=short] from the base RISC-V ISA and its
 extensions. The CSRs are renamed to reflect the fact that they are extended to
 CLEN+1 bits wide, as the *x* registers are renamed to *c* registers.
+Even though the name changes, the CSR index remains identical and the <<section-cheri-execution-mode>> determines whether CLEN+1 or XLEN bits are returned (see <<CSRRW>>).
```
andresag01, line 391: I like this sentence, but at this point, the spec hasn't really defined the CHERI mode and a CPU that only supports Zcheripurecap can only access the extended CSRs. I suggest to instead just replace the "rename" wording here with "extend and move this text to a "note" with a link to the section for CSR instructions in Zcherihybrid which says the same in expanded form:

> Zcherihybrid adds the concept of CSRs which contain a capability where the address field is visible in Integer Pointer Mode (e.g. mtvec) and the full capability is visible in Capability Pointer Mode through a different name (e.g. mtvecc). These are referred to as extended CSRs. Also, Zcherihybrid adds the new capability CSRs listed in Table 33.
Extended CSRs have only one address; the access width is determined by the execution mode.

tariqkurd-repo (approved):

src/riscv-integration.adoc
tariqkurd-repo: ok  -  fixed

src/riscv-integration.adoc
```diff
@@ -329,8 +329,9 @@ two tag bits in memory.
 === Zicsr, Control and Status Register (CSR) Instructions
 
 {cheri_base_ext_name} requires that RISC-V CSRs intended to hold addresses,
-like <<mtvec>>, are now able to hold capabilities. Therefore, such registers are
-renamed and extended to CLEN-bit in {cheri_base_ext_name}.
+are now able to hold capabilities. Therefore, such registers are
+extended to CLEN+1 bits in {cheri_base_ext_name} and a `c` suffix is added such
+as <<mtvec>> is extended to <<mtvecc>>.
```
arichardson, line 334:
```suggestion
extended to CLEN+1 bits in {cheri_base_ext_name} and a `c` suffix is added (e.g. <<mtvec>> is extended to <<mtvecc>>).
```

PR #546, opened 2025-02-24
fix typo in rep range section

tariqkurd-repo:


arichardson (approved):
Addresses is technically fine, but I agree bounds is clearer.

src/cap-description.adoc
```diff
@@ -663,7 +663,7 @@ or 'root' capability.
 
 The new address, after updating the address of a capability, is within the
 _representable range_ if decompressing the capability's bounds with the
-original and new addresses yields the same base and top addresses.
+original and new addresses yields the same base and top bounds.
```
arichardson, line 666:
```suggestion
original and new addresses yields the same base and top bound.
```

I wonder if singular is better?

andresag01: I think its referring to 2 bounds, base and top, so I would say plural

PR #547, opened 2025-02-26
Clarify CHERI PTE fault priority when 2-stage translation is in use

tariqkurd-repo:
CHERI PTE faults sometimes require a tag check which is not available until after two stage translation has completed and so can not be prioritised above guest page faults, even if other page faults are detected before completing translation.

There are occasions when a PTE permission fault can be taken, but the 2-stage translation hasn't finished, and so a load cannot have executed yet (as the physical address isn't available yet), and so the tag cannot be checked. In this case the page fault will be taken without reporting a CHERI PTE fault. It makes sense to say the same for the store.

the order is

possible page fault due to permissions
finish two-stage translation (possible guest page fault)
possible CHERI PTE fault

so it's not always possible to detect both the _normal_ and the CHERI PTE page faults at once

This affects all versions of the spec


nwf:
Seems right to me.

Issue #548, opened 2025-02-27
Clarify what happens if "Reserved 1" bits are clear in the acperm mask


chrehrhardt:
The latest PR that introduced bits that are "Reserved 1" int the `acperm`/`gcperm` masks does not really explain what happends if one of these reserved bits is zero in the mask passed to `acperm`. At least two options are possible that I can think of:

- Make the bits WARL as most riscv registers do, i.e. trying to clear the bits is simply ignored on hardware that has them set to "Reserved 1".
- Clear the tag on the capability.

I'm not sure which version I'd prefer, though.

tariqkurd-repo:
I vote for making the reserved bits WARL, so it doesn't matter what the written value is

martin-kaiser:
If the software understands more permissions than the CPU, the software may set one of them to 0 where the default is 1.
The CPU sees an acperm instruction with a reserved1 bit set to 0. The bit is ignored anyway and does not change acperm's result.

I think this case should be supported, the reserved bits should be WARL.

jrtc27:
If software is asking to clear a bit that doesn’t exist then it is broken. Turning it into a nop could be dangerous (eg if software expects mutable load to exist and wants to clear it when delegating a capability to some other compartment, it should not silently degrade into handing out a capability that permits stores). It should tag clear because you’ve asked for something that is not supported.

chrehrhardt:
I actually tried to prepare software (linux kernel and musl) for this change under the assumption that the tag is cleared and it is really horrible. You would have to probe for supported hardware features not only in the kernel but also in musl on every program startup. In fact each and every software component, that is doing something non-trivial with permission bits that may or may not be supported by hardware would have to probe. Otherwise, clearing a permission bits you _do_ know about but that happens to be unsupported on the current hardware will invalidate your capability.

And the whole _point_ of this entire change is to ensure that the same binary can run on hardware with different features.

Thus, even though I started out being in @jrtc27 's camp I don't think it will work and I tend to agree with @arichardson that software that depends on an optional feature should check for it during runtime and refuse to even start if the feature is missing.

So if we make the bits WARL it encourages software to adopt new security features because you can just pretend that the feature is there and your software will benefit from the feature if it is supported and continue to run if it is not.
On the other hand if the tag is cleared there is no easy path to adopt a new security feature because you would either do probing and replace constant bitmask with variables or you would have to ship different binaries on hardware that has the new feature in order to make use of it, i.e. break backward compatibility.

[...]

PR #551, opened 2025-02-28
note about masks for reserved bits in the ACPERM mask

tariqkurd-repo:
Fixes https://github.com/riscv/riscv-cheri/issues/548

### DIFF
--- a/src/insns/acperm_32bit.adoc
+++ b/src/insns/acperm_32bit.adoc
@@ -23,17 +23,19 @@ Encoding::
 include::wavedrom/acperm.adoc[]
 
 Description::
-ACPERM performs the following operations:
+<<ACPERM>> performs the following operations:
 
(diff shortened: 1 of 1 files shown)


arichardson:
I think this is the sensible behaviour here. I can see the argument that requesting to clear something that doesn't exist should tag clear but I'm not sure that is useful for software. If your software depends on being able to clear the level bits you can just check that it was possible to actually clear it (once on program startup) using GCPERM+ACPERM+GCPERM rather than rely on implicit tag invalidation.

arichardson:
Would like to hear from @LawrenceEsswood / @jrtc27 / @nwf as well though.

jrtc27:
I disagree. This is an invalid request and ignoring it could silently degrade security. See my comment on the issue.

Issue #552, opened 2025-03-03
Implicit reads from PTW aren't checked by a capability


tariqkurd-repo:
Reads from the page-table-walker are not checked by a capability.
I don't believe that this is stated anywhere in the spec.

tariqkurd-repo:
or writes - if  A and D bits are updated in hardware

PR #553, opened 2025-03-05
Add note about PTW

tariqkurd-repo:
Fixes https://github.com/riscv/riscv-cheri/issues/552


arichardson:
All of these loads happen outside of the checked scope (they read physical addresses directly), so adding this clarification seems fine. In the future I could imagine a CHERI-aware PTE format but that is definitely a future enhancement.

Issue #558, opened 2025-03-13
What happens if PTE.CW or PTE.UCRG are set and CRE=0?


tariqkurd-repo:
If CRE is 0, so software is running in effecitvely "RISC-V mode" with PCC/DDC checking, then all CHERI features are _reserved_ i.e. encodings, CSRs like DDC etc.

We haven't (clearly) specified what happens if software sees non-zero PTE.CW/PTE.UCRG bits in these cases.

I'm guessing that they shoudl be handled as _reserved_ and so there'll be a page fault / access fault (whatever is normal for a bad PTE entry) but it's worth a discussion about this.

jrtc27:
CHERI *features* are not reserved. CHERI *state* is. The premise of this is false and making this change is inconsistent with the rest of RISC-V.

jrtc27:
> If we run a hypothetical "RVA2x" compliance test suite that checks that all reserved instructions trap and reserved bits in the PTE fault

No? To quote profiles.adoc:

> Extensions that are not explicitly listed in the mandatory or optional categories are termed non-profile extensions, and are not considered parts of the profile. Some non-profile extensions can be added to an implementation without conflicting with the mandatory or optional components of a profile. In this case, the implementation is still compatible with the profile even though additional non-profile extensions are present. Other non-profile extensions added to an implementation might alter or conflict with the behavior of the mandatory or optional extensions in a profile, in which case the implementation would not be compatible with the profile.
>
> NOTE: Extensions that are released after a given profile is released are by definition non-profile extensions. For example, mandatory or optional profile extensions for a new profile might be prototyped as non-profile extensions on an earlier profile.

It is completely valid to implement additional extensions. CHERI would be a set of non-profile extensions, and would be non-conflicting (so long as you have and default to integer mode). You do not need to trap on every aspect that isn't part of the profile itself.

arichardson:
Ah thanks, was not aware of that. In that case I'd say we can just keep these bits as valid unconditionally since in CRE=0 operation the value of them does not matter.

@tariqkurd-repo

arichardson:
Fixed in https://github.com/riscv/riscv-cheri/commit/cf51732e0bf6b2775516d168642f05dcca54652d

PR #559, opened 2025-03-13
If CRE=0, PTE bits are reserved

tariqkurd-repo:
Fixes https://github.com/riscv/riscv-cheri/issues/558

### DIFF
--- a/src/cheri-pte-ext.adoc
+++ b/src/cheri-pte-ext.adoc
@@ -28,7 +28,7 @@ The minimum level of PTE support is to set CW to 1 in all PTEs intended for stor
 
 If <<section_cheri_disable,CRE>>=0 then the PTE.CW, PTE.CRG bits become _reserved_ when checking PTE permissions on memory accesses.
 
-NOTE: Hardware initiated memory accesses from the page-table walker are not checked by a capability, and also do not consider PTE.CW, PTE.CRG bits to be _reserved_.
+NOTE: Hardware initiated memory accesses from the page-table walker are not checked by a capability.
 
 [#cheri_pte_fault]
 === CHERI PTE page faults
--- a/src/cheri-pte-ext.adoc
+++ b/src/cheri-pte-ext.adoc
@@ -26,7 +26,9 @@ NOTE: The description below assumes that {cheri_pte_ext_name} has been implement
 
 The minimum level of PTE support is to set CW to 1 in all PTEs intended for storing capabilities (i.e. private anonymous mappings) and leave <<sstatusreg_pte,sstatus>>.UCRG and CRG in all PTEs set to 0, which will allow capabilities with their tags set to be loaded and stored successfully.
 
-NOTE: Hardware initiated memory accesses from the page-table walker are not checked by a capability.
+If <<section_cheri_disable,CRE>>=0 then the PTE.CW, PTE.CRG bits become _reserved_ when checking PTE permissions on memory accesses.
+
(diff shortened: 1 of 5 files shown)


jrtc27:
Why? They’re not new state that needs gating. It’s a bit pointless to set but does no harm.

tariqkurd-repo:
> Why? They’re not new state that needs gating. It’s a bit pointless to set but does no harm.

certainly possible - it needs stating one way or another

tariqkurd-repo:
For RISC-V compliance, they are reserved bits when CRE=0 so should take a page fault.

src/insns/atomic_exceptions.adoc
```diff
@@ -37,5 +35,13 @@ reported in the CAUSE field of <<mtval2>>, <<stval2>> or <<vstval2>>:
 | Invalid address violation  | The effective address is invalid according to xref:section_invalid_addr_conv[xrefstyle=short]
 | Bounds violation      | At least one byte accessed is outside the authority capability bounds, or the capability has <<section_cap_malformed,malformed>> bounds
 |==============================================================================
-
++
+If virtual memory is enabled on an RV64 hart, then the state of <<cheri_pte_ext,PTE>>.CW,
+and,  if {cheri_pte_ext_name} is implemented, <<cheri_pte_ext,PTE>>.CRG from the current virtual memory page may
```
arichardson, line 40:
```suggestion
and, if {cheri_pte_ext_name} is implemented, <<cheri_pte_ext,PTE>>.CRG from the current virtual memory page may
```

src/cheri-pte-ext.adoc
```diff
@@ -26,7 +26,9 @@ NOTE: The description below assumes that {cheri_pte_ext_name} has been implement
 
 The minimum level of PTE support is to set CW to 1 in all PTEs intended for storing capabilities (i.e. private anonymous mappings) and leave <<sstatusreg_pte,sstatus>>.UCRG and CRG in all PTEs set to 0, which will allow capabilities with their tags set to be loaded and stored successfully.
 
-NOTE: Hardware initiated memory accesses from the page-table walker are not checked by a capability.
+If <<section_cheri_disable,CRE>>=0 then the PTE.CW, PTE.CRG bits become _reserved_ when checking PTE permissions on memory accesses.
+
+NOTE: Hardware initiated memory accesses from the page-table walker are not checked by a capability, and also do not consider PTE.CW, PTE.CRG bits to be _reserved_.
```
arichardson, line 31: These only make sense on leaf entries, so there wouldn't be any hardware-initiated fetches that check them? If I understand correctly, the reserved check would only happen when inserting into the TLB?

arichardson, line 31:
```suggestion
NOTE: Hardware initiated memory accesses from the page-table walker are not checked by a capability.
```

Restoring old wording, this behaviour is no different from normal PTE reserved bit checking.

arichardson (approved):

jrtc27:
> For RISC-V compliance, they are reserved bits when CRE=0 so should take a page fault.

No it's not. Svnapot, Svpbmt etc do not have enable bits. You either use them by configuring the PTEs to something non-zero or you don't. I don't see why we should be any different here. CRE is about _state enable_, but this is not new state, it is existing state that is given a meaning. Therefore it does not need to be gated behind an enable bit. Yes, it is a waste of time to configure the bits, but it does absolutely no harm, and nowhere in the RISC-V spec does it say every single new feature must trap unless you explicitly opt into it in order to be conformant. That is absolutely not something that holds.

jrtc27:
Please revert this change.

PR #562, opened 2025-03-13
Revert "If CRE=0, PTE bits are reserved"

arichardson:
As noted in https://github.com/riscv/riscv-cheri/issues/558 we can actually have these bits be non-reserved and still be compatible with all profiles.

Reverts riscv/riscv-cheri#559


tariqkurd-repo:
closing this one as the other parts of it (slight rewording) were needed, I've made a new PR to remove the PTE bits being reserved

PR #564, opened 2025-03-14
remove PTE bits being reserved

tariqkurd-repo:


arichardson:
Don't we also need to revert the changes to store_exception?

arichardson:
And the src/insns/load_exceptions.adoc change that says it's reverted as well.

tariqkurd-repo:
> And the src/insns/load_exceptions.adoc change that says it's reverted as well.

yeah - sorry - I've removed those.

the problem was that I put some extra stuff in the original PR - some tidy-up of the load-exceptions and the atomic-exceptions was completely missing the text about the PTE bits.

I think this one is correct now

arichardson (approved):

PR #566, opened 2025-03-20
Initial integration with ISA manual

arichardson:
Huge refactoring of the current specification. There should not be any functional changes compared to 0.9.5 - purely changes to the structure of the document and splitting things between privileged and unprivileged more cleanly.

Using the `isa-manual` branch as the base to avoid having all the history from riscv/riscv-isa-manual show up in this PR.


src/riscv-unpriv-integration.adoc
```diff
@@ -0,0 +1,388 @@
...
+
+CHERI enhances the base ISA to add hardware memory access control.
+It has an additional memory access mechanism that protects _references to code and data_ (pointers), rather than the _location of code and data_ (integer addresses).
+This mechanism is implemented by providing a new primitive, called a *capability*, that software components can use to implement strongly protected pointers within an address space.
+
+=== CHERI overview
+
+{cheri_base_ext_name} introduces a new data type, the _capability_, generally used for architectural pointers.
+Capabilities are unforgeable and delegatable tokens of authority that grant software the ability to perform a specific set of operations.
+In CHERI, integer-based pointers can be replaced by capabilities to provide memory access control.
```
tariqkurd-repo, line 14:
```suggestion
In CHERI, integer-based pointers are be replaced by capabilities to provide memory access control.
```

is this better? you can't dereference memory without one....

src/riscv-unpriv-integration.adoc
```diff
@@ -0,0 +1,388 @@
...
+CHERI capabilities consist of:
+
+* An XLEN-bit address, describing where the capability currently points.
+* XLEN bits of metadata:
+** Bounds: a _base_ and a _top_ address, describing the range of addresses the capability can be used to access.
+** Permissions (read, write, execute, ...) describing the kinds of operations the capability can be used for.
+** Sealing information: a capability can be _sealed_, restricting it to only be used or modified in particular ways.
+* A one-bit integrity tag (the `V` field in <<cap_structure>>): this is maintained by hardware and cannot be directly modified by software.
+It indicates whether the capability is valid.
+An initial <<infinite-cap>> capability with access to all of memory (with all permissions) is provided in system registers on reset: all valid capabilities are derived from it.
```
tariqkurd-repo, line 46: MTVECC, MEPCC both have infinity as the reset value
I think these are both mandatory.

src/riscv-unpriv-integration.adoc
```diff
@@ -0,0 +1,388 @@
...
+----
+
+CHERI capabilities consist of:
+
+* An XLEN-bit address, describing where the capability currently points.
+* XLEN bits of metadata:
+** Bounds: a _base_ and a _top_ address, describing the range of addresses the capability can be used to access.
+** Permissions (read, write, execute, ...) describing the kinds of operations the capability can be used for.
+** Sealing information: a capability can be _sealed_, restricting it to only be used or modified in particular ways.
+* A one-bit integrity tag (the `V` field in <<cap_structure>>): this is maintained by hardware and cannot be directly modified by software.
```
tariqkurd-repo, line 44: Krste hated the name "tag" before. here's it's integrity tag, below it's just tag.
`as the tag will be cleared.`

I think we need to choose a good name - maybe in a variable? `Validity tag` perhaps to go with V in the diagram?

src/riscv-unpriv-integration.adoc
```diff
@@ -0,0 +1,388 @@
...
+Developers can use CHERI to build fine-grained spatial and temporal memory
+protection into their system software and applications and significantly
+improve their security.
+
+==== Added State
+
+A CHERI core adds state to allow capabilities to be used from within registers, and to ensure they are not corrupted as they flow through the system.
+This means the following state is added by {cheri_base_ext_name}:
+
+* Extended <<gprs,general purpose integer registers>>: These XLEN-wide registers (e.g. `sp`, `a0`) are all extended with another XLEN bits of capability metadata plus a validity bit.
```
tariqkurd-repo, line 87: here the tag is `validity bit` - maybe `validity tag`?

src/riscv-unpriv-integration.adoc
```diff
@@ -0,0 +1,388 @@
...
+When referring to the extended register, the `x` prefix is replaced with a `c`: i.e. the capability extended version of `x0` becomes `c0`.
+The existing integer register names refer to the address part of the capability.
+For the ABI register names the a `c` prefix is added, i.e. `csp`, `ca0`.
+The zero register is extended with zero metadata and a cleared tag: this is called the <<null-cap>> capability.
+
+* Extended system registers: system registers that store addresses are extended to contain capabilities.
```
tariqkurd-repo, line 93: mixing "system registers" and "CSRs"
what's more consistent with the main doc?

src/riscv-unpriv-integration.adoc
```diff
@@ -0,0 +1,388 @@
...
+In the base {cheri_base_ext_name}, the only such register is the program counter `pc` that is extended to the program counter capability (<<pcc>>). Extending `pc` ensures that the range of branches and jumps for currently executing code can be restricted.
+Additionally, CSRs that hold pointers are extended to capability, but since no CSRs are defined in the RV32I/RV64I base instruction set, the individual chapters that define new CSRs state whether they are extended to CLEN bits.
```
tariqkurd-repo, line 95:
```suggestion
Additionally, all CSRs that hold pointers are extended to CLEN bits, but since no CSRs are defined in the RV32I/RV64I base instruction set, the individual chapters that define new CSRs state whether they are extended to CLEN bits.
```

src/riscv-unpriv-integration.adoc
```diff
@@ -0,0 +1,388 @@
...
+(def boxes-per-row 65)
+(draw-column-headers {:height 30 :font-size 25 :labels (concat ["" "2*XLEN (CLEN)"] (repeat 31 "") ["XLEN"] (repeat 30 "") ["0"])})
+
+(draw-box "" [:box-first {:span 1}]) (draw-box "c0" [:box-related {:span 32}]) (draw-box "x0" [:box-last {:span 32 }])
+(draw-box "" [:box-first {:span 1}]) (draw-box "c1" [:box-related {:span 32}]) (draw-box "x1" [:box-last {:span 32 }])
+(draw-gap "c2-c30 (extending x2-x30)")
+(draw-box "" [:box-first {:span 1}]) (draw-box "c31" [:box-related {:span 32}]) (draw-box "x31" [:box-last {:span 32 }])
+(draw-box nil [:box-first {:span 1}]) (draw-box "pcc" [:box-related {:span 32}]) (draw-box "pc" [:box-last {:span 32 }])
+(draw-box "tag" {:span 1 :borders {}}) (draw-box "metadata" {:span 32 :borders {}}) (draw-box "address" {:span 32 :borders {}})
+----
```
tariqkurd-repo, line 113: this diagram has `tag`, as does the next section

src/riscv-unpriv-integration.adoc
```diff
@@ -0,0 +1,388 @@
...
+
+* Tags in registers, caches, and memory:
+
+** Every register has a one-bit tag, indicating whether the capability in the register is valid to be dereferenced.
+This tag is cleared whenever an invalid capability operation is performed. Such invalid operations include e.g. writing only the integer portion of the register or attempting to increase bounds or permissions.
+
+** The tags are also tracked through the memory subsystem: every aligned CLEN-bits wide region has a non-addressable one-bit tag, which the hardware manages atomically with the data.
+The tag is cleared if the memory region is ever written using an operation other than a store of a full capability register.
```
tariqkurd-repo, line 121:
```suggestion
The tag is cleared if the memory region is ever written using an operation other than a store of a full capability register with appropriate permission.
```

tariqkurd-repo: do we need a comment here about single-copy-atomicity about tag+data in the memory system?

src/riscv-unpriv-integration.adoc
```diff
@@ -0,0 +1,388 @@
...
+
+* Capability manipulations (e.g. <<CADD>>, <<SCBNDS>>): for security, capabilities can only be modified in restricted ways.
+  Special instructions are provided to perform these allowed operations, for example _shrinking_ the bounds or _reducing_ the permissions.
+  Any attempt to manipulate capabilities without using the instructions clears the tag, rendering them unusable for accessing memory.
+
+* Capability inspection (e.g. <<GCBASE>>, <<GCPERM>>): capability fields (for example the _bounds_ describing what addresses the capability gives access to) are stored compressed in registers and memory.
+  These instructions give convenient access to allow software to query them.
+
+* Memory access instructions (e.g. <<LC>>, <<SC>>): capabilities must be read from and written to memory atomically along with their tag.
+  Instructions are added to perform these wider accesses, allowing capability flow between the memory and the register file.
```
tariqkurd-repo, line 137:
```suggestion
  Instructions are added to perform these wider accesses, allowing capabilities to flow between the memory and the register file.
```

src/riscv-unpriv-integration.adoc
```diff
@@ -0,0 +1,388 @@
...
+This is called executing in {cheri_cap_mode_name}.
+In {cheri_cap_mode_name}, instructions that access memory, as well as branches and jumps are modified to check against the full capability from the base address register (e.g. `lw t0, 16(c3)`) instead of just the integer address (i.e. `lw t0, 16(x3)`).
+In some cases, they are also changed to return a full capability value, e.g. <<AUIPC>> will return the full <<pcc>> including the metadata.
+For the full list of instructions that change behavior with {cheri_base_ext_name}, see <<app_cheri_instructions>>.
+
+The new CHERI state and instructions can also be made available to a hart using RV32I/RV64I as the base ISA with full binary compatibility.
+The mechanics for this mode are described by the <<section_cheri_hybrid_ext,{cheri_default_ext_name}>> extension.
+When a hart supports CHERI but is using RV32I/RV64I as the base ISA, it is executing in {cheri_int_mode_name}.
+For instructions that have differing behavior in {cheri_cap_mode_name} and {cheri_int_mode_name} (memory access and pointer arithmetic instructions), the descriptions will list assembler syntax and behavior for both modes.
```
tariqkurd-repo, line 151: I think I'd remove this sentence
maybe just write something like "Integer pointer mode allows full compatibility with existing RV32I/RV32E binaries, but with with all data memory accesses and program counters checked against capabilities"

src/riscv-unpriv-integration.adoc
```diff
@@ -0,0 +1,388 @@
...
+==== Capability checks
+
+Every memory access performed by a CHERI core must be authorized by a capability.
+The authorizing capability is either named explicitly (the base register of a load/store operation) or implicitly (when executing a branch, <<pcc>> is used for authorization).
+For example, `lw t0, 16(csp)` loads a word from memory, getting the address, bounds, and permissions from the `csp` register.
+
+Instruction fetch is also authorized by a capability: the program counter capability (<<pcc>>) which extends PC.
+This allows code fetch to be bounded, preventing a wide range of attacks that subvert control flow with integer data.
+
+With {cheri_base_ext_name} instruction fetches, jumps, and memory accesses may result in a fatal trap if the authorizing capability does grant the required permissions or the access is out of bounds.
```
tariqkurd-repo, line 162:
```suggestion
With {cheri_base_ext_name}, instruction fetches, jumps, and memory accesses may result in a fatal trap if the authorizing capability does grant the required permissions or the access is out of bounds.
```

src/riscv-unpriv-integration.adoc
```diff
@@ -0,0 +1,388 @@
...
+* All bytes of the instruction must be in bounds
+
+NOTE: Operations that update <<pcc>>, such as changing privilege or executing
+jump instructions, unseal capabilities prior to writing. Therefore,
+implementations do not need to check that that <<pcc>> is unsealed when
+executing each instruction. However, this property has not yet been formally verified and may not hold if additional CHERI extensions beyond {cheri_base_ext_name} are implemented.
+
+.Program Counter Capability
+[#pcc-format]
+include::img/pccreg.edn[]
```
tariqkurd-repo, line 193: this says `tag` in the box instead of `V`

src/riscv-unpriv-integration.adoc
```diff
@@ -0,0 +1,388 @@
...
+
+<<pcc>> is a code capability, so it does not need to be able to hold all possible invalid addresses (see <<section_invalid_addr_conv>>).
```
tariqkurd-repo, line 195: The xref says "invalid address conversion" but I think it should say "CHERI invalid address conversion"

src/riscv-unpriv-integration.adoc
```diff
@@ -0,0 +1,388 @@
...
+include::insns/gchi_32bit.adoc[]
+include::insns/sceq_32bit.adoc[]
+include::insns/scss_32bit.adoc[]
+include::insns/cram_32bit.adoc[]
+
+=== Capability Manipulation Instructions
+
+These instructions allow software to manipulate the fields of a capability held
+in a *c* register. The output is a capability written to a *c* register with
+its fields modified. The output capability has its tag set to 0 if the
```
tariqkurd-repo, line 223:
```suggestion
its fields potentially modified. The output capability has its tag set to 0 if the
```

cmv doesn't modify anything for example

src/riscv-unpriv-integration.adoc
```diff
@@ -0,0 +1,388 @@
...
+
+Capability load and store instructions also cause load or store/AMO address
+misaligned exceptions if the address is not naturally aligned to a CLEN
+boundary.
+
+Misaligned capability loads and stores are errors.  Implementations must
+generate exceptions for misaligned capability loads and stores even if they
+allow misaligned integer loads and stores to complete normally.  Execution
+environments must report misaligned capability loads and stores as errors
+and not attempt to emulate them using byte access.  The Zicclsm extension
```
tariqkurd-repo, line 268:
```suggestion
and not attempt to emulate them using byte accesses.  The Zicclsm extension
```

src/riscv-unpriv-integration.adoc
arichardson: This needs to go into the privspec, not needed here.

arichardson: reworded.

src/riscv-unpriv-integration.adoc
arichardson: Updated to use V

src/riscv-unpriv-integration.adoc
arichardson: CSRs is more common. used that

src/riscv-unpriv-integration.adoc
```diff
@@ -0,0 +1,388 @@
...
+include::insns/cram_32bit.adoc[]
+
+=== Capability Manipulation Instructions
+
+These instructions allow software to manipulate the fields of a capability held
+in a *c* register. The output is a capability written to a *c* register with
+its fields modified. The output capability has its tag set to 0 if the
+input capability did not have a tag set, the output capability has more
+permissions or larger bounds compared to the input capability, or the operation
+results in a capability with malformed bounds. These instructions do not give
```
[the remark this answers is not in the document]
arichardson: I don't think so, unless maybe loads allow this?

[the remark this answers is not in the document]
tariqkurd-repo: yes they do - I guess CBLD does too if you're trying to authorise a cap with malformed bounds (with the tag cleared of course)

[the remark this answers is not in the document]
PeterRugg: It is now I think! SCBNDS of length INT_MAX on a cap with cursor pointing at INT_MAX. It should already fail the bounds check though.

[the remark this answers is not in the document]
arichardson: But that SCBNDS will clear the tag?

src/cheri/attributes.adoc
```diff
@@ -14,6 +14,7 @@
 :cheri_priv_m_dyn_xlen_ext: Smcherivarxlen
 :cheri_priv_s_ext: Sscheri
 :cheri_priv_h_ext: Shcheri
+:cheri_default_cap: Zcheridefaultcap
```
arichardson, line 17: Do we actually need this? Can't we say this is the format mandated if you support RV32CH or RV64CH if you don't have another extension (like `zcheriot`) that redefines it?

src/cheri/insns/gchi_32bit.adoc
```diff
@@ -21,6 +21,9 @@ include::wavedrom/gchi.adoc[]
 Description::
 Copy the metadata (bits [CLEN-1:MXLEN]) of capability `cs1` into `rd`.
 
-Operation::
+Prerequisites::
+{cheri_base_ext_name}
```
arichardson, line 25: I removed the Prequisites part since it's in the RV32CH/RV64CH chapter so clearly defined by that extension, but if you think we need it can add it back.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -79,20 +54,240 @@ Developers can use CHERI to build fine-grained spatial and temporal memory
...
+NOTE: While these bits are not interpreted by the hardware, modification follows the same rules as architectural permissions: SDP bits can only be cleared and never set.
+This property is required to ensure restricted programs cannot forge capabilities that would pass the software-enforced checks.
+
+[#section_special_caps]
+==== Special Capabilities
+
+[#infinite-cap,reftext="Infinite"]
+===== Infinite Capability
+
+The _Infinite_ capability grants all permissions while its bounds also cover the whole address space.
```
marnovandermaas, line 275: Not sure what we discussed about this, but there are conceivable systems where there is no infinite capability like CHERIoT. Should we mark the existence of this infinite capability as optional?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -109,280 +304,352 @@ Additionally, CSRs that hold pointers are extended to capability, but since no C
...
+For the ABI register names the `c` prefix is added, i.e. `csp`, `ca0`.
+The zero register is extended with zero metadata and a cleared validity tag: this is called the <<null-cap>> capability.
+
+NOTE: Instruction encodings refer to capability registers when they use `cs1`, `cd`, etc. instead of `rs1`, `rd` (which reads/writes the XLEN-bit subset of the register).
 
-** Every register has a one-bit tag, indicating whether the capability in the register is valid to be dereferenced.
-This tag is cleared whenever an invalid capability operation is performed. Such invalid operations include e.g. writing only the integer portion of the register or attempting to increase bounds or permissions.
+==== Extended CSRs
+All registers that store addresses are extended to contain capabilities.
+In the base {cheri_base_ext_name} ISA, the only such register is the program counter `pc` that is extended to the Program Counter Capability (<<pcc>>).
```
marnovandermaas, line 321:
```suggestion
In the base {cheri_base_ext_name} ISA, only the program counter (`pc`) register is extended to become the Program Counter Capability (<<pcc>>).
``

File: src/cheri/riscv-unpriv-integration.adoc [hunk 21]
```diff
@@ -109,280 +304,352 @@ Additionally, CSRs that hold pointers are extended to capability, but since no C
...
+===== The Program Counter Capability (`pcc`)
 
-{cheri_base_ext_name} adds new instructions to interact with the extended registers.
-These added instructions can be split into the following categories:
+The `pc` is extended to be the Program Counter Capability.
+Extending the `pc` allows the range of branches, jumps and linear execution for currently executing code to be restricted.
+
+The <<pcc>> address field is the `pc` in the base RISC-V ISA so that the
+hardware automatically updates it as instructions are executed. The <<pcc>>'s
+metadata and validity tag are reset to the <<infinite-cap>> capability values,
```
marnovandermaas, line 333: Again this is not exactly true for all CHERI implementations. Could we be more vague here and say that the PC should be reset to a capability that can execute from anywhere in memory?

src/cheri/zish4add.adoc
```diff
@@ -0,0 +1,39 @@
+[#sh4add_ext]
+== "Zish4add" Extension for 16-byte pointer addition, Version 0.9
+
+ifdef::cheri_standalone_spec[]
+WARNING: This chapter will appear in the unpriv spec (see https://github.com/riscv/riscv-isa-manual/pull/1923).
+endif::[]
+
+The B-extension adds `SH[1|2|3]ADD` instructions, and for RV64 all have an unsigned word .UW version.
```
marnovandermaas, line 8:
```suggestion
The B-extension adds `SH[1|2|3]ADD` instructions, and for RV64 all have an unsigned word `*.UW` version.
```

arichardson: I took this from https://github.com/riscv/riscv-isa-manual/pull/1923, made the same suggestion there.

src/cheri/riscv-unpriv-integration.adoc
arichardson: I agree that it should be optional, but not sure this should be mentioned here. Maybe in the appendix?

src/cheri/riscv-unpriv-integration.adoc
arichardson: Updated.

src/cheri/riscv-unpriv-integration.adoc
arichardson: Added note in appendix.

nwf:
It appears that the notes added by #390 and #528 seem to have gotten lost in the refactoring.  I don't _believe_ those were deliberate decisions?

arichardson:
> It appears that the notes added by #390 and #528 seem to have gotten lost in the refactoring. I don't _believe_ those were deliberate decisions?

No that was not intentional, we should restore those. I'm not sure we will be able to include the AUIPCC note in the main ISA document, but it should definitely remain in the standalone doc

davidchisnall:
> > It appears that the notes added by #390 and #528 seem to have gotten lost in the refactoring. I don't _believe_ those were deliberate decisions?
> 
> No that was not intentional, we should restore those. I'm not sure we will be able to include the AUIPCC note in the main ISA document, but it should definitely remain in the standalone doc

Now that it is a separate base, is there a reason to not reduce the shift by one for everyone? Having an overlap enables capability representations that don’t allow out-of-bounds representations, which is probably useful for GPUs and ML accelerators, as well as CHERIoT. For the bigger systems, the cost is zero: not being able to have a PCC bigger than 2 GiB isn’t a problem even for the largest codebase I’m aware of, which is vastly larger than anyone sensible uses (and the opposite of what compartmentalisation will encourage).

PR #567, opened 2025-03-24
rep range clarifications

tariqkurd-repo:
@PeterRugg 

note that most of this was written by @jamie-melling


tariqkurd-repo:
Yeah, step 1 get the rationale correct, step 2 restructure the doc as part of moving to the RISC-V spec

I'd like to merge soon so we don't have pending changes when migrating text

arichardson:
I am moving this section to an appendix in the restructure, so I think having a bit of rationale there should be fine.

PR #576, opened 2025-04-11
Clarify CHERI page fault dependence on PTE.U

### LINKED ISSUE #575: Clarify Zcheripte
In the spec for PTEs [here](https://riscv.github.io/riscv-cheri/#section_extending_pte) it says:

> A [CHERI load PTE page fault](https://riscv.github.io/riscv-cheri/#cheri_pte_fault) exception is raised when a capability load or AMO instruction is executed with [C-permission](https://riscv.github.io/riscv-cheri/#c_perm) granted and the virtual page’s CRG bit does not equal [sstatus](https://riscv.github.io/riscv-cheri/#sstatusreg_pte).UCRG **in user mode**.

It's not totally clear what "in user mode" means. Is it:

* When `pte[U]` is set?
* When the effective privilege for the access use User?
* When the actual current privilege is User?

I think the first two are equivalent but the third isn't because you can be in machine mode or supervisor mode and access a page with effective User mode and `pte[U]` set due to `mstatus[MPRV]` and `mstatus[SUM]`.

In other words, if you are in Machine/Supervisor mode but your effective privilege is User mode, should it raise a CHERI page fault?

Timmmm:
It is possible to be in machine or supervisor mode and still read pages with PTE.U set, because of mstatus.MPRV and mstatus.SUM.

The table makes clear the condition is PTE.U not the current privilege, so this updates the text to match it.

I also removed "optionally" from the second scheme because the optionality is why there are two schemes in the first place. I removed the footnote link from there too because that footnote isn't specific to the second scheme. Arguably it shouldn't be a footnote at all.

Fixes #575


arichardson:
Thanks, this LGTM, can you rebase on top of the latest version?

arichardson (approved):

PR #584, opened 2025-04-23
ARC feedback on unpriv

tariqkurd-repo:
rename all operands from C to Y
rename CLEN to YLEN
rename validity tag to valid tag

some other minor improvements

more feedback fixing to go


jrtc27:
We discussed this internally on Slack yesterday and believe this recommendation from ARC is a terrible idea. Extensions do not consistently use their name as a prefix for all their instructions, and the name of the data type is not inherently tied to an arbitrary letter chosen for the extension / base ISA. X registers are xN not iN/eN, and they're operated on with instructions that don't all start i/e. Similarly bitmanip instructions don't all start with b, nor do scalar crypto instructions all start with a k.

tariqkurd-repo:
> We discussed this internally on Slack yesterday and believe this recommendation from ARC is a terrible idea. Extensions do not consistently use their name as a prefix for all their instructions, and the name of the data type is not inherently tied to an arbitrary letter chosen for the extension / base ISA. X registers are xN not iN/eN, and they're operated on with instructions that don't all start i/e. Similarly bitmanip instructions don't all start with b, nor do scalar crypto instructions all start with a k.

I haven't renamed any instructions yet, and won't until we have agreement.

All vector begin with V, and apparently all P-extension begin with P.

It was an explicit request to change the name - so you'll have to discuss with ARC.

jrtc27:
> > We discussed this internally on Slack yesterday and believe this recommendation from ARC is a terrible idea. Extensions do not consistently use their name as a prefix for all their instructions, and the name of the data type is not inherently tied to an arbitrary letter chosen for the extension / base ISA. X registers are xN not iN/eN, and they're operated on with instructions that don't all start i/e. Similarly bitmanip instructions don't all start with b, nor do scalar crypto instructions all start with a k.
> 
> I haven't renamed any instructions yet, and won't until we have agreement.
> 
> All vector begin with V, and apparently all P-extension begin with P.

And that's fine because the extension names were chosen very early on when there was free choice to align with the names of the actual types. Y does not have much to do with Capability beyond being a letter somewhere in it, so the situation is quite different.

> It was an explicit request to change the names - so you'll have to discuss with ARC.
> 
> Going through the review was never going to be easy.....

ARC's perspective is welcome, but we shouldn't just accept everything they say without properly considering its implications and whether we agree with it. ARC have opinions on how things should fit into RISC-V, which is valuable, but they don't have the same experience that we do with CHERI, and those are sometimes going to be in conflict.

jrtc27:
My counterproposal is:

Base ISAs (currently) need to be single-letter. For CHERI with 32 general-purpose registers that arbitrary letter choice is Y. For CHERI with 16 general-purpose registers that arbitrary letter choice is something else. Or it's the strange YI + YE set. But, either way, that is all that it is, just a name for the extension. The data type is still a capability, so the registers still start with C, just as vectors start with V, floats start with F and packed SIMD registers start with P. The instructions operate on the data type, so start with a C, if anything (float instructions start with an F, but integer ones tend to do whatever they want, so either approach is fine here).

There's also been zero consideration given so far for G. Maybe it's obsoleted by profiles, but there isn't currently a proposal for how to describe a CHERI-extended IMAFD. That could fit in as YI + YE + YG, or we could just not care and say that it's not very useful, between profiles existing and the proliferation of multi-letter extensions for actually-useful systems.

tariqkurd-repo:
> My counterproposal is:
> 
> Base ISAs (currently) need to be single-letter. For CHERI with 32 general-purpose registers that arbitrary letter choice is Y. For CHERI with 16 general-purpose registers that arbitrary letter choice is something else. Or it's the strange YI + YE set. But, either way, that is all that it is, just a name for the extension. The data type is still a capability, so the registers still start with C, just as vectors start with V, floats start with F and packed SIMD registers start with P. The instructions operate on the data type, so start with a C, if anything (float instructions start with an F, but integer ones tend to do whatever they want, so either approach is fine here).
> 
Ok - so registers are still C (csp etc), and what about instruction names?

> There's also been zero consideration given so far for G. Maybe it's obsoleted by profiles, but there isn't currently a proposal for how to describe a CHERI-extended IMAFD. That could fit in as YI + YE + YG, or we could just not care and say that it's not very useful, between profiles existing and the proliferation of multi-letter extensions for actually-useful systems.

I was thinking that CHERI+IMAFD would be YG or YIMAFD, so whatever you havenow gets prefixed with a Y

And we we definitely need a CHERI profile.

arichardson:
I'd suggest we start out with splitting the naming changes into a separate PR and have this be the initial cleanup changes? I can give this a go later today.

tariqkurd-repo:
> I'd suggest we start out with splitting the naming changes into a separate PR and have this be the initial cleanup changes? I can give this a go later today.

yes, this makes sense

tariqkurd-repo:
ok - so I've kept the new architecture name and undone the YLEN, Ys1 etc. naming
Everything else is uncontroversial, but there are a lot of changes (mainly validity tag -> valid tag).
So I'll get this one merged and move on with the uncontraversial fixes

Issue #582, opened 2025-04-23
Include Zstid in the base architecture


tariqkurd-repo:
The hardware cost and complexity of Zstid is very low - only 3 CSRs.

However, it is difficult to justify the presence of ASR permission in the unpriv specification without it, because it's not needed for any user-mode accessible CSRs _except_ for utidc.

The ASR permission bit _is_ visible in user mode due to ACPERM and GCHI, but can't be used.

Given that we must include a compartmentalisation scheme in the spec, and it has the slightly odd side effect of allowing ASR-permission to be described in the unpriv spec (which will make the spec simpler), can anyone see a good reason _not_ to mandate it?

Including it doesn't prevent a new improved compartmentalisation scheme being added in future.

This was sticking point during the review with Krste.

tariqkurd-repo:
Does cheriot support user-mode? if it only supports M-mode then it's only one CSR
But yes it does also gate access to other CSRs, but in a very consistent way.

PR #588, opened 2025-04-25
Making zstid part of the main CHERI extension

### LINKED ISSUE #578: Zstid in the (new) privileged spec
I could spot the xTIDC registers in the privileged spec in the respective CSR lists, but not a description of them. Was that intentionally left out?

francislaus:
This integrates the xTIDC registers in the main CHERI extension.

This should close #578 and #582.


arichardson (commented):
Do we want to add a note that these registers are WARL so that tiny implementations that don't need them can make them zero-width? And the CHERI profile needs to mandate them as being full width? @nwf @davidchisnall I believe you didn't want these registers to be mandatory, would WARL (just return zeroes in your implementation) address that concern?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -308,6 +308,49 @@ The zero register is extended with zero metadata and a cleared valid tag: this i
 
 NOTE: Instruction encodings refer to *c* registers when they use `cs1`, `cd`, etc. instead of `rs1`, `rd` (which reads/writes the XLEN-bit subset of the register).
 
+==== Added CSRs
+
+{cheri_base32_ext_name} and {cheri_base64_ext_name} add the capability CSR shown in
```
arichardson, line 313: Could you add it to [#base_cap_registers]?

arichardson (commented):
Do we want to add a note to the ASR-perm section that this affects access to the utidc registers?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -308,6 +308,49 @@ The zero register is extended with zero metadata and a cleared valid tag: this i
...
+identifying the current software thread. While identifying the current software thread
+can be done by privileged code, e.g., the kernel, the implied performance
+overhead of this is not bearable for CHERI systems with many compartments.
+
+The RISC-V ABI includes a _thread pointer (tp)_ register, which is not
+usable for the purpose of reliably identifying the current software thread because
+the tp register is a general purpose register and can be changed arbitrarily
+by untrusted code. Therefore, this specification offers an additional CSR
+that facilitates a trusted source for the thread ID. The register is readable
+from U mode and writeable with <<asr_perm>>.
```
arichardson, line 352: I think all of this needs to be a non-normative note?

arichardson (commented):
Thanks for this, just a few comments.

nwf:
I think making them WARL and optionally unimplemented would be welcome, thank you.

arichardson:
Some further background here: having utidc in the base extension means we can describe ASR permission in the base extension, since otherwise it is completely irrelevant to unpriv and would have to be an enhancement of the capability format for the priv part.

We could certainly do this but IMO this adds a lot of complexity for the spec compared to just always having utidc but making it WARL.

arichardson:
> We had a discussion within the Cambridge architecture group. We currently believe that xTIDC registers need to be capability wide. After discussing different models, we concluded that using xTIDC only as an integer would bring two disadvantages:
> 
> 1. The need to handle concurrency
> 2. Performance slow down compared to the capability wide variant.
> 
> Having WARL wide xTIDC registers additionally means that legalisation needs to take place, which includes a representability check costing additional resources.
> 
> If we mandate capability width via a CHERI profile, then everybody not implementing that profile likely needs to come up with their own compartmentalisation primitives support.
> 
> We might need to be careful about avoiding that someone can claim to "implement CHERI", but not implementing the "CHERI profile" or the "CHERIoT" profile

I agree that you need capability width registers, the "WARL" suggestion was just to allow either readonly always zero (CHERIoT) or full register width.

src/cheri/riscv-unpriv-integration.adoc
francislaus: How would you want to fit it into the diagram? I am under the impression that we do not want to specify anything the actual integer register (utid), but only the capability register (utidc) and then make it WARL.

Would you want utid, then immediately extend it to utidc, and then make that WARL?

francislaus:
@arichardson @tariqkurd-repo I think this is ready for another round of review. I left out all the WARL related stuff as I understand that there is no need for that.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -375,6 +389,42 @@ The zero register is extended with zero metadata and a cleared valid tag: this i
...
+The <<utidc>> register is used to identify the current software thread in user mode.
+On reset the valid tag of <<utidc>> will be set to 0 and the remainder of the data is UNSPECIFIED.
+
+
+NOTE: Compartmentalization seeks to separate the privileges between different
+protection units, e.g., two or more libraries. Code can be separated by sentries,
+which allow for giving out code capabilities to untrusted code where the untrusted
+code can only call the code capability, but not modify it. Sentries can
+be called from different software threads and thus there needs to be a way of
+identifying the current software thread. While identifying the current software thread
```
tariqkurd-repo, line 417: why is there a need to identify the software thread, and what is done with that information?

I think we need a bit more usage information here

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -375,6 +389,63 @@ The zero register is extended with zero metadata and a cleared valid tag: this i
...
+|===
+|CLEN CSR|Address|Permissions|Description
+|<<utidc>>|0x480|RW, <<asr_perm>> required for writes, not reads|User Default Data Capability
+|===
+
+[#utidc,reftext="utidc"]
+===== User Thread Identifier Capability (utidc)
+
+The <<utidc>> register is used to identify the current software thread in user mode.
+On reset the valid tag of <<utidc>> will be set to 0 and the remainder of the data is UNSPECIFIED.
```
arichardson, line 409: We can't talk about reset in unpriv part of spec. And we also can't mandate what the priv execution environment sets this to, so best to drop it.
```suggestion
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -375,6 +389,63 @@ The zero register is extended with zero metadata and a cleared valid tag: this i
...
+code can only call the code capability, but not modify it. The <<utidc>> register supports
+a model where untrusted code is separated by trusted code and each call from one piece of untrusted
+code to another piece of untrusted piece of code goes through trusted code. Often, the trusted
+code is referred to as a *trampoline*. Sentries can
+be called from different software threads and thus there needs to be a way of
+identifying the current software thread. While identifying the current software thread
+can be done by privileged code, e.g., the kernel, the implied performance
+overhead of this is not bearable for CHERI systems with many compartments.
+
+NOTE: The <<utidc>> register is designed to hold a capability, which can only be accessed by
```
arichardson, line 424:
```suggestion
NOTE: The <<utidc>> register is designed to hold a capability, which can only be written by
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -375,6 +389,63 @@ The zero register is extended with zero metadata and a cleared valid tag: this i
...
+trusted code. In a commonly used model on CHERI systems, the trusted code's responsibility
+is only to switch between compartments, but not to switch threads. This responsibility is usually
+taken over by more privileged code, e.g., an operating system kernel running on a different privilege level.
+The privileged code switches software threads and writes the <<utidc>> register.
+
+NOTE: Every piece of code in the user space (and more privileged levels) can read the content of the <<utidc>>
```
arichardson, line 430: This is a lot of software guidance that is useful but probably shouldn't be part of this chapter.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -375,6 +389,63 @@ The zero register is extended with zero metadata and a cleared valid tag: this i
...
+register. However, the memory authorized by the capability in <<utidc>> must not be accessible to untrusted code,
+but only to trusted code. In order to protect this capability, it can be made a sentry (thus being sealed). The
+trusted code will be given a capability large enough so that it can recover the sentry stored in <<utidc>>.
+For the untrusted code, the sentry stored in <<utidc>> is inaccessible. The sentry itself is no secret, but the
+memory pointed to by the sentry is a secret and must not be accessed by any untrusted code.
+
+NOTE: Trusted code can use the sentry in <<utidc>> to implement secure compartment switches. Often, the sentry is
```
arichardson, line 437: Similarly this note is probably a better fit for a programmers guide

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -375,6 +389,63 @@ The zero register is extended with zero metadata and a cleared valid tag: this i
...
+used to implement a trusted stack. Whenever a compartment switch happens, the trusted code can pass arguments between
+the caller and callee compartment avoid capability leaks between the two compartments.
+The trusted code can store capabilities on the trusted stack when calling out of a compartment and can install them
+when returning the same compartment.
+
+NOTE: The RISC-V ABI includes a _thread pointer (tp)_ register, which is not
```
arichardson, line 443: We may be able to get away with just this last note and moving the remaining text to another document?

ARC doesn't want too much rationale in the main chapters so maybe as a supporting appendix or a separate document?

arichardson (approved):
I think we need to move some of the more detailed notes out of the main chapter but otherwise LGTM

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -375,6 +389,63 @@ The zero register is extended with zero metadata and a cleared valid tag: this i
...
+
+{cheri_base32_ext_name} and {cheri_base64_ext_name} add the capability CSR shown in
+xref:utidc-csrnames-added[xrefstyle=short].
+
+[[utidc-csrnames-added]]
+.Unprivileged capability CSRs added in {cheri_base32_ext_name} and {cheri_base64_ext_name}
+[%autowidth,float="center",align="center",cols="<,<,<,<",options="header"]
+|===
+|CLEN CSR|Address|Permissions|Description
+|<<utidc>>|0x480|RW, <<asr_perm>> required for writes, not reads|User Default Data Capability
```
arichardson, line 402: I don't think any other csr numbers are listed in the unpriv chapter, can probably drop this.

arichardson, line 402:
```suggestion
|<<utidc>>|0x480|RW, <<asr_perm>> required for writes, not reads|User Thread ID Capability
```

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: I do mention system init in unpriv as we have to talk about a source of infinite-cap.
But I think you're right - maybe the reset value guidance should be in priv (annoyingly)

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: I think that it needs to be in this document - we need to be able to give enough information so people can at least get a good idea of how to use the features - otherwise they won't know how to review it. It can go into an appendix certainly, it doesn't need to be part of the main spec.

Or.... we develop a programmers guide upfront and also submit that for ARC review.

The problem that we have is that it is very far from obvious how these features are actually used, and so they are very difficult to review. PTE.UCRG will have the same problem.

Once we get through ARC review then we still need to get signoff from the security HC, unpriv/priv HC, probably others, and then get to public review and the questions will be raised repeatedly about how these features are actually used.

tariqkurd-repo:
We have a problem - we can't talk about CSRs in this chapter because it's before the Zicsr chapter.
When we talk about CSRs we can _only_ talk about utid[c].
mtidc etc. must be in the priv spec

One solution is to move the RVxY chapter to _after_ Zicsr, and then we can talk about it.
This has the additional benefit of not having to change the Zicsr chapter at all, we can put all such details in the RVxY chapter.

francislaus:
I think it is not clean either way. The way I understand the unpriv spec is that base architectures come at the beginning and then come all the extensions. When we switch Zicsr and RVxxY then this not beautiful either.

The problem is that we are picking up a dependency on Zicsr with RVxxY, which is not beautiful given that it should be a base architecture.

That all being, I am not conclusive and what we should do. I believe reordering is the best thing I can currently think about even though I looks not so nice.

francislaus:
@arichardson @tariqkurd-repo I am also not fully decided on where the software facing notes about utidc should go. I am more torn to making a programmer's guide or something along these lines. I think such a guide could also be a good place to explain capability levels and other concepts. What are your thoughts?

Issue #577, opened 2025-04-14
Does the "implicit ACPERM" in LC do the full instruction semantics?


tomaird:
The description for `LC` (amongst others) mentions doing an "implicit ACPERM" to clear certain permissions of the loaded capability. e.g.:

> If the authorizing capability does not grant LM-permission, and the tag of cd is 1 and cd is not sealed, then an implicit ACPERM clearing W-permission and LM-permission is performed to obtain the intermediate permissions on cd.

What is not clear is if "implicit ACPERM" means to just do the permission clearing part of the ACPERM semantics, or apply the entire instruction.

The ACPERM description has the following steps:

![Image](https://github.com/user-attachments/assets/01d4768c-4de7-4ba7-9210-4cb9525c139c)

I think it's clear that steps 1-6 should be applied to clear the relevant permissions. But it's not clear if step 7 should also be applied as part of this process.

If so, this would mean that a loaded capability which is sealed or has non-zero reserved bits should have its tag cleared, but only if certain unrelated permission bits are cleared.

tariqkurd-repo:
If a sealed cap is loaded with write permission, which then can't be stripped as rule 7 isn't followed, then that probably isn't right?
Although the sealed cap can't be derefenced.
@arichardson @jrtc27 ? opinions?

tomaird:
Following internal discussions, we've agreed the "implicit ACPERM" should do the full instruction steps. Which is probably the most logical conclusion anyway.

However to make this clearer, I propose adding something like this to the `ACPERM` description:

```adoc
NOTE: Whenever an _implicit_ ACPERM is done (e.g. when clearing LM-permission), the full ACPERM instruction semantics are done, not just the permission clearing parts.
```

nwf:
In CHERIoT, `lc` does modulate permissions but is not a "full" `acperm`.  In particular, [`W` (AKA `SD`), `LM`, and `LG` permissions are "under seal" but the `GL`obal permission is not](https://github.com/CHERIoT-Platform/cheriot-sail/issues/14) and, as a consequence, while [`acperm` is allowed to clear `GL` on sealed capabilities if it clears no other bits](https://github.com/CHERIoT-Platform/cheriot-sail/issues/70) as if they had gone to memory and come back through a `LG`-absent authority, loading a sealed capability through a `LG`-bearing but `LM`-absent authority will leave the load result unaltered (that is, the capability loaded to the register file will be exactly the same as was in memory).

[...]

Issue #513, opened 2025-01-30
ACPERM and levels


LawrenceEsswood:
The Zcherilevels section states the capability level should be modified by ACPERM. However, ACPERM currently always tag clears if values are sealed. This is somewhat contradictory with how the implicit ACPERM happens on loads and stores, where the spec says:

"The Capability Level (CL) field can be reduced even if the capability is sealed, see Table 31."

I think we should bring these two in line and have ACPERM just be aware that some bits are not protected by sealing.

Currently the description of the last step of ACPERM is:

`Set cd.tag=0 if cs1 is sealed or if any reserved fields of cs1 are set.`

I propose we change this to:

`Set cd.tag=0 if cs1 is sealed _and any protected permissions have changed_ or if any reserved fields of cs1 are set.`

Without Zcherilevels all permissions are "protected". Level is unprotected (and we can extend the definition as more fields appear).

If we do have Zcherilevels, it is now possible to make a sealed capability local without having to store/load it back. For instance "acperm out, in, ~GLOBAL".

More importantly, I think this should be in the base spec as well so that binaries can be portable between platforms that do and do not have levels ("acperm out, in, ~GLOBAL" would become a move on such platforms, rather than an untag).

jrtc27:
I think this just comes back to what I was saying the other day with Alex, levels aren't permissions so shouldn't be shoehorned into ACPERM, just as we don't try to do the same with M.

LawrenceEsswood:
Conceptually I agree. But we do want to both burn instruction space, and require longer programs (as we expect changing level + EL concurrently to be an expected operation) just because it seems neater philosophically? Practically, ACPERM does fit the bill with a small tweak.

nwf:
FWIW, CHERIoT explicitly permits clearing G on sealed caps via its `candperm`; see https://github.com/CHERIoT-Platform/cheriot-sail/pull/83

arichardson:
I think we should either make this change to permit ACPERM on sealed caps if only the level was changed or introduce another instruction to get and restrict levels. But that is rather invasive and would need more design work, so IMO using the CHERIoT solution would be best.

LawrenceEsswood:
Worth noting that CheriOT took a subtly different direction than what I proposed in my original post. They went with any attempted change clears tag (i.e. based only on the mask), rather than the change actually happening (i.e. based on comparing new and old permissions)

@nwf was the choice arbitrary, or is there a good reason to go the CheriOT direction?

[...]

PR #590, opened 2025-04-25
clarify implicit ACPERM behaviour

tariqkurd-repo:
Fix https://github.com/riscv/riscv-cheri/issues/577

### DIFF
--- a/src/cheri/insns/acperm_32bit.adoc
+++ b/src/cheri/insns/acperm_32bit.adoc
@@ -36,8 +36,8 @@ IMPORTANT: Depending on the base ISA and supported extensions, some combinations
 In case the requested permissions combination falls into this category, <<ACPERM>> will return useful minimal sets of permissions, which may be no permissions.
 Therefore, it is possible that requesting to clear a permission also clears others, but <<ACPERM>> will never _add_ new permissions.
 . Copy `cs1` to `cd`, and update the AP and SDP fields with the newly calculated versions.
-. Set `cd.tag=0` if `cs1` is sealed only if <<ACPERM>> has been issued as an instruction.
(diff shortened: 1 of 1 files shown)


src/cheri/insns/acperm_32bit.adoc
```diff
@@ -36,7 +36,8 @@ IMPORTANT: Depending on the base ISA and supported extensions, some combinations
 In case the requested permissions combination falls into this category, <<ACPERM>> will return useful minimal sets of permissions, which may be no permissions.
 Therefore, it is possible that requesting to clear a permission also clears others, but <<ACPERM>> will never _add_ new permissions.
 . Copy `cs1` to `cd`, and update the AP and SDP fields with the newly calculated versions.
-. Set `cd.tag=0` if `cs1` is sealed.
+. Set `cd.tag=0` if `cs1` is sealed only if <<ACPERM>> has been issued as an instruction.
+.. Do not do the sealed check if <<ACPERM>> is _implicitly_ used to process data from <<LC>> or for any other reason.
```
arichardson, line 40: I think I'd prefer if we change the implicit ACPERMS notes in LC, since we're not really issuing an ACPERM instruction just following the same rules to derive the new encoded permissions set.

How about this wording:
```suggestion
. Set `cd.tag=0` if `cs1` is sealed and any permission bits were affected by <<ACPERM>>.
.. Do not invalidate the tag on sealed capability if only the <<section_cap_level,capability levels (CL)>> changed.
```

Kept the wording vague here to address #513, but still keeping it open whether affected means "final encoding changed" or "the mask could have resulted in a change".

arichardson (approved):

arichardson (approved):

Issue #597, opened 2025-05-02
amoswap.c needs to respect LC/SC rules


tariqkurd-repo:
The amoswap.c instruction page doesn't say that LC/SC rules are followed, so for example the stored tag will be cleared if cs1 doesn't have C permission.

(I'm pretty sure this was introduced by me recently - so it's a document bug not a spec bug)

PR #598, opened 2025-05-02
fix amoswap.c defn

tariqkurd-repo:
fix https://github.com/riscv/riscv-cheri/issues/597

### DIFF
--- a/src/cheri/insns/amoswap_32bit_cap.adoc
+++ b/src/cheri/insns/amoswap_32bit_cap.adoc
@@ -15,8 +15,14 @@ Mnemonic::
 Encoding::
 include::wavedrom/amoswap_cap.adoc[]
 
-//{cheri_cap_mode_name} Description::
+Description::
 Atomic swap of capability type, authorized by the capability in `cs1`.
++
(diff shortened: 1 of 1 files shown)


src/cheri/insns/amoswap_32bit_cap.adoc
```diff
@@ -15,8 +15,14 @@ Mnemonic::
 Encoding::
 include::wavedrom/amoswap_cap.adoc[]
 
-//{cheri_cap_mode_name} Description::
+Description::
 Atomic swap of capability type, authorized by the capability in `cs1`.
++
+The operation is equivalent to an atomically executed sequence of:
++
```
arichardson, line 22: This seems the wrong way around, we return the previous value, so LC first.

arichardson (commented):
LGTM with the order switched and ideally a better commit message.

src/cheri/insns/amoswap_32bit_cap.adoc
tariqkurd-repo: yes good point!

PR #600, opened 2025-05-06
Lots of unpriv spec reviewing 

tariqkurd-repo:
A big objective was to remove all instructions which are already defined such as C.LB, so that CHERI only lists instructions which are modified in unexpected ways. We can't / won't list all loads and stores as being different.

It turned in to a big PR with lots of proofreading and xref fixing.


src/cheri/app-cheri-instructions.adoc
```diff
@@ -285,28 +302,28 @@ include::insns/zcmp_cmva01s.adoc[]
...
 
-The jump vector table CSR <<jvt>> has a capability alias <<jvtc>> so that it can only be configured to point to accessible memory. All accesses to the jump table are checked against <<jvtc>> in {cheri_cap_mode_name}, and against <<pcc>> bounds in {cheri_int_mode_name}. This allows the jump table to be accessed
+The jump vector table CSR <<jvt_cheri>> has a capability alias <<jvtc>> so that it can only be configured to point to accessible memory. All accesses to the jump table are checked against <<jvtc>> in {cheri_cap_mode_name}, and against <<pcc>> bounds in {cheri_int_mode_name}. This allows the jump table to be accessed
 when the <<pcc>> bounds are set narrowly to the local function only in {cheri_cap_mode_name}.
 
 NOTE: Zcmt defines that the fetch from the jump table is from instruction memory.
- The overall instruction executed is effectively 48-bit, with 16-bits from <<CM.JALT>>/<<CM.JT>>, the other 32-bits (for RV32) from the table.
+ The overall instruction executed is effectively 48-bit, with 16-bits from <<CM_JALT_CHERI>>/<<CM_JT_CHERI>>, the other 32-bits (for RV32) from the table.
```
nwf, line 313: Tangential, but...  I'm really not sure that's a useful phrasing, even if it is in some sense true.  In at least one reading, it suggests that the table is somehow involved in compositing a 48-bit _instruction_.  Moreover, the 16 bits of the `cm.jt` or `cm.jalt` proper come from within the bounds of `pcc` and the other 32 bits come from within the bounds of `jvtc`.

src/cheri/riscv-hybrid-integration.adoc
```diff
@@ -50,6 +52,14 @@ RISC-V software is able to execute unchanged in implementations supporting both
 
 NOTE: The CHERI execution mode is always {cheri_cap_mode_name} on implementations that support {cheri_base_ext_name}, but not {cheri_default_ext_name}.
 
+NOTE: {cheri_base_ext_name} implementations which support {cheri_default_ext_name} are typically referred to as CHERI Hybrid, whereas implementations which do not support {cheri_default_ext_name} are typically referred to as CHERI purecap.
```
arichardson, line 55:
```suggestion
NOTE: {cheri_base_ext_name} implementations which support {cheri_default_ext_name} are typically referred to as CHERI hybrid, whereas implementations which do not support {cheri_default_ext_name} are typically referred to as CHERI purecap.
```

Either both lower or both capitalized?

src/cheri/riscv-hybrid-integration.adoc
```diff
@@ -25,19 +25,21 @@ The two execution modes are:
...
 +
 IMPORTANT: All instructions from existing RISC-V extensions can be run in {cheri_int_mode_name}.
-  This execution mode is fully binary compatible with a non-CHERI-aware programs.
+  This execution mode is fully binary compatible with non-CHERI-aware programs.
 
 {cheri_cap_mode_name}::
 Executing with the {cheri_base_ext_name} base ISA.
-Some instructions in the RV32I/RV64I base ISA (plus any extensions) use capabilities instead of integers for memory references (see xref:section_existing_riscv_insns[xrefstyle=short]).
++
+Some instructions in the RV32I/RV64I base ISA (and in some other extensions) use capabilities instead of integers for memory references (see xref:section_existing_riscv_insns[xrefstyle=short]). In these cases, such as <<SH1ADD_CHERI>>, the lower XLEN bits of the result is the same in either mode and the upper XLEN bits, and valid tag, are either handled as capabilities in {cheri_cap_mode_name}, or are ignored in the source and written to zero in {cheri_int_mode_name}
```
arichardson, line 37:
```suggestion
Some instructions in the RV32I/RV64I base ISA (and in some other extensions) use capabilities instead of integers for memory references (see xref:section_existing_riscv_insns[xrefstyle=short]).
In these cases, such as <<SH1ADD_CHERI>>, the lower XLEN bits of the result is the same in either mode and the upper XLEN bits, and valid tag, are either handled as capabilities in {cheri_cap_mode_name}, or are ignored in the source and written to zero in {cheri_int_mode_name}
```

Having one sentence per line makes (future) changes easier to review

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -616,6 +616,8 @@ In the base ISA this affects <<AUIPC_CHERI>> and all jumps and branches as well
 
 The rules for modifying the behavior of such instructions are:
 
+* The result written to the target register is always unchanged.
+** The lower XLEN bits are always written with the value from RV32I/RV64I.
```
arichardson, line 620:
```suggestion
* The lower XLEN bits of the result written to the target register are always the same as the value from RV32I/RV64I.
```

The first bullet is not quite true since there are extra bits that are written, how about just merging them?

src/cheri/insns/load_tag_perms.adoc
```diff
@@ -7,7 +7,7 @@ If `cd.tag=1`, `cd` is not sealed and `cs1` does not grant <<el_perm>>, then an
 +
 If `cd.tag=1`, `cd` is sealed and `cs1` does not grant <<el_perm>>, then an implicit <<ACPERM>> is performed restricting the <<section_cap_level>> of `cd` to the level of `cs1`
 +
-If {cheri_1levels_ext_name} is implemented, then also see <<cap_level_load_summary>>.
```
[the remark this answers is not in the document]
tariqkurd-repo: whoops! fixed.

src/cheri/app-cheri-instructions.adoc
tariqkurd-repo: I've tried to improve the comment in the spec

Issue #628, opened 2025-06-12
Split IFC/levels back out of the base


nwf:
I think we concluded in today's meeting that #611 / #568 should be unwound.  Concretely, I think that means...

- reserving, rather than defining, the presently-labeled `CL`, `SL`, and `EL` bits in the `acperm`/`gcperm` permissions bit field (figures 7 and 8).

- replacing the `EL`-referencing prose in the `LC` and `LR.C` instructions with something like

  > Extensions may define circumstances under which the successfully loaded result (now in `cd`) has a cleared tag or decreased permissions relative to the capability actually held in memory (at `offset(cs1)`), based on permissions carried by the authorizing `cs1` or other aspects of system state.

  (The latter bit is intended as a hook for CHERIoT's load filter.)

- similarly replacing the `SL`-referencing prose in the `SC` and `SC.C` instruction with something like

  > Extensions may define circumstances under which the capability successfully written to memory (at `offset(cs1)`) has a cleared tag or decreased permissions relative to the capability held in the source register (`cs2`), based on permissions carried by the authorizing `cs1`.

- replace the `CL` reference in the `SCSS` instruction prose with something like

  > if all other extensions in the current platform would leave `cs2` unmodified, were it to be loaded through an authority bearing the permissions of `cs1`

- Moving the replaced text, the definitions in "8.2.10. Architectural Permissions (AP)", all of "8.2.11. Capability Level (CL)" , and the reference to `CL` in "8.6.1. ACPERM" to a new specification.

- The reference to `CL` in the `CBLD` instruction prose also must move, but that probably follows `CBLD` over to its new extension (#627) and becomes a copy of what gets said for `SCSS`.

I'm not quite sure what to do about the permissions encoding in appendix D, but if that is formally an extension too, I suppose we can say something like "`SL` if Zylevels1 is supported by the platform" everywhere.

swm11:
Two levels (one bit of CL, SL and EL) has a great deal of utility for RV64Y and RV32Y and we might well end up with libraries or other code that is portable between RV64Y and RV32Y that depend on levels.  So to my mind it needs to be in the base spec. to avoid fragmenting all of the software that needs to be ported to CHERI.

PR #631, opened 2025-06-18
WIP: IFC to zylevels extensions (#628)

nwf:
This WIP PR is a first stab at #628.  At present, known deficiencies include:

- The encoding in the appendix is unmodified and so full of xref errors.
- Only Zylevels1 is well defined; ZylevelsN is not.  Please someone else (@LawrenceEsswood?) more versed in that idea contribute that section (or someone give me license to remove it).

But: feedback on approach, verbiage, &c more than welcome.  Attn especially @arichardson and @tariqkurd-repo.


nwf:
Erm, I appreciate fixes to whitespace, but please don't clobber my commit message?

src/cheri/zylevels1.adoc
```diff
@@ -0,0 +1,171 @@
...
+
+[#zylevels1_sl_perm,reftext="SL-permission"]
+Store Local Permission (SL):: This field allows limiting the propagation of _local_ capabilities.
++
+A capabilty without <<zylevels1_gl_perm>> set stored through an authorizing capability lacking <<zylevels1_sl_perm>> will be stored with a zero valid tag.
+
+<<zylevels1_sl_perm>> is a _refinement of_ <<c_perm>> and <<w_perm>>.
+That is, if either of the latter are clear, then <<zylevels1_sl_perm>> has no effect.
+The capability encoding might take advantage of this dependence for compression,
+and so clearing either <<c_perm>> or <<w_perm>> may result in
```
francislaus: I stumbled upon the "may result in". Would the spec need to be precise on whether SL will be cleared or not?

src/cheri/zylevels1.adoc
```diff
@@ -0,0 +1,171 @@
...
++
+A capability loaded through an authorizing capability lacking <<zylevels1_lg_perm>> will have its <<zylevels1_gl_perm>> bit cleared.
+Additionally, if the loaded capability is _unsealed_, its <<zylevels1_lg_perm>> is cleared.
++
+This permission is similar to the base <<lm_perm>> and its effects on loaded capabilties' <<w_perm>> and <<lm_perm>> (but note the difference in interaction with seals).
+
+<<zylevels1_lg_perm>> is a _refinement of_ <<c_perm>> and <<r_perm>>.
+That is, if either of the latter are clear, then <<zylevels1_lg_perm>> has no effect.
+The capability encoding might take advantage of this dependence for compression,
+and so clearing either <<c_perm>> or <<r_perm>> may result in
```
francislaus: Same as above

src/cheri/zylevels1.adoc
```diff
@@ -0,0 +1,171 @@
...
+The <<zylevels1_gl_perm>>, <<zylevels1_sl_perm>>, and <<zylevels1_lg_perm>> fields
+are mapped into the <<acperm_bit_field,capability permissions bitfield>> (<<acperm_bit_field>>),
+used by <<ACPERM>> and <<GCPERM>>,
+as shown in <<zylevels1_acperm_bit_field>>.
+
+=== Interaction with <<LC>> and <<zalrsc_cheri>>
+
+As outlined above, Zylevels1 introduces two new constraints on capabilities loaded from memory,
+as part of a <<LC>> instruction `lc cd, offset(cs1)` or <<LR_C>> instruction `lr.c cd, offset(cs1)`.
+These may be phrased as "implicit <<ACPERM>>-s" performed on the loaded capability thus:
```
francislaus, line 113: I could see how referring to this as "implicit ACPERM" might confuse people. An ACPERM does legalisation and thus can also clear other permission bits than the ones named. However a capability load does not clear those, but only local-global related bits in certain cases.

src/cheri/zylevels1.adoc
```diff
@@ -0,0 +1,171 @@
...
+
+=== Interaction with <<SCSS>>
+
+Implementations of Zylevels1 must ensure that a <<SCSS>> instruction `scss rd, cs1, cs2`
+indicates that `cs1` is a subset of `cs2` (that is, sets `rd` to 1) only if either
+
+* `cs2` 's <<zylevels1_gl_perm>> is clear or
+* `cs1` 's <<zylevels1_gl_perm>> is set.
+
+The existing permission subset logic applies to the new <<zylevels1_sl_perm>> and <<zylevels1_lg_perm>>.
```
francislaus, line 149: Is the GL bit irrelevant to SCSS? If so, it might be worth to state that here. In general, I am wondering whether we would want a paragraph somewhere earlier differentiating between the permissions and the GL bit. What do you think?

francislaus (changes requested):
Thank you Wes. I had a first look and had a few questions. I will try to get a more detailed look at some point later

src/cheri/zylevels1.adoc
nwf: That is a property of the capability encoding, in principle, but SL really is only meaningful in combination with w and c.

src/cheri/zylevels1.adoc
nwf: I (attempted to) keep the existing semantics.  In particular, the `GL` bit is handled by lines 129-130, I think?

arichardson:
> I think the n-level extension is far from maturity and in my understanding nobody has gotten close to evaluate it. I think I would be in favour of removing it for now to not confuse people. What do others think?

Yes I think we should completely drop it. It should already not be rendered in the current spec

nwf:
At long last, I have taken a stab at updating this.  Please advise.

src/cheri/zylevelsN.adoc
```diff
@@ -0,0 +1,143 @@
+[#section_zylevelsN,reftext="ZylevelsN"]
+== "ZylevelsN" Extension for CHERI Multi-level Information Flow Control
```
arichardson, line 2: Thanks for keeping this file, I agree it should not be part of the spec but it's useful to keep around.

src/cheri/app-cap-description.adoc
```diff
@@ -182,41 +169,63 @@ endif::[]
...
+| RV32-base-2     | <<x_perm>>   | (<<c_perm>> and <<lm_perm>>) or (not(<<c_perm>> and not(<<lm_perm>>)))
+|===
 
 
+When <<section_zylevels1>> _is_ present, the following new rows are added to <<acperm_rules>> table:
+
+[float="center",align="center",cols="2,2,4",options="header"]
+|===
+| Rule             | Permission   | Only valid if
+| RV32-Zylevels1-1 | <<zylevels1_sl_perm>>  | (<<lm_perm>> and (<<x_perm>> or <<w_perm>>))
```
arichardson: The problem we had with having the extension separate in the first version was subtle confusion about the order of these checks. I think maybe the best solution is to duplicate the table and put it in the Zylevels1 chapter?

nwf: I don't think duplicating the table is useful, because this is one of the things that needs to be a kind of scattered / tangled definition.  I _think_ it would be OK for the _specification_ (if not the _implementation_) to be "iterate to fixedpoint" the rules from the base and any rules added by extensions and the encoding, and I attempted to make the prose say that.

(Formally, each rule is monotonic nonincreasing on a finite set of permissions, and so the fixed-pointing operation must terminate.  As such, it's possible to flatten the fixed-pointing transform into a combinatorial circuit.)

arichardson:
> > so we'll have to hold off until then.
> 
> Can we please hold off not _too_ long? Because this PR moves prose around, it's annoying to rebase again and again.

We should hopefully have a conclusion by next Tuesday.

src/cheri/app-cap-description.adoc
```diff
@@ -261,24 +272,22 @@ When removing permissions with <<ACPERM>> the following rules are followed in or
...
 |  6  (RV32 only) | <<x_perm>>   | <<w_perm>> or <<c_perm>>
-|  7              | <<el_perm>>  | <<c_perm>> and <<r_perm>>
-|  8  (RV32 only) | <<el_perm>>  | <<lm_perm>>
 |  9              | <<lm_perm>>  | <<c_perm>> and <<r_perm>>
-| 10  (RV32 only) | <<lm_perm>>  | (<<w_perm>> or <<el_perm>>)
-| 11              | <<sl_perm>>  | <<c_perm>>
-| 12  (RV32 only) | <<sl_perm>>  | (<<lm_perm>> and (<<x_perm>> or <<w_perm>>))
-| 13  (RV32 only) | <<x_perm>>   | (<<c_perm>> and <<lm_perm>> and <<el_perm>> and (<<sl_perm>> == ∞)) or +
-(not(<<c_perm>> and not(<<lm_perm>>) and not(<<el_perm>>) and (<<sl_perm>>==0)))^1^
+| 10  (RV32 only) | <<lm_perm>>  | (<<w_perm>> or <<zylevels1_lg_perm>>)
```
arichardson: Need to renumber this.

src/cheri/insns/atomic_exceptions.adoc
```diff
@@ -8,10 +8,9 @@ If <<c_perm>> is not granted then store the memory tag as zero, and load `cd.tag
 +
 If the authorizing capability does not grant <<lm_perm>>, and the {ctag} of `cd` is 1 and `cd` is not sealed, then an implicit <<ACPERM>> clearing <<w_perm>> and <<lm_perm>> is performed to obtain the intermediate permissions on `cd` (see <<LC>>).
 +
-If the authorizing capability does not grant <<el_perm>>, and the {ctag} of `cd` is 1, then an implicit <<ACPERM>> restricting the <<section_cap_level>> to the level of the authorizing capability is performed.
-If `cd` is not sealed, this implicit <<ACPERM>> also clears <<el_perm>> to obtain the final permissions on `cd` (see <<cap_level_load_summary>> and <<LC>>).
+Extensions may further attenuate loaded or stored capabilities via "implicit <<ACPERM>>-s" and/or outright tag clearing;
```
arichardson: I'm not sure "attenuate" is a commonly understood word. Maybe worth being a bit more verbose?

src/cheri/insns/gcperm_32bit.adoc
```diff
@@ -19,22 +19,14 @@ Encoding::
...
+Convert the unpacked <<AP-field>>, <<SDP-field>>, and any other extension-defined permission-like fields of capability `cs1`
+into a bit field, with the same format as used by <<ACPERM>> (see <<acperm_bit_field>>),
 and write the result to `rd`.
++
 All bits in the `[23:0]` range that are reserved or assigned to extensions that are not implemented by the current hart always report 1.
 +
 All architectural permission bits in `rd` are set to 0 if any <<section_cap_integrity,integrity>> checks failed.
 +
-.Capability permissions bit field
-[#gcperm_bit_field]
```
arichardson, line 32: I know this is redundant, but I do kind of like seeing the bitfield values in both instructions. Don't really mind either way.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -429,7 +358,8 @@ NOTE: This property is required to ensure restricted programs cannot forge capab
 [#infinite-cap,reftext="Infinite"]
 ===== Infinite Capability
 
-The _Infinite_ capability grants all permissions, has the maximum <<section_cap_level,CL-field>>, and its bounds cover the whole 2^MXLEN^ address space.
+The _Infinite_ capability grants all permissions and its bounds cover the whole 2^MXLEN^ address space.
+Extensions that modify capability metadata must describe their metadata's state in Infinite capabilities.
```
```suggestion
Extensions that introduce new capability metadata must describe their metadata's state in Infinite capabilities.
```
I'd be tempted to just drop this sentence though.

src/cheri/zylevels1.adoc
```diff
@@ -0,0 +1,188 @@
+[#section_zylevels1,reftext="Zylevels1"]
+== "Zylevels1" Extension for CHERI 2-Level Information Flow Control
```
arichardson: Can we use a {cheri_ifc_ext_name} variable here?

src/cheri/app-cap-description.adoc
arichardson: Can we move the levels table to the levels chapter?

nwf: The rules about LG/SL/GL are indeed moved, I think?  But I don't think that's what you're asking about?

Issue #638, opened 2025-06-20
Taggable PMAs - do we need more than one


tariqkurd-repo:
Currently in the spec we have the option of the two behaviours for storing a valid tag to a non-taggable PMA

1. Cause an access fault exception
2. Ignore the stored valid tag

We need to deal with the options.

a. choose one, and say the other is a future extension
b. add an additional PMA now

In our (Codasip's) systems we ignore the stored valid tag.
Are there any objections to changing the text to:

```
When the hart attempts to store data with the valid tag set to memory regions that are not taggable, the implementation ignores the stored valid tag.

NOTE: A PMA from a future extension could raise an access fault exception when attempting to store a valid tag to non taggable memory.
```
is anyone taking or planning to take an access fault in this case?

tariqkurd-repo:
>Trapping isn't difficult to implement though?

Correct - but I think the question is more whether you really want that behaviour in all cases in a real system.
As it's still very early in the evolution of CHERI systems I suspect that the answer is to allow both so that people aren't overly constrained.

It could be a second PMA, or a bit in (maybe) menvcfg which controls the trapping behaviour.
As PMAs may be fixed in a real system, control from a CSR bit may be desirable.

tariqkurd-repo:
>If you make it trap on stores of capabilities then you don't get to decide whether it can retire until the stored operand is known. This introduces a lot of additional complexity on most pipeline designs.

We do have this for the MMU based revocation, but that's a very different design point to a small embedded core.

> most server pipelines don't want to do precise traps for ECC store failures: they do precise traps on loads and imprecise traps on stores

RISC-V doesn't say anything about handling of ECC faults, an so I don't believe that they are mandated to be precise or imprecise. I think all store errors (AXI or ECC) can be handled asynchronously as NMIs, so I don't think this part is an issue.

I think the overriding argument is not to force embedded cores to check the tag before deciding whether they can send a store transaction to the bus.

davidchisnall:
> The cap-dirtyable machinery -- what Morello calls CDBM -- is used by the (existing) revocation machinery to close a race between 1) a core that is attempting to stop the revoker from scanning a page in subsequent epochs, and 2) a core (and its TLB) that writes a capability to that page, meaning that the revoker must continue to scan that page.

Is it sufficient in this case to clear the cap-dirty bit in the PTE and then scan the page for capabilities?  Any store of a capability concurrently with the scan will set the cap-dirty bit.

[...]

PR #639, opened 2025-06-20
Add new taggable PMA

tariqkurd-repo:
Fixes https://github.com/riscv/riscv-cheri/issues/638

### DIFF
--- a/src/cheri/insns/load_32bit_cap.adoc
+++ b/src/cheri/insns/load_32bit_cap.adoc
@@ -20,9 +20,9 @@ Calculate the effective address of the memory access by adding `cs1.address` to 
 +
 Authorize the memory access with the capability in `cs1`.
 +
-Load a naturally aligned CLEN-bit data value and the associated valid tag from memory.
+Load a naturally aligned CLEN-bit data value from memory.
 +
-Write the CLEN-bit data and the valid tag to `cd` and conditionally update them as specified below.
+Use the CLEN-bit data and the valid tag to determine the value of `cd` as specified below.
 +
 include::malformed_no_check.adoc[]
 //+ remove as this is in the base ISA chapter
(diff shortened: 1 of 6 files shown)


jrtc27:
What is the benefit for removing this option that people might legitimately want?

tariqkurd-repo:
we now have three PMAs:

_CHERI Valid Tag_
_CHERI Valid Tag Strip_
_CHERI Valid Tag Fault_

_CHERI Valid Tag Fault_ required instruction page updates, and I noticed various things which needed fixing on those pages, so this PR has more in it than it strictly should (I could make another PR just for those fixes to separate them out)

davidchisnall:
> @nwf @davidchisnall is this ok for CHERIoT systems?

I don't think any of this matters particularly for CHERIoT.  We generally don't hand out capabilities that allow storing capabilities to things in the address space that can't hold capabilities.  We'd expect the auditing policy to prevent this in the firmware metadata, so if you skip the audit step in your deployment and ship something buggy... don't do that.

I'd have to check, but I don't think we're actually using PMAs to restrict capabilities at all.  If you do a store capability to a pair of 32-bit MMIO registers in a device, for example, the device will see it as two 33-bit stores and, if it doesn't know about CHERIoT, will simply discard the 33rd bit (it's in one of the user bits in TileLink, so things that don't know about it always ignore it on stores and set it to zero on loads).  I can imagine other memory bus designs where having a tagged value being sent to something that expected never to see tags would be a problem, but I don't imagine we (SCI) would use this functionality even if we do an application core.

Note that having the core be aware of the structure of physical memory is one of the big differences between Arm and RISC-V.  Arm makes an explicit architectural decision that the core should be oblivious to the structure of its physical address map.  Requiring this knowledge was a strange choice for RISC-V (x86 has this awareness because Intel owns everything on an Intel chip, AMD owns everything on an AMD chip, and the ISA was never designed to allow you to buy different IPs from different vendors and combine them) and leads to some odd knock-on effects in page table design and elsewhere, but I don't think that CHERI is making this any worse than it was already.

tariqkurd-repo:
I've already done a separate PR for the unrelated changes and hoped the
merging would remove them from this one but that didn't work. Feel free to
do whatever's necessary!

Issue #635, opened 2025-06-19
Does mstatus.MXR affect CHERI?


tariqkurd-repo:
A nice question from @ved-rivos during the ARC review is what happens on access with mstatus.MXR=1.
This is to Make eXecutable memory to be Readable (from the PTE permissions), so that emulation sequences can read the instruction which trapped.
This means that X or R permission are required to load from memory.

We haven't specified any behaviour for CHERI.

Should we check for X or R permission in this case, instead of R to be consistent?

The same question applies to HLVX.* - which need X permission not R permission in the PTE.
Should CHERI check X permission accordingly?

tariqkurd-repo:
We have a few opinions here:

- X not R memory only gives weak security, so doens't seem worthwhile with CHERI (so no spec change)
- It's seems reasonable to follow the existing spec and allwo R or X perm

so no conclusion here yet

arichardson:
It's also worth noting that this only affects RV64, since RV32 cannot encoding execute-only capabilities. 

Recapping from the discussion we had: It seems unlikely that allowing reads for X-only capabilities when MXR is set has security implications since code that is able to set MXR can do much more already.

The advantage of allowing MXR to affect the capability checks is that mepc can be used unconditionally to fetch the faulting instruction without having to rederive from another capability that R permission. But we are already in a slow exception path so one additional setaddr instruction shouldn't really have any noticeable impact.

We need to think a bit harder if there are any other implications of treating X-only as R+X. I believe either approach should be fine.
Personally, I slighlty prefer towards option 1: no change to capability permission checks, but would also be fine with checking MXR for X+R.

tariqkurd-repo:
We discussed this in the TG and decided that following MXR and HLXV isn't required for CHERI. PR to follow.

PR #643, opened 2025-06-25
tidy up xrefs, and spec MXR and HLVX

tariqkurd-repo:
fixes https://github.com/riscv/riscv-cheri/issues/635

### DIFF
--- a/src/cheri/app-cheri-instructions.adoc
+++ b/src/cheri/app-cheri-instructions.adoc
@@ -346,7 +346,7 @@ include::insns/zcmt_cmjt.adoc[]
 
 Hypervisor load and store instructions and the {cheri_base_ext_name} extensions.
 
-While HLVX.* only requires execute permission in the PTE, the authorizing CHERI capability must grant <<r_perm>> in addition to <<x_perm>>.
+While HLVX.* only requires execute permission in the PTE, the authorizing CHERI capability must grant <<r_perm>>.
 
 include::insns/hypv-virt-load-cap.adoc[]
 include::insns/hypv-virt-store-cap.adoc[]


src/cheri/app-cheri-instructions.adoc
```diff
@@ -346,5 +346,7 @@ include::insns/zcmt_cmjt.adoc[]
 
 Hypervisor load and store instructions and the {cheri_base_ext_name} extensions.
 
+HLVX.* permission checks are not affected by CHERI, they check <<r_perm>> not <<x_perm>>.
```
arichardson: This one is a bit ambiguous, maybe something like
```suggestion
While HLVX.* only requires execute permission in the PTE, the authorizing CHERI capability must grant <<r_perm>> in addition to <<x_perm>>.
```

arichardson (commented):
For MXR I believe this is the best approach since it is an ambient thing that would affects all checks.

However, for HLVX I am less sure about the correct behaviour since it explicitly requests X permission in the PTE, so we could change the semantics of the CHERI checks just for this one instruction.
But maybe it's better to not special case this and relax the checks in the future if needed?

Not sure what the best approach is here. Any opinions @jrtc27 @davidchisnall @nwf?

src/cheri/app-cheri-instructions.adoc
```diff
@@ -346,5 +346,7 @@ include::insns/zcmt_cmjt.adoc[]
 
 Hypervisor load and store instructions and the {cheri_base_ext_name} extensions.
 
+While HLVX.* only requires execute permission in the PTE, the authorizing CHERI capability must grant <<r_perm>> in addition to <<x_perm>>.
```
tariqkurd-repo: do we need x_perm here? I was thinking that it only needs r_perm......

arichardson: yeah it should only need r.

tariqkurd-repo: will fix as a result of our conversation

arichardson (approved):

Issue #636, opened 2025-06-19
should control flow changes to a faulty target except at the source or destination?


tariqkurd-repo:
It came out from the ARC review that faulting on branches and jumps due to a bad target is unpopular.
The proposal is to take them at the target.

One problem is how to indicate to the target what the problem is - maybe _always_ clear pcc.tag and have anothe CSR which said what went wrong?

How can we justify taking the exception at the source for security reasons?

tariqkurd-repo:
We discussed direct jumps and branches doing a representability check on the result of adding the offset to the PC, and making the result of that into the new PC, and _not_ doing the bounds check.

If the next instruction is out of bounds then it will naturally follow from the PCC check.

This conveniently removes the need to check for a "minimum sized instruction" at the target - which isn't a great check as it might be a 32-bit instruction (assuming 16-bit support) so the bounds check is effectively split in two in many cases.

For JALR, it will unseal (if sealed and the offset is zero), CADDI the offset to cs1 and redirect to the target. So again, if the result is unrepresentable, the tag will clear and so the fault will be taken at the target.

This will change some current bounds faults into tag faults, because if the offset takes the PCC out of the representable range the current spec would give a bounds fault.

Apparently Morello detectes PCC faults at the target like this

There are concerns about jumping out of a cmpartment before detecting the fault, or increasing the speculation window as a result of the proposed change.

vmurali:
> So again, if the result is unrepresentable, the tag will clear and so the fault will be taken at the target.

> There are concerns about jumping out of a compartment before detecting the fault, or potentially increasing the speculation window as a result of the proposed change.

While CHERIoT unsets the tag and takes the exception at the target (and as Wes points out, there's no security boundary crossing in existing SW platforms on a simple JALR), it might be prudent to throw the exceptions at the source for these things (untagged/sealed-but-not-zero-offset/unrepresentable/non-executable). We might very well have a CHERI system without trampolines that requires user code to clear the stack etc and use simple JALR to call and return across compartment boundaries. Fixing the ISA such that the exception is thrown only at the target will preclude such designs in the future.

nwf:
Just so it's written down somewhere...

The other aspect to this is that we apparently like to add spice to our `j{al,}r`s with CHERI.  "Sentries unseal" was probably the original sin (sorry) and is fine because it's self-contained, but then CHERIoT added IRQ-disposition-bearing sentries and we keep talking about adding indirect sentries which also commit other architectural changes as part of jumping.

[...]

PR #644, opened 2025-06-25
remove exception checks from branches and jumps

tariqkurd-repo:
https://github.com/riscv/riscv-cheri/issues/636

I was on the fence about whether we should make this change, so made the PR to explore the implications.
TBH it's quite a nice simplification, so I think  we should go for it.

If we find a specific security hole during compartment switching then we can always add a control bit somewhere to control exception behaviour for that specific case, without requiring exceptions on _every_ branch and jump.


src/cheri/insns/condbr_32bit.adoc
```diff
@@ -19,13 +19,8 @@ include::wavedrom/ct-conditional.adoc[]
...
-The target address is written into <<pcc>>.address.
+If taken, the <<pcc>> is incremented by the offset using <<SCADDR>> semantics.
 
-Exceptions::
-[%autowidth,options=header,align=center]
-|==============================================================================
-| Kind                      | Reason
-| Bounds violation          | Minimum size instruction at the target is not within the <<pcc>>'s bounds.
-|==============================================================================
+NOTE: A future extension may raise an exception on the branch instruction itself if a minimum sized instruction at the target <<pcc>> will raise a CHERI bounds exception.
```
arichardson:
```suggestion
NOTE: A future extension may raise an exception on the branch instruction itself if fetching a minimum sized instruction at the target <<pcc>> will raise a CHERI bounds exception.
```

src/cheri/insns/jal_32bit.adoc
```diff
@@ -18,17 +18,15 @@ Jump to the target <<pcc>>.
 +
 Increment <<pcc>>.address by the sign-extended offset to form the target <<pcc>>.
 The <<pcc>> of the next instruction is sealed and written to `cd`.
++
+Both address increments use <<SCADDR>> semantics to determine the result.
+
+NOTE: A future extension may raise an exception on the branch instruction itself if a minimum sized instruction at the target <<pcc>> will raise a CHERI bounds exception.
```
arichardson:
```suggestion
NOTE: A future extension may raise an exception on the branch instruction itself if fetching a minimum sized instruction at the target <<pcc>> will raise a CHERI bounds exception.
```

src/cheri/insns/jal_32bit.adoc
```diff
@@ -18,17 +18,15 @@ Jump to the target <<pcc>>.
...
```
arichardson, line 26:
```suggestion
  Peforming the <<pcc>> bounds check at the branch source instead of on instruction fetch is helpful for debugging and can simplify the implementation of CPUs with very short pipelines.
  
```

@davidchisnall Maybe something like this as a note until we have an actual extension specified?

arichardson (approved):
Based on the discussion we had, requiring checks at target and allowing a future extension performing the checks at the source sounds good to me.

However, I would like @davidchisnall @rmn30 @nwf and @jrtc27 to take a look at this before we go ahead and merge it.

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -25,20 +25,14 @@ Jump to the target capability in `cs1`.
 The target capability is unsealed if the `offset` is zero.
 Increment `cs1.address` by the sign-extended 12-bit `offset` and set the least-significant bit of the result to zero.
```
rmn30, line 26: Doesn't this need to say something about unsealing for sentries?

src/cheri/csv/CHERI_ISA.csv
```diff
@@ -41,12 +41,12 @@
...
-"C.JR","✔","✔","","","","✔","✔","Both","","","","✔","","","","","","","","","","","","","C2","","","","Indirect jump, bounds check minimum size target instruction.  set PCC=unseal(cs1)","MODE==D (optional)","","","","","","",""
+"C.J","✔","✔","","","","✔","✔","Both","","","","✔","","","","","","","","","","","","","C2","","","","Jump to PC+offset","MODE==D (optional)","","","","","","",""
+"C.JAL","✔","","","","","✔","✔","Both","","","","✔","","","","","","","","","","","","","C2","","","","Jump to PC+offset, link to cd","MODE==D (optional)","","","","","","",""
+"JAL_CHERI","✔","✔","","","","✔","✔","Both","","","","✔","","","","","","","","","","","","","JAL","","","","Jump to PC+offset, link to cd","MODE==D (optional)","","","","","","",""
+"JALR_CHERI","✔","✔","","","","✔","✔","Both","","","","","","","","","","","","","","","","","JALR","","","","Indirect jump and link.  set PCC=unseal(cs1),cd=seal(nextPCC)","MODE==D (optional)","","","","","","",""
```
rmn30: Does `JAL_CHERI` not seal the link capability? Why not write it the same way as `JALR_CHERI` i.e. `cd=seal(nextPCC)`?

rmn30 (commented):
I think it's fine, although I'm not sure whether we'll actually do the minimum instruction size check. Other checks (tag etc) on JALR much more likely.

src/cheri/csv/CHERI_ISA.csv
tariqkurd-repo: ahhhh you're right - will fix

tariqkurd-repo:
> I think it's fine, although I'm not sure whether we'll actually do the minimum instruction size check. Other checks (tag etc) on JALR much more likely.

I'm supportive of also skipping that check if it's ok for CHERIoT

src/cheri/insns/jalr_32bit.adoc
tariqkurd-repo: yep - fixed!

PR #648, opened 2025-07-01
Define Zicond instructions to copy the full capability

arichardson:
This allows for a branchless conditional move of capabilities. If we only copy the integer part and zero the metadata+tag, we would need to rely on macro-op fusion for conditional capability moves.

### DIFF
--- a/src/zicond.adoc
+++ b/src/zicond.adoc
@@ -90,6 +90,8 @@ Encoding::
 
 Description::
 If _rs2_ contains the value zero, this instruction writes the value zero to _rd_.  Otherwise, this instruction copies the contents of _rs1_ to _rd_.
+When {cheri_base_ext_name} is supported, this instruction copies all bits of _rs1_ to _rd_ including capability metadata and valid tag.
+If _rs2_ is zero, this instructions writes a <<null-cap>> (with zero metadata and valid tag).
 
 This instruction carries a syntactic dependency from both _rs1_ and _rs2_ to _rd_.
 Furthermore, if the Zkt extension is implemented, this instruction's timing is independent of the data values in _rs1_ and _rs2_.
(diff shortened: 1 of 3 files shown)


src/zicond.adoc
```diff
@@ -90,6 +90,8 @@ Encoding::
 
 Description::
 If _rs2_ contains the value zero, this instruction writes the value zero to _rd_.  Otherwise, this instruction copies the contents of _rs1_ to _rd_.
+When {cheri_base_ext_name} is supported, this instruction copies all bits of _rs1_ to _rd_ including capability metadata and valid tag.
```
tariqkurd-repo, line 93:
```suggestion
When the base ISA is {cheri_base_ext_name}, this instruction copies all bits of _rs1_ to _rd_ including capability metadata and valid tag.
```

in the wavedrom diagram, to be consistent with other places, we should have rs1/cs1 as the source and rd/cd as the dest, so this should be consistent
unless we separate out the CHERI version.... which would seem a shame because in both case it is just test the LSB of rs2, and copy _all_ of rs1.

src/zicond.adoc
```diff
@@ -129,6 +131,8 @@ Encoding::
 
 Description::
 If _rs2_ contains a nonzero value, this instruction writes the value zero to _rd_.  Otherwise, this instruction copies the contents of _rs1_ to _rd_.
+When {cheri_base_ext_name} is supported, this instruction copies all bits of _rs1_ to _rd_ including capability metadata and valid tag.
```
tariqkurd-repo, line 134:
```suggestion
When the base ISA is {cheri_base_ext_name}, this instruction copies all bits of _rs1_ to _rd_ including capability metadata and valid tag.
```

tariqkurd-repo:
closing this one as we need to add new encodings instead of repurposing them

PR #650, opened 2025-07-07
Clarify ADD/CADD behaviour

tariqkurd-repo:


src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -806,6 +806,22 @@ include::insns/store_32bit_cap.adoc[]
...
+They must be used for all address incrementing.
+
+[NOTE]
+====
+
+Integer add (`ADD`) and capability add (`CADD`) have separate encodings.
+Using a single encoding for both is undesirable:
+
+ . For high performance implementations which can issue multiple ADDs, it means that the integer ADD units don't need the upper halves of the operands, and don't need the capability check logic on the result.
+ . The compiler and/or programmer would have to execute another metadata clearing instruction after each ADD to ensure that capabilities don't leak capabilities.
```
```suggestion
 . The compiler and/or programmer would have to execute another metadata clearing instruction after each ADD to ensure that compartments don't leak capabilities.
```

arichardson: Should we add a third point that this is consistent with `ADDIW`?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -806,6 +806,22 @@ include::insns/store_32bit_cap.adoc[]
...
 
+`ADD` and `ADDI` are not affected by the rule above.
+Even though they _are_ used for handling addresses, they also have other uses.
+New encodings are used for capability addition: <<CADD>> and <<CADDI>>.
+They must be used for all address incrementing.
+
+[NOTE]
+====
+
+Integer add (`ADD`) and capability add (`CADD`) have separate encodings.
```
```suggestion
Integer add (`ADD`) and capability add (<<CADD>>) have separate encodings.
```
so the name updates once we have finalized naming?

tariqkurd-repo: yes - the idea is all the names update afterwards! so keep using the old ones for now

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -806,6 +806,22 @@ include::insns/store_32bit_cap.adoc[]
...
+New encodings are used for capability addition: <<CADD>> and <<CADDI>>.
+They must be used for all address incrementing.
+
+[NOTE]
+====
+
+Integer add (`ADD`) and capability add (<<CADD>>) have separate encodings.
+Using a single encoding for both is undesirable:
+
+ . For high performance implementations which can issue multiple ADDs, it means that the integer ADD units don't need the upper halves of the operands, and don't need the capability check logic on the result.
```
tariqkurd-repo, line 821:
```suggestion
 . Integer ADD is most commonly used for purposes other than address calculations.
 . For high performance implementations which can issue multiple ADDs, it means that the integer ADD units don't need the upper halves of the operands, and don't need the capability check logic on the result.
```

arichardson (approved):

PR #653, opened 2025-07-08
Clarified legalisation around M-bit

### LINKED ISSUE #649: Impact of M-bit on legalisation
The definition of `gcperm` (and others) requires legalisation of the AP field. Specifically it says:

https://github.com/riscv/riscv-cheri/blob/a4527b8bc9edab55c775c7253b0ba065da2ca191/src/cheri/insns/gcperm_32bit.adoc?plain=1#L29-L30

This definition does not include the `M` bit as it is not an architectural permission. However the Sail model seems to zero the architectural permissions if the `M` bit is illegally set (only valid if X-permission set and Zcherihybrid is implemented).

I propose to add to the specification that an illegally set M bit impacts the return of legalised values where applicable.

francislaus:
This should close #649 

I also fixed a minor typo I encountered.

### DIFF
--- a/src/cheri/app-cap-description.adoc
+++ b/src/cheri/app-cap-description.adoc
@@ -578,7 +578,7 @@ CHERI enforces the following rules for all valid capabilities:
 
 . The bounds are not <<section_cap_malformed,malformed>>.
 . No reserved bit in the capability encoding is set.
-. The permissions can be legally produced by <<ACPERM>>.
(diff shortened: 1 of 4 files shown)


tariqkurd-repo:
the problem is that we can't include the definition of the M-bit yet......
Zcherihybrid is another extension which ARC aren't willing to look at until we have the base ISAs stable
So the description of GCPERm etc. can't make any reference to the M-bit in the main chapters....

PR #656, opened 2025-07-09
JALR clarification

tariqkurd-repo:
JALR needed updating now that exception checks have been removed

### DIFF
--- a/src/cheri/insns/jalr_32bit.adoc
+++ b/src/cheri/insns/jalr_32bit.adoc
@@ -20,14 +20,19 @@ include::wavedrom/ct-unconditional-2.adoc[]
 include::base_isa_extension.adoc[]
 
 Description::
-Jump to the target capability in `cs1`.
+Indirect jump to the target capability in `cs1` with an address offset.
 +
-If the `cs1` is a sentry, unseal it if the `offset` is zero.
-Increment `cs1.address` by the sign-extended 12-bit `offset` and set the least-significant bit of the result to zero.
-The <<pcc>> of the next instruction is sealed and written to `cd`.
+Copy `cs1` to the target <<pcc>>
 +
-Both address increments use <<SCADDR>> semantics to determine the result.
-
+. If the target <<pcc>> is a sentry, and the `offset` is zero, then unseal the target <<pcc>>.
+. Set bit zero of the `offset` to zero.
+. Increment the address of the target <<pcc>> by the sign-extended 12-bit `offset` using <<SCADDR>> semantics.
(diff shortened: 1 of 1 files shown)


src/cheri/insns/jalr_32bit.adoc
```diff
@@ -20,14 +20,19 @@ include::wavedrom/ct-unconditional-2.adoc[]
...
+Copy `cs1` to the target <<pcc>>
++
+. If the target <<pcc>> is a sentry, and the `offset` is zero, then unseal the target <<pcc>>.
+. Set bit zero of the `offset` to zero.
+. Increment the address of the target <<pcc>> by the sign-extended 12-bit `offset` using <<SCADDR>> semantics.
+. Jump to the target <<pcc>>
++
+The <<pcc>> of the next instruction is calculated using <<SCADDR>> semantics, sealed and written to `cd`.
++
+NOTE: Because of the unsealing and <<SCADDR>> behavior, it is not possible for the target <<pcc>> to be sealed.
```
```suggestion
Copy `cs1` to the target <<pcc>>
+
If the target <<pcc>> is a sentry, and the `offset` is zero, then unseal the target <<pcc>>.
Otherwise, increment the address of the target <<pcc>> by the sign-extended 12-bit `offset`.
+
Then set bit zero of <<pcc>>.address to zero, check the address against the original <<pcc>> using <<SCADDR>> sematics and then and jump to the target <<pcc>>
+
The <<pcc>> of the next instruction is calculated using <<SCADDR>> semantics, and written to `cd` as a sentry.
+
NOTE: Because of the unsealing and <<SCADDR>> behavior, it is not possible for the target <<pcc>> to be sealed.
```

How about this wording?

arichardson: Actually this is not quite correct:
```suggestion
Copy `cs1` to the target <<pcc>>
+
. If the target <<pcc>> is a sentry, and the `offset` is zero, then unseal the target <<pcc>>.
Otherwise, increment the address of the target <<pcc>> by the sign-extended 12-bit `offset`.
. Set bit zero of the target address to zero and update <<pcc>> with the new target address using <<SCADDR>> semantics
. Jump to the target <<pcc>>
+
The <<pcc>> of the next instruction is calculated using <<SCADDR>> semantics, and written to `cd` as a sentry.
+
NOTE: Because of the unsealing and <<SCADDR>> behavior, it is not possible for the target <<pcc>> to be sealed.
```

How about this wording?

tariqkurd-repo:
fixes https://github.com/riscv/riscv-cheri/issues/658

Issue #661, opened 2025-07-14
ACPERM no longer has integrity checks


tariqkurd-repo:
Since 0.9.5 ACPERM has lost the tag clearing if any reserved bits are set or the perms cannot have been produced by ACPERM
These need to be reinstated

nwf:
Those are properties of the capability (in-memory) encoding; please be sure that, when reinstated, these predicates are suitably parametric.

PR #662, opened 2025-07-15
add missing integrity check into ACPERM

tariqkurd-repo:
Fix https://github.com/riscv/riscv-cheri/issues/661

### DIFF
--- a/src/cheri/insns/acperm_32bit.adoc
+++ b/src/cheri/insns/acperm_32bit.adoc
@@ -37,7 +37,7 @@ In these cases, <<ACPERM>> will return a minimal sets of permissions, which may 
 Therefore, it is possible that requesting to clear a permission also clears others, but <<ACPERM>> will never _add_ new permissions.
 . Copy `cs1` to `cd`, and update the <<AP-field>>, <<section_cap_level,CL-field>> and <<SDP-field>> with the newly calculated versions.
 . Set `cd.tag=0` if `cs1` is sealed and any bits in the <<AP-field>> or <<SDP-field>> were affected by <<ACPERM>>.
-. Set `cd.tag=0` if the <<AP-field>> in `cs1` could not have been produced by <<ACPERM>>, or has any reserved bits set.
+. Set `cd.tag=0` if the <<AP-field>> in `cs1` could not have been produced by <<ACPERM>>, or `cs1` has any reserved bits set.
 .. Do _not_ set the valid tag to zero on a sealed capability if _only_ the <<section_cap_level,CL-field>> changed.
 
 .Capability permissions bit field


src/cheri/insns/acperm_32bit.adoc
```diff
@@ -38,6 +38,7 @@ Therefore, it is possible that requesting to clear a permission also clears othe
 . Copy `cs1` to `cd`, and update the <<AP-field>>, <<section_cap_level,CL-field>> and <<SDP-field>> with the newly calculated versions.
 . Set `cd.tag=0` if `cs1` is sealed and any bits in the <<AP-field>> or <<SDP-field>> were affected by <<ACPERM>>.
 .. Do _not_ set the valid tag to zero on a sealed capability if _only_ the <<section_cap_level,CL-field>> changed.
+. Set `cd.tag` to 0 if the <<AP-field>> in `cs1` could not have been produced by <<ACPERM>>.
```
arichardson: Could move this higher but here is fine too

arichardson (approved):

src/cheri/insns/acperm_32bit.adoc
tariqkurd-repo: yes - and I've added in that reserved bits shouldn't be set

src/cheri/insns/acperm_32bit.adoc
```diff
@@ -37,6 +37,7 @@ In these cases, <<ACPERM>> will return a minimal sets of permissions, which may
 Therefore, it is possible that requesting to clear a permission also clears others, but <<ACPERM>> will never _add_ new permissions.
 . Copy `cs1` to `cd`, and update the <<AP-field>>, <<section_cap_level,CL-field>> and <<SDP-field>> with the newly calculated versions.
 . Set `cd.tag=0` if `cs1` is sealed and any bits in the <<AP-field>> or <<SDP-field>> were affected by <<ACPERM>>.
+. Set `cd.tag=0` if the <<AP-field>> in `cs1` could not have been produced by <<ACPERM>>, or has any reserved bits set.
```
tariqkurd-repo:
```suggestion
. Set `cd.tag=0` if the <<AP-field>> in `cs1` could not have been produced by <<ACPERM>>, or `cs1` has any reserved bits set.
```

jrtc27:
I mean, my perspective is it’s architecturally impossible, unless you have broken memory or heterogenous cores (something that needs to be forbidden architecturally) when it comes to capability encodings, and so the architecture need not specify it at all. If you ever hit this case your architecture is already broken and so who cares in what way it’s broken. Then you deal with it as a RAS problem.

But if we really must specify it then it should be done in as sensible a way as possible. I’m concerned though that mandating such checks is going to add restrictions on the implementation that will hurt PPA, and that seems like a bad idea at least for small cores where you wouldn’t normally bother with RAS concerns.

I would love to know ARC’s perspective on this.

arichardson:
I agree that always checking all requirements or none at all makes the most sense. Tiny cores that don't care about memory corruption could always just omit the checks and say it's impossible so no need to check

tariqkurd-repo:
ok - merging this and I filed https://github.com/riscv/riscv-cheri/issues/665 for the ongoing improvements

arichardson:
> > I agree that always checking all requirements or none at all makes the most sense. Tiny cores that don't care about memory corruption could always just omit the checks and say it's impossible so no need to check
> 
> But that's true of any core, so then why bother to put it in as a requirement?

The spec needs to define what happens for all possible input bits so that DV can deal with it consistently. We could leave it up to each implementation but defining the behaviour in the spec is much cleaner IMO

Issue #665, opened 2025-07-17
clean up integrity checks


tariqkurd-repo:
An outcome from https://github.com/riscv/riscv-cheri/pull/662 is that we should clean up the integrity checks, so instructions do _all_ or _none_ to simplify it and get the actual details of the checks out of the instruction descriptions.
It will also be easier to modify the checks for different encoding formats if the details haven't been spread across the instruction pages.

So instead of checking a subset of:

* malformed bounds
* invalid perms
* reserved bits set

just check all or none

PR #666, opened 2025-07-17
make integrity checks more consistent

tariqkurd-repo:
fix https://github.com/riscv/riscv-cheri/issues/665

### DIFF
--- a/src/cheri/app-cap-description.adoc
+++ b/src/cheri/app-cap-description.adoc
@@ -591,6 +591,10 @@ NOTE: These checks are much less rigorous than parity or ECC protection, and are
 NOTE: Even though valid capabilities which fail the integrity check could not have been legally generated,
  defining the handling in the architecture allows the behavior to be precisely specified for all 2^(CLEN+1)^ input values.
 
+Implementing these checks is optional, as all detected cases are theoretically unreachable.
+
+IMPORTANT: The default is to _include_ integrity checks, and they are _excluded_ in all cases to save logic area if {cheri_nointegrity_ext} is implemented.
+
 [#section_special_caps_encoding]
 === Encoding of Special Capabilities
 
--- a/src/cheri/attributes.adoc
+++ b/src/cheri/attributes.adoc
@@ -19,6 +19,7 @@
 :cheri_priv_h_ext: Shcheri
 :cheri_priv_vmem_ext: Svcheri
 :cheri_priv_debug_ext: Sdcheri
+:cheri_nointegrity_ext: Zcherinointegrity
 // Extension for CHERI CRG bits
 :cheri_priv_crg_ext:          Svucrg
 :cheri_priv_crg_load_tag_ext: Svucrglvt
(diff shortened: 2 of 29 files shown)


src/cheri/app-cap-description.adoc
```diff
@@ -591,6 +591,10 @@ NOTE: These checks are much less rigorous than parity or ECC protection, and are
 NOTE: Even though valid capabilities which fail the integrity check could not have been legally generated,
  defining the handling in the architecture allows the behavior to be precisely specified for all 2^(CLEN+1)^ input values.
 
+Implementing these checks is optional, as all detected cases are theoretically unreachable.
```
```suggestion
Implementing these checks is optional, as all detected cases are the result of misbehaving hardware.
```

Maybe something like this? Or "cannot occur in correctly functioning hardware".

src/cheri/app-cap-description.adoc
```diff
@@ -591,6 +591,10 @@ NOTE: These checks are much less rigorous than parity or ECC protection, and are
...
+
+IMPORTANT: The default is to _include_ integrity checks, and they are _excluded_ in all cases to save logic area if {cheri_nointegrity_ext} is implemented.
```
arichardson: I feel like we shouldn't introduce a new extension. Maybe something like this?
```suggestion
IMPORTANT: It is strongly recommended to _include_ integrity checks, but they may be omitted to save area in systems that do not wish to be more resilient to bit flips.
```

src/cheri/app-cap-description.adoc
tariqkurd-repo: it can be incompatible CHERI IPs though in the same system, where one has more features than another, so not necessarily misbehaving.

arichardson: That's true, I guess that would be more like "misconfiguration".
```suggestion
Implementing these checks is optional, as all detected cases are not reachable in a correctly functioning system.
```

src/cheri/app-cap-description.adoc
arichardson: And maybe some note on needing to define behaviour for all combination of input bits for DV?

tariqkurd-repo: I don't think resistance to bit flips is what we're aiming for here - if you want that you put ECC everywhere.
I think what we're aiming for is capabilities written by an incompatible IP.
I'll have a go at rewording soon....

jrtc27: If it’s possible to do so then that is an integration error that does not conform to the RISC-V specification.

tariqkurd-repo: I'm not sure we want to deny such systems, I've heard people talking of (e.g.) mixed 32-bit/64-bit CHERI systems before.
I'm advocating for such a system, but I don't see why we should prevent it by definition.

So these checks will be saying that a capability has been read which can't be handled, which is probably an error in a heterogenous system, but better to have a definition of the behaviour in these cases.

jrtc27: You can totally do that if you design your system very carefully, but the tags must be distinct. What you cannot do is interpret a valid 64-bit capability as a 128-bit one, or vice-versa, and get a valid tag.

tariqkurd-repo: yes indeed and I guess what this is trying to point out that what has been read is non-canonical and so is obviously wrong, and so giving people a standard set of checks to use to reject the capability

tariqkurd-repo: I think the current wording reflects that as it talks about `IP compatibility issues` so any objections to merging?

arichardson: New wording is vague enough. Sounds good to me

PR #676, opened 2025-07-29
ASR is always from the PCC

tariqkurd-repo:


jrtc27:
I don't think so; the closest that comes to that is something like CID. Anything that's not tied to PCC likely needs to be much more fine-grained, otherwise you might as well just run with ASR.

davidchisnall:
Not for v1.

Longer term, on big CHERI, I want ASR to go away entirely and be replaced with the system-access capability model.  In this model, we have an additional capability format where all bits except the disciminator are used to represent a bitmap of permissions and a SCR that holds it to authorise access to specific privileged features.  For example, the pmap code would hold the permission to modify the page-table base registers, but no other privileged registers.  For example, no ability to change the interrupt vector.  Similarly, early-boot code may set up the interrupt vector and then drop the permission to modify it.  

We couldn't do this on Arm because they had a *lot* of MSRs and they were all interleaved in exciting ways that made decomposition hard.  RISC-V currently has a smaller set and so there's probably an easier path.

Issue #660, opened 2025-07-11
If CBO.INVAL always issues as CBO.FLUSH does it need PCC ASR permission?


tariqkurd-repo:
CBO.INVAL needs PCC ASR permission, due to the danger of exposing stale capabilities in memory.

In an implementation which always executes CBO.INVAL as CBO.FLUSH then I think we should specify that the ASR permission check isn't required, due to the difficulty of making PCC.ASR appear in the LSU of a large OoO implementation.

We could go further and state that CBO.INVAL _must_ be implemented as CBO.FLUSH in a CHERI system.

Right now the spec requires that the LSU see PCC.ASR and I think we need a way around that

tariqkurd-repo:
See https://github.com/riscv/riscv-cheri/pull/676 which clarifies that this check uses ASR from the PCC

tariqkurd-repo:
My suggestion is that ASR perm is needed if:

1. CBO.INVAL is _not_ executing as CBO.FLUSH (either through CSR programming or by always implementing it as CBO.FLUSH)
2. The PMA is _CHERI Valid Tag_

PR #677, opened 2025-07-29
CBO.INVAL checks PCC.ASR in decode, prioritised above Load/store exceptions

tariqkurd-repo:
fixes https://github.com/riscv/riscv-cheri/issues/660


src/cheri/insns/cbo_exceptions.adoc
```diff
@@ -11,7 +11,9 @@ ifdef::cbo_clean_flush[]
 endif::cbo_clean_flush[]
 
 ifdef::cbo_inval[]
-| {cheri_excep_name_st} | {cheri_excep_desc_perm} <<w_perm>>, <<r_perm>> and <<asr_perm>> are all required.
+| {cheri_excep_name_st} | {cheri_excep_desc_perm} <<w_perm>>, <<r_perm>> are both required.
+
+If the PMA is _CHERI Valid Tag_ and CBO.INVAL is not executing as CBO.FLUSH then <<asr_perm>> is required in the <<pcc>>.
```
arichardson: Doesn't checking the PMA now introduce a dependency on the address that is being accessed? I'd just keep it simple and always require PCC.ASR?

arichardson: I'm less concerned about the CBO.INVAL->CBO.FLUSH mapping since that is just checking a single CSR bit, but the first one does not make sense to me.

jrtc27:
> I'm not sure it makes sense to make this check conditional?
> 
> Really unprivileged software should never issue an invalidate, only a flush. So even if the inval happens to map to a flush it's still user error?
> 
> @jrtc27 @davidchisnall @nwf what do you think?

I agree, and I believe we've had this discussion in the past, even. CBO.INVAL without ASR is a bug, period.

src/cheri/insns/cbo_exceptions.adoc
tariqkurd-repo: ok - the point was that if it's a FLUSH it's not a problem (in general) and if the PMA doesn't support tags then it's no worse then in non-CHERI

but OK - I'm trying to avoid connecting PCC.ASR further than it needs to be connected. We can also solve this by changing the fault type to `CHERI Instruction Access Fault` and then the fault can be raised in the decoder - it's a PCC fault after all.

tariqkurd-repo:
now in-line with https://github.com/riscv/riscv-cheri/pull/690 XRET and CBO.INVAL all take illegal instruction exceptions if PCC.ASR=0

arichardson (approved):

PR #684, opened 2025-07-30
Use "C" for capability tag in diagrams

arichardson:


arichardson:
I guess we could also use T?

tariqkurd-repo:
`CT` for Capability Tag? Which is the full name used in the text...

tariqkurd-repo:
V is ok by me

arichardson:
Okay will abandon this one.

PR #685, opened 2025-07-30
Fix incorrect operand name in SCBNDS description

arichardson:


tariqkurd-repo (approved):
this was correct in 0.9.5, this is a bug which crept in

PR #687, opened 2025-07-30
Use y-prefixed general purpose register names instead of c-prefix

arichardson:
This was requested by ARC and one benefit here is that we no longer have
the ambiguity in the c.jalr expansion: `jalr y1, 0(cs1)` makes it
obvious that the first register argument is a fixed `ra` value and the
latter is the `cs1` variable from the encoding.

Use attributes so we can easily change this depending on ARC feedback.

### DIFF
--- a/src/cheri/riscv-unpriv-integration.adoc
+++ b/src/cheri/riscv-unpriv-integration.adoc
@@ -621,7 +621,7 @@ This allows code fetch to be bounded, preventing a wide range of attacks that su
 
 The authorizing capability is either named explicitly (the base register of a load/store operation) or implicitly (when executing a branch, <<pcc>> is used for authorization).
 
-// E.g., `lw t0, 16({abi_creg}sp)` loads a word from memory, getting the address, bounds, and permissions from the `{abi_creg}sp` (capability stack pointer) register.
+E.g., `lw t0, 16({abi_creg}sp)` loads a word from memory, getting the address, bounds, and permissions from the `{abi_creg}sp` (capability stack pointer) register.
 
 No other exception paths are added by {cheri_base_ext_name}: in particular, capability manipulations do not raise an exception, but may set {ctag} of the resulting capability to zero if the operation is not permitted.
 


src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -622,18 +619,18 @@ This allows code fetch to be bounded, preventing a wide range of attacks that su
 
 The authorizing capability is either named explicitly (the base register of a load/store operation) or implicitly (when executing a branch, <<pcc>> is used for authorization).
 
-E.g., `lw t0, 16(csp)` loads a word from memory, getting the address, bounds, and permissions from the `csp` (capability stack pointer) register.
+// E.g., `lw t0, 16(ysp)` loads a word from memory, getting the address, bounds, and permissions from the `ysp` (capability stack pointer) register.
```
tariqkurd-repo: why comment this?

src/a-st-ext.adoc
```diff
@@ -505,7 +505,7 @@ All atomic operations follow the rules in xref:section_cheri_int_load_store_insn
 where the atomic operation is considered to be a load followed by a store for the
 capability rules.
 
-Therefore _all_ atomic operations are _reserved_ if `cs1=c0`.
+Therefore _all_ atomic operations are _reserved_ if `cs1=y0`.
```
tariqkurd-repo:
```suggestion
Therefore _all_ atomic operations are _reserved_ if `rs1=y0`.
```

I'm expecting to change these to rs1/rs2/rd as per normal risc-v - to be confirmed on tuesday

arichardson: I think using a separate "placeholder name" is still useful even if ARC decides we have to use x register names.

src/cheri/riscv-unpriv-integration.adoc
arichardson: I felt like it didn't really add too much to this section but we can restore it.

tariqkurd-repo: I don't mind either way - uncomment or delete

src/a-st-ext.adoc
tariqkurd-repo: yeah - let's keep cs1 until the tuesday meeting

PR #689, opened 2025-07-30
Rename load and store instructions to use Y as the data type

arichardson:
ARC requested this to ensure all RVY instruction include a Y character.
It also resolves the ARC concern that SC is very similar to the existing
store conditional instruction name.


tariqkurd-repo (approved):

arichardson:
I noticed that using `,subs=attributes+` in the wavedrom works, so we can avoid any raw uses of the instruction name and always use the macros.

tariqkurd-repo (approved):

Issue #617, opened 2025-05-28
ARC review changes since 0.9.5


tariqkurd-repo:
This issue is to track changes in the specification since 0.9.5.
There have been many extensive PRs during the review process, so spec changes can easily get lost or forgotten.

### Removal of options

1. Include Zstid in the base architecture (issue https://github.com/riscv/riscv-cheri/issues/582, PR https://github.com/riscv/riscv-cheri/pulls/588)
2. Include Zcherilevels in the base architecture, and remove >2 levels scheme (https://github.com/riscv/riscv-cheri/issues/568, https://github.com/riscv/riscv-cheri/pull/611, also see PR https://github.com/riscv/riscv-cheri/pull/631)
3. Loaded capabilities which are modified go to data trace _after_ modification (the option of tracing the raw data has been removed), (no issue, https://github.com/riscv/riscv-cheri/pull/596)
4. Remove support for dynamic XLEN, dynamic endianess, and for different XLEN or endianess between privilege levels (PR https://github.com/riscv/riscv-cheri/pull/736)

### Exceptions

1. Misaligned cap width memory accesses take access faults, not misaligned faults as they are not to be emulated (no issue, https://github.com/riscv/riscv-cheri/commit/12d8002ee5592443178278c23552f3485b9deb18)
2. Remove exceptions on branches, so that faults are taken at the target (issue https://github.com/riscv/riscv-cheri/issues/636, PR https://github.com/riscv/riscv-cheri/pull/644). A future extension may optionally add them back in (issue https://github.com/riscv/riscv-cheri/issues/647)
4. Add a PMA CHERI_VALID_TAG_FAULT to give the option of tag stripping or faulting to non-taggable memory (issue https://github.com/riscv/riscv-cheri/issues/638, PR https://github.com/riscv/riscv-cheri/pull/646)
5. Collapse exception types (issue https://github.com/riscv/riscv-cheri/issues/657, PR https://github.com/riscv/riscv-cheri/pull/654)
6. ASR checks are always PC.ASR (PR: https://github.com/riscv/riscv-cheri/pull/676)
7. CBO.INVAL detects PCC.ASR before any load/store checks (issue https://github.com/riscv/riscv-cheri/issues/660, PR https://github.com/riscv/riscv-cheri/pull/677)
[...]

PR #690, opened 2025-07-30
XRET should take an illegal instruction if PCC.ASR is missing, instead of PCC fault

tariqkurd-repo:
This one seems to have been an old thing left in the CSV file
I also removed pointless illegal instruction table as it's duplicate information in the main risc-v spec


tariqkurd-repo:
yes - I'm on board with that - if we say that XRET and CBO.INVAL take an illegal instruction fault if PCC.ASR is missing then it fits well with being detected in the decoder.

tariqkurd-repo:
fixes https://github.com/riscv/riscv-cheri/issues/692

src/cheri/tables.adoc
```diff
@@ -158,10 +158,3 @@ include::generated/xlen_dependent_encoding_insns_table_body.adoc[]
 NOTE: <<MODESW_CAP>>, <<MODESW_INT>> and <<SCMODE>> only exist in {cheri_cap_mode_name} if
 {cheri_int_mode_name} is _also_ present. A hart does not support the <<m_bit>>
 if it does not implement the {cheri_default_ext_name} extension.
-
-.Conditions for detecting illegal CHERI instructions
```
arichardson, line 162: Why did you remove this table? It still looks relevant?

arichardson:
It looks like there is also some text in the debug integration that needs updating:

> `As a result, the value of <<pcc>> is UNSPECIFIED in debug mode according
to this specification. The <<pcc>> metadata has no architectural effect in debug
mode. Therefore <<asr_perm>> is implicitly granted for access to all CSRs and
no _{cheri_excep_name_pc}_ is possible.`

arichardson:
And the following also needs to be illegal instruction:
```
4+| *CSR/Xret additional exception check*
| CSR*, <<MRET_CHERI>>, <<SRET_CHERI>>           | {cheri_excep_cause_pc}     | {cheri_excep_name_pc} | <<pcc>> does not have <<asr_perm>> when required for CSR access or execution of <<MRET_CHERI>>/<<SRET_CHERI>>
```

arichardson:
The title of this PR seems wrong? More like "take illegal instruction instead of CHERI PC fault"?

src/cheri/tables.adoc
tariqkurd-repo: it's mainly other illegals which are listed elsewhere in the RISC-V spec
the one CHERI exception table has all the necessary information

tariqkurd-repo:
I think one is good to go now......

arichardson:
Should probably add this to https://github.com/riscv/riscv-cheri/issues/617

src/cheri/tables.adoc
arichardson: Fair that sounds good to me.

arichardson:
> Should probably add this to #617

Done now.

PR #691, opened 2025-07-30
remove stale notes

tariqkurd-repo:


tariqkurd-repo:
I'm tracking changes relative to 0.9.5 in https://github.com/riscv/riscv-cheri/issues/617
No need to list them in the document.

PR #694, opened 2025-07-31
Move to using variables for mnemonics

tariqkurd-repo:
All non-load/store mnemonics are now set as variables so we can change them easily during ARC review
Load/store mnemonics have the same treatment in another PR


nwf:
I haven't checked exhaustively that this covers all cases, but this is probably a good plan.

tariqkurd-repo:
> I haven't checked exhaustively that this covers all cases, but this is probably a good plan.

yes - I expect the names to take a while to settle so this way we can respond easily

arichardson:
Pushed two tiny fixes, thanks for this! I'll go ahead and merge it to make it easier to settle on final names.

arichardson (approved):

PR #700, opened 2025-08-06
ISA renaming

tariqkurd-repo:
no specification changes - only ISA renaming and using the same ISA descriptions in the spreadsheet and the instruction pages


src/cheri/insns/cmv_16bit.adoc
```diff
@@ -15,8 +15,6 @@ Expansion::
 Suggested assembly syntax::
 `mv cd, cs2`
 
-NOTE: This encoding is shared with integer `c.mv`.
```
arichardson, line 18: I think it's still useful to say that we remap this encoding? There is no integer move in compressed opcode space.

src/cheri/attributes.adoc
```diff
@@ -95,61 +95,107 @@ endif::[]
...
-:SCMODE_LC:    ywmode
+:SCSS_DESC:    Capability less than comparison including {ctag}
+
+:SCMODE:       YMODEW
+:SCMODE_LC:    ymodew
+:SCMODE_DESC:  Set capability execution mode
+
+:C_MV_CAP:     C.MV
+:C_MV_CAP_LC:  c.mv
+:C_MV_CAP_DESC: Capability register copy, 16-bit encoding
```
arichardson, line 166: I don't think we can use this one since it  is confusing  compared to `mv` for integers. The addi*sp ones are non ambiguous since we are adding to so so it needs to be a pointer operand but this one I do think we need to rename.

src/cheri/insns/cmv_16bit.adoc
tariqkurd-repo: Maybe the note should say something like "In RVY c.mv copy YLEN-bit registers"

src/cheri/attributes.adoc
tariqkurd-repo: but surely they'll all the same? C.ADDI4SPN is going to refer to `sp` without stating whether it's XLEN or YLEN
so from the disassembly you won't know without looking at the surrounding code for C.ADDI4SPN, C.ADDI16SP, C.MV

tariqkurd-repo: or maybe you can argue that C.ADDI4SPN, C.ADDI16SP will be listed as ADDYI

src/cheri/insns/cmv_32bit.adoc
```diff
@@ -16,9 +16,7 @@ Mnemonic::
 `{cmv_lc} cd, cs1`
 
 Suggested assembly syntax::
-`mv cd, cs1`
-
-NOTE: The suggested assembly syntax distinguishes from integer `mv` by operand type.
+`{cmv_lc} cd, cs1`
```
marnovandermaas, line 19: Is there a difference between `CMV_LC` and `cmv_lc`?

arichardson: asciidoc is case-insensitive with the macros, I've just been using "lowercase" attribute names for the "_lc" lowercase macros.

src/cheri/attributes.adoc
```diff
@@ -95,61 +95,107 @@ endif::[]
...
-:SCMODE:       YWMODE
-:SCMODE_LC:    ywmode
+:SCSS_DESC:    Capability less than comparison including {ctag}
+
+:SCMODE:       YMODEW
+:SCMODE_LC:    ymodew
+:SCMODE_DESC:  Set capability execution mode
+
+:C_MV_CAP:     C.MV
+:C_MV_CAP_LC:  c.mv
```
```suggestion
:C_MV_CAP:     C.YMV
:C_MV_CAP_LC:  c.ymv
```

arichardson (approved):
LGTM with one small change. We also need to raise an issue for ACPERM.

PR #703, opened 2025-08-06
rename execution modes

tariqkurd-repo:
fixes https://github.com/riscv/riscv-cheri/issues/704


jrtc27:
This feels like a real case of "The missile knows where it is at all times. It knows this because it knows where it isn't."

On a more serious note, describing things by what they are not is generally less helpful than describing things by what they are. The previous names told you what they did. The new names don't so much. "CHERI-mode" also isn't accurate; both modes are CHERI, the processor is a CHERI one and CHERI is always enabled (PCC and DDC are always there constraining things), the difference is in whether loads and stores take an integer or capability operand. Which is what the existing names capture.

Issue #699, opened 2025-08-05
Introduce a CSR that defines the capability format


arichardson:
This allows future extensibility of the capability format and allows CHERIoT to be RVY+Xcheriot.

I imagine the register should be unprivileged rather than m-mode only so that userspace can detect the format easily without kernel support.

Naming: How about `ycfg`? I initially thought we would use `yfmt` but if we use a more generic name such as `ycfg` we could add other discoverable properties to the register in the future.

For now we could reserve the low N bits for format and keep space for future fields.
Not sure how many bits we need - maybe just reserve 8?

The spec can define following values of the CSR:
- 1 = current appendix in the spec
- 2 = CHERIoT

Or should we start allocating from 0?

CC: @davidchisnall @nwf @tariqkurd-repo

jrtc27:
How does anyone ever write portable code when the encoding of AUIPC can change? How do toolchains (compilers, assemblers and/or linkers) implement that? How do RELATIVE-style relocations work when you don't know the capability format?

I guess the real question is: in what world do you have code that actually dynamically queries the encoding and works with both? If never, then what's the point?

tariqkurd-repo:
Fixed by https://github.com/riscv/riscv-cheri/pull/709

PR #709, opened 2025-08-14
capability encoding version CSR

nwf:
Is this the kind of thing people had in mind?

### DIFF
--- a/src/cheri/hypervisor-integration.adoc
+++ b/src/cheri/hypervisor-integration.adoc
@@ -141,6 +141,20 @@ of the data is UNSPECIFIED.
 .Virtual supervisor thread identifier capability register
 include::img/vstidcreg.edn[]
 
+[#vsycfg,reftext="vsycfg"]
+=== Virtual Supervisor CHERI Capability Encoding (vsycfg)
+The <<vsycfg>> register is used to identify which CHERI capability encoding is used by the platform.
+The capability encoding both determines the in-memory representation of a CHERI capability
+and entails a set of extensions present atop {cheri_base_ext_name}.
+The CSR exists in the _writable_ namespace and all bits are defined to be WARL,
+but, absent any future extensions, it is expected that each implementation has exactly one legal value.
+For the meaning of the individual fields, refer to the <<uycfg>> register in the unprivileged specification.
+
+.Virtual Supervisor CHERI capability encoding CSR (`vsycfg`) format
+[#vsycfg_csr_structure]
+include::img/ycfgreg.edn[]
+
+
 === "Smstateen/Ssstateen" Integration
 The new TID bit controls access to the <<vstidc>> CSR.
 
--- a/src/cheri/riscv-unpriv-integration.adoc
+++ b/src/cheri/riscv-unpriv-integration.adoc
@@ -163,10 +163,10 @@ In a typical load/store unit, the expansion of the bounds from `{cs1}` and bound
 A compressed format is used to encode the bounds with a scheme similar to floating-point using an exponent and a mantissa.
 Therefore small exponents can allow byte granularity on the bounds, but larger exponents give coarser granularity.
 
-The bounds encoding format is defined in <<section_cap_bounds>>.
+The {cheri_base_ext_name} bounds encoding format is defined in <<section_cap_bounds>>.
 
-NOTE: Future extensions may redefine the bounds encoding format.
-  #ARC-COMMENT: if they redefine, then they wouldn't be a straightforward extension, but an incompatible alternative?#
+NOTE: Future extensions may use alternative bounds encoding formats.
+Software can query the capability format by reading the <<ycfg>> CSR.
 
 Software can query the capability bounds:
 
@@ -452,14 +452,23 @@ include::img/pccreg.edn[]
 ==== Added CSRs
 
 {cheri_base32_ext_name} and {cheri_base64_ext_name} add the YLEN-bit CSR shown in
-xref:utidc-csrnames-added[xrefstyle=short].
+xref:csrnames-added-y[xrefstyle=short] and the XLEN-bit CSR shown in
+xref:csrnames-added-x[xrefstyle=short].
 
-[[utidc-csrnames-added]]
+[[csrnames-added-y]]
 .Unprivileged capability CSRs added in {cheri_base_ext_name}
 [%autowidth,float="center",align="center",cols="<,<,<",options="header"]
 |===
 |YLEN CSR|Permissions|Description
 |<<utidc>>|RW, <<asr_perm>> required for writes, not reads|User Thread ID Capability
+|===
+
+[[csrnames-added-x]]
+.Unprivileged integer CSRs added in {cheri_base_ext_name}
+[%autowidth,float="center",align="center",cols="<,<,<",options="header"]
+|===
+|XLEN CSR|Permissions|Description
+|<<ycfg>>|WARL; see section for details|Capability encoding version
 |===
 
 [#utidc,reftext="utidc"]
@@ -508,6 +517,45 @@ used to implement a trusted stack. Whenever a compartment switch happens, the tr
 the caller and callee compartment avoid capability leaks between the two compartments.
 The trusted code can store capabilities on the trusted stack when calling out of a compartment and can install them
 when returning to the same compartment.
+=====
+
+[#ycfg,reftext="__x__ycfg"]
+===== CHERI Capability Encoding (__x__ycfg)
+
+The <<ycfg>> register is used to identify which CHERI capability encoding is used by the platform.
+The capability encoding both determines the in-memory representation of a CHERI capability
+and entails a set of extensions present atop {cheri_base_ext_name}.
+The CSR exists in the _writable_ namespace and all bits are defined to be WARL,
+but, absent any future extensions, it is expected that each implementation has exactly one legal value.
+This CSR is always present in the machine mode read/write CSR space;
+if the hart supports user mode, a `uycfg` register exists,
+if the hart supports supervisor mode, an `sycfg` register exists, and
+if the hart supports hypervisor mode, a `vsycfg` register exists.
+`sycfg` is always read-only, as is `vsycfg`, if it exists.
+`uycfg` is a read-only alias of `sycfg`, if it exists, or of `mycfg` otherwise.
+If the hypervisor extension is enabled, then `vsycfg` is WARL,
+updatable from M-mode or HS-mode with <<pcc>> granting <<asr_perm>>.
+
+The base specification defines the following values for `mycfg`, `sycfg`, and `uycfg`:
+|===
+|Version value|Encoding
+|0|Reserved (unspecified capability encoding)
+|0x001|<<app_cap_description>> capabilities, without <<section_zylevels1>> features
+|0x101|<<app_cap_description>> capabilities, with <<section_zylevels1>> features
+|0x002|Reserved for CHERIoT's (RV32-only) encoding
+|===
+
+Future extensions that change capability encoding are expected to define new values.
+Extensions that define encodings derived from <<app_cap_description>>
+are encouraged to preserve the bottom 8 bits as `0x01`.
+
+[NOTE]
+=====
+We suggest that only low-level, system software query this value directly.
+Operating systems and/or runtime environments should provide a mechanism for software
+to determine if the platform's capability encoding is backwards-compatible with a given, presumed encoding.
+Among other things, such a facility should be used as part of the platform executable loader
+to reject executables that expect unknown or known-incompatible CHERI features or capability encodings.
 =====
 
 ==== Extended CSRs
--- a/src/cheri/riscv-unpriv-integration.adoc
+++ b/src/cheri/riscv-unpriv-integration.adoc
@@ -561,8 +561,8 @@ The following values for `mycfg`, `sycfg`, and `uycfg` are defined:
 |===
 |Base Encoding|Features|Encoding
 | 0x0         | -      | Reserved (unspecified capability encoding)
-.2+|0x1       | 0x0    | <<app_cap_description>> capabilities, without <<section_zylevels1>> features
-              | 0x1   <| <<app_cap_description>> capabilities, with <<section_zylevels1>> features
+.2+|0x1       | 0x0    | <<app_cap_description, Default Encoding>> capabilities, without <<section_zylevels1>> features
+              | 0x1   <| <<app_cap_description, Default Encoding>> capabilities, with <<section_zylevels1>> features
 |0x2          | 0x0    | Reserved for CHERIoT's (RV32-only) encoding
 |===
 
--- a/src/cheri/riscv-unpriv-integration.adoc
+++ b/src/cheri/riscv-unpriv-integration.adoc
@@ -166,7 +166,7 @@ Therefore small exponents can allow byte granularity on the bounds, but larger e
 The {cheri_base_ext_name} bounds encoding format is defined in <<section_cap_bounds>>.
 
 NOTE: Future extensions may use alternative bounds encoding formats.
-Software can query the capability format by reading the <<ycfg>> CSR.
+Software can query the capability format by reading the <<uycfg>> CSR.
 
 Software can query the capability bounds:
 
@@ -468,7 +468,7 @@ xref:csrnames-added-x[xrefstyle=short].
 [%autowidth,float="center",align="center",cols="<,<,<",options="header"]
 |===
 |XLEN CSR|Permissions|Description
-|<<ycfg>>|WARL; see section for details|Capability encoding version
+|<<uycfg>>|RO; see section for details|Capability encoding version
 |===
 
 [#utidc,reftext="utidc"]
@@ -519,43 +519,20 @@ The trusted code can store capabilities on the trusted stack when calling out of
 when returning to the same compartment.
 =====
 
-[#ycfg,reftext="ycfg"]
-===== CHERI Capability Encoding (__x__ycfg)
+[#uycfg,reftext="uycfg"]
+===== CHERI Capability Encoding (uycfg)
 
-The <<ycfg>> register is used to identify which CHERI capability encoding is used by the platform.
+The <<uycfg>> register is a read-only XLEN CSR used to identify which CHERI capability encoding is used by the platform.
 The capability encoding both determines the in-memory representation of a CHERI capability
 and entails a set of extensions present atop {cheri_base_ext_name}.
-The CSR exists in the _writable_ namespace and all bits are defined to be WARL,
-but, absent any future extensions, it is expected that each implementation has exactly one legal value.
-This CSR is always present in the machine mode read/write CSR space;
-if the hart supports user mode, a `uycfg` register exists,
-if the hart supports supervisor mode, an `sycfg` register exists, and
-if the hart supports hypervisor mode, a `vsycfg` register exists.
-`sycfg` is always read-only, as is `vsycfg`, if it exists.
-`uycfg` is a read-only alias of `sycfg`, if it exists, or of `mycfg` otherwise.
-If the hypervisor extension is enabled, then `vsycfg` is WARL,
-updatable from M-mode or HS-mode with <<pcc>> granting <<asr_perm>>.
 
-.CHERI capability encoding CSR (`__x__ycfg`) format
+NOTE: This CSR is read-only, but future extensions may provide a mechanism to modify the capability encoding currently in use.
+
+.CHERI capability encoding CSR (`uycfg`) format
 [#ycfg_csr_structure]
-[bytefield]
-----
-(defattrs :plain [:plain {:font-family "M+ 1p Fallback" :font-size 24}])
-(def row-height 40)
-(def row-header-fn nil)
-(def left-margin 100)
-(def right-margin 100)
-(def boxes-per-row 32)
-(draw-column-headers {:height 20 :font-size 18 :labels (reverse (concat ["0"] (repeat 6 "") ["8" "9"] (repeat 22 "") ["XLEN-1"]))})
+include::img/ycfgreg.edn[]
 
-(draw-box "Features" {:span 24})
-(draw-box "Base Encoding" {:span 8})
-
-(draw-box "MXLEN-8"  {:span 24 :borders {}})
-(draw-box "8"        {:span 8 :borders {}})
-----
-
-The following values for `mycfg`, `sycfg`, and `uycfg` are defined:
+The following values for `uycfg` are defined:
 
 [%autowidth,options=header,cols="^,^,<"]
 |===
--- a/src/cheri/riscv-unpriv-integration.adoc
+++ b/src/cheri/riscv-unpriv-integration.adoc
@@ -519,7 +519,7 @@ The trusted code can store capabilities on the trusted stack when calling out of
 when returning to the same compartment.
 =====
 
-[#ycfg,reftext="__x__ycfg"]
+[#ycfg,reftext="ycfg"]
 ===== CHERI Capability Encoding (__x__ycfg)
 
 The <<ycfg>> register is used to identify which CHERI capability encoding is used by the platform.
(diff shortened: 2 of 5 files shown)


src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -501,16 +501,25 @@ include::img/pccreg.edn[]
...
 |===
 |YLEN CSR|Permissions|Description
 |<<utidc>>|RW, <<asr_perm>> required for writes, not reads|User Thread ID Capability
 |===
 
+[[csrnames-added-x]]
+.Unprivileged integer CSRs added in {cheri_base_ext_name}
+[%autowidth,float="center",align="center",cols="<,<,<",options="header"]
+|===
+|YLEN CSR|Permissions|Description
```
```suggestion
|XLEN CSR|Permissions|Description
```

arichardson (commented):
Yes this is generally what I was thinking. In https://github.com/riscv/riscv-cheri/issues/699 I suggested `ycfg` as the name in case we decide to add writable bits later, but maybe it's better to just have a small read-only csr rather than conflating things.

nwf:
> I think _vendor_ was the wrong word. I guess that Cambridge Uni may end up with more than one and CHERIoT? So `cheri family`? Maybe?
> 
> It just seems neater for Cambridge Uni formats to be together and CHERIoT formats to be together doesn't it?

I'm not sure it matters in any meaningful way?

> > I'd prefer it RO in the base spec and leave it to the hypervisor extension to split it into [msu]ycfg and make [us]ycfg WARL (to support VM migration; mycfg remains RO), but if there's strong preference for that already in base, so be it.
> 
> For VM migration with varying `ycfg` values we'd need to allow support multiple formats on one CPU - are we going to allow that? Sounds like a nightmare to write tests for. This why I prefer it to be writeable with no writeable fields - i.e. the only field is WPRI - with only one valid value per implementation - but there's scope to allow more values in future (at which point you'd want the higher privilege versions as you say).

That doesn't match my understanding of WPRI.  In any case, I think server-class cores are going to eventually need to understand VM migration and somehow prohibit software from seeing features that are present in the actual underlying core but not in the emulated, cross-fleet-normalized CPU.  It doesn't _have_ to involve `ycfg`, but it seems like it should.

tariqkurd-repo:
>In any case, I think server-class cores are going to eventually need to understand VM migration and somehow prohibit software from seeing features that are present in the actual underlying core but not in the emulated, cross-fleet-normalized CPU. It doesn't have to involve ycfg, but it seems like it should.

I think I'd prefer to avoid this if we can and settle on a single RV64 format..... seems like a lot less pain the long run.

nwf:
> To reiterate some of the discussion from the TG call today, we need to be careful to separate the encoding from extensions to the encoding.

I had _thought_ I was advocating for that kind of a thing and had been referring to the aggregate of encoding and extensions as a "CHERI platform", but that generally seemed to be unpopular and I thought we'd been told to squash the distinction, with "capability encoding" _determining_ the set of extensions and "facets" of CHERI (permissions, IFC, ...) that were in play.

Can you please articulate what you mean by "encoding" and "extensions to the encoding"?

jrtc27:
> > Can you please articulate what you mean by "encoding" and "extensions to the encoding"?
> 
> In the meeting with Krste, we had agreed that:
> 
> * Encodings carried semantics.
> * Changes to the encoding including new backwards-compatible versions, were _new_ encodings from the ISA.
> * Enumerating them with this CSR let us avoid needing multiple base architectures for new encoding versions.
> * If we don't do it this way, we need a new base for every new encoding version.

So if I decide to, say, allocate some of the reserved capability bits to mean "seal permission" and "otype", that's a new encoding? Even though it's done in a backwards-compatible way? That seems untenable, and contrary to how things work in ISAs. If you add a PMP which adds new semantics to the ISA that's not a new base ISA that's just a more featureful, i.e. extended, version of the existing ISA.

Anything that's specified as part of the capability format in the base ISA is the encoding. Anything that's in a later ISA extension which does not change the meaning of any of those bits but defines the meaning for reserved bits in a backwards-compatible manner is an extension to the encoding. Such as all the IFC stuff, so long as if you leave all the reserved bits as their reserved values you don't observe any new behaviour.

arichardson:
I wonder if we should split this CSR into two fields: one "base encoding" field and a "encoding features" field? E.g. "base encoding" in the LSBs and the rest of the CSR for feature bits that get allocated from the MSB? So RVY+nolevels=`1`, RVY+levels=`1|1<<31`, CHERIoTv1=`2` and CHERIoTv2=`3`.

The former would be RVY, CHERIoT, some-accelerator-format, etc. and the encoding features field could be used to identify extensions that assign meaning to reserved bits (eg. RVY+levels, vs RVY+nolevels). Similary addition of otypes could just set another feature flag but not change the "base encoding" field.

davidchisnall:
I think you will rapidly run out of feature bits.  Understanding the matrix of backwards-compatible formats out of band is likely to be the only long-term workable model.

nwf:
I suppose I'm happy with "Big CHERI with IFC and Seals" -- basically the semantics of the university machines -- and none of "Big CHERI without IFC and/or Seals" and leave things that want such a substrate for later.

davidchisnall:
> Or are we saying that a cpu with levels and a cpu with no levels are fundamentally incompatible so it may as well be a whole new encoding format?

There are two notions of compatibility that matter here:

 - Backwards compatibility in terms of software.  If I wrote software assuming A, can I run it on B?
 - Exactly compatible for multicore big.LITTLE systems.

The register should define the second.  Software can then determine the first compatibility model based on knowledge of multiple formats.  

To my mind, and based on what Krste said in the call, a CPU that does support levels and one that doesn't would have different CSR values here.

arichardson:
> > Or are we saying that a cpu with levels and a cpu with no levels are fundamentally incompatible so it may as well be a whole new encoding format?
> 
> There are two notions of compatibility that matter here:
> 
> * Backwards compatibility in terms of software.  If I wrote software assuming A, can I run it on B?
> * Exactly compatible for multicore big.LITTLE systems.
> 
> The register should define the second. Software can then determine the first compatibility model based on knowledge of multiple formats.
> 
> To my mind, and based on what Krste said in the call, a CPU that does support levels and one that doesn't would have different CSR values here.

This sounds good to me. I do wonder if we should split the CSR into "major+minor" fields to make this easier to maintain in the future?

For example the low 8 bits are CHERIoT/Big CHERI, and then the next 8 bits are the "minor/compatible encoding differences", i.e. 0=no levels, 1=levels, and in the future we can have 2=levels+otypes and so on. I think it makes sense for the minor field to just be a number rather than bitflags so we don't run out of space.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -559,6 +568,38 @@ The trusted code can store capabilities on the trusted stack when calling out of
...
+if the core supports user mode, a `uycfg` alias exists in the appropriate space,
+and, if the core supports supervisor mode, a `sycfg` alias exists, again in the appropriate space.
+
+The base specification defines the following values for <<ycfg>>:
+|===
+|Version value|Encoding
+|0|Reserved (unspecified capability encoding)
+|1|<<app_cap_description>> capabilities, with <<section_zylevels1>> features
+|2|<<app_cap_description>> capabilities, without <<section_zylevels1>> features
+|3|Reserved for CHERIoT's (RV32-only) encoding
```
```suggestion
|0x11|<<app_cap_description>> capabilities, with <<section_zylevels1>> features
|0x01|<<app_cap_description>> capabilities, without <<section_zylevels1>> features
|0x02|Reserved for CHERIoT's (RV32-only) encoding
```
If we were to split into a major+minor scheme without having to add a new diagram for this register?

nwf:
I don't think I mind a (semi-)structured `ycfg`, but it's still going to be on software to understand the interrelation of the values, in terms of "software expecting ycfg X can run on a core with ycfg Y".  I don't know that the structure actually buys us much?

arichardson (commented):
Since we are making this WALR, I guess we need to allocate it in the m-mode RW region. Should it be mycfg?

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: I think that providing that adding the 0x10 value doesn't always have to mean levels then we're good to go - i.e. 0x1, 2, 3 in the first digit can have levels, and so adding 0x10 means levels. But maybe 0x4 in the first digit can never have levels, and so adding 0x10 in that case could mean something else (if that makes any sense).

So adding 0x10 to the "first option" which for the currently defined formats is "levels", it's not saying that adding 0x10 _always_ means levels.

arichardson: Yes exactly, that is my proposal. Having the second hex digit be 0x1 (i.e. 0x10) does not mean generally mean levels, only when combined with 0x1 in the least significant hex digit does it means levels (0x011). A hypothetical future change to the CHERIoT format would be 0x12 and the second hex digit would mean something completely different.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -512,6 +521,44 @@ The trusted code can store capabilities on the trusted stack when calling out of
...
+===== CHERI Capability Encoding (ycfg)
+
+The <<ycfg>> register is used to identify which CHERI capability encoding is in use.
+The capability encoding determines both the in-memory representation of a CHERI capability
+and which CHERI features are available for software's use.
+The CSR exists in the _writable_ namespace and all bits are defined to be WARL,
+but, absent any future extensions, it is expected that each implementation has exactly one legal value.
+This CSR is always present in the machine mode read/write CSR space;
+if the core supports user mode, a `uycfg` register exists,
+if the core supports supervisor mode, a `sycfg` register exists.
```
```suggestion
if the core supports supervisor mode, an `sycfg` register exists.
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -512,6 +521,44 @@ The trusted code can store capabilities on the trusted stack when calling out of
...
+`mycfg` is always read-only.
+`uycfg` is a read-only alias of `sycfg`, if it exists, or of `mycfg` otherwise.
+If the hypervisor extension is enabled, then `sycfg` is WARL,
+updatable from M or H mode with <<pcc>> bearing <<asr_perm>>.
```
```suggestion
updatable from M or H mode with <<pcc>> granting<<asr_perm>>.
```

I've been using "granting" elsewhere in the spec, hope that's ok

tariqkurd-repo: plus H needs vsycfg, so that should be added

src/cheri/csv/CHERI_CSR.csv
```diff
@@ -46,3 +46,6 @@ direct write if address didn't change","","✔","","{cheri_default_ext_name}","U
 "stidc","0x580","","S","Read: S, Write: S, <<asr_perm>>","tag=0, otherwise undefined","Update the CSR using <<SCADDR>>.","direct write","","","","{cheri_base_ext_name}","Supervisor thread ID","","","","","","","","","","","","","","","","","","","","",""
 "vstidc","0xA80","","H","Read: VS, Write: VS, <<asr_perm>>","tag=0, otherwise undefined","Update the CSR using <<SCADDR>>.","direct write","","","","{cheri_base_ext_name}","Virtual supervisor thread ID","","","","","","","","","","","","","","","","","","","","",""
 "mtidc","0x780","","M","Read: M, Write: M, <<asr_perm>>","tag=0, otherwise undefined","Update the CSR using <<SCADDR>>.","direct write","","","","{cheri_base_ext_name}","Machine thread ID","","","","","","","","","","","","","","","","","","","","",""
+"mycfg","0x781","","M","RO","Platform specified","","","","","","","Supported CHERI capability encoding","","","","","","","","","","","","","","","","","","","","",""
+"sycfg","0x581","","S","Read: any, Write: (M or VS), <<asr_perm>>","Copies mycfg","WARL","WARL","","","","","","","","","","","","","","","","","","","","","","","","","",""
```
tariqkurd-repo: +vsycfg?

nwf: The additions to this file are erroneous, I think, and should be removed.  (`ycfg` is a XLEN-sized register, while this file describes YLEN-sized registers.)

src/cheri/riscv-unpriv-integration.adoc
nwf: As a read-only alias of `mycfg`?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -512,6 +521,45 @@ The trusted code can store capabilities on the trusted stack when calling out of
...
+Extensions that define encodings derived from <<app_cap_description>>
+are encouraged to preserve the bottom 8 bits as `0x01`.
+
+[NOTE]
+=====
+We suggest that only low-level, system software query this value directly.
+Operating systems and/or runtime environments should provide a mechanism for software
+to determine if the platform's capability encoding is backwards-compatible with a given, presumed encoding.
+Among other things, such a facility should be used as part of the platform executable loader
+to reject executables that expect unknown or known-incompatible CHERI features or capability encodings.
```
arichardson, line 556: We may get feedback that it should be m-mode only and be exposed via software to lower levels. But I think we should go with the current approach and let ARC decide if this is acceptable.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -512,6 +521,45 @@ The trusted code can store capabilities on the trusted stack when calling out of
 when returning to the same compartment.
 =====
 
+[#ycfg,reftext="ycfg"]
+===== CHERI Capability Encoding (ycfg)
+
+The <<ycfg>> register is used to identify which CHERI capability encoding is used by the platform.
+The capability encoding determines both the in-memory representation of a CHERI capability
+and which CHERI features are available for software's use.
```
```suggestion
The capability encoding determines the in-memory representation of a CHERI capability.
```
I'm not sure we should be promising here that any new "CHERI feature" gets a new value in the ycfg register. For example, adding a future CRelocate extension would be a new CHERI feature but not change the encoding.

Maybe best to be a bit less prescriptive here and just say it determines the in-memory format of the capability?

arichardson (approved):
Thanks! This looks great to me. I have one minor comment but if there are no objections I will go ahead and merge this tomorrow.

jrtc27:
I'd like to take a look at this first, hopefully tomorrow

nwf:
> This note needs removing in this PR:

Eliminating that note is, I think, part of a much more extensive refactoring that needs to be done.  Right now `<<section_cap_bounds>` is in `app-cap-description.adoc`, which should not be (normatively, anyway) referred to by the rest of the specification document.  (It's fine to point at it _informatively_, of course).

arichardson:
I think the note can be removed separately. @jrtc27 are you okay with this change?

nwf:
OK, note removed; amended and pushed.

src/cheri/riscv-unpriv-integration.adoc
nwf: How's this new prose as a compromise?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -512,6 +520,45 @@ The trusted code can store capabilities on the trusted stack when calling out of
...
+The <<ycfg>> register is used to identify which CHERI capability encoding is used by the platform.
+The capability encoding both determines the in-memory representation of a CHERI capability
+and entails a set of extensions present atop {cheri_base_ext_name}.
+The CSR exists in the _writable_ namespace and all bits are defined to be WARL,
+but, absent any future extensions, it is expected that each implementation has exactly one legal value.
+This CSR is always present in the machine mode read/write CSR space;
+if the hart supports user mode, a `uycfg` register exists,
+if the hart supports supervisor mode, an `sycfg` register exists, and
+if the hart supports hypervisor mode, a `vsycfg` register exists.
+`mycfg` is always read-only, as is `vsycfg`, if it exists.
```
```suggestion
`sycfg` is always read-only, as is `vsycfg`, if it exists.
```

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: I think we should go for it like this and get feedback.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -512,6 +520,45 @@ The trusted code can store capabilities on the trusted stack when calling out of
...
+`uycfg` is a read-only alias of `sycfg`, if it exists, or of `mycfg` otherwise.
+If the hypervisor extension is enabled, then `sycfg` is WARL,
+updatable from M or H mode with <<pcc>> granting <<asr_perm>>.
+
+The base specification defines the following values for `mycfg`, `sycfg`, and `uycfg`:
+|===
+|Version value|Encoding
+|0|Reserved (unspecified capability encoding)
+|0x001|<<app_cap_description>> capabilities, without <<section_zylevels1>> features
+|0x101|<<app_cap_description>> capabilities, with <<section_zylevels1>> features
```
jrtc27: I don't think this is going to be maintainable. If you want to capture features in this then it should be split between a top-level encoding field (i.e. effectively a base ISA in disguise) and a bitfield of feature bits (which may be top-level encoding-specific).

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -512,6 +520,45 @@ The trusted code can store capabilities on the trusted stack when calling out of
...
+|0x002|Reserved for CHERIoT's (RV32-only) encoding
+|===
+
+Future extensions that change capability encoding are expected to define new values.
+Extensions that define encodings derived from <<app_cap_description>>
+are encouraged to preserve the bottom 8 bits as `0x01`.
```
jrtc27: Not just encouraged. If it is the same base encoding, with the extension done in a compatible manner, then that *is* 0x01 and listing anything else is gratuitous.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -512,6 +520,45 @@ The trusted code can store capabilities on the trusted stack when calling out of
...
+The CSR exists in the _writable_ namespace and all bits are defined to be WARL,
+but, absent any future extensions, it is expected that each implementation has exactly one legal value.
+This CSR is always present in the machine mode read/write CSR space;
+if the hart supports user mode, a `uycfg` register exists,
+if the hart supports supervisor mode, an `sycfg` register exists, and
+if the hart supports hypervisor mode, a `vsycfg` register exists.
+`mycfg` is always read-only, as is `vsycfg`, if it exists.
+`uycfg` is a read-only alias of `sycfg`, if it exists, or of `mycfg` otherwise.
+If the hypervisor extension is enabled, then `sycfg` is WARL,
+updatable from M or H mode with <<pcc>> granting <<asr_perm>>.
```
jrtc27: H mode does not exist. HS-mode does.

jrtc27:
So who's going to do the proof-of-concept for this to demonstrate that such a CSR works and serves a purpose?

src/cheri/riscv-unpriv-integration.adoc
arichardson: This is effectively what is being done here, although we don't require the "features" field to be bitfields and it's up to the base encoding to allocate them as deemed sensible.

src/cheri/riscv-unpriv-integration.adoc
```suggestion
updatable from M-mode or HS-mode with <<pcc>> granting <<asr_perm>>.
```

src/cheri/riscv-unpriv-integration.adoc
arichardson: Once we have more "compatible extensions" and if those are representable as orthogonal feature bits, we can convert this table into a more maintainable representation.

I don't think we need to change the representation of this data until there is a combinatorics explosion. Does that sound acceptable?

src/cheri/riscv-unpriv-integration.adoc
```suggestion
Future extensions that change capability encoding must define new values.
Extensions that define encodings derived from <<app_cap_description>>
and are generally compatible with this encoding (e.g. they only assign meaning to previously reserved bits) must preserve the bottom 8 bits as `0x01`.
```

I agree that any "compatible enhancement" *must* keep the 0x01 value. Not sure how best to phrase that. @jrtc27 does the following (verbose) wording work?

arichardson:
> > > So who's going to do the proof-of-concept for this to demonstrate that such a CSR works and serves a purpose?
> > 
> > 
> > I don't think anyone has plans to PoC this as a writable CSR. But this PR addresses the ARC comment that the format must be easily queryable by SW, so I don't think we need to solve this future hypothetical use case.
> 
> And somebody needs to go write software to demonstrate that you can meaningfully use this CSR even in that capacity. I am skeptical that you would ever actually be able to write software that did anything other than "if major encoding != supported value then die".

I agree you can't do much with an unexpected value unless SW is designed to handle both. Being able to report a sensible error/panic on startup is still valuable. This addresses an ARC comment and it's only a handful of hardwired bits in a CSR, so I believe we should add the CSR even if it has limited use cases.

src/cheri/riscv-unpriv-integration.adoc
nwf: I agree that this is not ideal, but **AFAIK** it is mandated by RVI (as per https://github.com/riscv/riscv-cheri/pull/709#issuecomment-3201452144).

arichardson:
I added a new diagram to show the layout of the ycfg register. Hopefully ready to be merged now.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -512,6 +521,66 @@ The trusted code can store capabilities on the trusted stack when calling out of
 when returning to the same compartment.
 =====
 
+[#ycfg,reftext="ycfg"]
```
```suggestion
[#ycfg,reftext="__x__ycfg"]
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -512,6 +521,66 @@ The trusted code can store capabilities on the trusted stack when calling out of
...
+and entails a set of extensions present atop {cheri_base_ext_name}.
+The CSR exists in the _writable_ namespace and all bits are defined to be WARL,
+but, absent any future extensions, it is expected that each implementation has exactly one legal value.
+This CSR is always present in the machine mode read/write CSR space;
+if the hart supports user mode, a `uycfg` register exists,
+if the hart supports supervisor mode, an `sycfg` register exists, and
+if the hart supports hypervisor mode, a `vsycfg` register exists.
+`sycfg` is always read-only, as is `vsycfg`, if it exists.
+`uycfg` is a read-only alias of `sycfg`, if it exists, or of `mycfg` otherwise.
+If the hypervisor extension is enabled, then `sycfg` is WARL,
```
```suggestion
If the hypervisor extension is enabled, then `vsycfg` is WARL,
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -512,6 +521,66 @@ The trusted code can store capabilities on the trusted stack when calling out of
...
+----
+
+The following values for `mycfg`, `sycfg`, and `uycfg` are defined:
+
+[%autowidth,options=header,cols="^,^,<"]
+|===
+|Base Encoding|Features|Encoding
+| 0x0         | -      | Reserved (unspecified capability encoding)
+.2+|0x1       | 0x0    | <<app_cap_description>> capabilities, without <<section_zylevels1>> features
+              | 0x1   <| <<app_cap_description>> capabilities, with <<section_zylevels1>> features
```
```suggestion
.2+|0x1       | 0x0    | <<app_cap_description, Default Encoding>> capabilities, without <<section_zylevels1>> features
              | 0x1   <| <<app_cap_description, Default Encoding>> capabilities, with <<section_zylevels1>> features
```

we need to come up with a name for them - and ideally set it on the heading of that appendix too - then this is good to go

arichardson:
I have split this into unpriv and priv sections now. Please take another look and if there are no objections I will merge this afternoon and send the spec out for ARC review.

src/cheri/hypervisor-integration.adoc
```diff
@@ -141,6 +141,20 @@ of the data is UNSPECIFIED.
 .Virtual supervisor thread identifier capability register
 include::img/vstidcreg.edn[]
 
+[#vsycfg,reftext="vsycfg"]
+=== Virtual Supervisor CHERI Capability Encoding (sycfg)
```
```suggestion
=== Virtual Supervisor CHERI Capability Encoding (vsycfg)
```

src/cheri/hypervisor-integration.adoc
```diff
@@ -141,6 +141,20 @@ of the data is UNSPECIFIED.
...
+The <<sycfg>> register is used to identify which CHERI capability encoding is used by the platform.
```
```suggestion
The <<vsycfg>> register is used to identify which CHERI capability encoding is used by the platform.
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -510,6 +519,43 @@ The trusted code can store capabilities on the trusted stack when calling out of
...
+The capability encoding both determines the in-memory representation of a CHERI capability
+and entails a set of extensions present atop {cheri_base_ext_name}.
+
+NOTE: This CSR is read-only, but future extensions may provide a mechanism to modify the capability encoding currently in use.
+
+.CHERI capability encoding CSR (`uycfg`) format
+[#ycfg_csr_structure]
+include::img/ycfgreg.edn[]
+
+The following values for `mycfg`, `sycfg`, and `uycfg` are defined:
```
```suggestion
The following values for `uycfg` are defined:
```

it's in unpriv, so no mention of priv versions

tariqkurd-repo (approved):

PR #712, opened 2025-08-18
Fix column heading for Table 21. Encoding of architectural permissions for MXLEN=32

david-henley:
This column lists the values for bits[2:0] within each quadrant, the heading was incorrect. Using "Encoding[2:0]" is consistent with the version of the table provided when cheri_nolevels is set.


david-henley:
@arichardson Sorry only just saw this. It looks like the commit on `main` has the correct email so I guess @tariqkurd-repo fixed it.

Issue #706, opened 2025-08-06
allocate new encodings for SH[123]ADD instead of remapping existing ones


tariqkurd-repo:
The ARC feedback is that instead of remapping SH[123]ADD is to allocate SH[123]ADDY
Also for the .UW versions and ADD.UW

At the same time it makes sense to merge SH4ADDY into RVY and drop the integer version.

tariqkurd-repo:
https://github.com/riscv/riscv-cheri/pull/715

PR #715, opened 2025-08-19
delete Zish4add, allocate new encodings for sh123add

tariqkurd-repo:
SH[1234]ADDY are now part of RVY only (RV64Y only for SH4ADD)
All have new encodings
also added .UW versions

Fixes https://github.com/riscv/riscv-cheri/issues/706


arichardson:
For reference here are the non-purecap numbers:
```
sh1add,31761
sh1add.uw,2620
sh2add,59792
sh2add.uw,11373
sh3add,80030
sh3add.uw,16346
sh4add,11630
sh4add.uw,5111
```

tariqkurd-repo:
ok - I've removed add.uw - I think this one is good to go now and when merged I'll commit this PR https://github.com/riscv/riscv-opcodes/pull/328 with the equivalent changes

Issue #672, opened 2025-07-24
Future enhancement: distinct call (forward) & return (backward) sentries, and distinct call & return jumps


vmurali:
Let me preface by saying that this doesn't require any changes to the base spec and instead I am proposing this as a potential future extension to enforce stronger CFI guarantees.

A call/forward sentry is exported explicitly by a compartment/security domain and can be called from outside the compartment. A callee is expecting such a call to the start of the function at any time, and so is prepared with whatever checks it needs to do on the arguments. But a return/backward sentry is created on the fly when a new call is made so that the callee can return to the next instruction after completing the call. If these sentries are stashed by the callee and/or exposed to other threads which previously didn't have access to that sentry, then that thread can use the backward sentry to unexpectedly jump to the corresponding address - the caller that created the backward sentry has no way to defend against such an attack.

This attack would be prevented if a backward sentry cannot be stashed in the global memory and hence cannot be passed to other threads. This is similar to the Local/Global mechanism (where Local caps can only live in the stack, and would be destroyed with the stack on a function return). In order to implement such a guardrail, one has to distinguish between call/forward sentries and return/backward sentries - possibly using otypes (similar to how CHERIoT does this, ignoring the interrupt disposition in CHERIoT).

Once such a distinction is made between call/forward sentries and return/backward sentries, it would be good to have explicit instructions to distinguish between call jumps and return jumps. If there are no new instructions, then the jump instruction's intent becomes ambiguous when the link register is `c0`: the instruction does not know if it's a call jump or a return jump and instead relies on the data present in the register to reason about the jump. While this doesn't pose any immediate problems with the current RISCV-CHERI spec, it prevents CHERIoT-like extensions such as changing the interrupt disposition (see https://github.com/CHERIoT-Platform/cheriot-sail/issues/100#issue-3133336496 and https://github.com/CHERIoT-Platform/cheriot-sail/pull/101 for a detailed explanation of why this is a problem).

The opcode for JALR has a funct3 field (bits 12 to 14) of value 0. Another value can be reserved for a backward jump.

vmurali:
Right - that's another thing we should talk about (in the future); I am not convinced not having "pseudolocality" is a good idea :).

tariqkurd-repo:
Generic JALR support forward and backward-edge sentries - so closing this - please reopen if necessary

PR #716, opened 2025-08-20
DNM WIP: capability mediated sealing

nwf:
**This is an incomplete draft, raised for early review to assess screams and any inbound vegetable projectiles**

- Remove sentries from RVnnY bases
- Define `cbld` to always build unsealed forms
- Introduce Zyseal and Zysentry parametric extensions that particular encodings can mix in
- Zyseal is parametric in which abstract capability types the encoding can actually encode (other than the mandatory `0`).
- Almost all of the interesting aspects of Zysentry's behavior is delegated to the capability encoding.

The concrete changes to the "big CHERI" encoding section are not yet present; imagine them here in in the most straightforward way.


src/cheri/zyseal_amb.adoc
```diff
@@ -0,0 +1,33 @@
...
+
+Capability encodings may define a set of <<sec_cap_type>> values that do not
+need capability-based authority to be used as seals.
+Thus, the right to seal at these types (recall <<section_zyseal>>)
+becomes "ambiently available" to all software in the system.
+
+Concretely, extensions may define either or both:
+
+* One or more <<sec_cap_type>> values that will be used for <<YSEAL>>
+  instructions with `null` authority:
```
vmurali, line 14: If I understand this correctly, ambient sealing cannot be represented with just encoding changes using YSEAL and YUNSEAL. That is, null in ambient sealing doesn't need to have a valid cap, and doesn't represent any bounds?

src/cheri/zysentry.adoc
```diff
@@ -0,0 +1,67 @@
+[#section_zysentry,reftext="{cheri_sentry_ext_name}"]
+== "{cheri_sentry_ext_name}" Extension for CHERI Sentry Capabilities
+
+[#sentry_cap,reftext="sentry capability"]
+=== Sentry Capabilities
+
```
vmurali, line 6: I think this requires defining another extension to add a new backward jump type to keep the semantics of return sentries simple (as discussed in https://github.com/riscv/riscv-cheri/issues/672

vmurali (approved):
LGTM but would like to add the new-instruction-for-return extension also in the future

src/cheri/zyseal_amb.adoc
nwf: Right.  YSEAL ordinarily uses the address field of the authority to determine the type of the resulting capability's seal.

vmurali: Hmm that means a null cap should necessarily have a seal type of 1 (that means c0 cannot be used) or seal type of 0 should mean sealed which is problematic. Don't you mean for ambient extension you just use c0 as the authority that magically has the required authority even if it's not a valid cap? Or do you mean something else?

If it's indeed the case that c0 magically has all the required authority even when it's not a valid cap, it essentially means this extension doesn't need to force YSEAL to take authority, right? Sorry, I am not sure I completely follow this extension's semantics as written.

src/cheri/zysentry.adoc
nwf: The intent is that Zysentry is the minimal architectural support needed for a variety of future semantics (see, for example, the section on sentries over in `xcheriot.adoc` of #743) while also being possible to trivially instantiate with only one sealing type as per the existing RISC-V CHERI specification's approach.

src/cheri/zyseal_amb.adoc
nwf: Correct: zyseal_amb takes instructions that would certainly return an untagged cap (`yseal rd, null, cs2` or `yseal rd, cs1, cs2` with untagged `cs1`) and redefines their semantics.  The 2nd option is nic_er_, or at least less invasive, in that it preserves the "address becomes otype" aspect of `yseal`; the 1st option does not attempt to interpret the all-zeroes capability bit pattern and instead reinterprets the meaning of "register 0".  This is, at least somewhat, akin to making loads and stores through `null` authority -- which cannot mean anything in CHERI and are pretty dubious even in the base integer-address architecture -- instead mean some other authority (say, `mtidc`).

src/cheri/zysentry.adoc
vmurali: Sure; just saying it would be good to add the other extension (that adds a new instruction) also.

src/cheri/zyseal_amb.adoc
vmurali: Reread the section again. It looks like for certain sealing types (as opposed to certain cs1 register indices), an untagged authority with arbitrary address would suffice. Is that what you rely on for using c0 for "instruction sentry" sealing?

nwf:
Once #761 lands, I will rebase this to instead keep that sentry extension available alongside the Zysealambient herein.

Issue #705, opened 2025-08-06
make AND-PERM into CLEAR-PERM


tariqkurd-repo:
The suggestion is to change the semantics from AND-PERM to CLEAR-PERM to be consistent with CSRRC

This means that the mask is inverted - so et the bits you want to remove instead of setting the bits you want to keep

PR #721, opened 2025-08-20
change ACPERM to bit clearing

tariqkurd-repo:
https://github.com/riscv/riscv-cheri/issues/705


nwf:
Does this mean we can do away with the "reserved 0" and "reserved 1" distinction in the bitfield?

arichardson:
@davidchisnall for visibility. 

I think this is a good change. Clear and mask are the same since we can just invert the mask and from what I've seen so far removing W+LM or similar seems more common than masking with a specific set of permissions.

One benefit of moving to clearing is that we are able to clear more combinations of permissions vs and-perm where we have to ensure all other bits of the register are set to 1.
Quoting from @LawrenceEsswood at https://github.com/riscv/riscv-cheri/pull/351#issuecomment-2327088281, we can now
> clear in [0..4] with a c.li+acperm
> clear in [12..16] with a c.lui + acperm
> clear in [12..30] with a lui + acperm
> clear in [0..10] with a li + acperm

With andperm we can only do
- clear in [0..4] with a c.li+acperm (since bit 5 needs to be 1 to avoid affecting others)
- clear in [0..10] with a li + acperm (since bit 11 needs to be 1 to avoid affecting others)
The two lui ones can't be used since it would clear all lower permission bits.

src/cheri/insns/acperm_32bit.adoc
```diff
@@ -16,8 +16,6 @@ include::new_encoding_note.adoc[]
 Synopsis::
 {ACPERM_DESC}
```
arichardson: I think we should probably also change the macro names from ACPERM to CLRPERM or similar to show that it's different now.

tariqkurd-repo: Yes, as it's not a simple renaming of ACPERM any more

arichardson:
> > One benefit of moving to clearing is that we are able to clear more combinations of permissions vs and-perm where we have to ensure all other bits of the register are set to 1.
> 
> I don't think that's true: the top bits the sign-extending load-immediates had better be `0` for use with bit-clearing-ACPERM just like they'd needed to be `1` for bit-ANDing-ACPERM, surely? What am I missing?

Yes, that's why the bit 5/11 of (c.)li need to be zero for clearing, but it allows use of (c.)lui since the low bits are now zero.

PR #726, opened 2025-08-20
Small typo and xref fixes

arichardson:


src/cheri/insns/gchi_32bit.adoc
```diff
@@ -13,7 +13,7 @@ Encoding::
 include::wavedrom/gchi.adoc[]
 
 Description::
-Logical right shift of Y register `rs1`, writing the lower YLEN bits of the result to `rd`.
+Logical right shift of Y register `rs1`, writing the lower XLEN bits of the result to `rd`.
```
tariqkurd-repo: YLEN is intended - to be an arbitrary right shift in future

arichardson: But then there is no need to talk about "the lower YLEN bits"? It's just zero-filling the higher bits?

arichardson: How about the new wording?

PR #727, opened 2025-08-20
Rename YBNDSIW to YBNDSWI

arichardson:
This matches CSRW/CSRWI/CSRCI and all other instructions with immediates that follow the `<mnemonic>I<size>` pattern (e.g. ADDIW).

### DIFF
--- a/src/cheri/attributes.adoc
+++ b/src/cheri/attributes.adoc
@@ -107,8 +107,8 @@
 :SCBNDS_LC:    ybndsw
 :SCBNDS_DESC:  Write capability bounds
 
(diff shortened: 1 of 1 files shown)


tariqkurd-repo:
it's not `<mnemonic>I<size>` Where is Write not Word in this case
so it's still a bit inconsistent
I'm sure ARC will have a say in this

tariqkurd-repo:
actually this is consistent with CSRRWI so why not 😄

tariqkurd-repo (approved):
I'll update riscv-opcodes

PR #728, opened 2025-08-20
massive operand renaming

tariqkurd-repo:
cs1->rs1
cs2->rs2
cd->rd
cn->xn
csp->sp

I've reverted some wavedrom files back to the copy from riscv-isa-sim as the whitespace ended up different (as we no longer are writing rs1/cs1)


src/c-st-ext.adoc
```diff
@@ -548,14 +544,14 @@ These instructions use the CR format.
 
 C.JR (jump register) performs an unconditional control transfer to the
 address in register _rs1/cs1_. C.JR expands to `jalr x0, 0(rs1)`, or
-`jalr {creg}0, 0(cs1)` for {cheri_base_ext_name}.
+`jalr {creg}0, 0({creg_pref}s1)` for {cheri_base_ext_name}.
```
arichardson: Aren't these the same now? Both expand to `jalr x0, 0(rs1)`

tariqkurd-repo: yep - it _is_ a lot simpler - will fix

arichardson (approved):
Made a few fixes, but LGTM now.

PR #729, opened 2025-08-20
Fix permission compression with Zylevels1

nwf:
Some of the rows in the encoding table still had multi-bit SLs.

### DIFF
--- a/src/cheri/app-cap-description.adoc
+++ b/src/cheri/app-cap-description.adoc
@@ -196,7 +196,7 @@ is supported, otherwise such encodings are reserved. Despite being encoded here 
 
 When MXLEN=32, this compressed permission format introduces
 new dependencies between permisison bits.
-When <<section_zylevels1>> _is not_ present, the following new rows are added to <<acperm_rules>> table:
+When <<section_zylevels1>> _is not_ present, the following new rows are added to <<acperm_rules>>:
 
 [float="center",align="center",cols="2,2,4",options="header"]
 |===
@@ -206,7 +206,7 @@ When <<section_zylevels1>> _is not_ present, the following new rows are added to
 |===
 
 
-When <<section_zylevels1>> _is_ present, the following new rows are added to <<acperm_rules>> table:
+When <<section_zylevels1>> _is_ present, the following new rows are added to <<acperm_rules>>:
 
 [float="center",align="center",cols="2,2,4",options="header"]
 |===


arichardson (commented):
Thanks this looks good. I do wonder if we should re-encode the RV32 permissions now that 2-bit levels is definitely out?

@LawrenceEsswood for visibility.

src/cheri/app-cap-description.adoc
```diff
@@ -175,30 +175,27 @@ orthogonal to permissions as it can vary arbitrarily using <<SCMODE>>.
 | 4-5   | ✔ | ✔ | ✔ | ✔  | ✔  | 1   | ✔ |     | Mode^1^  | Execute + Data & Cap RW
 | 6-7   | ✔ | ✔ |   |    |    | 0^2^| ✔ |     | Mode^1^  | Execute + Data RW
 11+| *Quadrant 2: Restricted capability data read/write*
-11+| bit[2] = write, bit[1:0] = store level. R and C implicitly granted, LM dependent on W permission.
+11+| bit[2] = write, bit[1] reserved, bit[0] = SL. R and C implicitly granted, LM dependent on W permission.
 |Encoding[2:0]| R | W | C | LM | LG | SL    | X | ASR | Mode^1^ |
 | 0-2   10+| reserved
-| 3       | ✔ |   | ✔ |    |    | 0^1^  |   |     | N/A | Data & Cap R0 (without <<lm_perm>>)
```
arichardson, line 181: Looks like the change from bits[]->encoding has made this table misaligned and a bit harder to read. I can fix this but don't want to conflict.

src/cheri/app-cap-description.adoc
```diff
@@ -169,36 +169,30 @@ orthogonal to permissions as it can vary arbitrarily using <<SCMODE>>.
...
+11+| bit[2] = write, bit[1] reserved, bit[0] = SL. R and C implicitly granted, LM dependent on W permission.
 | 0-2   10+| reserved
-| 3       | ✔ |   | ✔ |    |    | 0^1^  |   |     | N/A | Data & Cap R0 (without <<lm_perm>>)
-| 4       | ✔ | ✔ | ✔ | ✔  |    | _(3)_ |   |     | N/A | _Reserved_^3^
-| 5       | ✔ | ✔ | ✔ | ✔  |    | _(2)_ |   |     | N/A | _Reserved_^3^
-| 6       | ✔ | ✔ | ✔ | ✔  |    | 1     |   |     | N/A | Data & Cap RW (with <<zylevels1_sl_perm>>, no <<zylevels1_lg_perm>>)
-| 7       | ✔ | ✔ | ✔ | ✔  |    | 0     |   |     | N/A | Data & Cap RW (no <<zylevels1_sl_perm>>, no <<zylevels1_lg_perm>>)
+| 3       | ✔ |   | ✔ |    |    | 0     |   |     | N/A | Data & Cap R0 (without <<lm_perm>>)
+| 4       | ✔ | ✔ | ✔ | ✔  |    | 0     |   |     | N/A | Data & Cap RW (no <<zylevels1_sl_perm>>, no <<zylevels1_lg_perm>>)
+| 5       | ✔ | ✔ | ✔ | ✔  |    | 1     |   |     | N/A | Data & Cap RW (with <<zylevels1_sl_perm>>, no <<zylevels1_lg_perm>>)
```
arichardson: hmm I wonder if we should keep the existing order to avoid changes relative to 0.9.5. We were storing storelevel as the negated value, so it's really ~bit[1].

nwf:
There's going to need to be _some_ changes made for #716, but I was going to do that as part of that PR and have this just be correcting things left from #631.

src/cheri/app-cap-description.adoc
```diff
@@ -169,40 +169,34 @@ orthogonal to permissions as it can vary arbitrarily using <<SCMODE>>.
...
 11+| *Quadrant 3: Capability data read/write*
-11+| bit[2] = write, bit[1:0] = store level. R and C implicitly granted.
+11+| bit[2] = write, bit[1] reserved, bit[0] = !SL. R and C implicitly granted.
 11+| _Reserved bits for future extensions must be 1 so they are implicitly granted_
-|Encoding[2:0]| R | W | C | LM | LG | SL    | X | ASR | Mode^1^ |
 | 0-2   10+| reserved
 | 3       | ✔ |   | ✔ | ✔  | ✔  | 0^2^  |   |     | N/A | Data & Cap R0
-| 4       | ✔ | ✔ | ✔ | ✔  | ✔  | _(3)_ |   |     | N/A | _Reserved_^3^
-| 5       | ✔ | ✔ | ✔ | ✔  | ✔  | _(2)_ |   |     | N/A | _Reserved_^3^
+| 4-6   10+| reserved
```
arichardson, line 188:
```suggestion
| 4-5   10+| reserved
```

arichardson (approved):
One typo I think otherwise ready to merge

src/cheri/app-cap-description.adoc
arichardson: @nwf do you plan to push an updated version or should I apply the suggestion and merge this PR?

nwf:
This branch conflicts with #731.  In the interest of getting bigger changes in first, I've just pushed this sat atop that.

arichardson (changes requested):
LGTM, but marking as request changes so we don't accidentally merge this before #731.

arichardson (approved):

PR #730, opened 2025-08-21
define new ISA extensions

tariqkurd-repo:
Define new extensions for RVYI (the base), RVYIC (cmpressed) etc.
I've done separate extensions for _modified_ instructions e.g. AUIPC - not sure if this is needed


tariqkurd-repo:
> Why are these all new extensions? If they're part of the base then they're not extensions. This seems to be a major step backwards (and also against the expressed desire to not have a proliferation of extensions, though these new "extensions" don't look like they should be anyway). So a big -1 from me on the direction of this so far.

This was an explicit request from Krste to define extensions, although previously we was talking about minising them. He also agreed that the levels should be an extension for example. I think this shows a lot of progress with getting the understanding across of what we're trying to achieve.

For the _modified_ instructions, there was no such request, but... we had criticism before about it not being clear _which_ instructions had modified behaviour so I thought it was better this way (although we still have the blanket statement of all loads and stores) but the others are explicit now. Those aren't extensions really, just lists. I'm sure ARC will have an opinion on the terminology.

So for an RVY system you get all `RVI instructions + new RVYI instructions + modify some RVI instructions`, and for a hybrid system you add the additional `Hybrid` extension on top
With optional `levels` too.

jrtc27:
> > Why are these all new extensions? If they're part of the base then they're not extensions. This seems to be a major step backwards (and also against the expressed desire to not have a proliferation of extensions, though these new "extensions" don't look like they should be anyway). So a big -1 from me on the direction of this so far.
> 
> This was an explicit request from Krste to define extensions, although previously we was talking about minising them. He also agreed that the levels should be an extension for example. I think this shows a lot of progress with getting the understanding across of what we're trying to achieve.
> 
> For the _modified_ instructions, there was no such request, but... we had criticism before about it not being clear _which_ instructions had modified behaviour so I thought it was better this way (although we still have the blanket statement of all loads and stores) but the others are explicit now. Those aren't extensions really, just lists. I'm sure ARC will have an opinion on the terminology.

Yeah but that's my point, they're not extensions, they are part of RVY. Categorising *within* the specification of RVY is fine, but there should be absolutely no confusion about them being part of RVY.

There's also the separate issue that this is now making optional some features which should never be optional. CHERI without a sealing mechanism is not something that should be allowed, for example.

> So for an RVY system you get all `RVI instructions + new RVYI instructions + modify some RVI instructions`, and for a hybrid system you add the additional `Hybrid` extension on top With optional `levels` too, and optional `Zyba` (reflecting that not all cores have `Zba`).
?
> The other comment is that I've created a CHERI version of other extensions, so for normal RISC-V you get `Zba` but for CHERI you need `Zyba`. Also for the others we've defined like CBO.FLUSH, they have an equivalent extension: `Zcbo* / Zycbo*`.

This is untenable. We can't create a new name for every extension that exists in the wild. It's going to get silly at every level: architecture, toolchain and software stacks on top. You don't have separate Zba and Zeba, you have just Zba, and it is a given that Zba on RVE means the E version of Zba. Otherwise you're going to end up in an absurd situation where you need to duplicate everything, and on the software side you're going to need to do moronic things like `#if defined(__riscv_zba) || defined(__riscv_zyba)` which serves absolutely zero good other than to require adding pointless code to all manner of third-party packages and make everyone miserable in the process.

> We can ask ARC for opinions before merging if that's helpful (replies can be very slow though - what they're expecting is an updated spec to review).

I will vote no on any specification that requires a new name for every existing ISA extension.

tariqkurd-repo:
>There's also the separate issue that this is now making optional some features which should never be optional. CHERI without a sealing mechanism is not something that should be allowed, for example.
Agreed - this is a precursor to @nwf 's next PR with a solution for sealing, from the dsicussion you missed last week

>Yeah but that's my point, they're not extensions, they are part of RVY.
So what is RVY? That what I'm trying to resolve:

1. _all_ of RVI base (or RVE base)
2. it's the new instructions we've added (which I'm currently calling RVYI)
3. It's the modified instructions (which I'm currently calling RVYI_MOD - bad name this I know)

So for the toolchain - what do you actually need? A single definition of RVY which includes those three things I guess.

(also the small extension for hybrid mode of course)

>This is untenable. We can't create a new name for every extension that exists in the wild.

Yes ok - makes sense - and we've stopped redefining 32-bit instructions too.

We do need a 16-bit extension for C.LY etc., which goes with Zca

We have a blanket statement about _all_ memory accesses being authorized, so I think that is clear and understood.

That leaves:

1. 16-bit encodings which we are redefining e.g. C.ADDI16SP - the compiler must know about this as they _are_ different
2. CMOs which we are adding permission checks too (CBO.FLUSH etc.) - in this case only the programmer cares - the compiler has no knowledge


It's all about how to represent this informatin clearly so people know what we're trying to achieve, it's not simple.

arichardson:
I don't think we need new extensions for the modified instructions, but we need named extensions for new instructions that aren't part of the base, such as the AMO ones.

I don't think any of those should be optional though! I think we just need to say, if you have RVY and Zaamo, then you must also implement Zyaamo (AMOSWAP.Y).

jrtc27:
RV64I's Zaamo includes AMOSWAP.D but RV32I's Zaamo does not. I don't see why RV64Y's Zaamo shouldn't also include AMOSWAP.Y in the exact same vein.

arichardson:
> Also please do something about the commit history in here, it's unreviewable

I'll take a go at cleaning this up this afternoon.

tariqkurd-repo:
> Huge PR, still not managed to get through it all. I think the new tables are great and we should definitely keep them, but not sure we should be removing GCMODE/SCMODE from hybrid?

resolve in different PR for hybrid mode

tariqkurd-repo:
> You can’t just disagree with me and go ahead and merge. I do not agree with your rebuttal.

that wasn't the intention - it was to put this mammoth PR to bed and continue with a new one - there was so much dead wood in the spec which needed to be trimmed which was exposed as part of making the lists of instructions - and I just had to get it on the main line

jrtc27:
The better approach is to not entangle the two so that the non-contentious cleanups can land even earlier and the contentious bits are more limited, and stay unmerged until consensus is reached.

Issue #391, opened 2024-10-01
Multi-root capability systems vs. The Infinite Cap and CSR initialization


nwf:
## Problem Statement

At present, the Zcheripurecap draft presumes the existence of a single all-permissive Infinite cap (§2.3.2).  This is incompatible with multi-root capability encodings (CHERIoT is once again the motivating example).  This could have just been a nuisance, as in many cases the specification is careful to test relevant permissions or phrase certain comparisons as "ha[ving] infinite bounds" or similar, but then...

1. The spec requires that the following CSRs can initialize to this Infinite capability: `pcc` (§3.2.1), `mtvecc` (§3.7.3), `mepcc` (§3.7.7), `stvecc` (§3.8.2, if S), `sepcc` (§3.8.6, if S), `dinfc` (§3.14.9, if Sdext), `ddc` (§5.8.7, if Zcherihybrid), `vstvecc` (§6.7, if H), `vsepcc` (§6.11, if H).

2. In Zcherihybrid and with Sdext, `ddc` becomes Infinite on entry to debug mode, with the new CSR `dddc` holding the old value (§5.6).

There is some provision for multiple roots in an informative note in §3.14.9:
> A future version of this specification may add writeable fields to allow creation of other capabilities, if, for example, a future extension requires multiple formats for the Infinite capability.

However, I think some more broadly-scoped verbiage is appropriate, and in particular, there should be provision for other initialization behaviors.  A hypothetical Zcheriot would like something slightly tweaked from CHERIoT's initialization, say...
- `pcc` initializes to the executable memory root (with cursor address set to the boot offset; the Infinite-like form, with `cursor = base`, can be obtained with an `AUIPC 0; SCADDR` sequence).
- `mtvecc` initializes to the read-write memory root
- `mepcc` initializes to the sealing type root

## Straw proposal

I think I'd like to have §2.3.2 define Root Capabilities and The Infinite Capability, when it exists, something like this:

> Capabilities form a partial order, defined by subset inclusion on authorized actions.  A **root capability** is maximal in this partial order; a root capability is not an attenuated form of another capability.  In general, capability encodings (recall [the note before 2.1; that probably wants a name and link target]) may have several root capabilities, but many, including the encoding of §2.1, have exactly one, which we call the **Infinite** capability.

We can then require CSR initialization to _root_ capabilities:

* In §3.2.1,
[...]

PR #731, opened 2025-08-21
Primordial, not Infinite, capabilities

nwf:
Rampage through the spec and attempt to expunge the notion of a singular infinite capability in favor of explicit requirements.

Partially addresses https://github.com/riscv/riscv-cheri/issues/391

### DIFF
--- a/src/cheri/debug-integration.adoc
+++ b/src/cheri/debug-integration.adoc
@@ -57,7 +57,7 @@ ebreak
 
 # Write the {ctag} (if nonzero)
 csrrw   {creg}2, dscratch0, {creg}2
-csrr    {creg}2, dinfc
+csrr    {creg}2, drootc
 {cbld_lc}    {creg}1, {creg}2, {creg}1
 csrrw   {creg}2, dscratch0, {creg}2
 ebreak
@@ -126,7 +126,7 @@ Upon entry to debug mode, the _RISC-V Debug Specification_, does not specify how
 update the PC, and says PC relative instructions may be illegal. This concept
 is extended to include any instruction which reads or updates <<pcc>>, which refers to
 all jumps, conditional branches and <<AUIPC_CHERI>>. The exceptions are <<MODESW_CAP>> and <<MODESW_INT>>
-which _are_ supported if {cheri_default_ext_name} is implemented, see <<dinfc>>
+which _are_ supported if {cheri_default_ext_name} is implemented, see <<drootc>>
 for details.
 
 As a result, the value of <<pcc>> is UNSPECIFIED in debug mode according
@@ -164,30 +164,44 @@ The <<dscratch1>> register is extended to hold a capability.
 .Debug scratch 1 capability register
 include::img/dscratch1creg.edn[]
 
-[#dinfc,reftext="dinfc"]
-==== Debug Infinite Capability Register (dinfc)
+[#drootcsel,reftext="drootcsel"]
+==== Debug Root Capability Selector (drootcsel)
 
-<<dinfc>> is a debug mode accessible capability CSR. The address and access details are shown in
+<<drootcsel>> is a debug mode accessible integer CSR.  The address and access details are shown in
 xref:default-csrnames-added[xrefstyle=short].
 
-The reset value is the <<infinite-cap>> capability.
+It selects which <<root-cap>> capability is exposed through <<drootc>>.
+The reset value is `0`, which must cause <<drootcsel>> to expose a
+<<root-rx-cap>> capability.
 
-If {cheri_default_ext_name} is implemented:
+Other capability values may be defined for exposure through <<drootc>>
+by the capability encoding,
+and may be selected by having the debugger write to this register.
+Writes are WARL, so the debugger may confirm that its selection has been applied.
+
+.Debug root capability register
+include::img/drootcselreg.edn[]
+
+[#drootc,reftext="drootc"]
+==== Debug Root Capability Register (drootc)
+
+<<drootc>> is a debug mode accessible capability CSR. The address and access details are shown in
+xref:default-csrnames-added[xrefstyle=short].
+It exposes the capability selected by <<drootcsel>>.
+
+If {cheri_default_ext_name} is implemented,
+the <<root-rx-cap>> exposed when <<drootcsel>> is `0` is further specified thus:
 
 * The <<m_bit>> is reset to {cheri_int_mode_name} ({INT_MODE_VALUE}).
 * The debugger can set the <<m_bit>> to {cheri_cap_mode_name} ({CAP_MODE_VALUE}) by executing <<MODESW_CAP>> from the program buffer.
 ** Executing <<MODESW_CAP>> causes subsequent instructions execution from the program buffer, starting from the next instruction, to be executed in {cheri_cap_mode_name}. It also sets the CHERI execution mode to {cheri_cap_mode_name} on future entry into debug mode.
 ** Therefore to enable use of a CHERI debugger, a single <<MODESW_CAP>> only needs to be executed once from the program buffer after resetting the core.
-** The debugger can also execute <<MODESW_INT>> to change the mode back to {cheri_int_mode_name}, which also affects the execution of the next instruction in the program buffer, updates the <<m_bit>> of <<dinfc>> and controls which CHERI execution mode to enter on the next entry into debug mode.
+** The debugger can also execute <<MODESW_INT>> to change the mode back to {cheri_int_mode_name}, which also affects the execution of the next instruction in the program buffer, updates the <<m_bit>> of this capability and controls which CHERI execution mode to enter on the next entry into debug mode.
 
-The <<m_bit>> of <<dinfc>> is _only_ updated by executing <<MODESW_CAP>> or <<MODESW_INT>> from the program buffer.
+The <<m_bit>> of this capability is _only_ updated by executing <<MODESW_CAP>> or <<MODESW_INT>> from the program buffer.
 
-NOTE: A future version of this specification may add writeable fields to allow creation
- of other capabilities, if, for example, a future extension requires multiple formats for
- the <<infinite-cap>> capability.
-
-.Debug infinite capability register
-include::img/dinfcreg.edn[]
+.Debug root capability register
+include::img/drootcreg.edn[]
 
 [#section_hybrid_debug_integration]
 === Integrating {cheri_default_ext_name} with Sdext
@@ -198,15 +212,15 @@ shown in xref:default-csrnames-added[xrefstyle=short].
 {cheri_default_ext_name} allows <<MODESW_CAP>> and <<MODESW_INT>> to execute in debug mode.
 
 When entering debug mode, whether the core enters {cheri_int_mode_name} or
-{cheri_cap_mode_name} is controlled by the <<m_bit>> in <<dinfc>>.
+{cheri_cap_mode_name} is controlled by the <<m_bit>> in the <<drootc>> capability selected by <<drootcsel>> value 0.
 
-The current mode can be read from <<dinfc>>.
+The current mode can be read by setting <<drootcsel>> to `0` and then reading <<drootc>>.
 
 The following sequence executed from the program buffer will write {CAP_MODE_VALUE} for {cheri_cap_mode_name} and {INT_MODE_VALUE} for {cheri_int_mode_name} to `x1`:
 
 [source,subs=attributes+]
 ----
-csrr   {creg}1, dinfc
+csrr   {creg}1, drootc
 {gcmode_lc} x1, {creg}1
 ----
 
@@ -228,9 +242,8 @@ xref:default-csrnames-added[xrefstyle=short].
 .Debug default data capability
 include::img/dddcreg.edn[]
 
-Upon entry to debug mode, <<ddc>> is saved in <<dddc>>. <<ddc>>'s metadata is
-set to the <<infinite-cap>> capability's metadata (with tag set) and <<ddc>>'s
-address remains unchanged.
+Upon entry to debug mode, <<ddc>> is saved in <<dddc>>. <<ddc>> is set to a
+<<root-rw-cap>> capability such that <<ddc>>'s address remains unchanged.
 
 When debug mode is exited by executing <<DRET_CHERI>>, the hart's <<ddc>> is updated to
 the capability stored in <<dddc>>. A debugger may write <<dddc>> to change the
--- a/src/cheri/img/drootcreg.edn
+++ b/src/cheri/img/drootcreg.edn
@@ -6,17 +6,17 @@
 (def left-margin 100)
 (def right-margin 100)
 (def boxes-per-row 34)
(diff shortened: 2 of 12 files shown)


src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -353,17 +354,35 @@ NOTE: This property is required to ensure restricted programs cannot forge capab
...
+(and so on the primordial capabilities held therein).
+
+Because particular sets of requirements recur throughout the specification,
+we define some useful short-hand terminology.
+
+[#root-rx-cap,reftext="Root Executable"]
+Root Executable Capability::
+A capability that has bounds covering all addresses
+and bears at least all of <<x_perm>>, <<r_perm>>, <<c_perm>>, <<lm_perm>>, and <<asr_perm>>.
+Extensions may also specify additional requirements.
```
arichardson: Requirements sounds ambiguous to me. Is the intention behind this sentence that an extension could say they have fewer permissions or smaller bounds? Or do you mean if we have a new permission relevant to executable caps then that must also be present?

If it's the latter I would just drop the "Extensions may also specify additional requirements.", since you did say it needs to have at least those permissions. Just to keep this simpler for readers.

EDIT: Just saw your change further down the diff, maybe the simpler phrasing something like "Extensions introducing new permissions may require these to be provided by the root executable capability."? And the same for the data one?
I think this ensures there is no ambiguity regarding "requirements".

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -353,17 +354,35 @@ NOTE: This property is required to ensure restricted programs cannot forge capab
...
+In some systems (and capability encodings), there is a single "Infinite" capability value,
+which grants all permissions and has bounds covering the whole 2^MXLEN^ address space;
+in such systems, primordial capabilities are often Infinite.
+More generally, the _set_ of primordial capabilities collectively grant all permissions to all addresses.
+By way of example, an encoding may prohibit one capability from authorizing both write and execute;
+a system using such an encoding would typically make available maximally permissive
+read-write and read-execute capabilities as part of their primordial set.
+How unprivileged software receives its primordial capabilities is largely an ABI question;
+the privileged specification will detail requirements of capability registers' reset state
+(and so on the primordial capabilities held therein).
```
```suggestion
[NOTE]
====
How unprivileged software receives its primordial capabilities is largely an ABI question;
the privileged specification will detail requirements of capability registers' reset state
(and so on the primordial capabilities held therein).
====
```

I'd make this a non-normative note.

src/cheri/zylevels1.adoc
```diff
@@ -65,10 +65,16 @@ The _Capability Global_ flag holds one of two values:
 
 As with permissions, the _Capability Global_ flag of a valid capability can be cleared but never set (without reconstruction from a global superset capability).
 
-=== Interaction with <<infinite-cap,Infinite Capabilities>>
+=== Interaction with Primordial Capabilities
```
arichardson:
```suggestion
=== Interaction with <<primordial-cap>> Capabilities
```

src/cheri/debug-integration.adoc
```diff
@@ -165,12 +165,16 @@ The <<dscratch1>> register is extended to hold a capability.
 include::img/dscratch1creg.edn[]
 
 [#dinfc,reftext="dinfc"]
-==== Debug Infinite Capability Register (dinfc)
+==== Debug Execution Capability Register (dinfc)
```
arichardson: Maybe we could keep the existing dinfc name by using "infinite code" as the name. Or we just change it to `dec`, `dxc`, `drxc` (debug root execute capability) or the other way around `dxrc` (debug execute root capability)?

`dec` is probably not a great name, but do any of the others sound reasonable?
```suggestion
==== Debug Infinite Code Capability Register (dinfc)
```

src/cheri/riscv-unpriv-integration.adoc
nwf: Fair point!  I will steal the verbiage from your ETA.

src/cheri/riscv-priv-integration.adoc
```diff
@@ -506,7 +506,7 @@ in connection with the <<section_cheri_disable,{CRE}>> and the <<cheri_execution
 ^1^ M-bit is irrelevant when {CRE}=0. +
 ^2^ See xref:legacy_mnemonics[xrefstyle=short] for a list of remapped
 instructions. +
-^3^ The hart is fully compatible with standard RISC-V when {CRE}=0 provided that <<pcc>>, <<mtvec_y>>, <<mepc_y>>, <<stvec_y>>, <<sepc_y>>, <<vstvec_y>>, <<vsepc_y>> and <<ddc>> have not been changed from the default reset state (i.e., hold the <<infinite-cap>> capability).
+^3^ The hart is fully compatible with standard RISC-V when {CRE}=0 provided that <<pcc>>, <<mtvec_y>>, <<mepc_y>>, <<stvec_y>>, <<sepc_y>>, <<vstvec_y>>, <<vsepc_y>> and <<ddc>> have not been changed from the default reset state (i.e., hold <<root-rx-cap>> capabilities).
```
tariqkurd-repo: ddc need a root-rw cap, not root-rx

src/cheri/riscv-priv-integration.adoc
```diff
@@ -281,7 +281,7 @@ include::img/sscratchcreg.edn[]
 ==== Supervisor Exception Program Counter Capability (sepc)
 
 The <<sepc>> register is extended to hold a capability.
-Its reset value is the <<infinite-cap>> capability.
+Its reset value is a <<root-rx-cap>> capability.
```
tariqkurd-repo: we can't say `reset` for s* CSRs - we need to say `initialized to XXX by the supervisor mode execution environment`
as it could be a VM
(not a criticism of this PR - but better to fix it now?)
also for stidc

src/cheri/debug-integration.adoc
```diff
@@ -228,9 +228,8 @@ xref:default-csrnames-added[xrefstyle=short].
 .Debug default data capability
 include::img/dddcreg.edn[]
 
-Upon entry to debug mode, <<ddc>> is saved in <<dddc>>. <<ddc>>'s metadata is
-set to the <<infinite-cap>> capability's metadata (with tag set) and <<ddc>>'s
-address remains unchanged.
+Upon entry to debug mode, <<ddc>> is saved in <<dddc>>. <<ddc>> is reset to a
```
```suggestion
Upon entry to debug mode, <<ddc>> is saved in <<dddc>>. <<ddc>> is set to a
```

src/cheri/debug-integration.adoc
tariqkurd-repo: I'm worried about this - in a purecap system with no DDC, how do you get access to a root-data cap in debug mode? previously dfinc served both purposes.
Don't we need another CSR for root-data?

nwf: I'd meant for that to be addressed with
```
The reset value is a <<root-rx-cap>> capability.
If rights beyond those provided by this reset value are required,
root capabilities may be obtained by the debugger immediately after reset,
and/or system software make make them available in memory, loadable through
the capability held in <<dinfc>>.
```

On systems where root executable caps are not also root data caps, extra work is required.  On systems where they are (where, for example, there is an Infinite capability, as is true of the encoding in the appendix), there's nothing to worry about.

tariqkurd-repo: how do you get the root-data capability from the debugger and set the tag so it's usable?
the whole point of dinfc is to have an infinite cap available - if the debugger can send them on request then maybe it's not necessary at all?

arichardson: Maybe the easiest solution is to have code and data infinite CSR dinfxc/dinfdc for a cheriot system where ddc does not exist?

nwf: > how do you get the root-data capability from the debugger and set the tag so it's usable?

Even if root-data is different from root-execute, root-execute...
* provides ASR, so a CSR whose reset state is root-data is accessible; it can be fetched and stored in memory (through itself).
* provides R, C, LM (and LG) permission, so root-data could be loaded from memory.

I think I'd be happy to introduce `dinfxc`/`dinfdc`, but I somewhat suspect we will need to make other changes for making a properly CHERI-aware RISC-V debug unit?

jrtc27: You cannot expect a debugger to be attached at reset time in order to work correctly, nor is it reasonable to go stomp on a random bit of memory for the debugger to use for the entire time you want to be able to debug the system.

nwf: I think I can, however, expect a little bit of cooperation from the software stack if I'm not attached at reset.

Nevertheless, this doesn't matter for the big-CHERI encoding, because `dinfc` is going to initialize to the Infinite encoding.

jrtc27: One of the points of a hardware debugger is to be able to poke at a system when the software stack _isn't_ cooperating.

jrtc27:
You can always extend this with a dinfsel CSR to select which root you want on systems without a single root.

tariqkurd-repo:
That conveniently limits the need to 2 CSRs which is nice

src/cheri/debug-integration.adoc
```diff
@@ -164,30 +164,44 @@ The <<dscratch1>> register is extended to hold a capability.
 .Debug scratch 1 capability register
 include::img/dscratch1creg.edn[]
 
-[#dinfc,reftext="dinfc"]
-==== Debug Infinite Capability Register (dinfc)
+[#drootcsel,reftext="drootcsel"]
```
arichardson, line 167: Think we need to add this to the CSV as well

src/cheri/debug-integration.adoc
```diff
@@ -164,30 +164,44 @@ The <<dscratch1>> register is extended to hold a capability.
...
+<<drootcsel>> is a debug mode accessible integer CSR.  The address and access details are shown in
 xref:default-csrnames-added[xrefstyle=short].
 
-The reset value is the <<infinite-cap>> capability.
+It selects which <<root-cap>> capability is exposed through <<drootc>>.
+The reset value is `0`, which must cause <<drootcsel>> to expose a
+<<root-rx-cap>> capability.
 
-If {cheri_default_ext_name} is implemented:
+If {cheri_default_ext_name} is implemented, this <<root-rx-cap>> is further
```
arichardson: I feel this text would be better as part of drootc

nwf: OK.

arichardson (approved):
Thanks! This looks good to me now. Will give others a bit of time to review as well, but if there are no objections I plan to merge this by Thursday.

src/cheri/debug-integration.adoc
```diff
@@ -164,30 +164,44 @@ The <<dscratch1>> register is extended to hold a capability.
...
-The reset value is the <<infinite-cap>> capability.
+It selects which <<root-cap>> capability is exposed through <<drootc>>.
+The reset value is `0`, which must cause <<drootcsel>> to expose a
+<<root-rx-cap>> capability.
 
-If {cheri_default_ext_name} is implemented:
+Other capability values may be defined for exposure through <<drootc>>
+by the capability encoding,
+and may be selected by having the debugger write to this register.
+Writes are WARL, so the debugger may confirm that its selection has applied.
```
```suggestion
Writes are WARL, so the debugger may confirm that its selection has been applied.
```

src/cheri/debug-integration.adoc
```diff
@@ -228,9 +242,8 @@ xref:default-csrnames-added[xrefstyle=short].
 .Debug default data capability
 include::img/dddcreg.edn[]
 
-Upon entry to debug mode, <<ddc>> is saved in <<dddc>>. <<ddc>>'s metadata is
-set to the <<infinite-cap>> capability's metadata (with tag set) and <<ddc>>'s
-address remains unchanged.
+Upon entry to debug mode, <<ddc>> is saved in <<dddc>>. <<ddc>> is set to a
+<<root-rw-cap>> capability with its address set to that of the just-updated <<dddc>>.
```
tariqkurd-repo: I quite like the previous wording of "address remains unchanged" - is there anything wrong with that?

src/cheri/debug-integration.adoc
```diff
@@ -198,15 +212,15 @@ shown in xref:default-csrnames-added[xrefstyle=short].
 {cheri_default_ext_name} allows <<MODESW_CAP>> and <<MODESW_INT>> to execute in debug mode.
 
 When entering debug mode, whether the core enters {cheri_int_mode_name} or
-{cheri_cap_mode_name} is controlled by the <<m_bit>> in <<dinfc>>.
+{cheri_cap_mode_name} is controlled by the <<m_bit>> in the capability selected by <<drootcsel>> value 0.
```
```suggestion
{cheri_cap_mode_name} is controlled by the <<m_bit>> in the <<drootc>> capability selected by <<drootcsel>> value 0.
```

src/cheri/img/drootcreg.edn
```diff
@@ -11,11 +11,11 @@
...
-(draw-box "dinfc (Metadata)" {:span 32})
+(draw-box "drootc (Metadata)" {:span 32})
 
 (draw-box "" {:span 2 :borders {}})
 
-(draw-box "dinfc (Address)"  {:span 32})
+(draw-box "drootc (Address)"  {:span 32})
 
 (draw-box "" {:span 2 :borders {}})
 (draw-box "MXLEN" {:span 32 :borders {}})
```
tariqkurd-repo:
```suggestion
(draw-box "DXLEN" {:span 32 :borders {}})
```

src/cheri/img/drootcselreg.edn
```diff
@@ -0,0 +1,14 @@
+[bytefield]
+----
+(defattrs :plain [:plain {:font-family "M+ 1p Fallback" :font-size 21}])
+(def row-height 40)
+(def row-header-fn nil)
+(def left-margin 100)
+(def right-margin 100)
+(def boxes-per-row 32)
+(draw-column-headers {:height 20 :font-size 18 :labels (reverse ["0" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "MXLEN-1"])})
```
tariqkurd-repo:
```suggestion
(draw-column-headers {:height 20 :font-size 18 :labels (reverse ["0" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "DXLEN-1"])})
```

src/cheri/img/drootcselreg.edn
```diff
@@ -0,0 +1,14 @@
...
+
+(draw-box "drootcsel"  {:span 32})
+
+(draw-box "MXLEN" {:span 32 :borders {}})
```
tariqkurd-repo:
```suggestion
(draw-box "DXLEN" {:span 32 :borders {}})
```

tariqkurd-repo (approved):

src/cheri/debug-integration.adoc
nwf: It's a little unusual to my mind to talk about "setting the capability metadata", in that it feels like violating abstraction boundaries.  But no, it's fine and I've put it back (I think acceptably).

PR #737, opened 2025-08-27
Rework Svucrg extension to use a 3-bit PTE implementation.

qwattash:
This PR modifies the Svucrg extension to use a 3-bit PTE format with the CW, CRG and CD bits.

This is understood to be sufficient to implement the existing Cornucopia Reloaded revoker, a validation prototype is ongoing.

 - Remove the "pre-CW" state with CW=0 CRG=1 and the associated CRG update rule.
 - Introduce a new CD bit (Capability Dirty), independent from CW and CRG.
 - Describe operation of the CD bit and update capability-dirty tracking specification.
 - Move the PTE bits to bits {58, 56} to avoid conflicting with RVS-3647 _Svrsw60t59b_.

This is an initial cut at the 3-bit PTE solution.

The assignment of PTE bits is debatable. There is an argument for using one or all of the new SWR bits from _Svrsw60t59b_ for CHERI PTE bits, which would reduce the PTE footprint with the understanding that OSes that make use of these bits will need some modification.
Similarly, we may not have exclusive use of the 3 reserved PTE bits. It might be worth considering whether in the future we may want the CD bit to be optional, on top of the basic CW/CRG bits and/or some other mechansim to reduce the CHERI PTE footprint.

### DIFF
--- a/src/cheri/cheri-pte-ext.adoc
+++ b/src/cheri/cheri-pte-ext.adoc
@@ -105,10 +105,13 @@ or AMO instruction is executed and:
 * the virtual page's CRG bit does not equal <<sstatusreg_pte,sstatus>>.UCRG, and
 * the virtual page's PTE.U is set.
 
-When the _{cheri_priv_crg_load_tag_ext}_ extension is present, the
+When the _{cheri_priv_crg_load_tag_ext}_ extension is implemented, the
 {cheri_excep_name_pte_ld} is raised when capability load or AMO instruction
 satisfies the conditions above and the capability read from memory
 has its {ctag} set.
+
+{cheri_priv_crg_load_tag_ext} implies {cheri_priv_crg_ext}.
+
 
 When CRW is set, the CD bit indicates that a capability was stored to the
 virtual page since the last time the CD bit was cleared.


tariqkurd-repo:
> The assignment of PTE bits is debatable. There is an argument for using one or all of the new SWR bits from _Svrsw60t59b_ for CHERI PTE bits, which would reduce the PTE footprint with the understanding that OSes that make use of these bits will need some modification. Similarly, we may not have exclusive use of the 3 reserved PTE bits. It might be worth considering whether in the future we may want the CD bit to be optional, on top of the basic CW/CRG bits and/or some other mechansim to reduce the CHERI PTE footprint.

We were told that the othe rPTE bits would only be used to extend the physical page number, and so it's not a problem for us to use more of the bits (given that CHERI servers will ahve a more limited PPN). So I believe that's the only consideration here.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -79,45 +83,63 @@ as usual, and capability reads are controlled by the CRG bit.
...
+ local hart. Furthermore, the PTE update must appear in the global memory order
+ no later than the explicit memory access, or any subsequent explicit memory
+ access to that virtual page by the local hart. The ordering on loads and
+ stores provided by FENCE instructions and the acquire/release bits on atomic
+ instructions also orders the PTE updates associated with those loads and
+ stores as observed by remote harts.
+ The PTE update is not required to be atomic with respect to the explicit memory
+ access that caused the update, and the sequence is interruptible. However, the
+ hart must not perform explicit memory accesses before the PTE update is globally
+ visible.
```
tariqkurd-repo: do we really need all this text? isn't it already specified by Svadu, so we can just use that?

src/cheri/cheri-pte-ext.adoc
```diff
@@ -79,45 +83,63 @@ as usual, and capability reads are controlled by the CRG bit.
...
+
+NOTE: _{cheri_priv_crg_ext}_ requires that when CW, CD and CRG are all clear,
+the implementation clears loaded tags and raises {cheri_excep_name_pte_st} when
+the stored {ctag} is set. Other bit combinations when CW=0 are reserved
+for future extensions.
```
tariqkurd-repo: the table don't show that - the second one shows PTE.CW=0 and PTE.CD=X
I think you need a row for PTE.CW=0 and PTE.CD=0 separately and if you hit the reserved case what happens? what exception type?

similarly for the first table which was CW=0 CRG=X U=X, I think you need a separate line for CW=0 CRG=0 U=X

src/cheri/cheri-pte-ext.adoc
```diff
@@ -79,45 +83,63 @@ as usual, and capability reads are controlled by the CRG bit.
...
-access that caused the update, and the sequence is interruptible. However, the
-hart must not perform explicit memory access before the PTE update is globally
-visible.
+* When a capability store or AMO instruction is executed and the {ctag} bit
+ of the capability being written is set, the implementation raises a
+ {cheri_excep_name_pte_st}.
+
+When the CW bit is set, the "capability state", then the behaviour of
+capability loads is controlled by CRG and the behaviour of capability stores
+is controlled by CD.
```
nwf, line 97: Gating CRG behavior on CW means we would be in trouble if ever we wanted PTEs that grant W and not CW for (aliased/shared) pages that hold caps (such as those that have a W and CW PTE elsewhere).  I don't know that we want such a thing, but it'd be good to (informatively) `[NOTE]` that this is a constraint on software behavior.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -79,45 +83,63 @@ as usual, and capability reads are controlled by the CRG bit.
...
 
 When CW is set, the CRG bit indicates the current generation of the virtual memory page with
 regards to the ongoing capability revocation cycle. Two schemes are permitted:
 
 * A {cheri_excep_name_pte_ld} is raised when a capability load or AMO instruction is executed
-with <<c_perm>> granted and the virtual page's CRG bit does not equal <<sstatusreg_pte,sstatus>>.UCRG and PTE.U is set.
+ with:
+** <<c_perm>> granted, and
+** the virtual page's CRG bit does not equal <<sstatusreg_pte,sstatus>>.UCRG, and
+** PTE.U is set.
```
nwf: Purely grammatical suggestion:
```suggestion
 and:
** the authorizing capability grants <<c_perm>>,
** the virtual page's CRG bit does not equal <<sstatusreg_pte,sstatus>>.UCRG, and
** the virtual page's U bit is set.
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -79,45 +83,63 @@ as usual, and capability reads are controlled by the CRG bit.
...
 * A {cheri_excep_name_pte_ld} is raised when a capability load or AMO instruction is executed
-with <<c_perm>> granted and the virtual page's CRG bit does not equal <<sstatusreg_pte,sstatus>>.UCRG and PTE.U is set
-and the capability read from memory has its tag set.
+ with:
+** <<c_perm>> granted, and
+** the virtual page's CRG bit does not equal <<sstatusreg_pte,sstatus>>.UCRG, and
+** PTE.U is set, and
+** the capability read from memory has its {ctag} set.
```
nwf:
```suggestion
 wherein:
** the authorizing capability grants <<c_perm>>,
** the virtual page's CRG bit does not equal <<sstatusreg_pte,sstatus>>.UCRG,
** the virtual page's U bit is set, and
** the capability read from memory has its {ctag} set.
```

nwf (commented):
Generally seems sensible to me.  Just a few nits from this read-thru.

src/cheri/cheri-pte-ext.adoc
qwattash: Yeah, I debated myself about doing something similar to what you are suggesting. After re-reading it, I agree that it is much better. I will address that.

src/cheri/cheri-pte-ext.adoc
qwattash: Dropped and explicitly referenced Svadu.

src/cheri/cheri-pte-ext.adoc
qwattash: I think I clarified this now.

src/cheri/cheri-pte-ext.adoc
qwattash: Why is this a problem though? I  think that you can have a {W, ~CW} alias of a {W, CW} page, because in Svucrg CW gates both capability reads and writes, so you can't ever load a valid capability through the ~CW alias anyway. I think CW is a somewhat misleading name for the bit in this model, perhaps simply calling it `C` bit might be better and mirrors what we do in the capability HW permission bits with the "C-permission bit".

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: "the authorizing capability grants" is a phrase we use elsewhere, so better to use it

src/cheri/cheri-pte-ext.adoc
```diff
@@ -79,87 +83,101 @@ as usual, and capability reads are controlled by the CRG bit.
...
-The PTE update is not required to be atomic with respect to the explicit memory
-access that caused the update, and the sequence is interruptible. However, the
-hart must not perform explicit memory access before the PTE update is globally
-visible.
+* When a capability store or AMO instruction is executed and the {ctag} bit
+ of the capability being written is set, the implementation raises a
+ {cheri_excep_name_pte_st}.
+
+When the CW bit is set, the "capability state", then the behaviour of
+capability loads is controlled by CRG and the behaviour of capability stores
```
```suggestion
When the CW bit is set, the "capability state", then the behavior of
capability loads is controlled by CRG and the behavior of capability stores
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -79,87 +83,101 @@ as usual, and capability reads are controlled by the CRG bit.
...
 * A {cheri_excep_name_pte_ld} is raised when a capability load or AMO instruction is executed
-with <<c_perm>> granted and the virtual page's CRG bit does not equal <<sstatusreg_pte,sstatus>>.UCRG and PTE.U is set.
+ and:
+** the authorizing capability grants <<c_perm>>,
+** the virtual page's CRG bit does not equal <<sstatusreg_pte,sstatus>>.UCRG, and
+** the virtual page's PTE.U is set.
 * A {cheri_excep_name_pte_ld} is raised when a capability load or AMO instruction is executed
-with <<c_perm>> granted and the virtual page's CRG bit does not equal <<sstatusreg_pte,sstatus>>.UCRG and PTE.U is set
-and the capability read from memory has its tag set.
+ wherein:
```
```suggestion
 and:
```

for consistency with the first scheme above

src/cheri/cheri-pte-ext.adoc
```diff
@@ -79,87 +83,101 @@ as usual, and capability reads are controlled by the CRG bit.
...
+of the capability being written is set and the CD bit is clear,
+two schemes are permitted (also see <<section_hardware_pte_updates>>):
+
+* The same behavior as when CW is clear, allowing software interpretation
+ of this state.
+* The implementation sets the CD bit. The PTE update behaves in the same way as
+ the D bit update described by the _Svadu_ extesion.
+
+When CW, CD and CRG are all clear, the implementation is required to clear
+loaded tags and raise {cheri_excp_name_pte_st} when the stored {ctag} is set.
```
```suggestion
loaded tags and raise {cheri_excep_name_pte_st} when the stored {ctag} is set.
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -79,87 +83,101 @@ as usual, and capability reads are controlled by the CRG bit.
...
+
+NOTE: Other CD and CRG combinations when CW=0 are reserved for future extension.
+The reserved PTE states behave as the CW=0 CD=0 CRG=0, unless a future extension
```
```suggestion
NOTE: Other CD and CRG combinations when CW=0 are reserved for future extensions.
The reserved PTE states behave as the CW=0, CD=0 and CRG=0 state, unless a future extension
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -79,87 +83,101 @@ as usual, and capability reads are controlled by the CRG bit.
...
+[%autowidth,float="center",align="center",cols="<,<,<,<,<",options="header"]
 |===
-|PTE.CW |PTE.U|PTE.CRG                                    |Load/AMO
-| 0     | X   | X                                         | Clear loaded tag
-| 1     | 1   |&#8800; <<sstatusreg_pte,sstatus>>.UCRG    | {cheri_excep_name_pte_ld}, or {cheri_excep_name_pte_ld} if tag is set for _{cheri_priv_crg_load_tag_ext}_^1^
-| 1     | 1   |=       <<sstatusreg_pte,sstatus>>.UCRG^1^ | Normal operation
-| 1     | 0   | X                                         | Normal operation^2^
+|PTE.CW | PTE.CD | PTE.CRG                                 | PTE.U | Load/AMO
+| 0     | 0      | 0                                       | X     | Clear loaded tag
+| 0     | X      | 1                                       | X     | Reserved
```
```suggestion
| 0     | 0      | 1                                       | X     | Reserved
| 0     | 1      | X                                       | X     | Reserved
```

I think this covers all reserved states now

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: in fact I think it's better not to list the first 3 points twice, and just add the optional check to the end of the first list.

src/cheri/cheri-pte-ext.adoc
qwattash: Yeah, there should be 3 reserved states.

src/cheri/cheri-pte-ext.adoc
qwattash: I added a tentative refactor that renames CW -> CRW as in "capability read-write". This seems more in line with the actual behaviour of that bit, but I'm very much open to alternative names (or keeping to CW).

tariqkurd-repo (approved):

src/cheri/cheri-pte-ext.adoc
qwattash: This should be better now.

src/cheri/riscv-priv-integration.adoc
```diff
@@ -544,10 +544,10 @@ In supervisor and user modes, capability addresses are interpreted as dictated b
 
 {cheri_priv_vmem_ext} requires that the <<pcc>> grants the <<asr_perm>> to change the page-table root *satp* and other virtual-memory parameters as described in xref:supervisor-level-csrs-section[xrefstyle=short].
 
-[#section_cw_bit]
-=== Capability Write (CW) Bit
+[#section_crw_bit]
+=== Capability (CRW) Bit
```
```suggestion
=== Capability Read-Write (CRW) Bit
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -69,97 +73,109 @@ include::img/sv57pte_crg.edn[]
...
+of the capability being written is set and the CD bit is clear,
+two schemes are permitted (also see <<section_hardware_pte_updates>>):
+
+* The same behavior as when CRW is clear, allowing software interpretation
+ of this state.
+* The implementation sets the CD bit. The PTE update behaves in the same way as
+ the D bit update described by the _Svadu_ extesion.
+
+When CRW, CD and CRG are all clear, the implementation is required to clear
+loaded tags and raise {cheri_execp_name_pte_st} when the stored {ctag} is set.
```
```suggestion
loaded tags and raise {cheri_excep_name_pte_st} when the stored {ctag} is set.
```

tariqkurd-repo (approved):

src/cheri/cheri-pte-ext.adoc
```diff
@@ -69,97 +73,109 @@ include::img/sv57pte_crg.edn[]
...
+or AMO instruction is executed and:
+
+* the authorizing capability grants <<c_perm>>,
+* the virtual page's CRG bit does not equal <<sstatusreg_pte,sstatus>>.UCRG, and
+* the virtual page's PTE.U is set.
+
+When the _{cheri_priv_crg_load_tag_ext}_ extension is present, the
+{cheri_excep_name_pte_ld} is raised when capability load or AMO instruction
+satisfies the conditions above and the capability read from memory
+has its {ctag} set.
```
```suggestion
When the _{cheri_priv_crg_load_tag_ext}_ extension is implemented, the
{cheri_excep_name_pte_ld} is raised when capability load or AMO instruction
satisfies the conditions above and the capability read from memory
has its {ctag} set.

{cheri_priv_crg_load_tag_ext} implies {cheri_priv_crg_ext}.

```

the real question here is do we have load tag checking in addition to, or instead of Svucrg. I think it's simpler to have it in addition.

Issue #592, opened 2025-04-25
Non-maximal result of ACPERM rules


chrehrhardt:
If, after applying the ACPERM mask, the requested permission before applying the rules are `R + C + LM + EL +  X`  the SL permission is irrelevant (because `W` is not set). Thus this combination of permissions can always be encoded in quadrant 1 for 32-bit. However, if the (irrelevant) SL permission after applying the acperm mask is clear, Rule 13 will needlessly remove the `X` permission because the SL permission is not maxed out. I think this is unfortunate because the resulting permissions are not maximal in the partial ordering induced by the subset relation.

While playing around with this I found that the current 32-bit rules are equivalent to the following (which may help when trying to phrase the acperm rules without relying directly on the encoding):
Given a requested set of permission the actual permissions are
* A subset of the requested permissions
* Maximal among the representable combinations of permissions that are a subset of the requested permissions
* Ties are broken be preferring encodings that include the `C` bit set
* If there are still ties encodings with the `W` bit are preferred.
The rules for 64-bit would be the same except that more combinations of permissions can be represented.

It may be possible to replace all of the acperm rules with something along the lines of the above.

tariqkurd-repo:
@chrehrhardt can you have another look at this one? @nwf has updated the rules: https://github.com/riscv/riscv-cheri/pull/742

PR #742, opened 2025-08-29
Refactor CLRPERM rules

nwf:
Move common rules out into the architectural specification, not the encoding, and clarify that extensions and encodings may each add new rules.  Adjust cross-references to follow along.

### DIFF
--- a/misc/rv32y_perms.hs
+++ b/misc/rv32y_perms.hs
@@ -1,0 +1,190 @@
+{-# OPTIONS_GHC -Wall #-}
+{-# LANGUAGE OverloadedLists #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+-- | Check that the RV32 encodings, with and without Zylevels1, are sensible and
+--   that the procedure we give for computing permissions has the intended
+--   effect.
+--
+--   1. Each defined compressed permission set holds only defined permissions
+--
+--   2. For each defined compressed permission set, it had better be stable
+--      under the transformation rules (that is, a fixed point thereof).
+--
+--   3. For each subset of all defined permissions, the result of evaluating
+--      the rules is a defined compressed permission set.
+--
+--   4. Removing permissions from a defined compressed permission set and
+--      evaluating the rules selects the best result, in that all other defined
+--      compressed permission sets that could have been chosen are subsets of
+--      the one picked.
+module Main where
+
+import           Control.Monad
+import qualified Data.Set  as S
+import qualified Data.Text as T
+
+data P = PR | PW | PC | PLM | PLG | PSL | PX | PASR | PM
+ deriving (Bounded, Enum, Eq, Ord, Show)
+
+type SP = S.Set P
+
+allPermsL1, allPermsL0 :: SP
+allPermsL1 = S.fromList [minBound .. maxBound]
+allPermsL0 = S.difference allPermsL1 [PLG, PSL] -- remove Zylevel perms
+
+validsL1, validsL0 :: S.Set SP
+validsL0 = S.fromList $
+  [ -- Quadrant 0
+    {- No permissions             -}     [                         ]
+  , {- Data RO                    -}     [PR                       ]
+  , {- Data WO                    -}     [    PW                   ]
+  , {- Data RW                    -}     [PR, PW                   ]
+
+  , -- Quadrant 2
+    {- Data & Cap RO              -}     [PR,     PC               ]
+
+  , -- Quadrant 3
+    {- Data & Cap RO              -}     [PR,     PC, PLM          ]
+  , {- Data & Cap RW              -}     [PR, PW, PC, PLM          ]
+  ]
+  -- Quadrant 1
+  ++ {- Exec, Data & Cap RW & ASR -} wPM [PR, PW, PC, PLM, PX, PASR]
+  ++ {- Exec, Data & Cap RO       -} wPM [PR,     PC, PLM, PX      ]
+  ++ {- Exec, Data & Cap RW       -} wPM [PR, PW, PC, PLM, PX      ]
+  ++ {- Exec, Data RW             -} wPM [PR, PW,          PX      ]
+ where
+  wPM s = [s, S.insert PM s]
+
+validsL1 = S.fromList $
+  [ -- Quadrant 0
+    {- No permissions             -}     [                                   ]
+  , {- Data RO                    -}     [PR                                 ]
+  , {- Data WO                    -}     [    PW                             ]
+  , {- Data RW                    -}     [PR, PW                             ]
+
+  , -- Quadrant 2
+    {- Data & Cap RO, !SL, !LG    -}     [PR,     PC                         ]
+  , {- Data & Cap RW, SL, !LG     -}     [PR, PW, PC, PLM,      PSL          ]
+  , {- Data & Cap RW, !SL, !LG    -}     [PR, PW, PC, PLM                    ]
+
+  , -- Quadrant 3
+    {- Data & Cap RO, !SL, LG     -}     [PR,     PC, PLM, PLG               ]
+  , {- Data & Cap RW, SL, LG      -}     [PR, PW, PC, PLM, PLG, PSL          ]
+  , {- Data & Cap RW, !SL, LG     -}     [PR, PW, PC, PLM, PLG               ]
+  ]
+  -- Quadrant 1
+  ++ {- Exec, Data & Cap RW & ASR -} wPM [PR, PW, PC, PLM, PLG, PSL, PX, PASR]
+  ++ {- Exec, Data & Cap RO       -} wPM [PR,     PC, PLM, PLG,      PX      ]
+  ++ {- Exec, Data & Cap RW       -} wPM [PR, PW, PC, PLM, PLG, PSL, PX      ]
+  ++ {- Exec, Data RW             -} wPM [PR, PW,                    PX      ]
+ where
+  wPM s = [s, S.insert PM s]
+
+(||^), (&&^) :: Applicative f => f Bool -> f Bool -> f Bool
+(||^) = liftA2 (||)
+(&&^) = liftA2 (&&)
+
+sAny, sAll :: SP -> SP -> Bool
+sAny s = not . S.null . S.intersection s
+sAll   = S.isSubsetOf
+
+sImplies :: P -> P -> SP -> Bool
+sImplies p q = (not <$> S.member p) ||^ S.member q
+
+data Rule = Rule T.Text P (SP -> Bool)
+
+rulesL0, rulesL1 :: [Rule]
+rulesL0 =
+  [ Rule "C on R"         PC   $ S.member PR
+  , Rule "X on R"         PX   $ S.member PR
+  , Rule "W on LM | !C"   PW   $ S.member PLM ||^ (not <$> S.member PC)
+  , Rule "X on W | C"     PX   $ sAny [PC, PW]
+  , Rule "LM on C"        PLM  $ S.member PC
+  , Rule "X dependence"   PX   $ sAll [PC, PLM] ||^ (not <$> sAny [PC, PLM])
+  , Rule "ASR on W C X"   PASR $ sAll [PW, PC, PX]
+  , Rule "M on X"         PM   $ S.member PX
+  ]
+
+rulesL1 =
+  [ Rule "C on R"         PC   $ S.member PR
+  , Rule "X on R"         PX   $ S.member PR
+  , Rule "W on LM | !C"   PW   $ S.member PLM ||^ (not <$> S.member PC)
+  , Rule "X on W | C"     PX   $ sAny [PC, PW]
+  , Rule "LM on C"        PLM  $ S.member PC
+  , Rule "LM on W | LG"   PLM  $ sAny [PW, PLG]
+  , Rule "LG on LM"       PLG  $ S.member PLM
+  , Rule "SL on W LM"     PSL  $ sAll [PW, PLM]
+  , Rule "X dependence"   PX   $ (sAll [PC, PLM, PLG] &&^ sImplies PW PSL)
+                                 ||^ (not <$> sAny [PC, PLM, PLG, PSL])
+  , Rule "ASR on W C X"   PASR $ sAll [PW, PC, PX]
+  , Rule "M on X"         PM   $ S.member PX
+  ]
+
+data RuleResult = RRFire   -- Rule fired and removed target permission
+                | RRAbsent -- Rule fired, but permission not present
+                | RRNoFire -- Rule did not fire
+ deriving (Eq, Ord, Show)
+
+type LogEntry = (T.Text, RuleResult)
+
+interpRule :: Rule -> SP -> (LogEntry, SP)
+interpRule (Rule n p f) s = case () of
+  _ | f s                -> ((n, RRNoFire), s)
+  _ | not (S.member p s) -> ((n, RRAbsent), s)
+  _ | otherwise          -> ((n, RRFire  ), S.delete p s)
+
+interpRules :: [Rule] -> SP -> ([LogEntry], SP)
+interpRules = \rs s0 -> let (l, s') = foldl interp (id, s0) rs in (l [], s')
+ where
+  interp :: ([LogEntry] -> [LogEntry], SP)
+         -> Rule
+         -> ([LogEntry] -> [LogEntry], SP)
+  interp (l, s) r = let (le,s') = interpRule r s in (l . (le:), s')
+
+stepsUnder :: [Rule] -> SP -> Maybe ([LogEntry], SP)
+stepsUnder rs s = let e@(_, s') = interpRules rs s in
+                  if s == s' then Nothing else Just e
+
+conditions :: [(T.Text, SP, [Rule], S.Set SP)]
+conditions = [("no levels", allPermsL0, rulesL0, validsL0)
+             ,("zylevels1", allPermsL1, rulesL1, validsL1)
+             ]
+
+main :: IO ()
+main = forM_ conditions $ \(cn, cperms, crules, cdefines) -> do
+  putStrLn $ "[*] Checking condition " ++ (show cn)
+
+  putStrLn "[*] Checking that all defined encodings are subsets of all perms..."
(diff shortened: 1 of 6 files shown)


src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -341,6 +341,33 @@ NOTE: This permission is important in privileged execution environments.
...
+===== Permission Transitions
+
+Not all capability permissions are _orthogonal_
+(that is, some permissions inherently _depend_ on others).
+As such, using <<CLRPERM>> to clear some permissions
+may have the effect of clearing others as well,
+such that a permission bit being set in the result of <<GCPERM>>
+implies that all bits for permissions upon which it depends will also be set.
+
+For the base set of permissions just defined, the follow rules apply.
```
```suggestion
For the base set of permissions just defined, the following rules apply.
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -341,6 +341,33 @@ NOTE: This permission is important in privileged execution environments.
...
+Extensions that define new permission bits (such as
+<<section_cheri_hybrid_ext,{cheri_default_ext_name}>> and <<section_zylevels1>>)
+may also introduce new dependency constraints.
+Capability _encodings_ may impose additional constraints,
+even non-dependency constraints,
```
```suggestion
or even non-dependency constraints,
```

tariqkurd-repo: or maybe those words just aren't needed I think it reads fine without them (delete ,even non-dependency constraints,)

src/cheri/insns/acperm_32bit.adoc
```diff
@@ -30,7 +30,9 @@ format shown in xref:acperm_bit_field[xrefstyle=short].
 . The initial value in register `rs2[XLEN-1:0]` is treated as a bit mask that specifies bit positions to be cleared in the bit field. Any bit that is high in `rs2` will cause the corresponding bit to be cleared in the bit field.
 +
 NOTE: Future extensions may include hardwired permission bits, in which case they are not cleared by set bits in `rs2`.
-. Encode the resulting architectural permissions following the rules in <<sec_permission_transitions>>.
+. Encode the resulting architectural permissions,
+  by iterating the rules in <<sec_permission_transitions>>
```
tariqkurd-repo:
```suggestion
  by iterating over the rules in <<sec_permission_transitions>>
```

src/cheri/insns/acperm_32bit.adoc
```diff
@@ -30,7 +30,9 @@ format shown in xref:acperm_bit_field[xrefstyle=short].
...
+  as well as any rules added by extensions or the capability encoding, to fixed point.
```
tariqkurd-repo:
```suggestion
  as well as any rules added by extensions or the capability encoding, to a fixed point.
```

or maybe "a stable set of permissions"?

src/cheri/zylevels1.adoc
```diff
@@ -96,13 +96,13 @@ Implementations _are permitted but not mandated_ to require that an explicit <<C
...
 
-.Additional <<CLRPERM>> rules (see <<acperm_rules>>)
+.Additional <<CLRPERM>> Rules
 
 As mentioned, the <<zylevels1_sl_perm>> and <<zylevels1_lg_perm>> permissions
 are dependent on (refinements of) base permissions.
 <<CLRPERM>> (including "implicit <<CLRPERM>>" operations) and/or the capability encoding
 therefore clear these permissions when their dependencies clear.
-Specifically, we add the following rules to <<acperm_rules>>:
+Specifically, we add the following rules those of <<sec_permission_transitions>>:
```
tariqkurd-repo:
```suggestion
Specifically, we add the following rules after those of <<sec_permission_transitions>>:
```

I think we need to be careful here that we clearly specify which order the rules are applied in. We can say that Zylevels1 is first (after the base) and then future extensions come after?!

I'm also nervous about changing functionality here as it took a long time to get the permission transformations to work correctly the first time......

tariqkurd-repo: it's possible we push the order of applying the rules from different extensions to a CHERI profile

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -341,6 +341,33 @@ NOTE: This permission is important in privileged execution environments.
...
+to reduce the number of bits necessary to represent permissions.
+
+.<<CLRPERM>> base rules
+[float="center",align="center",cols="2,2,4",options="header"]
+|===
+| Rule            | Permission   | Valid only if
```
```suggestion
| <<CLRPERM>> Rule            | Permission   | Valid only if
```

I was trying to think of a more descriptive title for the rules

src/cheri/zylevels1.adoc
nwf: I'd like the specification to jettison any notion of ordering of these rules, which is why I phrased it as "iterate to fixed point".  Because there are only finitely many bits and the rules are monotonic non-increasing (in the number of granted permissions; that is, they can clear but never set bits), this approach is guaranteed to be strongly normalizing (that is, iteration eventually yields a stable set of permissions).

In this fixed-pointing world, encodings would have to show that their aggregate set of rules -- the base's, their own, and those from extensions they entail -- is also confluent (all ways of normalizing a common input yield the same output) on all sets reachable from their root capabilities.  (And stability of roots under these rules is probably also a good idea.)  This is trivial for the RV64 encoding proposed in the spec and CHERIoT, but the current RV32 proposal is not, due (at least) to the negation in the "RV32-Zylevels-3" rule (consider removing `C` from `X R W C LM LG SL`).  IMHO, this is a deficiency of the (unvalidated) RV32 encoding and not something that should necessarily torpedo a more modular approach to rules.

tariqkurd-repo: well - we certainly had a lot of trouble formulating the rules to converge nicely last time - so I'm very worried that we're re-opening that can of worms with this. I think this needs careful consideration see how it applies to the existing RV32 encoding format. Everything you say in principle sounds sensible but we had a lot of trouble with this before... I need to remind myself exactly what the problems were....

arichardson: I recall there was an issue where not applying them in order resulted in two possible outcomes, with one of them being "worse" (i.e. not the "most permissive" subset). 
https://github.com/riscv/riscv-cheri/issues/426 and https://github.com/riscv/riscv-cheri/pull/427

nwf: Yes, that's what I mean by lack of confluence.  It comes from the non-monotonic logic of the present RV32 encoding's rules.

I suppose it would be OK for the RV32 encoding (or, indeed, any other encoding) to additionally require an ordering of the rules of the extensions it entails, but in general, making _extensions_ have to contemplate each other's rules seems like a maintenance nightmare in the making...  FWIW, so far we haven't seen extensions that wanted non-monotonic logic in their permission rules, just the proposed RV32 encoding's compression mechanism.  I _suspect_ this will be the case going forward, too.

All that said, I'll changed the prose of `<<CLRPERM>>` to be more tolerant of encoding's needs or whims and will put an ordering back in the RV32 encoding discussion.

arichardson: Thanks!

src/cheri/app-cap-description.adoc
```diff
@@ -243,36 +243,29 @@ implementation supports {cheri_default_ext_name} _and_ <<x_perm>> is set.
...
-|  7              | <<lm_perm>>  | <<c_perm>> and <<r_perm>>
-|  8  (RV32 only) | <<lm_perm>>  | (<<w_perm>> or <<zylevels1_lg_perm>>)
-|  9              | <<asr_perm>> | <<x_perm>>
-| 10^1^           | <<m_bit>>    | <<x_perm>> and {cheri_default_ext_name} is implemented
+| RV32-1 | <<asr_perm>> | All other permissions are set.
+| RV32-2 | <<c_perm>>   | <<r_perm>>
+| RV32-3 | <<x_perm>>   | <<r_perm>>
+| RV32-4 | <<w_perm>>   | not(<<c_perm>>) or <<lm_perm>>
+| RV32-5 | <<x_perm>>   | <<w_perm>> or <<c_perm>>
+| RV32-6 | <<lm_perm>>  | (<<w_perm>> or <<zylevels1_lg_perm>>)
```
```suggestion
```
I wonder if this one should just be a levels rule? Since we can have LM without W (Executable Data+Cap RO) when levels are not supported.

nwf: I'd rather not make LM generally dependent on LG.  We didn't in CHERIoT, at least.

arichardson:
> Can you point me at said script?

I don't have it, I believe @tariqkurd-repo should know where to find it.

tariqkurd-repo:
> > Can you point me at said script?
> 
> I don't have it, I believe @tariqkurd-repo should know where to find it.

I've sent it to @nwf separately

arichardson:
> > > Can you point me at said script?
> > 
> > 
> > I don't have it, I believe @tariqkurd-repo should know where to find it.
> 
> I've sent it to @nwf separately

If there is nothing confidential in there it might be worth checking into the repository?

arichardson:
I believe this should fix https://github.com/riscv/riscv-cheri/issues/592 (assuming I read the script correctly?).

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -339,6 +339,32 @@ NOTE: This permission is important in privileged execution environments.
...
+Capability _encodings_ may impose additional constraints.
+to reduce the number of bits necessary to represent permissions.
+
+.<<CLRPERM>> base rules
+[float="center",align="center",cols="2,2,4",options="header"]
+|===
+| <<CLRPERM>> Rule | Permission   | Valid only if
+| base-1           | <<c_perm>>   | <<r_perm>> or <<w_perm>>
+| base-2           | <<lm_perm>>  | <<c_perm>> and <<r_perm>>
+| base-3           | <<asr_perm>> | <<x_perm>>
```
```suggestion
| [[perm_req:base:c:r-or-w,base-1]]base-1   | <<c_perm>>   | <<r_perm>> or <<w_perm>>
| [[perm_req:base:lm:c-and-r,base-2]]base-2 | <<lm_perm>>  | <<c_perm>> and <<r_perm>>
| [[perm_req:base:asr:x,base-3]]base-3      | <<asr_perm>> | <<x_perm>>
```

Maybe something like this to allow cross-references to the rule being replaced?

src/cheri/app-cap-description.adoc
```diff
@@ -169,54 +169,92 @@ orthogonal to permissions as it can vary arbitrarily using <<SCMODE>>.
...
+|       ^3^  | <<asr_perm>>          | <<w_perm>> and <<c_perm>> and <<x_perm>>
+|            | <<m_bit>>             | <<x_perm>> and {cheri_default_ext_name} is implemented
 |===
 
+^1^ This rule supersedes <<sec_permission_transitions,base-1>>,
+    granting <<c_perm>> in fewer circumstances. +
+^2^ This rule is <<sec_permission_transitions,base-2>>,
+    simplified by the earlier strict dependence of <<c_perm>> on <<r_perm>>. +
+^3^ This rule supersedes <<sec_permission_transitions,base-3>>,
+    granting <<asr_perm>> in fewer circumstances.
```
```suggestion
| Rule        | Permission            | Valid only if
| RV32-base-1 | <<c_perm>>            | <<r_perm>> (supersedes rule <<perm_req:base:c:r-or-w>>, granting <<c_perm>> in fewer circumstances)
| RV32-base-2 | <<x_perm>>            | <<r_perm>>
| RV32-base-3 | <<w_perm>>            | not(<<c_perm>>) or <<lm_perm>>
| RV32-base-4 | <<x_perm>>            | <<w_perm>> or <<c_perm>>
| RV32-base-5 | <<lm_perm>>           | <<c_perm>> (simplified rule <<perm_req:base:c-and-r>> due to earlier strict dependence of <<c_perm>> on <<r_perm>>)
| RV32-base-6 | <<x_perm>>            | +
  (<<c_perm>> and <<lm_perm>>) or not (<<c_perm>> or <<lm_perm>>)
| RV32-base-7 | <<asr_perm>>          | <<w_perm>> and <<c_perm>> and <<x_perm>> (supersedes <<perm_req:base:asr:x>>, granting <<asr_perm>> in fewer circumstances.)
| RV32-base-8 | <<m_bit>>             | <<x_perm>> and {cheri_default_ext_name} is implemented
|===
```

Not sure if this is better without the footnotes? xrefs should now use automatic reference text and link to the actual cell.

nwf:
> I believe this should fix #592 (assuming I read the script correctly?).

Yes, that's one of the things the script checks, though I continue to hope that there's a simpler way to specify the encoding procedure.

arichardson (approved):
Thanks! Will merge this once @tariqkurd-repo has also taken a look.

src/cheri/app-cap-description.adoc
arichardson: With this comment I meant <<zylevels1_lg_perm>> should not be referenced in the base RV32-no-levels rules. But that has since been fixed :)

src/cheri/app-cap-description.adoc
```diff
@@ -243,44 +262,12 @@ implementation supports {cheri_default_ext_name} _and_ <<x_perm>> is set.
...
-^1^ For RV32, the encodings which have the <<m_bit>> set to {int_mode_value} for {cheri_int_mode_name}
-are only valid if {cheri_default_ext_name} is implemented. Otherwise those encodings represent invalid permissions.
-
 NOTE: Future extensions may allow more combinations of permissions, especially for MXLEN=64.
 
 NOTE: Future extensions may define new dependent permissions and, if so, must augment the above table.
 
-If any of these transitions (or those added by extensions or the encoding) are applied to a sealed capability,
-the {ctag} of the resulting capability is cleared.
+If any of these transitions (or those added by extensions or the encoding) alter a sealed capability,
```
tariqkurd-repo: we got complaints about referring to capabilities as variables - so I think we need to say "if the source capability is sealed"

nwf: We can just remove this sentence, since it's already captured by the YPERMC semantics.

src/cheri/insns/acperm_32bit.adoc
```diff
@@ -30,7 +30,11 @@ format shown in xref:acperm_bit_field[xrefstyle=short].
 . The initial value in register `rs2[XLEN-1:0]` is treated as a bit mask that specifies bit positions to be cleared in the bit field. Any bit that is high in `rs2` will cause the corresponding bit to be cleared in the bit field.
 +
 NOTE: Future extensions may include hardwired permission bits, in which case they are not cleared by set bits in `rs2`.
-. Encode the resulting architectural permissions following the rules in <<sec_permission_transitions>>.
+. Encode the resulting architectural permissions as specified by the encoding in use.
+  This may involve iterating over the rules in <<sec_permission_transitions>>,
+  as well as any rules added by extensions or the capability encoding,
+  to a fixed point (that is, a set of permissions not further reduced by any rule);
+  encodings may, however, specify other encoding procedures.
```
tariqkurd-repo, line 37:
```suggestion
  extensions may, however, specify other encoding procedures.
```

I guess this was the intention?

src/cheri/riscv-hybrid-integration.adoc
```diff
@@ -87,6 +87,14 @@ The <<m_bit>> of a capability can be read and written by the instructions listed
...
 
+For capabilities that do not grant <<x_perm>>,
+<<m_bit>> must always be interpreted and reported as {cap_mode_value}.
+
+NOTE: While this is not phrased as an additional rule for <<CLRPERM>> to follow,
+beyond those of <<sec_permission_transitions>>,
+capability encodings may nevertheless take advantage of this implication
+in their representation of architectural CHERI capabilities.
+
 NOTE: The mode can also be updated by setting the <<m_bit>> of a target capability using <<SCMODE>> followed by a <<JALR_CHERI>>.
```
tariqkurd-repo: I think this note needs to move up to line 89, as it relates to the table above it.

tariqkurd-repo:
apologies for this slightly unusual reference (need to reboot before I can access my VM again)

image


Is it the case that these rules are run in order _instead_ of the rules specified by YPERMC?
Or are the YPERMC rules run first, and then these after.

Because you're replacing some rules, I guess it's the former? So this set is a full replacement. It could be good to clarify that.

src/cheri/insns/acperm_32bit.adoc
nwf: No: it's specifically the capability encoding that has the final word.

src/cheri/zylevels1.adoc
```diff
@@ -102,13 +102,15 @@ As mentioned, the <<zylevels1_sl_perm>> and <<zylevels1_lg_perm>> permissions
...
```
arichardson:
```suggestion
Specifically, we add the following rules to those of <<sec_permission_transitions>>:
```

arichardson (approved):
Thank you for the thorough work on this one!!

misc/rv32y_perms.hs
```diff
@@ -0,0 +1,190 @@
...
+--   2. For each defined compressed permission set, it had better be stable
+--      under the transformation rules (that is, a fixed point thereof).
+--
+--   3. For each subset of all defined permissions, the result of evaluating
+--      the rules is a defined compressed permission set.
+--
+--   4. Removing permissions from a defined compressed permission set and
+--      evaluating the rules selects the best result, in that all other defined
+--      compressed permission sets that could have been chosen are subsets of
+--      the one picked.
```
chrehrhardt, line 20: AFAICS, this statement is not true and not what the script checks. The script checks below that the selected result is maximal in that there is _no_ other set that could have been selected which is a strict _super_set of the one picked. I don't think we can promise more than this unless we add additional, somewhat arbitrary tie breaker rules.

Please keep in mind that set inclusion is only a partial ordering, i.e. there may be (an in fact are) elements A and B where neither A<=B nor B<=A.

arichardson: That is a good point and I agree that if there were cases where we have to choose between two options we'd have to weight e.g. C higher than other permissions. 
I believe with the current set of permissions combinations we don't run into this problem but I haven't looked through all combinations in detail.

chrehrhardt: `R+C` and `R+W` are representable but `R+W+C` is not. Thus if you start with `R+W+C` there are two maximal elements wrt. set inclusion odering.

Issue #734, opened 2025-08-26
should we allow MISA.Y to be writeable even if hybrid mode isn't supported?


tariqkurd-repo:
Currently we have purecap cores, we then add hybrid mode and as a result we actually get 3 modes:

1. cap mode
2. int mode
3. cheri-disabled mode (formerly CRE=0, now misa.Y=0 or menvcfg.Y=0 or senvcfg.Y=0)

What I'm wondering about is if there's any reason not to allow 1. and 3 even without hybrid mode being implemented.
Are there useful scenarios where an OS can make use of standard RISC-V code without hybrid?
I guess the problem here is how you return to cap mode - it will have to be through a change in exception level only (exception or interrupt).

So the question is - does making any of the three Y bits writeable really depend on implementing hybrid mode?

arichardson:
While I'm not sure how useful it is to have them be writable if you don't have hybrid, I agree that disabling it for lower modes could be done without hybrid mode support.

And I guess for M-mode, the write to misa.Y would switch between capability and integer interpretation?

The problem I see is that we don't have ddc without hybrid mode, so would all integer-based accesses be unrestricted? I guess we can allow people to shoot themselves in the foot but this does not seem ideal.

arichardson:
I agree there is no real motivating use case for it now. If someone in the future has a compelling use cases where they want to use CHERI for say M-mode, but rely on something else (PMP??) for S-mode isolation, we could always relax the restriction to make it writable even without DDC being implemented?

The one use case I could see is having an early M-mode decision whether to run code as purecap fall back to non-CHERI only based on e.g. a fuse value. Such a system would have no intention of running hybrid code and would always either be purecap, or fall back to the previous approach in case a serious HW bug implementation bug was found or a vendor only wants to tape out one chip and have the provisioning process take care of CHERI on/off?


I'd vote for keeping it RO unless DDC is implemented. I can also see an argument for having it be one-time writable by the boot ROM, but dynamic changes seem like a bad idea without hybrid mode.

jrtc27:
A vendor can always provide some external config pin or custom M-mode CSR to do whatever they like in that regard.

vmurali:
[...]

Issue #659, opened 2025-07-11
add extension to report more exception cause information


tariqkurd-repo:
The standard exception cause values have been collapsed to the minimum (just instruction / load / store): see 
https://github.com/riscv/riscv-cheri/issues/657

CHERIoT needs more exception cause information, including:

1. the detailed cause type (tag, sealed, perm, bounds, others?)
2. the source register which caused the exception, so they don't have to (or are not able to) decode the faulting instruction

Therefore we should define a new extension which adds a CSR to report this information.

PR #743, opened 2025-08-29
WIP: Xcheriot

nwf:
This is mostly an exercise to verify that the base spec and its constellation of extensions is meeting its goal of being a suitable foundation for a standard-based CHERIoT encoding.  It is very much a work in progress but hopefully is a useful exercise and comes to at least show the kind of thing we'd like to have by the end of the process.

Presently this is sat on top of other PRs, and that will probably remain true for the foreseeable future.

All the same, commentary welcome.


src/cheri/xcheriot.adoc
```diff
@@ -0,0 +1,397 @@
...
+* <<section_cheri_disable,{cheri_priv_m_reg_enable_ext}>>,
+
+* <<section_zyseal_amb>>, and
+
+* ...
+
+[NOTE]
+=====
+While {cheriot_ext_name} is nominally compatible with
+<<section_cheri_hybrid_ext,{cheri_default_ext_name}>> and
```
vmurali: This might be useful (for instance, see https://github.com/riscv/riscv-cheri/issues/734)

src/cheri/xcheriot.adoc
```diff
@@ -0,0 +1,397 @@
...
+
+=== New Instructions
+
+==== Loads and Stores With `x0` Authority
+
+Instructions such as `{load_cap_name_lc} {cd}, n(x0)` would,
+in {cheri_base_ext_name},
+necessarily fault due to the clear {ctag} of the all-zeros `x0`.
+{cheriot_ext_name} defines these instructions
+to use <<mtidc>> as the authority instead.
```
vmurali: I dont think I understand this. Why are you using mtidc as the authority for cheriot's Load Cap?

nwf: This is a half-baked thought, but the intent is to use `mtidc` in place of `cgp` in the RTOS.  `mtidc` is not ASR-gated for reads (unlike `mscratch`, which makes the latter a more directly analogous replacement for our `mtdc`) but is for writes, so it could be a good choice for holding `cgp` outside of the GPRs.  If we go down this path, our `auicgp` instruction would operate on `mtidc` and not `x2`.

src/cheri/xycheriot.adoc
```diff
@@ -0,0 +1,511 @@
...
+=====
+TODO: setbounds register that fixes base and rounds length down for
+representability.
+=====
+
+==== {cheriot_gettop_name}
+
+[NOTE]
+=====
+TODO: Two-operand instruction that fuses <<GCBASE>>, <<GCLEN>>, and ADD.
```
arichardson: I imagine this should be a fast-track extension if we can show some benchmark/code pattern this helps with a lot since I believe it is generally useful. The reason don't have it in the base (anymore?) is that we were trying to keep the number of newly added instructions as small as possible if they can easily be emulated by a short sequence.

src/cheri/xycheriot.adoc
```diff
@@ -0,0 +1,511 @@
...
+which adds its shifted (see <<section_cheriot_auipc_shift>>) U-immediate
+to the address of <<mtidc>> and stores the result in a target register,
+analogously to <<AUIPC_CHERI>> and <<pcc>>.
+
+==== {cheriot_setbounds_big_imm_name}
+
+[NOTE]
+=====
+TODO: setbounds immediate as an I-type instruction w/ 12-bit immediate rather
+than the base's R-type w/ (scaled) 5-bit immediate.
```
arichardson: This would of course be nice to have in the base but we can't have more than one I-type. So leaving it for custom space sounds good to me.

src/cheri/xycheriot.adoc
```diff
@@ -0,0 +1,511 @@
...
+=====
+
+==== {cheriot_setbounds_round_down_name}
+
+[NOTE]
+=====
+TODO: setbounds register that fixes base and rounds length down for
```
arichardson: Would also really like to see this one as a non-CHERIoT-specific extension in the future.

src/cheri/xycheriot.adoc
```diff
@@ -0,0 +1,511 @@
...
+How hard is it to explain such a thing to LLVM?  Would we need to push type
+refinements up onto C / C++ function pointers?
+
+=====
+
+=== Miscellaneous Changes
+
+[NOTE]
+=====
+TODO: What else am I forgetting?
```
arichardson: I believe you have the bounds checking on branch rather than at target? Or will that be removed from CHERIoTv2?

src/cheri/xycheriot.adoc
```diff
@@ -0,0 +1,511 @@
...
+:cheriot_ot_x_max: 7
+:cheriot_ot_absent: 8
+:cheriot_ot_nx_min: 9
+:cheriot_ot_nx_max: 15
+:cheriot_ot_max: 15
+
+[appendix]
+[#section_cheriot,reftext="{cheriot_ext_name}"]
+== "{cheriot_ext_name}" Extension and Capabilities
+
```
arichardson, line 31:
```suggestion
IMPORTANT: This chapter is not part of the to-be ratified package. We include it in this specification since it is important to illustrate how the base RVY ISA can be augmented and why certain extension hooks exist in the current prose.

```

Maybe something like this?

arichardson (approved):
I believe including this chapter in the spec document will make it clear to ARC why we have made certain decisions and I will merge this once the PRs it depends on have been merged.

Thank you for the huge amount of effort that you have put into all of these pull requests!

vmurali:
Would it be better to include just the concepts instead of concrete encodings? I mean, the CHERIoT Spec already has the current (v1?) encodings and there are some improvements that can be made that can directly go into this RISC-V extension once we finalize them?

nwf:
> Would it be better to include just the concepts instead of concrete encodings? I mean, the CHERIoT Spec already has the current (v1?) encodings and there are some improvements that can be made that can directly go into this RISC-V extension once we finalize them?

That's basically what I've been doing, admittedly in a much more "notes to the cognoscenti" way than "justify the shape of RVY".  If people think having this for RVI's review of RVY is sufficiently worth while, I can spend a bit of time trying to close that gap.

src/cheri/zyseal.adoc
```diff
@@ -0,0 +1,104 @@
...
+Seal Permission (SE):: Permit the bearer to <<YSEAL>> capabilities
+at the in-bound types of this capability.
+
+[#zyseal_us_perm,reftext="US-permission"]
+Unseal Permission (US):: Permit the bearer to <<YUNSEAL>> capabilities
+at the in-bound types of this capability.
+
+=== Interaction with <<CLRPERM>> and <<GCPERM>>
+
+.Extended capability permissions bit field (see <<acperm_bit_field>>)
```
vmurali, line 79: Is it a good idea to show an example encoding for SE and US perms?

src/cheri/xycheriot.adoc
```diff
@@ -0,0 +1,678 @@
...
+a few proposed incremental changes to the existing CHERIoT ISA;
+these are marked as "TODO, research required"
+and can be ignored by readers focused on {cheri_base_ext_name}.
+
+=== Required Extensions
+
+{cheriot_ext_name} is specific to {cheri_base32_ext_name}.
+It requires the presence of all of:
+
+* <<section_priv_cheri,{cheri_priv_m_ext}>>,
```
vmurali, line 54: Just to clarify: this doesn't bring in S-mode CSRs, right? Like scause, etc?

src/cheri/xycheriot.adoc
```diff
@@ -0,0 +1,678 @@
...
+
+CHERIoT's bounds encodings use 9-bit top and bottom mantissa fields
+and a compressed 4-bit exponent value.
+The resulting encoding significantly differs from the bounds
+of the <<app_cap_description>> encoding; CHERIoT's encoding...
+
+- has byte alignment only up to length 511.
+
+- does not guarantee a representable region beyond the authorized base and top
+  (except for "one past the end", where the address is equal to the limit).
```
vmurali: "where the address is equal to the limit" is unclear.

src/cheri/xycheriot.adoc
```diff
@@ -0,0 +1,678 @@
...
+  from {cheri_base_ext_name};
+
+* <<zylevels1_lg_perm>>, <<zylevels1_sl_perm>>, and <<zylevels1_gl_perm>>
+  from <<section_zylevels1>>; and
+
+* <<zyseal_se_perm>> and <<zyseal_us_perm>> from <<section_zyseal>>.
+
+The aggregate permissions bit field (<<acperm_bit_field>>) used by <<GCPERM>>
+and <<CLRPERM>> for {cheriot_ext_name} is thus:
+include::./img/xycheriot-acperm_bit_field.edn[]
```
vmurali, line 182: Do we have to specify this encoding?

src/cheri/xycheriot.adoc
```diff
@@ -0,0 +1,678 @@
...
+* The <<zyseal_se_perm>>, <<zyseal_us_perm>>, and `U0`
+  collectively conflict with
+  either of the <<r_perm>> and <<w_perm>> permissions.
++
+That is, a capability may grant permissions
+from at most one of these two sets;
+this serves to partition capabilities that refer to memory addresses
+from those that refer to <<sec_cap_type>> values
+(those granting <<zyseal_se_perm>> and/or <<zyseal_us_perm>>)
+or uninterpreted integers (those granting `U0`).
```
vmurali, line 219: I think all this is probably unnecessary for this draft; writing this down ties us down to this encoding right away.

Is the goal of this section to give an example encoding of caps, but not specify the encoding of instructions? I assumed we are just specifying the concepts, not any encoding.

src/cheri/xycheriot.adoc
```diff
@@ -0,0 +1,678 @@
...
+
+[NOTE]
+=====
+The lack of ambient types has implications:
+
+* The <<section_rvy_sentry_insn_ext>> extension (and its <<SENTRY>> instruction)
+  requires an ambiently available type,
+  and so <<section_rvy_sentry_insn_ext>> is incompatible with {cheriot_ext_name}.
+
+* If the result of a <<CBLD>> instruction has a set {ctag},
```
vmurali, line 347: This is a weird way of stating it. The easier way to parse is if the input is sealed, then the output will be untagged.

src/cheri/xycheriot.adoc
```diff
@@ -0,0 +1,678 @@
...
+Permission masks should thus be permuted and bitwise negated.
+
+==== Elimination of Special Capability Registers
+
+CHERIoTv1 inherited from University CHERI a notion of
+"special capability registers" (SCRs; its §7.10) in addition to CSRs.
+The base {cheri_base_ext_name} ISA has instead extended some CSRs to YLEN.
+CHERIoTv1's `MTCC`, `MScratchC`, and `MEPCC` SCRs directly correspond with
+{cheri_base_ext_name}'s <<mtvec>>, <<mscratch>>, and <<mepc>> YLEN-wide CSRs.
+CHERIoTv1 additionally had a `MTDC` SCR, which has proven to be redundant,
```
vmurali: Hmm, I don't think you should state that here, especially with all the discussions we are having to preserve MTDC anyway

src/cheri/xycheriot.adoc
```diff
@@ -0,0 +1,678 @@
...
+
+Perhaps this should be some non-{cheriot_ext_name} extension.
+=====
+
+[NOTE]
+=====
+TODO, research required: can we use this to simplify trap entry, too?
+Say, by having the architecture copy <<mscratch>> over <<mtidc>> on trap entry?
+Or by having the authority used by `x0`-authorized operations
+be dependent on whether <<pcc>> grants <<asr_perm>>?
```
vmurali: Let's remove this comment? Or at least the copying part?

vmurali (approved):
Thanks a ton Wes! I can imagine how tedious it would be to write all this down.

src/cheri/zyseal.adoc
nwf: I'm not sure what you mean, sorry?  I don't really want to tacitly claim enough expertise with the RV32 default encoding to suggest which compressed permissions quadrant they should end up in, and for the RV64 default encoding, one presumes "just more bits" a la what's there and University CHERI.

Maybe I should take the time to Asciidoctor-ify the CHERIoT encoding table?

src/cheri/xycheriot.adoc
nwf: That is my intent, even if that's the wrong section to point to, but I don't think it is.

src/cheri/xycheriot.adoc
nwf: "where a capability's address is equal to its upper bound", perhaps?

src/cheri/xycheriot.adoc
nwf: Ah, no, sorry, this was originally in a [NOTE] and I moved it out to the main text without indicating that it is purely a consequence of other parts of the spec.

src/cheri/xycheriot.adoc
nwf: Since this isn't being ratified, I don't think it ties us to anything... and I think it's useful to show root-based capability (permission) invariants in addition to permission-dependency rules-based invariants?

src/cheri/xycheriot.adoc
nwf: After #767, the `YBLD` semantics are https://github.com/riscv/riscv-cheri/blob/c93971925926714489aa14946adba9ca2d75e104/src/cheri/insns/cbld_32bit.adoc?plain=1#L39-L43 .  That is, `YBLD` will prefer to build a tagged and unsealed thing rather than an untagged thing if the input bit patterns call for a non-ambient sealing type.

src/cheri/xycheriot.adoc
nwf: "had a `MTDC` SCR, which may yet make an appearance in {cheriot_ext_name}", perhaps?

src/cheri/xycheriot.adoc
nwf: OK

src/cheri/xycheriot.adoc
nwf: I added a paragraph noting that we might allow-but-not-mandate it.  We'll see how the uarches evolve. :)

src/cheri/zyseal.adoc
arichardson: That sounds like it would be quite a lot of pain - maybe easier to just include the LaTex (Tikz) image?

vmurali: I was originally under the impression that this chapter is cap encoding and instruction encoding agnostic, but I don't think that was the intent. So ignore my comment.

nwf: > That sounds like it would be quite a lot of pain - maybe easier to just include the LaTex (Tikz) image?

Too late.

src/cheri/xycheriot.adoc
vmurali: Yeah, that's better. (It would be ideal if you can state it as research-to-be-done but it's way too confusing for this chapter.)

arichardson (approved):
I am happy to merge this now - it's only part of the standalone document and it nicely documents why the hooks in the base ISA exist.

@tariqkurd-repo what do you think?

src/cheri/xycheriot.adoc
```diff
@@ -0,0 +1,712 @@
...
+
+:cheriot_ycfg_value: `0x02`
+
+:cheriot_v1_isa: link:https://cheriot.org/cheriot-sail/cheriot-architecture.pdf[The CHERIoT ISA]
+
+[appendix]
+[#section_xycheriot,reftext="{cheriot_ext_name}"]
+== "{cheriot_ext_name}" Extension and Capabilities
+
+IMPORTANT: This chapter is not part of the to-be ratified package.
```
tariqkurd-repo:
```suggestion
IMPORTANT: This chapter is not part of the v1.0 ratification package.
```

maybe like this? I'm going to add the same label in other places too.

src/cheri/xycheriot.adoc
```diff
@@ -0,0 +1,712 @@
...
+{cheriot_ext_name} extends the semantics of `{load_cap_name_lc}`
+to introduce a _capability load filter_.
+The memory block(s) in which software's dynamic allocation heap(s) are to reside
+are paired with a _revocation bitmap_,
+a bit-vector wherein each bit corresponds to a capability-sized memory granule.
+(That is, with each capability-sized and -aligned region of the primal memory.)
+Such memory block(s) are said to be _revokable_.
+These revocation bitmaps are also exposed (as memory) to software.
+
+When `{load_cap_name_lc}` transfers a capability with set {ctag} from memory
```
tariqkurd-repo:
```suggestion
When `{load_cap_name_lc}` transfers a capability with a set {ctag} from memory
```

src/cheri/xycheriot.adoc
```diff
@@ -0,0 +1,712 @@
...
+The concrete encoding table is shown in <<tab_cheriot_perms_enc>>.
+
+[NOTE]
+=====
+This second rule _adds_ <<r_perm>> as a dependency of <<zylevels1_sl_perm>>.
+Its sole effect is to exclude local and global capabilities that would otherwise
+grant exactly all of <<w_perm>>, <<c_perm>>, and <<zylevels1_sl_perm>>.
+These capabilities were judged to be the least useful of the combinations
+left after the by-construction properties of the roots and the myriad
+<<CLRPERM>> rules had winnowed the possible combinations.
```
tariqkurd-repo:
```suggestion
<<CLRPERM>> rules had pruned the possible combinations.
```

maybe this word is more familiar?

src/cheri/xycheriot.adoc
```diff
@@ -0,0 +1,712 @@
...
+* <<section_priv_cheri,{cheri_priv_m_ext}>>,
+
+* <<section_zylevels1>>, and
+
+* <<section_zyseal>>.
+
+
+The present specification presumes the *absence* of all of:
+
+* H-, S-, or U-mode,
```
tariqkurd-repo:
```suggestion
* HS-, VS-, S-, or U-mode,
```

src/cheri/xycheriot.adoc
```diff
@@ -0,0 +1,712 @@
...
+
+* <<section_rvy_sentry_insn_ext>>.
+
+[NOTE]
+=====
+While {cheriot_ext_name} is nominally compatible with
+<<section_cheri_hybrid_ext,{cheri_default_ext_name}>> and
+<<section_cheri_disable,{cheri_priv_m_reg_enable_ext}>>,
+and particular instantiations may opt to permit disabling CHERI,
+we have not yet found compelling reason to formally specify this composition.
```
tariqkurd-repo:
```suggestion
we have not yet found a compelling reason to formally specify this composition.
```

src/cheri/xycheriot.adoc
```diff
@@ -0,0 +1,712 @@
...
+
+=== Sentries
+
+{cheriot_ext_name} defines _five_ <<sentry_cap>> <<sec_cap_type>> values
+({cheriot_sentry_ot_min} through {cheriot_sentry_ot_max}, inclusive):
+
+* Any of these values may be passed as inputs to <<JALR_CHERI>>.
+  When such a <<JALR_CHERI>> instruction retires,
+  the hart's `mstatus.MIE` bit is updated as per <<tab_cheriot_sentry_irq>>.
+
```
tariqkurd-repo, line 398: So it's impossible to execute JALR out of M-mode then.
I know above that it says U-mode isn't supported, but I find this a bit surprising.

src/cheri/xycheriot.adoc
nwf: Sorry, I guess I meant [the hypervisor extension](https://riscv.github.io/riscv-isa-manual/snapshot/privileged/#hypervisor) and [the supervisor extension](https://riscv.github.io/riscv-isa-manual/snapshot/privileged/#supervisor) and U-mode (which doesn't seem to have a corresponding extension)?

src/cheri/xycheriot.adoc
nwf: CHERIoT runs entirely in M mode, using ASR to gate the very few things that resemble traditional rings.

src/cheri/xycheriot.adoc
```diff
@@ -0,0 +1,712 @@
...
+we can readily additionally require not-<<x_perm>>
+of one of <<mtidc>> and <<mscratch>>,
+whichever is not involved in the `csrrw` at the start of trap entry.
+If we can revise trap entry suitably, we could further require
+not-<<x_perm>> of both and not-sealing-root of <<mtidc>>.
+=====
+
+==== <<JALR_CHERI>> Exceptions
+
+We permit, but do not require, <<JALR_CHERI>> to raise a {cheri_excep_name_ld}
```
tariqkurd-repo: why a load access fault? why not instruction access fault?

src/cheri/xycheriot.adoc
```diff
@@ -0,0 +1,712 @@
...
+CHERIoTv1 differentiated between register selector operands intended to access
+an integer value (`xN`) from those intended to access a capability value (`cN`).
+ABI names for registers were similarly differeniated,
+using a prefix of `c` when the full capability was intended (`csp`, `ca0`, &c).
+{cheri_base_ext_name} does away with the distinction:
+all register selectors are `xN` and all ABI names are un-prefixed.
+
+==== `CGetAddr` and `CSub` Replacement
+
+CHERIoTv1 inherited from University CHERI and subsequently offered
```
tariqkurd-repo: maybe a note on what University CHERI is?

src/cheri/xycheriot.adoc
tariqkurd-repo: It's probably enough to say that CHERIoT is Machine-mode only, and currently incompatible with any other privilege level.

src/cheri/xycheriot.adoc
tariqkurd-repo: yeah - it might be nice to make it clearer that it's M-mode only above (following the comment above)

src/cheri/xycheriot.adoc
nwf: Probably a typo; happy to use whatever, so long as we can distinguish the case of "the jump instruction could not execute" and "the jump instruction computed a nonsense target".  (Probably we can, based on `mtval` not equaling `mepc`?)

tariqkurd-repo: but in both cases you're taking the exception on the JALR aren't you?
so you either need a different exception type to distinguish or this: https://github.com/riscv/riscv-cheri/issues/659

nwf: Please let's not rabbit hole on this in this PR.

tariqkurd-repo: maybe just say "an exception" and we can work out the rest later?

nwf: Sure

PR #747, opened 2025-09-01
tidy up atomic pages - no spec changes

tariqkurd-repo:
small restructuring, and making them all refer to the semantics of LY, SY to avoid replicating text


nwf (approved):
Seems OK by me

PR #749, opened 2025-09-02
Move SH*ADDY from RVY base chapter to Zba

arichardson:
These are part of the Zba extension and the CHERI versions should also
be defined there.


arichardson:
I believe #715 should only have defined the new encodings but kept them as part of Zba.

PR #750, opened 2025-09-02
Drop separate names for M/S/H extensions

arichardson:
ARC feedback was that instead of a named extension we can simply state
that this is the behaviour when RVY is implemented on the hart.


src/cheri/hypervisor-integration.adoc
```diff
@@ -1,5 +1,5 @@
 [#section_hypervisor_integration]
-== "{cheri_priv_h_ext}", Integrating {cheri_base_ext_name} and {cheri_default_ext_name} with the Hypervisor Extension
+== Integrating {cheri_base_ext_name} and {cheri_default_ext_name} with the Hypervisor Extension
```
```suggestion
== Integrating {cheri_base_ext_name} and {cheri_default_ext_name} with the "H" Extension
```

src/cheri/riscv-priv-integration.adoc
```diff
@@ -1,19 +1,19 @@
 [#section_priv_cheri]
-== "{cheri_priv_m_ext}/{cheri_priv_s_ext}" Extensions, Version 1.0
+== "Machine/Supervisor_level ISA ({cheri_base_ext_name}" Extensions, Version 1.0
```
```suggestion
== "Machine/Supervisor-Level ISA ({cheri_base_ext_name})" Extensions, Version 1.0
```

PR #751, opened 2025-09-02
Priv spec proofreading

tariqkurd-repo:


src/cheri/riscv-priv-integration.adoc
```diff
@@ -665,3 +665,160 @@ compatible with RISC-V harts provided that <<pcc>> and <<ddc>> are set to
 ==== Memory Accesses
 
 If the effective address of the memory access is invalid, and the authorizing capability's bounds do not cover all addresses, then raise a {cheri_excep_name_ld} or {cheri_excep_name_st} exception because the bounds cannot be reliably decoded.
+
+==== Glossary
```
nwf: Given the common refrain of "see the unprivileged manual", perhaps this is more a "list of terms defined in the unprivileged manual" rather than a "Glossary" per se?

tariqkurd-repo: yes good point!

src/cheri/riscv-priv-integration.adoc
```diff
@@ -429,35 +429,33 @@ NOTE: *CHERI v9 Note:* The rules for excepting have been tightened here. Also,
...
+//* XLEN in the current mode is less than MXLEN, or
+//* the endianness in the current mode is not the reset value of <<mstatus_cheri>>.MBE, or
+//* the effective CHERI-enable for the current privilege is 0.
 
 The effective CHERI-enable for the current privilege is:
 
-* Machine: `<<misa>>.{CRE}`
-* Supervisor: `<<misa>>.{CRE} & <<menvcfg>>.{CRE}`
-* User: `<<misa>>.{CRE} & <<menvcfg>>.{CRE} & <<senvcfg>>.{CRE}`
+* Machine: `<<misa,misa>>.{CRE}`
```
marnovandermaas, line 451: Why do you have to write "misa,misa" instead of just "misa"?

marnovandermaas (approved):
Nice changes!

src/cheri/riscv-priv-integration.adoc
tariqkurd-repo: to make it render correctly, otherwise it comes up with a section number instead of the name

arichardson: Could also add a reftext to the "misa" section?

arichardson: Will do in a separate PR.

PR #755, opened 2025-09-03
Distinguish between capability values and variables (registers)

arichardson:
This should address the final outstanding ARC comment.

### DIFF
--- a/src/cheri/debug-integration.adoc
+++ b/src/cheri/debug-integration.adoc
@@ -138,7 +138,7 @@ On debug mode entry <<dpc_y>> is updated with the
 capability in <<pcc>> whose address field is set to the address of the next
 instruction to be executed upon debug mode exit as described in the _RISC-V Debug Specification_.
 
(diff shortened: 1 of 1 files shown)


src/cheri/debug-integration.adoc
```diff
@@ -138,9 +138,8 @@ On debug mode entry <<dpc_y>> is updated with the
 capability in <<pcc>> whose address field is set to the address of the next
 instruction to be executed upon debug mode exit as described in the _RISC-V Debug Specification_.
 
-When leaving debug mode, the capability in <<dpc_y>> is unsealed and written
-into <<pcc>>.  A debugger may write <<dpc_y>> to change where the hart resumes
-and its mode, permissions, sealing or bounds.
+When leaving debug mode, a unsealed capability value is created from the value in <<dpc_y>> and written into <<pcc>>.
```
```suggestion
When leaving debug mode, a unsealed capability value is copied from the value in <<dpc_y>> and written into <<pcc>>.
```

PR #756, opened 2025-09-03
make Svucrglct more visible in the extension list

tariqkurd-repo:


arichardson:
probably not worth listing as a separate extension, so this sounds good to me.

Issue #199, opened 2024-04-12
Instruction for unsealing sentries with a superset


arichardson:
In previous versions of the CHERI ISA (v7/v8?) and in Morello CBuildCap was usable for unsealing sentries since it cleared the otype field. This is no longer the case, so unsealing a sentry with a superset is somewhat awkward. I believe we need the following sequence:

```
gctag t0, <sentry>
bnez t0, .Lfail
gchi t0, <sentry>
and t0, t0, ~SEALED_BIT
schi <sentry>, <sentry>, t0
cbld <unsealed>, <superset>, <sentry>
// Start using result if tag is set
gctag t0, <unsealed>
bnez t0, .Lfail
```

It would be much nicer to have a variant of cbld that did the otype clearing so we can use
```
unsentry <unsealed>, <superset>, <sentry>
gctag t0, <unsealed>
bnez t0, .Lfail
```

The checks this new instruction would need to do are:
- Is tag set?
- Is otype == sentry
- If so, clear otype and use remaining cbld logic
- else return input with tag=0

Almost all the logic is shared with cbld so I don't think this would add much complexity, only need to reserve one 3 operand instruction.

Naming is hard, could use something like `unsentry` or `ssu` (superset unseal) depending on how we think it will interact with a future otypes extension.

jrtc27:
That’s not true for ISAv9; our CBuildCap creates sentries. Morello has the behaviour you’re talking about.

arichardson:
> That’s not true for ISAv9; our CBuildCap creates sentries. Morello has the behaviour you’re talking about.

I'm pretty sure that at some point it was possible to feed a tagged sentry to cbuildcap and receive a tagged unsealed cap. Probably changed when ccopytype was added? Will update the description.

jrtc27:
It's a straight up specification bug that it takes so many instructions to unseal, rendering sealing pretty useless with regards to performance. This needs to be fixed prior to ratification (we don't seem to have a label for that?).

davidchisnall:
[...]

PR #760, opened 2025-09-03
cbld: always build unsealed caps

nwf:


arichardson:
Fixes: #199

src/cheri/insns/cbld_32bit.adoc
```diff
@@ -37,9 +39,22 @@ Otherwise, set `{cd}.tag=0`
...
+and having checked for a set tag and an expected <<sec_cap_type>> field value,
+can present, in `{cs1}`,
+a capability to a superset of the authority borne by an unsealed `c`,
+and thereby reconstruct, into `{cd}`, a modified copy of `c`, presented in `{cs2}`,
+with a zeroed <<sec_cap_type>> field and a set tag.
+
+NOTE: If software wishes to permit conditional reconstructions of sentries,
+it suffices to conditionally execute a <<SENTRY>> instruction
+on the result of a <<CBLD>>
+depending on the result of <<GCTYPE>> on the input to said <<CBLD>>.
```
arichardson, line 81: Should we add a code listing here?

```
# Assuming x1=output, x2=authority cap, x3=input bits, x4=temp
ybld x1, x2, x3
ytyper x4, x3
beqz x4, .Lunsealed
ysentry x1, x
.Lunsealed:
```
YBLD; YTYPER; BEQZ; YSEAL.

arichardson (commented):
I believe this approach is better than what we have currently: it allows superset unsealing without needing to do an awkward set-high to clear the type field. This removes the last place from the spec where user code would have to know the raw encoding.

The only downside is that we need a branch in the swap restore code, but that could be avoided with a conditional-seal extension.

I also believe this is fine for capability initialization using cbld, we either do the branch (slow) or tell the linker to sort sentry and non-sentry capabilities into two separate lists of relocations (or two relocation types) and do `YBLD` for the non-sentry ones and `YBLD;YSENTRY` for the function pointers.

src/cheri/insns/cbld_32bit.adoc
nwf: Sure, happy to if it's easier to read. :)

nwf: OK, took a stab.

src/cheri/insns/cbld_32bit.adoc
```diff
@@ -23,11 +23,13 @@ include::wavedrom/cbld.adoc[]
...
 Copy `{cs2}` to `{cd}`.
 +
+Set `{cd}.ct=0` (that is, <<CBLD>> always builds _unsealed_ capabilities).
++
 Set `{cd}.tag=1` if:
 +
 . `{cs1}.tag` is set, and
 . `{cs1}` passes all <<section_cap_integrity,integrity>> checks, and
-. `{cs1}` is not sealed, and
+. `{cs1}` is not sealed (that is, `{cs1}.ct=0`), and
```
tariqkurd-repo: I'm confused - what is cs1.ct if it's not cs1.tag?

nwf: The capability type field.

tariqkurd-repo: ok - it's not defined at such in the doc. I've been using {cs1}.tag which I thin is obvious but didn't know what this one was.

nwf: It's called CT in https://github.com/riscv/riscv-cheri/blob/7c130255e1ad9489132e43f59f3d0cec44138347/src/cheri/riscv-unpriv-integration.adoc?plain=1#L229-L230 but I'm open to using `.otype` or something instead.

vmurali: `.ct` defines the sealed or unsealed bit, not whether it is tagged or not. Maybe one can use a different name, like `.otype`, because `.ct` is really hard to parse.

jrtc27:
I don't understand what you perceive the problem to be. CHERI without any sealing has no issue, as there are no sealed caps to create. CHERI with ambient sealing but no non-ambient sealing has no issues, as it's fine to create sealed capabilities. CHERI with non-ambient sealing, whether with or without ambient sealing, has no issues, as the extension that adds new those new non-ambient sealing types can just state that they are not created by CBLD.

jrtc27:
> Specifically, one wants the following (to avoid inadvertent ambient sealing):

I disagree with this. CBLD or a CBLD-like creating sentries is a feature not a bug; it poses no security risk to create sentries instead of unsealed capabilities, and is arguably a more secure default.

vmurali:
> CHERI with non-ambient sealing, whether with or without ambient sealing, has no issues, as the extension that adds new those new non-ambient sealing types can just state that they are not created by CBLD.

The problem is because the sealing type created by JALR can also be created using CBLD using the corresponding bit-vector encoding. To stop this, the non-ambient sealing extension should basically say "CBLD's behavior is different".

vmurali:
> Yes, that's my point. If you add non-ambient types, you say that, as part of their definition, CBLD does not create them. And that's it.

Yes, that could also work (I think restricting behaviors is allowed in extensions). So we either restrict CBLD in the non-ambient sealing extension or keep CBLD as it is (with this PR's change) and add a new instruction in the ambient sealing extension to support preserving the seal

nwf:
Would the preference be then that we instead...

- leave `YBLD` as is (with #716 merely making some prose changes to formally define ambient seals)

- add an instruction (`YSUNSEAL`?) as proposed by #199 which requires tagged input and does superset unsealing (basically fusing together `YTAGR; BEQZ; YHIR; ANDI; YPACK; CBLD`)?

jrtc27:
> Would the preference be then that we instead...
> 
> * leave `YBLD` as is (with [DNM WIP: capability mediated sealing #716](https://github.com/riscv/riscv-cheri/pull/716) merely making some prose changes to formally define ambient seals)
> * add an instruction (`YSUNSEAL`?) as proposed by [Instruction for unsealing sentries with a superset #199](https://github.com/riscv/riscv-cheri/issues/199) which requires tagged input and does superset unsealing (basically fusing together `YTAGR; BEQZ; YHIR; ANDI; YPACK; CBLD`)?

That was the proposal I made earlier to Alex privately. I proposed CRBLDU (Capability ReBuiLD Unsealed), or I guess YRBLDU, as a mnemonic that reflects both its behaviour and its close ties to ~C~YBLD, since most of the implementation is the same. I also noted that it's nice to have the instruction itself require the tag on the input, so there's no risk of screwing up and forgetting to check that, since accepting untagged inputs is quite dangerous and easy to accidentally make an oracle.

PR #762, opened 2025-09-04
make C and Zca equivalent for RV64Y

tariqkurd-repo:
Zca is the subset of C without fp load/store, so C and Zca are equivalent for RVY

### DIFF
--- a/src/zc.adoc
+++ b/src/zc.adoc
@@ -55,13 +55,14 @@ The C extension is the superset of the following extensions:
 
 * Zca
 * Zcf if F is specified (RV32 only)
-* Zcd if D is specified (RV64, but not {cheri_base64_ext_name})
+* Zcd if D is specified (not supported for {cheri_base64_ext_name})
 
 As C defines the same instructions as Zca, Zcf, and Zcd, the rule is that:
 
 * C always implies Zca
 * C+F implies Zcf (RV32 only)
-* C+D implies Zcd (RV64, but not {cheri_base64_ext_name})
+* C+D implies Zcd (except for {cheri_base64_ext_name})
+
 
 [reftext="Zce"]
 === Zce


src/zc.adoc
```diff
@@ -63,6 +63,8 @@ As C defines the same instructions as Zca, Zcf, and Zcd, the rule is that:
 * C+F implies Zcf (RV32 only)
 * C+D implies Zcd
 
+NOTE: Specifying C and Zca are equivalent on {cheri_base_ext_name} architectures, as Zcf and Zcd are incompatible with that base architecture.
```
arichardson: Doesn't RV32Y still support Zcd (unless you have zcmp etc.)?

src/zc.adoc
```diff
@@ -63,7 +63,7 @@ As C defines the same instructions as Zca, Zcf, and Zcd, the rule is that:
 * C+F implies Zcf (RV32 only)
 * C+D implies Zcd
 
-NOTE: Specifying C and Zca are equivalent on {cheri_base_ext_name} architectures, as Zcf and Zcd are incompatible with that base architecture.
+NOTE: Specifying C and Zca are equivalent on {cheri_base64_ext_name}, as Zcd is incompatible with that base architecture.
```
jrtc27: Maybe architecturally, but there is weirdness due to the history of C coming before Zc* when it comes to toolchain behaviour and ABIs

jrtc27: (Though it looks like we might have cleaned that up?)

tariqkurd-repo: The point is that on RV32 you can say F+Zca and not get Zcf, or C+F and get Zcf on RVY
On RV64Y there's simply no difference - C=Zca.

src/zc.adoc
```diff
@@ -63,6 +63,8 @@ As C defines the same instructions as Zca, Zcf, and Zcd, the rule is that:
 * C+F implies Zcf (RV32 only)
 * C+D implies Zcd
 
+NOTE: Specifying C and Zca are equivalent on {cheri_base64_ext_name}, as Zcd is incompatible with that base architecture.
```
jrtc27: Regardless, this isn't non-normative. The bullet points above need to be updated with the correct predicates, and then this text becomes redundant.

tariqkurd-repo: ok - try this

src/zc.adoc
```diff
@@ -55,13 +55,13 @@ The C extension is the superset of the following extensions:
 
 * Zca
 * Zcf if F is specified (RV32 only)
-* Zcd if D is specified
+* Zcd if D is specified (RV64, but not {cheri_base64_ext_name})
```
jrtc27: RV32 includes this?..

arichardson:
```suggestion
* Zcd if D is specified (not supported for {cheri_base64_ext_name})
```

src/zc.adoc
```diff
@@ -55,13 +55,13 @@ The C extension is the superset of the following extensions:
...
 
 As C defines the same instructions as Zca, Zcf, and Zcd, the rule is that:
 
 * C always implies Zca
 * C+F implies Zcf (RV32 only)
-* C+D implies Zcd
+* C+D implies Zcd (RV64, but not {cheri_base64_ext_name})
```
arichardson:
```suggestion
* C+D implies Zcd (except for {cheri_base64_ext_name})

```

PR #767, opened 2025-09-05
Split reconstruction and superset unsealing

nwf:
This is an alternative to #760 as per discussion over there.  It introduces a notion of ambient sealing types, defines `YBLD` explicitly as sealing those types, and defines type `1` as ambient.  Separately, it introduces a `YSUNSEAL` instruction for explicit superset unsealing, which is a lot like `YBLD` but requires a tagged input.

There are two other commits here that are purely clerical: adding xrefs and changing the variable names, not their rendered values, of the sentry instruction extension.

If this is agreeable, I will rework #716 accordingly.

Fixes: #199.


src/cheri/insns/ysunseal_32bit.adoc
```diff
@@ -0,0 +1,74 @@
...
+a capability to a superset of the authority borne by an unsealed `c`,
+and thereby reconstruct, into `{cd}`, an unsealed copy of `c` (from `{cs2}`).
+For example, to unseal from `x3` into `x1`
+using the superset in `x2` with scratch `x4`, such software might look like:
+
+[subs="+attributes"]
+----
+{GCTYPE_LC} x4, x3 // check that input is sealed
+beqz x4, fail
+{CBLD_LC} x1, x2, x3
```
arichardson:
```suggestion
{YSUNSEAL} x1, x2, x3
```

nwf: D'oh!

src/cheri/insns/ysunseal_32bit.adoc
```diff
@@ -0,0 +1,75 @@
...
+. any extension-specific constraints on <<YSUNSEAL>> hold.
+
++
+Otherwise, set `{cd}.tag=0`
+
+NOTE: When `{cs1}` is `{creg}0` <<YSUNSEAL>> will copy `{cs2}` to `{cd}` and clear `{cd}.tag` and `{cd}.ct`.
+However future extensions may add additional behavior to update currently reserved fields,
+and so software should not assume `{cs1}==0` to be a pseudo-instruction for {ctag} and type clearing.
+
+NOTE: {YSUNSEAL} is essentially <<CBLD>> except that it
```
tariqkurd-repo:
```suggestion
NOTE: <<YSUNSEAL>> is essentially <<CBLD>> except that it
```

src/cheri/insns/ysunseal_32bit.adoc
```diff
@@ -0,0 +1,75 @@
...
+requires `{cs2}` to be tagged and always sets `{cd}.ct` to `0`.
+
+[NOTE]
+=====
+<<YSUNSEAL>> is intended to enable "superset unsealing":
+wherein software, given a capability in `{cs2}`, can present, in `{cs1}`,
+a capability to a superset of the authority granted were the capability in `{cs2}` to be unsealed,
+and thereby reconstruct, into `{cd}`, an unsealed copy of `{cs2}`.
```
tariqkurd-repo:
```suggestion
a capability with a superset of the authority granted of the capability in `{cs2}` to be unsealed,
and thereby reconstruct, into `{cd}`, an unsealed copy of `{cs2}`.
```

I find it easier to read this way..?!

nwf: "granted of" is probably not right, and the _sealed_ capability in `{cs2}` by definition grants no authority to its bearer, so some hypothetical -- the subjunctive "were" -- seems required?

I took a stab at better text.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -696,6 +715,7 @@ Special instructions are provided to copy capabilities or derive a new capabilit
 |<<SCBNDS>>   |Set register bounds on capability with rounding, clear {ctag} if rounding is required
 |<<SCBNDSR>>  |Set bounds on capability with rounding up as required
 |<<CBLD>>     |Set {cd} to {cs2} with its {ctag} set after checking that {cs2} is a subset of {cs1}
+|<<YSUNSEAL>> |Set {cd} to {cs2} with a zero <<sec_cap_type>> if {cs2} is otherwise a subset of {cs1}
```
```suggestion
|<<YSUNSEAL>> |Set {cd} to {cs2} with a zero <<sec_cap_type>> if {cs2} is sealed otherwise a subset of {cs1}
```

src/cheri/zylevels1.adoc
```diff
@@ -156,13 +156,17 @@ indicates that `{cs1}` is a subset of `{cs2}` (that is, sets `rd` to 1) only if
...
+=== Interaction with <<CBLD>> and <<YSUNSEAL>>
 
 A <<CBLD>> instruction `{CBLD_LC} {cd}, {cs1}, {cs2}` may yield `{cd}.tag=1` only if either
 
 * `{cs2}` 's <<zylevels1_gl_perm>> is clear or
 * `{cs1}` 's <<zylevels1_gl_perm>> is set.
 
+Identical requirements hold for a <<YSUNSEAL>> instruction `{YSUNSEAL_LC} {cd}, {cs1}, {cs2}`.
+
+#TODO: Or would we prefer that `{cd}` 's <<zylevels1_gl_perm>> is the AND of those of `{cs1}` and `{cs2}`?#
```
tariqkurd-repo: to be resolved still?

nwf: Yes, probably.  Since I don't have experience with superset unsealing in practice, I'll defer to @jrtc27, @arichardson, or anyone else with more insight.

ETA: tentatively, I've gone and done the "same as ISAv9 thing" here and have updated https://github.com/riscv/riscv-cheri/pull/780 to un-drop some text and do the same thing.

src/cheri/insns/ysunseal_32bit.adoc
```diff
@@ -0,0 +1,76 @@
...
+
+[#YSUNSEAL,reftext="{YSUNSEAL}"]
+==== {YSUNSEAL}
+
+#New since last ARC review#
+
+ifdef::cheri_v9_annotations[]
+NOTE: *CHERI v9 Note:* {YSUNSEAL} does not use ddc if {cs1}==0
+endif::[]
+
```
arichardson:
```suggestion
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -257,6 +257,25 @@ Capabilities with `CT≠0` are sealed against modification and cannot be derefer
 Instructions that operate on capabilities will produce a result with a cleared {ctag} if the source capability is sealed and the operation would alter its address, bounds, or permissions.
 Extensions that augment capability metadata must describe their interaction(s) with sealed capabilities.
 
+Given a capability with `CT=0`,
```
arichardson, line 262:
```suggestion
#START: New since ARC review#
Given a capability with `CT=0`,
```

Maybe worth adding a start + end to make it obvious which part is new?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -257,6 +257,25 @@ Capabilities with `CT≠0` are sealed against modification and cannot be derefer
...
+extensions will specify how software expresses this authority
+for types not defined above.
+
+[#sec_cap_type_ambient]
+Ambient sealing type::
+Some capability types are said to be "ambiently available" (or just "ambient")
+if they do not require specific authority to seal a capability (with that type).
+Type 1 is considered ambiently available on a given platform
+if <<section_rvy_sentry_insn_ext>> is available on that platform.
+#New since last ARC review#
```
```suggestion
#END: New since last ARC review#
```

src/cheri/insns/cbld_32bit.adoc
```diff
@@ -34,6 +34,13 @@ Set `{cd}.tag=1` if:
 +
 Otherwise, set `{cd}.tag=0`
 
+#New since last ARC review#
```
arichardson: Maybe worth adding a start+end for the new since review? Or we highlight the whole region in yellow?

arichardson (approved):
This looks great, my only comment is that we should probably somehow delimit the newly added regions. Either make the whole thing highlighted or add a start+end annotation.

src/cheri/insns/cbld_32bit.adoc
nwf: I've no idea how asciidoctor would compose formatting, so I've gone with begin/end.

src/cheri/insns/ysunseal_32bit.adoc
nwf: Removed.

src/cheri/insns/ysunseal_32bit.adoc
```diff
@@ -0,0 +1,74 @@
...
+Set `{cd}.ct=0`.  (That is, unseal `{cd}`.)
++
+Set `{cd}.tag=1` if:
++
+. `{cs1}.tag` is set, and
+. `{cs1}` passes all <<section_cap_integrity,integrity>> checks, and
+. `{cs1}` is not sealed (that is, `{cs1}` has zero <<sec_cap_type>>), and
+. `{cs2}.tag` is set, and
+. `{cs2}` passes all <<section_cap_integrity,integrity>> checks, and
+. `{cs2}` is sealed (that is, `{cs2}` has non-zero <<sec_cap_type>>), and
```
jrtc27, line 31: I wonder if this is something we want to require or not. It's a no-op (or CClearTag if cs1 doesn't have sufficient authority) if you give it an unsealed cap. Wondering if there are any cases where you'd want to not have to conditionally YUNSEAL.

src/cheri/insns/ysunseal_32bit.adoc
```diff
@@ -0,0 +1,74 @@
...
+. any extension-specific constraints on <<YSUNSEAL>> hold.
+
++
+Otherwise, set `{cd}.tag=0`
+
+NOTE: When `{cs1}` is `{creg}0` <<YSUNSEAL>> will copy `{cs2}` to `{cd}` and clear `{cd}.tag` and `{cd}.ct`.
+However future extensions may add additional behavior to update currently reserved fields,
+and so software should not assume `{cs1}==0` to be a pseudo-instruction for {ctag} and type clearing.
+
+NOTE: <<YSUNSEAL>> is essentially <<CBLD>> except that it
```
jrtc27: I worry this might undermine the justification for its addition

src/cheri/insns/ysunseal_32bit.adoc
```diff
@@ -0,0 +1,74 @@
...
+requires `{cs2}` to be tagged and always sets `{cd}.ct` to `0`.
+
+[NOTE]
+=====
+<<YSUNSEAL>> is intended to enable "superset unsealing":
+wherein software, given a capability in `{cs2}`, can present, in `{cs1}`,
+a capability granting a superset of the authority
+that would be granted by an unsealed version of the capability in `{cs2}`,
+and thereby reconstruct, into `{cd}`, that unsealed version of `{cs2}`.
+For example, to unseal from `x3` into `x1`
```
jrtc27: Don't we need `{creg}` instead of `x` throughout the rest of this?

src/cheri/insns/ysunseal_32bit.adoc
```diff
@@ -0,0 +1,74 @@
...
+using the superset in `x2` with scratch `x4`, such software might look like:
+
+[subs="+attributes"]
+----
+{YSUNSEAL_LC} x1, x2, x3
+{GCTAG_LC} x4, x1  // check that output is tagged
```
jrtc27: I mean, depending on what you're doing you may pass it on (e.g. to a system call that could return EFAULT) or just dereference at risk of SIGPROT.

src/cheri/insns/ysunseal_32bit.adoc
```diff
@@ -0,0 +1,74 @@
...
+beqz x4, fail
+----
+
+Additional instructions (a <<GCTYPE>> with `{cs1}=x3` and comparisons)
```
jrtc27: To the uninitiated it's not clear what this is referring to. I *assume* the point is "you may wish to pick a different authority based on the type"? But that may also be true based on the address alone if you have multiple allocation pools. So this is weirdly specific.

src/cheri/insns/ysunseal_32bit.adoc
```diff
@@ -0,0 +1,74 @@
...
+may be required if the encoding implements
+multiple <<sec_cap_type>> values distinguishing (un)sealed capabilities.
```
jrtc27: Not sure what the (un) is doing here, there's only ever one unsealed type

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -257,6 +257,28 @@ Capabilities with `CT≠0` are sealed against modification and cannot be derefer
...
+to operate at the non-zero type;
+extensions will specify how software expresses this authority
+for types not defined above.
+
+[#sec_cap_type_ambient]
+Ambient sealing type::
+Some capability types are said to be "ambiently available" (or just "ambient")
+if they do not require specific authority to seal a capability (with that type).
+Type 1 is considered ambiently available on a given platform
+if <<section_rvy_sentry_insn_ext>> is available on that platform.
```
jrtc27, line 278: I still strongly believe this should be a mandatory feature in the base ISA that CHERIoT's software platform can choose to not take advantage of and that CHERIoT's extra jump-and-assert-for-CFI instructions can choose to forbid as inputs (and perhaps forbid for the baseline RISC-V ret too). But that's not for this PR...

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -936,11 +960,21 @@ If the target of a taken branch lies outside the bounds of <<pcc>>, the next ins
 
 NOTE: A future extension may raise an exception on the branch instruction itself if fetching a minimum sized instruction at the target <<pcc>> will raise a _{cheri_excep_name_pc}_.
 
+[#section_rvy_sentry_insn_ext,reftext="{rvy_sentry_insn_ext_name}"]
+== "{rvy_sentry_insn_ext_name}" Extension for Creation of Sentry Capabilities
+
+#Begin changed since last ARC review#
+
+In the {cheri_base_ext_name} base ISA the only ways to create a <<sentry_cap>>
```
jrtc27: This is odd. If you don't have ambient sentries, CBLD will never create a sentry, so it's dependent on the sentry extension. JALR also won't give you a sentry if you have neither the sentry extension nor CHERIoT's sentry family extension.

src/cheri/zylevels1.adoc
```diff
@@ -156,13 +156,20 @@ indicates that `{cs1}` is a subset of `{cs2}` (that is, sets `rd` to 1) only if
...
 
-=== Interaction with <<CBLD>>
+=== Interaction with <<CBLD>> and <<YSUNSEAL>>
 
 A <<CBLD>> instruction `{CBLD_LC} {cd}, {cs1}, {cs2}` may yield `{cd}.tag=1` only if either
 
 * `{cs2}` 's <<zylevels1_gl_perm>> is clear or
 * `{cs1}` 's <<zylevels1_gl_perm>> is set.
 
+A <<YSUNSEAL>> instruction `{YSUNSEAL_LC} {cd}, {cs1}, {cs2}`
```
jrtc27, line 170: I trust someone who's thought in detail about levels has checked that this makes sense? Noting that this differs in description to CBLD.

src/cheri/insns/ysunseal_32bit.adoc
arichardson: That's fair we can drop the note?

arichardson: Or maybe be more explicit in that it allows unsealing without having to know the exact bit representation of capabilities to mess with the raw CT encoding.

jrtc27: I don't mind which, so long as it's not this. If we can articulate the justification well then it can stay.

src/cheri/riscv-unpriv-integration.adoc
nwf: In principle, a capability encoding could define ambient types without the platform also opting into the `Zys` extension.

(This prose is also updated over in #768, where JALR's sentry behavior is split out to Zysentry.)

src/cheri/zylevels1.adoc
nwf: It is the behavior of ISAv9 CUnseal (and so CHERIoT's, by inheritance).

ETA: And CBLD's behavior is that of ISAv9's CBuildCap.

src/cheri/insns/ysunseal_32bit.adoc
nwf: Not requiring this here means that software that wanted to enforce it would need to `YTYPER` and branch prior to `YSUNSEAL`, and "I mean to be unsealing a sealed handle" seems more common than "I am willing to accept either a sealed or unsealed thing".  I'm open to claims in the other direction, though, since I have experience exclusively with mediated seals.

src/cheri/riscv-unpriv-integration.adoc
nwf: A consequence of this PR's refactoring is that all non-zero otypes are entirely a function of the encoding in use rather than needing something more akin to a central registry of otypes/concepts in the base spec that may or may not be supported by a particular platform.

src/cheri/insns/ysunseal_32bit.adoc
nwf: Took a stab with new prose.

src/cheri/insns/ysunseal_32bit.adoc
nwf: Deleted instead.

src/cheri/insns/ysunseal_32bit.adoc
nwf: Sure; deleted code in favor of prose.

src/cheri/insns/ysunseal_32bit.adoc
nwf: No, the intent was more "perhaps you wish to restrict which otype(s) are acceptable on handles you're unsealing", since superset unsealing is not, itself, sensitive to which otype is in use.

Hopefully the new wording is more clear.

src/cheri/insns/ysunseal_32bit.adoc
arichardson: I'd lean towards requiring a branch if you want to accept either sealed or unsealed and have the "require sealed" case be the one that can be done without a branch prior.

src/cheri/riscv-unpriv-integration.adoc
arichardson: Since this is adjusted by #768 again, I think the current proposed wording is fine.

src/cheri/insns/ysunseal_32bit.adoc
arichardson: This looks fine to me. But since we mention that it's useful for superset unsealing in the NOTE below I don't think we necessarily need the rationale for why it exists in addition to CBLD. I'd lean towards dropping/commenting this note but don't mind either way.

src/cheri/insns/ysunseal_32bit.adoc
arichardson: Followed up offline and conclusion is that either is fine. We just have to pick one. Keeping the current one should be good.

arichardson (approved):
Thanks for the updates! will merge once @jrtc27 confirms it all looks good.

src/cheri/riscv-unpriv-integration.adoc
jrtc27: #768 doesn't really make this any better. I just believe this block of text is more harmful than helpful. If you have an ambient type that CBLD can create then you should have an instruction that can do so on an existing (unsealed) capability; not doing so makes no sense. So why do we need to call out CBLD here? And the only time when JALR is writing back capabilities with ambient types is when CBLD can also create that capability, because it's ambient, and thus there should also be instruction for it. If you don't have Zys then I don't understand what world you would still want ambient types in. It's just weird to say "hey look how you can create this type in the base ISA despite it not existing, and let's ignore that once the type exists there's also a new instruction that can do so".

arichardson: Are you okay with merging this with this paragraph removed?

jrtc27: Sure, if there's nothing here that others believe needs to be salvaged

src/cheri/insns/ysunseal_32bit.adoc
nwf: Hrm.  This note wasn't really meant as rationale so much as just noting their similarities.  I've commented it out in case there's something later useful there.

src/cheri/riscv-unpriv-integration.adoc
nwf: Right, great, out it goes.

PR #768, opened 2025-09-05
Split out JALR's unsealing behavior to Zysentry

nwf:
This is part of #716 sat atop #767 (go read that first).  It should have no semantic change, but it starts to set the stage for Xcheriot.  The Zyseal extension from #716 to follow.


src/cheri/debug-integration.adoc
```diff
@@ -203,7 +203,15 @@ The <<m_bit>> of this capability is _only_ updated by executing <<MODESW_CAP>> o
...
-The `DRET` instruction reads the full YLEN bits of the <<mepc_y>>/<<sepc_y>> register and unseals it prior to exception return if it is a <<sentry_cap>>.
+If the platform incorporates <<section_zysentry>>,
+the `DRET` instruction reads the full YLEN bits of the <<mepc_y>>/<<sepc_y>> register and unseals it prior to exception return
+if the read capability's type
+is defined as acceptable for <<JALR_CHERI>> to unseal.
+(If there are conditions on <<JALR_CHERI>>'s willingness
+to unseal particular types, those do not apply here;
+that is, if there is any circumstance
+under which <<JALR_CHERI>> would unseal the type in question,
+`DRET` unseals the capability in question.)
```
vmurali: Are you trying to point out that DRET will not unseal non-sentry seals?

nwf: Yes, and simultaneously intending to answer "what about Xcheriot forward/backward sentries?" with "it will unseal any of those five, because JALR would unseal each of them in some circumstance".

src/cheri/app-cap-description.adoc
```diff
@@ -294,7 +294,21 @@ WARNING: *CHERI v9:* There is now a 1-bit otype (sentry or unsealed) and the old
...
 endif::[]
 
-For {cheri_base_ext_name} the CT field is one bit wide and maps 1:1 to the capability types listed in <<tab_cap_types>>: it is a sentry capability if the bit is 1 or unsealed if it is 0.
+The capabilities of this chapter define a 1-bit field for <<sec_cap_type>> values;
+this field directly encodes the values `0` and `1`.
+The value `1` is...
+
+* considered <<sec_cap_type_ambient,ambiently available>> for <<CBLD>>,
+
+* used as the sealing type for <<SENTRY>> instructions,
```
tariqkurd-repo: should this say `type of the sealed capability`? It's not involved in the sealing process - it's the output (just trying to be more specific).

src/cheri/app-cap-description.adoc
```diff
@@ -294,7 +294,21 @@ WARNING: *CHERI v9:* There is now a 1-bit otype (sentry or unsealed) and the old
...
+  if <<section_rvy_sentry_insn_ext>> is present in the platform.
+
+Additionally, if the platform includes the <<section_zysentry>> extension, then
+<<JALR_CHERI>> both
+
+* unseals input capabilities of type 1 and
+
+* seals its return capabilities at type 1.
```
tariqkurd-repo:
```suggestion
* seals its return capabilities with type 1.
```

reads a bit more naturally to me...

src/cheri/introduction.adoc
```diff
@@ -76,7 +77,8 @@ CAUTION: The extension names are provisional and subject to change.
 |=============================================================================================================================================================
 | Extension                                                 | Status        | Comment
 |<<rv32y,{cheri_base_ext_name}>>                            | Stable        | This extension is a candidate for freezing
-|<<SENTRY,{rvy_sentry_ext_name}>>                           | Stable        | This extension is a candidate for freezing
+|<<section_rvy_sentry_insn_ext>>                            | Stable        | This extension is a candidate for freezing
+|<<section_rvy_sentry_jalr_ext>>                            | Stable        | This extension is a candidate for freezing
```
tariqkurd-repo: this comes up as unresolved

src/cheri/zysentry.adoc
```diff
@@ -0,0 +1,58 @@
+[#section_zysentry,reftext="{cheri_sentry_ext_name}"]
+== "{cheri_sentry_ext_name}" Extension for CHERI Sentry Capabilities
```
tariqkurd-repo: this variable is unresolved - I think it should be `rvy_sentry_ext_insn_name`

src/cheri/riscv-priv-integration.adoc
```diff
@@ -447,7 +447,16 @@ NOTE: A future extension may add CHERI checks to the page table walker.
 
 === Modified Trap-Return Instructions Behavior
 
-When the {cheri_base_ext_name} base ISA is implemented, the <<otherpriv,trap-return instructions>> (`MRET` and `SRET`) read the full YLEN bits of the <<mepc_y>>/<<sepc_y>> register and unseal it prior to exception return if it is a <<sentry_cap>>.
+When the {cheri_base_ext_name} base ISA is implemented
+and the platform incorporates <<section_zysentry>>,
```
tariqkurd-repo: comes up as unresolved

src/cheri/debug-integration.adoc
```diff
@@ -203,7 +203,15 @@ The <<m_bit>> of this capability is _only_ updated by executing <<MODESW_CAP>> o
 include::img/drootcreg.edn[]
 
 === Modified Trap-Return Instruction Behavior
-The `DRET` instruction reads the full YLEN bits of the <<mepc_y>>/<<sepc_y>> register and unseals it prior to exception return if it is a <<sentry_cap>>.
+If the platform incorporates <<section_zysentry>>,
+the `DRET` instruction reads the full YLEN bits of the <<mepc_y>>/<<sepc_y>> register and unseals it prior to exception return
+if the read capability's type
+is defined as acceptable for <<JALR_CHERI>> to unseal.
+(If there are conditions on <<JALR_CHERI>>'s willingness
```
```suggestion
(If there are conditions on <<JALR_CHERI>>'s ability
```
willingness seems strange - it's not a teenager who we're trying to persuade to do something 😉

src/cheri/debug-integration.adoc
```diff
@@ -203,7 +203,15 @@ The <<m_bit>> of this capability is _only_ updated by executing <<MODESW_CAP>> o
 include::img/drootcreg.edn[]
 
 === Modified Trap-Return Instruction Behavior
-The `DRET` instruction reads the full YLEN bits of the <<mepc_y>>/<<sepc_y>> register and unseals it prior to exception return if it is a <<sentry_cap>>.
+If the platform incorporates <<section_zysentry>>,
+the `DRET` instruction reads the full YLEN bits of the <<mepc_y>>/<<sepc_y>> register and unseals it prior to exception return
```
```suggestion
the `DRET` instruction reads the full YLEN bits of the <<mepc_y>>/<<sepc_y>> register and unseals it prior to exception return,
```

src/cheri/riscv-priv-integration.adoc
```diff
@@ -447,7 +447,16 @@ NOTE: A future extension may add CHERI checks to the page table walker.
...
+the <<otherpriv,trap-return instructions>> (`MRET` and `SRET`) read the full YLEN bits of the <<mepc_y>>/<<sepc_y>> register and unseal it prior to exception return
```
```suggestion
the <<otherpriv,trap-return instructions>> (`MRET` and `SRET`) read the full YLEN bits of the <<mepc_y>>/<<sepc_y>> register and unseal it prior to exception return,
```

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -19,12 +19,21 @@ Indirect jump to the target capability in `{cs1}` with an address offset.
 +
 Copy `{cs1}` to the target <<pcc>>
 +
-. If the target <<pcc>> is a sentry, and the `offset` is zero, then unseal the target <<pcc>>.
-Otherwise, increment the address of the target <<pcc>> by the sign-extended 12-bit `offset`.
+. If the platform includes the <<section_zysentry>> extension,
```
tariqkurd-repo: unresolved

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -925,14 +927,38 @@ If the target of a taken branch lies outside the bounds of <<pcc>>, the next ins
...
 
-NOTE: Future extensions to {cheri_base_ext_name} (such as CHERIoT, see <<uycfg>>) support additional <<sec_cap_type>> values as alternatives to the <<sentry_cap>> (<<sec_cap_type>>=1), so this instruction is part of an optional extension instead of the {cheri_base_ext_name} base ISA.
+* careful bit manipulation and <<CBLD>>.
+
+The {rvy_sentry_insn_ext_name} extension adds the <<SENTRY>> instruction to allow sealing arbitrary capabilities
+at a <<sec_cap_type>> value specified by the capability encoding.
+
+NOTE: In principle, the type used by <<SENTRY>>
+may not be any of the type(s) used by <<section_Zysentry>>,
+though that is not true of the capabilities of <<app_cap_description>>.
```
tariqkurd-repo: <<app_cap_description>> renders as `Appendix A` which is not very helpful. In `uycfg` we call it `default encoding` and I think that the reftext for that heading should say that.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -255,19 +253,23 @@ Capabilities with `CT≠0` are sealed against modification and cannot be derefer
...
+(or "unsealing at ``CT``" when a particular input ``CT`` is meant).
+In general, each of these actions may require *authority*
+to operate at the non-zero type;
+extensions will specify how software expresses this authority
+for types not defined above.
+
+[#sec_cap_type_ambient]
+Ambient sealing type::
+Some capability types are said to be "ambiently available" (or just "ambient")
+if they require no particular authority to use to seal a capability.
```
```suggestion
if they don't require a specific permission to be granted to seal a capability.
```

I find "no particular authority" quite vague - I think we're saying that there's no specific sealing permission required.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -229,21 +229,19 @@ image::../cheri/img/cap-bounds-map.png[width=80%,align=center]
...
-|Type |Hardware interpretation
-
-|0    | Unsealed capability
-|1    | Sealed entry (sentry) capability
-|===
-
-NOTE: Future extensions will add more types of sealed capabilities by extending the width of the CT-field.
- In the base ISA, sentries are the only type of sealed capability and the CT field is only 1-bit.
+This metadata value indicates the type of the capability.
+The sole type defined in {cheri_base_ext_name} is `0`.
```
```suggestion
The sole type defined in the {cheri_base_ext_name} base ISA is `0`.
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -255,19 +253,23 @@ Capabilities with `CT≠0` are sealed against modification and cannot be derefer
...
+to operate at the non-zero type;
+extensions will specify how software expresses this authority
+for types not defined above.
+
+[#sec_cap_type_ambient]
+Ambient sealing type::
+Some capability types are said to be "ambiently available" (or just "ambient")
+if they require no particular authority to use to seal a capability.
+Type 1 is considered ambiently available on a given platform
+if <<section_rvy_sentry_insn_ext>> is available on that platform.
```
tariqkurd-repo: an extra xref here to the default encoding would help

src/cheri/insns/ysunseal_32bit.adoc
```diff
@@ -0,0 +1,77 @@
...
+
+[NOTE]
+=====
+<<YSUNSEAL>> is intended to enable "superset unsealing",
+wherein software, given a capability `c`
+and having checked for an expected <<sec_cap_type>> field value,
+can present, in `{cs1}`,
+a capability to a superset of the authority borne by an unsealed `c`,
+and thereby reconstruct, into `{cd}`, an unsealed copy of `c` (from `{cs2}`).
+For example, to unseal from `x3` into `x1`
```
tariqkurd-repo: this reads ok spread over multiple lines like this but is really difficult to read when in a single paragraph, so please make it into a list so that the rendered version is easy to read.

src/cheri/introduction.adoc
```diff
@@ -76,7 +77,8 @@ CAUTION: The extension names are provisional and subject to change.
 |=============================================================================================================================================================
 | Extension                                                 | Status        | Comment
 |<<rv32y,{cheri_base_ext_name}>>                            | Stable        | This extension is a candidate for freezing
-|<<SENTRY,{rvy_sentry_ext_name}>>                           | Stable        | This extension is a candidate for freezing
+|<<section_rvy_sentry_insn_ext>>                            | Stable        | This extension is a candidate for freezing
```
tariqkurd-repo, line 79: should the YSUNSEAL extension be in this list?

src/cheri/insns/ysunseal_32bit.adoc
```diff
@@ -0,0 +1,77 @@
...
+
+#TODO: Should {YSUNSEAL} also require the input to be sealed?#
+
+[NOTE]
+=====
+<<YSUNSEAL>> is intended to enable "superset unsealing",
+wherein software, given a capability `c`
+and having checked for an expected <<sec_cap_type>> field value,
+can present, in `{cs1}`,
+a capability to a superset of the authority borne by an unsealed `c`,
```
tariqkurd-repo: borne -> granted for consistency?

src/cheri/insns/ysunseal_32bit.adoc
nwf: A list of what, exactly, sorry?

nwf:
I had sincerely hoped to get something like this in before (re)submission to ARC precisely because it is moderately invasive while also being requisite (AFAICT) for the "avoid multiple base ISAs" outcome we were looking for.  I'll add #changed since last ARC review# everywhere I've touched.

src/cheri/insns/ysunseal_32bit.adoc
tariqkurd-repo: lines 50-54 turns into a really long hard to read sentence, so either could be a bulleted list, or broken into more sentences.

nwf: I took a stab at it.  Please use #767 to discuss future YSUNSEAL-specific changes and leave this PR for the topmost commit?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -257,19 +255,25 @@ Capabilities with `CT≠0` are sealed against modification and cannot be derefer
...
-They are used as the target of <<JALR_CHERI>>, which must have a zero offset for the use of a sentry to be valid.
-
-Sentry capabilities can establish a form of control-flow integrity between mutually distrusting code.
-<<JALR_CHERI>> with zero offset automatically unseals a sentry target capability and installs it in the <<pcc,program counter capability>>.
-
-<<JALR_CHERI>> also seals the return register, so the callee can return to the caller but cannot access the memory pointed at by the return register, and cannot return elsewhere.
+Given a capability with `CT=0`,
+deriving a capability with `CT≠0` is termed "sealing"
+(or "sealing with type ``x``" when a particular output ``CT=x`` is meant).
+Dually,
```
```suggestion
In addition,
```

Thinking about non-native english speakers here

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -257,19 +255,25 @@ Capabilities with `CT≠0` are sealed against modification and cannot be derefer
...
+Ambient sealing type::
+Some capability types are said to be "ambiently available" (or just "ambient")
+if they require no particular authority to use to seal a capability.
+For example, if <<section_rvy_sentry_insn_ext>> is available on a given platform,
+the type with which it seals capabilities is considered ambiently available.
+(With the capabilities of <<app_cap_description>>, that would be type `1`.)
+#New since last ARC review#
 
 NOTE: In addition to using sealed capabilities as sentries for secure entry points, sealed capabilities can also be useful to software as secure software tokens.
  They can be converted to an unsealed capability by extracting the metadata with <<GCHI>>, zeroing the `CT` field, reapplying it with <<SCHI>> and then rebuilding via a superset capability with the <<CBLD>> instruction.
```
tariqkurd-repo: YSUNSEAL does this doesn't is? so we don't need the sequence anymore?
I think it's fixed in the other PR.....

src/cheri/riscv-priv-integration.adoc
```diff
@@ -447,7 +447,18 @@ NOTE: A future extension may add CHERI checks to the page table walker.
 
 === Modified Trap-Return Instructions Behavior
 
-When the {cheri_base_ext_name} base ISA is implemented, the <<otherpriv,trap-return instructions>> (`MRET` and `SRET`) read the full YLEN bits of the <<mepc_y>>/<<sepc_y>> register and unseal it prior to exception return if it is a <<sentry_cap>>.
+#Changed since last ARC review#
+
+When the {cheri_base_ext_name} base ISA is implemented
+and the platform incorporates <<section_rvy_sentry_jalr_ext>>,
```
tariqkurd-repo: unresolved xref (in a few places)

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -933,13 +939,26 @@ If the target of a taken branch lies outside the bounds of <<pcc>>, the next ins
 
 NOTE: A future extension may raise an exception on the branch instruction itself if fetching a minimum sized instruction at the target <<pcc>> will raise a _{cheri_excep_name_pc}_.
 
+[#section_rvy_sentry_insn_ext,reftext="{rvy_sentry_insn_ext_name}"]
+== "{rvy_sentry_insn_ext_name}" Extension for Creation of Sentry Capabilities
+
+#Changed since last ARC review#
+
+In the {cheri_base_ext_name} base ISA the only ways to create
+a sealed capability are either through
```
```suggestion
a sealed capability are either through:
```

src/cheri/insns/sentry_32bit.adoc
```diff
@@ -19,7 +19,10 @@ Encoding::
 include::wavedrom/sentry.adoc[]
 
 Description::
-Copy `{cs1}` to `{cd}`. Set the capability type of `{cd}` to <<sentry_cap>>.
+Copy `{cs1}` to `{cd}`.
++
+Set the capability type (<<sec_cap_type>>) of `{cd}` to
+the <<sec_cap_type_ambient,ambient>> value specified by the capability encoding.
```
tariqkurd-repo, line 27: I think that here, as there's a depency on the encoding, we should have a table specifying the encoding (uycfg value) and the value / values which are legal.
This will be necessary for AUIPC to specify the shift value too, so better have it here first.
This method was suggested during the review with Krste.

src/cheri/riscv-unpriv-integration.adoc
nwf: Indeed.  I've updated #767.

src/cheri/insns/sentry_32bit.adoc
nwf: I don't really like the idea of duplicating information like that.  "Single source of truth" &c.

src/cheri/riscv-priv-integration.adoc
nwf: It surely shouldn't be; the anchor is at the top of `src/cheri/zysentry.adoc`... oh.  Where should `include::zysentry.adoc[]` go in `src/riscv-privileged.adoc`, or should this be replaced with `{rvy_sentry_jalr_ext_name}` so as to not be a hyperlink?

I guess the latter is more in keeping with the rest of the privileged document, so I did that.

src/cheri/insns/sentry_32bit.adoc
tariqkurd-repo: fair point - but when you've included the CHERIoT chapter the <<sec_cap_type>> link will be a problem as it links to the default encoding. Wouldn't you want a second link to the CHERIoT version?

maybe something like:

uycfg.base={default}: values in <<sec_cap_type>>
uycfg.base={cheriot}: values in <<cheriot_sec_cap_type>>

something like this so the information is easy to find?
that way we keep the single source of truth through the links.

nwf: > fair point - but when you've included the CHERIoT chapter the <<sec_cap_type>> link will be a problem as it links to the default encoding. Wouldn't you want a second like to the CHERIoT version?

No, `<<sec_cap_type>>` links to the architectural, non-encoding-specific https://github.com/riscv/riscv-cheri/blob/d39b6ea4f78d6dbcaf9e33e0f98477008c59e82f/src/cheri/riscv-unpriv-integration.adoc?plain=1#L231 .

tariqkurd-repo: yes ok - but it does say the value when you follow the link - I guess we'll cross the CHERIoT bridge when we come to it (soon...!)

tariqkurd-repo (approved):

src/cheri/attributes.adoc
```diff
@@ -72,6 +72,8 @@ endif::support_varxlen[]
 // Extension for capability levels (flow control)
 :cheri_levels1_ext_name:             Zylevels1
 
+:rvy_sentry_jalr_ext_name: Zysentry
```
arichardson: Should we just merge this with the Zys extension now that the jalr sealing is conditional on ambient sentries being available? I am not sure we need a separate extension, can't we say Zys introduces type=1 and updates JALR to seal with type 1? The base spec can add a note that extensions can define that JALR seals with a given type. E.g. if you have Zys then type=1 and the future Zycfi (forward/backward sentries) uses type 2? Of course we need to also have some kind of priority - I imagine Zycfi overrides the Zys type?

src/cheri/debug-integration.adoc
```diff
@@ -204,7 +204,20 @@ The <<m_bit>> of this capability is _only_ updated by executing <<MODESW_CAP>> o
 include::img/drootcreg.edn[]
 
 ==== Modified Trap-Return Instruction Behavior
-The `DRET` instruction reads the full YLEN bits of the <<mepc_y>>/<<sepc_y>> register and unseals it prior to exception return if it is a <<sentry_cap>>.
+
+#Begin changed since last ARC review#
+
+If the platform incorporates <<section_rvy_sentry_jalr_ext>>,
+the `DRET` instruction reads the full YLEN bits of the <<mepc_y>>/<<sepc_y>> register and unseals it prior to exception return,
```
arichardson: I think this can be independent of this new extension. Can we say: "DRET unseals all ambient sentry types"? Once we have Zycfi, we could have a table of capability types that are unsealed by JALR+xRET?

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -14,22 +14,35 @@ include::wavedrom/ct-unconditional-2.adoc[]
...
 +
-. If the target <<pcc>> is a <<sentry_cap>> and the `offset` is zero, then unseal the target <<pcc>>.
-Otherwise, increment the address of the target <<pcc>> by the sign-extended 12-bit `offset`.
+. If the platform includes the <<section_rvy_sentry_jalr_ext>> extension,
+  then impose any conditional behavior specified and,
+  if the target <<pcc>> is a sentry (that is, sealed with an acceptable type),
+  ensure that the `offset` immediate is zero and unseal the target <<pcc>>.
+. Increment the address of the target <<pcc>> by the sign-extended 12-bit `offset`.
 . Set bit zero of the target address to zero and update <<pcc>> with the new target address using the semantics of the <<SCADDR>> instruction
+. If `rd` is not `null`, add the width of this instruction
```
arichardson: Unrelated to the text but I've generally been using 1 line per sentence and relying on editor wrapping since that's recommended by asciidoctor: https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line. This ensures that editing one sentence doesn't affect multiple lines.

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -14,22 +14,35 @@ include::wavedrom/ct-unconditional-2.adoc[]
...
+  to the current <<pcc>> and store the result into `rd`.
+. If the platform includes the <<section_rvy_sentry_jalr_ext>> extension,
+  and return capabilities are to be sealed, do so as specified.
 . Jump to the target <<pcc>>
+  and, if the capability encoding entails the presence of <<section_rvy_sentry_jalr_ext>>
+  with additional architectural semantics for the type of `cs1`, apply those.
 +
-The <<pcc>> of the next instruction is calculated using the semantics of the <<SCADDR>> instruction, and written to `{cd}` as a <<sentry_cap>>.
+The <<pcc>> of the next instruction is calculated using the semantics of the <<SCADDR>> instruction.
```
arichardson: This should be part of the "store the result into `rd`" step.

src/cheri/introduction.adoc
```diff
@@ -37,6 +37,7 @@ extensions to support CHERI alongside RISC-V
 
 <<rv32y,{cheri_base_ext_name}>>:: Unprivileged specification for {cheri_base_ext_name} to support CHERI.
 <<section_rvy_sentry_insn_ext>>:: Unprivileged extension for capability ambient sealing.
+<<section_rvy_sentry_jalr_ext>>:: Unprivileged extension for sealed function entry points (for CFI).
```
arichardson: Can we have this extension be a hook in the JALR/xRET instructions instead of a separate named extension?

src/cheri/riscv-priv-integration.adoc
```diff
@@ -447,7 +447,20 @@ NOTE: A future extension may add CHERI checks to the page table walker.
 
 === Modified Trap-Return Instructions Behavior
 
-When the {cheri_base_ext_name} base ISA is implemented, the <<otherpriv,trap-return instructions>> (`MRET` and `SRET`) read the full YLEN bits of the <<mepc_y>>/<<sepc_y>> register and unseal it prior to exception return if it is a <<sentry_cap>>.
+#Begin changed since last ARC review#
+
+When the {cheri_base_ext_name} base ISA is implemented
+and the platform incorporates {rvy_sentry_jalr_ext_name},
```
arichardson: Would just drop this condition. If it would be unsealed by JALR then we unseal in XRET as well.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -274,25 +276,12 @@ for types not defined above.
 Ambient sealing type::
 Some capability types are said to be "ambiently available" (or just "ambient")
 if they do not require specific authority to seal a capability (with that type).
-Type 1 is considered ambiently available on a given platform
-if <<section_rvy_sentry_insn_ext>> is available on that platform.
+For example, if <<section_rvy_sentry_insn_ext>> is available on a given platform,
+the type with which it seals capabilities is considered ambiently available.
+(With the capabilities of <<app_cap_description>>, that would be type `1`.)
```
arichardson, line 281: I would quite like the architectural otypes to be defined centrally rather than dependent on the encoding. I.e. 1 is always ambient sentry, but does not need to be encodable by CHERIoT. That way the YTYPER result is independent of the encoding.

src/cheri/zysentry.adoc
```diff
@@ -0,0 +1,62 @@
...
+3. Conditional behavior of <<JALR_CHERI>> based on the _register selectors_
+   used in an instance of the instruction and the <<sec_cap_type>> value of the
+   input capability.
+
+NOTE: All type(s) defined for sealing return capabilities must be acceptable
+for unsealing as inputs to <<JALR_CHERI>> as well (in at least some conditions).
+
+NOTE: Unless other mechanisms to construct sealed capabilities are available,
+all type(s) defined for sealing return capabilities should be
+<<sec_cap_type_ambient,ambiently available>> for (re)construction by <<CBLD>>.
```
arichardson: I think it would be nice to have all of this in the base chapter since this is essentially all a hook for other extension. Maybe we can get by with just adding a note that the instructions to create sentry types are defined by separate extensions.

arichardson (commented):
Sorry for another round of comments. Overall I think this is great. Just wondering if we can avoid having to define a whole new extension for these JALR/xRET hooks instead of having them in the base.

src/cheri/riscv-unpriv-integration.adoc
nwf: I would prefer not to have a central registry for two reasons:

- Suppose I build a machine that has University CHERI mediated seals and a plethora of types as well as the default encoding's (ambient, jalr) semantics for otype 1.  Because I built this machine in the near term, there was no meaning for otype 2, so it was perfectly reasonable to use as a software type.  In the future, suppose otype 2 is given meaning, then some later-written software might see an otype 2 capability on my machine and presume it has these new semantics.  In order to prevent that, the software has to be aware of the set of (architectural) types on my machine anyway.

- Having a central registry suggests that _extensions_ are the right place to define new types, and this means that coordination across extensions is required to avoid conflicts.  Despite that, encodings still have to pick a mapping from their bit patterns to supported architectural types.

If (architectural) otypes and their semantics instead are functions of the encoding and the ABI, it should be pretty natural to write software using symbolic names for particular otype semantics (further allowing for easy compile-time testing of the availability of particular semantics, at that) and we don't have to worry about conflicts between extensions.  But maybe I'm worried about the wrong things?

tariqkurd-repo: Maybe I'm missing something, but if otype 2 isn't a hardware defined type then you can't have a valid capability including it, so what's the meaning behind the first point? How can you use otype 2 as a software defined type if the capability is invalid?

nwf: Sorry, wasn't clear.  Suppose otype 2 _is_ a hardware type, but doesn't have any semantics beyond "sealed" when I build the hypothetical machine in question.

tariqkurd-repo:
Can we merge into Zys please?

src/cheri/riscv-unpriv-integration.adoc
arichardson: I don't feel strongly either way so happy to just keep what you currently have and we can revisit this discussion once we add further extensions to the spec.

src/cheri/riscv-priv-integration.adoc
```suggestion
When the {cheri_base_ext_name} base ISA is implemented,
```

src/cheri/debug-integration.adoc
nwf: In the case of `Xycheriot`, the types used by `JALR` are _not_ ambient: we explicitly do not want `YBLD` to build them.  More generally, even if you believe _forward_ arc sentry types should be ambient, _backward_ arc sentry types are probably not, on the grounds that one wants to ensure that they come about only from `JALR` itself (and TCB code via `YSEAL`, perhaps).

~~I'll try to make this less chatty, especially with `Zysentry`'s hooks moving into the base ISA.~~  Sorry, it's late.  It should suffice to say simply "unseals all sentry types", omitting "ambient" from your suggestion.

src/cheri/attributes.adoc
nwf: The existing prose in Zysentry.adoc specifically is parametric in the type(s) used for forward and backward sentries so that all of these options are possible:
- Zys is omitted, no types are ambient, and a single sentry type is used for both forward and backward arcs.
- Zys is omitted, no types are ambient, and many sentry types are in flight (CHERIoT).
- Zys is present and its type coincides with the single sentry type in use (default encoding).
- Zys is present and its type coincides with the/a _forward_ arc sentry type, but separate type(s) is/are used for backward sentries.
- Zys is present and its type is distinct from any sentry type.

The nomenclature has become muddied, but it's important to distinguish between _ambient sealing_ (or "sealing as attenuation") and _sentry_ behavior, which has "entry" in the name specifically because it refers to unsealing on control transfer.  I will attempt to clarify in the specification itself, which will admittedly be easier if the hooks are in the base ISA.  (The instruction's name is an unfortunate conflation, but that was also true in ISAv9 and I am not going to bikeshed it.)

src/cheri/debug-integration.adoc
arichardson: Thank you that sounds great!

src/cheri/attributes.adoc
arichardson: Thank you this sounds great!

src/cheri/insns/jalr_32bit.adoc
arichardson: By the way, no need to make this change. this was just an observation.

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -14,22 +14,47 @@ include::wavedrom/ct-unconditional-2.adoc[]
...
+1. Architectural interpretation of zero or more particular <<sec_cap_type>> values when a capability is passed as the input to <<JALR_CHERI>>.
+   These <<sec_cap_type>> values are, by definiton, <<sentry_cap>> types.
+
+2. A subset of those <<sec_cap_type>> values for use by <<JALR_CHERI>> to seal the return capability installed in the link register.
+   A sentry return capability grants the right to return exactly to the instruction after the caller's call without relative offset.
+   Moreover, a sentry return capability does not grant the callee access to the memory pointed to by the caller's <<pcc>>.
+
+3. Conditional behavior of <<JALR_CHERI>> based on the _register selectors_ (`{cd}`, `{cs1}`) used in an instance of the instruction together with the <<sec_cap_type>> value of the capability in `{cs1}`.
+
+In all cases, when a sealed capability is used as the input to <<JALR_CHERI>>, the instruction must have a zero offset or the target <<pcc>> will become unsealed, since the `offset` is interpreted with <<SCADDR>> semantics.
```
```suggestion
In all cases, when a sealed capability is used as the input to <<JALR_CHERI>>, the instruction must have a zero offset or the target <<pcc>> will have its {ctag} set to zero, since the `offset` is interpreted with <<SCADDR>> semantics.
```

I believe this is a typo?

arichardson (approved):
Thanks, this looks great to me

src/cheri/insns/jalr_32bit.adoc
tariqkurd-repo: I changed it from `untagged` as I've purged that word from the spec

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -14,22 +14,47 @@ include::wavedrom/ct-unconditional-2.adoc[]
...
 Indirect jump to the target capability in `{cs1}` with an address offset.
 +
 Copy `{cs1}` to the target <<pcc>>
 +
-. If the target <<pcc>> is a <<sentry_cap>> and the `offset` is zero, then unseal the target <<pcc>>.
-Otherwise, increment the address of the target <<pcc>> by the sign-extended 12-bit `offset`.
+. Increment the address of the target <<pcc>> by the sign-extended 12-bit `offset`.
+. If the target <<pcc>> is a <<sentry_cap>>, then
+  - clear the {ctag} of the target <<pcc>> if the triple of its <<sec_cap_type>> value, `{cd}`, and `{cs1}` is not acceptable, as defined by the capability encoding, and
+  - and unseal the target <<pcc>>.
```
```suggestion
  - unseal the target <<pcc>>.
```

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -14,22 +14,47 @@ include::wavedrom/ct-unconditional-2.adoc[]
...
+. If the capability encoding defines <<sentry_cap>> <<sec_cap_type>> values for return pointers, seal the capability in `{cd}` as specified.
+. Jump to the target <<pcc>>.
+. Apply any capability encoding specified architectural semantics for the type of `{cs1}`.
 +
-The <<pcc>> of the next instruction is calculated using the semantics of the <<SCADDR>> instruction, and written to `{cd}` as a <<sentry_cap>>.
+The <<pcc>> of the next instruction is calculated using the semantics of the <<SCADDR>> instruction.
 +
 NOTE: Because of the unsealing and <<SCADDR>> behavior, it is not possible for the target <<pcc>> to be sealed.
 +
 NOTE: A future extension may raise an exception on the JALR instruction itself if the target <<pcc>> will raise a CHERI exception at the target.
```
```suggestion
NOTE: A future extension may raise an exception on the JALR instruction itself if the target <<pcc>> will raise a {cheri_excep_name_pc} at the target.
```

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -14,22 +14,47 @@ include::wavedrom/ct-unconditional-2.adoc[]
...
 Description::
 Indirect jump to the target capability in `{cs1}` with an address offset.
 +
 Copy `{cs1}` to the target <<pcc>>
 +
-. If the target <<pcc>> is a <<sentry_cap>> and the `offset` is zero, then unseal the target <<pcc>>.
-Otherwise, increment the address of the target <<pcc>> by the sign-extended 12-bit `offset`.
+. Increment the address of the target <<pcc>> by the sign-extended 12-bit `offset`.
+. If the target <<pcc>> is a <<sentry_cap>>, then
+  - clear the {ctag} of the target <<pcc>> if the triple of its <<sec_cap_type>> value, `{cd}`, and `{cs1}` is not acceptable, as defined by the capability encoding, and
```
tariqkurd-repo: I don't understand the "{cd} is not acceptable" bit.

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -14,22 +14,47 @@ include::wavedrom/ct-unconditional-2.adoc[]
...
+  - and unseal the target <<pcc>>.
 . Set bit zero of the target address to zero and update <<pcc>> with the new target address using the semantics of the <<SCADDR>> instruction
-. Jump to the target <<pcc>>
+. If `{cd}` is not `null`, add the width of this instruction to the current <<pcc>> and store the result into `{cd}`.
```
tariqkurd-repo: does this mean rd != x0? Or something else?

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -14,22 +14,47 @@ include::wavedrom/ct-unconditional-2.adoc[]
...
+. If the capability encoding defines <<sentry_cap>> <<sec_cap_type>> values for return pointers, seal the capability in `{cd}` as specified.
```
tariqkurd-repo: this is treating rd as a variable unfortunately - I think we need a single step which increments the PC using YADDRW semantics, seals if necessary and stores the result to rd

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -14,22 +14,47 @@ include::wavedrom/ct-unconditional-2.adoc[]
...
+. Jump to the target <<pcc>>.
+. Apply any capability encoding specified architectural semantics for the type of `{cs1}`.
```
tariqkurd-repo, line 38: I don't know what this is referring to (I know it's a forwards hook for CHERIoT) but can you at least say "nothing for the default encoding"?

This is really making me wonder if we need a different JALR page for each encoding........?

Either a generic description on this page following by specifics "for the default encoding.... and list everything the designer needs to know" or an xref to a JALR section for each specific encoding. I'm finding this page very hard to understand now.

if I give this spec to someone and ask them to type in JALR they won't know what to type.

src/cheri/insns/jalr_32bit.adoc
nwf: Yes, exactly that.

src/cheri/insns/jalr_32bit.adoc
nwf: The entire wad of text treats `the target <<pcc>>` as a variable already.  Repeated updates to an architectural register's value _has_ to be fine.

In any case, I have endeavored to group the computational steps appropriately.

src/cheri/insns/jalr_32bit.adoc
nwf: This is meant to refer to point 3 of the note immediately below:

> Conditional behavior of <<JALR_CHERI>> based on the _register selectors_ (`{cd}`, `{cs1}`) used in an instance of the instruction together with the <<sec_cap_type>> value of the capability in `{cs1}`.

src/cheri/insns/jalr_32bit.adoc
nwf: I've attempted to make more clear what the hook is meant for and that it is trivial with the default capability encoding.

arichardson (approved):

PR #769, opened 2025-09-05
Fix comment in misc/rv32y_perms.hs

nwf:
Brown paper bag fix on https://github.com/riscv/riscv-cheri/pull/742 as noted by https://github.com/chrehrhardt

### DIFF
--- a/misc/rv32y_perms.hs
+++ b/misc/rv32y_perms.hs
@@ -14,10 +14,12 @@
 --   3. For each subset of all defined permissions, the result of evaluating
 --      the rules is a defined compressed permission set.
 --
---   4. Removing permissions from a defined compressed permission set and
---      evaluating the rules selects the best result, in that all other defined
---      compressed permission sets that could have been chosen are subsets of
---      the one picked.
+--   4. Removing permissions H from a defined compressed permission set S and
+--      evaluating the rules selects a maximal result C, in that no other
+--      compressed permission D that could have been chosen (that is, such that
+--      D is a subset of S \ H) is a (strict) superset of C.  There may,
+--      however, be other subsets of S \ H, D', that are incomparable to the
+--      selected C.
 module Main where
 
 import           Control.Monad


chrehrhardt:
This fully addresses my comment and LGTM.

misc/rv32y_perms.hs
```diff
@@ -14,10 +14,12 @@
...
 --
---   4. Removing permissions from a defined compressed permission set and
---      evaluating the rules selects the best result, in that all other defined
---      compressed permission sets that could have been chosen are subsets of
---      the one picked.
+--   4. Removing permissions H from a defined compressed permission set S and
+--      evaluating the rules selects a maximal result C, in that no other
+--      compressed permission D that could have been chosen (that is, such that
+--      D is a subset of S \ H) is a (strict) superset of C.  There may,
+--      however, be other subsets of S \ H, D', that are incomparable to the
```
tariqkurd-repo, line 21: By `incomparable` do you mean mutually-exclusive / non-overlapping? Or something else, I'm not quite sure.

nwf: Incomparable is a bit of jargon for partial order relations in general; here, it means that "D is not a subset of D' nor is D' a subset of D" and is different from non-overlapping.

To give a concrete example, consider the RV32 compressed encoding (without Zylevels1) of Appendix A and removing `H = {LM}` from the set `S = {R, W, C, LM}`.  The difference `S \ H = {R, W, C}` cannot be compressed/encoded, as there's no entry for it in the table.  So now we're faced with a choice: the expressible subsets of `S \ H` are `{R, W}` (and its subsets) and `{R, C}`.  We pretty clearly don't want to pick any of `{}`, `{R}`, or `{W}`, because `{R, W}` is obviously better, but it's not _a priori_ clear which of the incomparable `{R, W}` or `{R, C}` we should pick.  Rule `RV32-base-3` arbitrarily "breaks the tie", by insisting that we remove `W` rather than `C`.

chrehrhardt: Non-comparable simply means that neither permission set is a subset of the other.
E.g. if the requested permissions are `r+w+C` which is  not representable (on 32-bit). Possible representable permission sets are `r+C` or `r+w`. These two are not comparable in the sense that neither is clearly better than the other because neither is a subset of the other. Another representable subset of `r+w+C` would e.g. be `r` but it is a bad idea to choose this because `r+w` is clearly better than just `r` (as is `r+C`). I.e. `r+w` is comparable to `r`.

PR #771, opened 2025-09-09
Update name of default encoding, and a bit of tidy-up around that

tariqkurd-repo:


src/cheri/app-cap-description.adoc
```diff
@@ -1,16 +1,21 @@
 [appendix]
-[#app_cap_description]
-== Anatomy of Capabilities in {cheri_base_ext_name}
+[#app_cap_description, reftext="Default capability encoding"]
+== Default Capability Encoding in {cheri_base_ext_name}
```
nwf, line 3: I confess to not being a fan of "default" as the adjective here, but I have nothing better to offer.

src/cheri/app-cap-description.adoc
```diff
@@ -1,16 +1,21 @@
...
+
+This chapter describes the default capability encoding for both {cheri_base64_ext_name} and {cheri_base32_ext_name}.
+It covers the following configurations:
+
+* <<uycfg>>.base=0, and
+* <<uycfg>>.features=0 (no levels support) or 1 (<<section_zylevels1>> support)
```
nwf, line 9: Those were, to be clear, strawperson proposals for encoding values.  It may be that we don't wish to offer the levels-less configurations?  (That can be a separate PR, tho'.)

src/cheri/app-cap-description.adoc
```diff
@@ -1,16 +1,21 @@
...
+
+If {cheri_default_ext_name} is supported, the specification includes encoding the <<cheri_execution_mode>> in the capability format.
```
nwf: I've usually tried to say "supported by the platform" or something, to make it clear who's doing the relevant support.  Is there standard verbiage for this kind of composition in RISC-V specs?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -552,6 +552,11 @@ The <<uycfg>> register is a read-only XLEN CSR used to identify which CHERI capa
 The capability encoding both determines the in-memory representation of a CHERI capability
 and entails a set of extensions present atop {cheri_base_ext_name}.
 
+NOTE: Alternative capability encoding
+specifications must provide key primitives, such as permissions and bounds,
+while using a different encoding that, for example,
+changes the granularity of bounds or adds new features.
```
nwf: I recognize that this is just moved text, but while here...

```suggestion
NOTE:  Alternative capability encoding specifications,
with, for example, new features and/or different bounds granularity,
are possible and would not change how CHERI integrates with the RISC-V ISA,
provided that said encodings still provide the primitives
detailed in <<sec_capability_registers>>.
```

nwf (changes requested):
Just some nits; mostly seems fine to me.

src/cheri/app-cap-description.adoc
tariqkurd-repo: it's covered by the chapter though - it has the RV32 permissions table with and without levels
I guess the point is that the one without levels might be the basis for a future extension with more levels

I also don't think it's advisable - I can add that as a note

src/cheri/app-cap-description.adoc
tariqkurd-repo: To be honest `default` seems to be the least contentious name

src/cheri/app-cap-description.adoc
tariqkurd-repo: https://riscv-specs.timhutt.co.uk/spec/riscv-isa-release-4931695-2025-07-16/riscv-unprivileged.html#_risc_v_hardware_platform_terminology

Standard terminology is to say "if extension XYZ is implemented", looking through the unpriv spec

tariqkurd-repo:
```suggestion
If {cheri_default_ext_name} is implemented, the specification includes encoding the <<cheri_execution_mode>> in the capability format.
```

src/cheri/app-cap-description.adoc
tariqkurd-repo, line 9:
```suggestion
* <<uycfg>>.features=0 (no levels support) or 1 (<<section_zylevels1>> support)

NOTE: No levels support is specified as an option to allow flexibility for future extensions to reuse the architectural permissions allocated to <<section_zylevels1>> in {cheri_base32_ext_name}. It is not recommended to exclude <<section_zylevels1>> until a better alternative exists.

If no capability encoding format extension is specified then the selected format is always <<uycfg>>.base=0, <<uycfg>>.features=1. The explicit name for this is `Zydefaultcap`.

```

nwf: Seems fine.

arichardson (approved):

PR #773, opened 2025-09-09
Dont use caps as variables rep check

tariqkurd-repo:
Found more text still referring to caps as variables


src/cheri/app-cap-description.adoc
```diff
@@ -683,31 +683,30 @@ architecture: normal bounds check rules should be followed.
...
 
-In other words, given a capability with address _a_ and the
-new address `a' = a + x`, the bounds _b_ and _t_ are decoded using _a_ and the
-new bounds _b'_ and _t'_ are decoded using _a'_. The new address is within the
-capability's _representable range_ if `b == b' && t == t'`.
+This is the _representable range_ check that checks the new address is valid for use with the source capability metadata.
 
-Changing a capability's address to a value outside the _representable range_
-unconditionally clears the capability's tag. Examples are:
+This check is the fundamental building block for combining addresses and capability metadata, and is used by instructions such as <<SCADDR>> and <<CADD>>/<<CADDI>>.
```
nwf: This check is fundamental for _compression_.  The historic, uncompressed CHERI-MIPS 256-bit capability encoding _also_ combined addresses and metadata, just not quite as intricately as the CHERI Concentrate and derivative schemes.

src/cheri/app-cap-description.adoc
```diff
@@ -683,31 +683,30 @@ architecture: normal bounds check rules should be followed.
...
-The new address, after updating the address of a capability, is within the
-_representable range_ if decompressing the capability's bounds with the
-original and new addresses yields the same base and top bounds.
+When a valid source capability is read, a new address is applied, and then the resulting capability is written to the destination register, a check is applied to the new address to ensure that the decoding of the bounds is the same for the source and destination capabilities.
 
-In other words, given a capability with address _a_ and the
-new address `a' = a + x`, the bounds _b_ and _t_ are decoded using _a_ and the
-new bounds _b'_ and _t'_ are decoded using _a'_. The new address is within the
-capability's _representable range_ if `b == b' && t == t'`.
+This is the _representable range_ check that checks the new address is valid for use with the source capability metadata.
```
nwf: It's probably worth-while to keep "representable range" defined as a concept separate from the check that an address is within that range.  I think this rewrite loses the distinction.

src/cheri/app-cap-description.adoc
```diff
@@ -683,31 +683,30 @@ architecture: normal bounds check rules should be followed.
...
-Changing a capability's address to a value outside the _representable range_
-unconditionally clears the capability's tag. Examples are:
+This check is the fundamental building block for combining addresses and capability metadata, and is used by instructions such as <<SCADDR>> and <<CADD>>/<<CADDI>>.
 
-* Instructions such as <<CADD>> which include pointer arithmetic.
-* The <<SCADDR>> instruction which updates the capability address field.
+Given:
+
+* A source capability with address _a_, metadata _m_ that decodes to give the bounds _b_ and _t_.
+* An arbitrary address _a'_ with the same metadata _m_ decodes to give the bounds _b'_ and _t'_.
```
```suggestion
* An arbitrary address _a'_ with the same metadata _m_ that decodes to give the bounds _b'_ and _t'_.
```

src/cheri/app-cap-description.adoc
```diff
@@ -683,31 +683,30 @@ architecture: normal bounds check rules should be followed.
 
 ==== Concept
 
-The new address, after updating the address of a capability, is within the
-_representable range_ if decompressing the capability's bounds with the
-original and new addresses yields the same base and top bounds.
+When a valid source capability is read, a new address is applied, and then the resulting capability is written to the destination register, a check is applied to the new address to ensure that the decoding of the bounds is the same for the source and destination capabilities.
```
```suggestion
Whenever a capability with a new different address is derived from a valid source capability, a check is applied to the new address to ensure that the decoding of the bounds is the same for the source and destination capabilities.
```

arichardson (approved):

src/cheri/app-cap-description.adoc
arichardson: That's a good point, we should keep those separate.

tariqkurd-repo:
there were mulitple descriptions of the representable range, and multiple citations to CHERI concentrate
I've tried to rationalise them so the rep range is described in the intro, and the specifics in the default encoding just talk about the actual check

PR #778, opened 2025-09-10
pointer masking and CHERI clarification

tariqkurd-repo:
The spec doesn't specify the effect of RVY on pointer masking.
Is that what people are thinking?

### DIFF
--- a/src/zpm.adoc
+++ b/src/zpm.adoc
@@ -235,6 +235,17 @@ Each extension introduces a 2-bit WARL field (`PMM`) that may take on the follow
 
 All of these fields are read-only 0 on RV32 systems.
 
+==== PMM for {cheri_base64_ext_name}
+
+For {cheri_base64_ext_name}, PMM is read-only 0.
+
+If {cheri_default_ext_name} is implemented:
+
+* PMM can be written to non-zero values in {cheri_int_mode_name} only.
+* Switching into {cheri_cap_mode_name} causes PMM to always read as 0, and therefore pointer masking is never enabled in that mode.
+
+Therefore non-zero state programmed in {cheri_int_mode_name} is retained in {cheri_cap_mode_name}, but is only visible or takes effect in {cheri_int_mode_name}.
+
 ==== Ssnpm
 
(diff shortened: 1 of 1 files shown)


src/zpm.adoc
```diff
@@ -235,6 +235,17 @@ Each extension introduces a 2-bit WARL field (`PMM`) that may take on the follow
...
 All of these fields are read-only 0 on RV32 systems.
 
+==== PMM for {cheri_base64_ext_name}
+
+For {cheri_base64_ext_name}, PMM is read-only 0.
+
+If {cheri_default_ext_name} is implemented:
+
+* PMM can be written to non-zero values in {cheri_int_mode_name} only.
+* Switching into {cheri_cap_mode_name} causes PMM to always read as 0, and therefore pointer masking is never enabled in that mode.
```
tariqkurd-repo:
```suggestion
* PMM can be written to non-zero values in {cheri_int_mode_name} only.
* All writes to PMM in {cheri_cap_mode_name} set the field to 0.
* Switching into {cheri_cap_mode_name} causes PMM to always read as 0, and therefore pointer masking is never enabled in that mode.
```

src/zpm.adoc
```diff
@@ -235,6 +235,18 @@ Each extension introduces a 2-bit WARL field (`PMM`) that may take on the follow
...
+For {cheri_base64_ext_name}, PMM is read-only 0.
+
+If {cheri_default_ext_name} is implemented:
+
+* PMM can be written to non-zero values in {cheri_int_mode_name} only.
+* All writes to PMM in {cheri_cap_mode_name} set the field to 0.
+* Switching into {cheri_cap_mode_name} causes PMM to always read as 0, and therefore pointer masking is never enabled in that mode.
+
+Therefore non-zero state programmed in {cheri_int_mode_name} is retained in {cheri_cap_mode_name}, but is only visible or takes effect in {cheri_int_mode_name}.
+
```
tariqkurd-repo, line 249:
```suggestion

NOTE: How to enable pointer masking in {cheri_cap_mode_name} is a research topic for a future extension.
```

PR #782, opened 2025-09-12
Tidy up and clarify pointer masking spec

tariqkurd-repo:
I've integrated pointer masking into the correct chapter
I've started using _transformed_effective_address_ from the pointer masking spec (set PMLEN upper bits to zero)
I've tried to summarise the actual effect of pointer masking on the existing RVY ISA
I've highlighted the current comment saying that further evaluation is required.
I've removed all references to hybrid mode, as I don't believe it needs an explicit mention - specifying the rules is enough.
I've moved from MXLEN to XLEN for consistency as that's what used for the transformed effective address definition.

And I don't understand how changing PMLEN dynamically, which changes bounds decoding, isn't a massive security hole.
I think this needs some text explaining why it's ok.

### DIFF
--- a/src/cheri/ptrmask-integration.adoc
+++ b/src/cheri/ptrmask-integration.adoc
@@ -31,7 +31,7 @@ After any CHERI operations, the final access address is still subject to as much
 
 In summary, for data accesses only:
 
-* When encoding bounds ({SCBNDS}/{SCBNDSI}/{SCBNDSR}), bits [XLEN-1:XLEN-PMLEN] of the base address are set to zero.
+* When setting bounds ({SCBNDS}/{SCBNDSI}/{SCBNDSR}), bits [XLEN-1:XLEN-PMLEN] of the address are set to zero and therefore the resulting capability will have a base with the PMLEN high address bits set to zero.
 * When decoding bounds, the address used for decoding has bits [XLEN-1:XLEN-PMLEN] set to zero.
 * When checking the <<section_cap_representable_check, representable range>> for {CADDI}/{CADD}/{SCADDR}, the old address and new addresses both have bits [XLEN-1:XLEN-PMLEN] set to zero *before* the check.
 


src/cheri/ptrmask-integration.adoc
```diff
@@ -1,32 +1,49 @@
...
+After any CHERI operations, the final access address is still subject to as much sign extension as the pointer masking extensions mandate.
+
+The summary below makes use of the _transformed_effective_address_ defined above where the upper PMLEN bits of the effective address are set to 0.
+
+In summary:
+
+* When encoding bounds (<<SCBNDS>>/<<SCBNDSI>>/<<SCBNDSR>>), the base address is set to the _transformed_effective_address_.
+** `base = {PMLEN{0}},rs1.address[XLEN-PMLEN-1:0]`.
+* When decoding bounds (<<GCBASE>>), `{cd}` is set to the _transformed_effective_address_.
+** `{cd} = {PMLEN{0}},decoded_base[XLEN-PMLEN-1:0]`.
```
arichardson: I'm not sure this is always true, we just mask off the high address bits when computing the base, I think with a sufficiently large exponent the computed base could be larger than the masked address?

arichardson (commented):
I believe this is not quite true, we only transform the address, not the base.

src/cheri/ptrmask-integration.adoc
```diff
@@ -1,32 +1,41 @@
+ifdef::cheri_standalone_spec[]
 [#section_pointer_masking_integration]
-== Integrating {cheri_base_ext_name} and {cheri_default_ext_name} with Pointer Masking
+== Integrating {cheri_base64_ext_name} with Pointer Masking
 
-ifdef::cheri_standalone_spec[]
-WARNING: This chapter should appear as a section in the pointermasking chapter. Exact location TBD.
+WARNING: This chapter should appear as a section in the pointer masking chapter. Exact location TBD.
+The pointer masking extensions Smmpm, Smnpm, SSnpm, Sspm and Supm are compatible with {cheri_base64_ext_name}.
```
tariqkurd-repo:
```suggestion
```

I've located it in zpm.adoc

src/cheri/ptrmask-integration.adoc
```diff
@@ -1,32 +1,41 @@
...
 
-Any address that is checked against a capability (whether via loads/stores in {cheri_int_mode_name} or {cheri_cap_mode_name}),
+Any address that is checked against a capability
 is also first subject to the same masking as bounds decode (masking without extension).
-After any CHERI operations, the final access address is still subject to as much sign extension as the masking extensions mandate.
+After any CHERI operations, the final access address is still subject to as much sign extension as the pointer masking extensions mandate.
+
+In summary, for data accesses only:
+
+* When encoding bounds ({SCBNDS}/{SCBNDSI}/{SCBNDSR}), bits [XLEN-1:XLEN-PMLEN] of the base address are set to zero.
```
arichardson:
```suggestion
* When setting bounds ({SCBNDS}/{SCBNDSI}/{SCBNDSR}), bits [XLEN-1:XLEN-PMLEN] of the address are set to zero and therefore the resulting capability will have a base with the PMLEN high address bits set to zero.
```

src/cheri/ptrmask-integration.adoc
```diff
@@ -1,32 +1,41 @@
...
+* When decoding bounds, the address used for decoding has bits [XLEN-1:XLEN-PMLEN] set to zero.
+* When checking the <<section_cap_representable_check, representable range>> for {CADDI}/{CADD}/{SCADDR}, the old address and new addresses both have bits [XLEN-1:XLEN-PMLEN] set to zero *before* the check.
+
+Also note that:
+
+* <<pcc>> bounds decoding isn't affected by pointer masking.
+* Any <<section_cap_representable_check, representable range check>> triggered by a write to a CSR (e.g. <<mtvec_y>>) are not affected by pointer masking.
```
arichardson, line 41: Hm this is just because it only affects data pointers and only affects user mode? What about jvtc?

tariqkurd-repo: _No pointer masking operations are applied when software reads/writes to CSRs, including those meant to hold addresses._

from the pointer masking spec, so following that
(also JVTC is RV32 only and pointer masking in RV64 only)

PR #783, opened 2025-09-12
Codify the AUIPC shift being a function of encoding

nwf:
We/I'd never actually (re)written the prose for this, oops.  It was there at points past, but it seems to have gotten lost?  Here's an attempt at background, rationale, and the actual codification of AUIPC's parametric behavior.

### DIFF
--- a/src/cheri/insns/auipc_32bit.adoc
+++ b/src/cheri/insns/auipc_32bit.adoc
@@ -14,9 +14,15 @@ include::wavedrom/rv64_lui-auipc.adoc[]
 
 include::base_isa_extension.adoc[]
 
+#Begin changed since last ARC review#
+
 Description::
-Form a 32-bit offset from the 20-bit immediate filling the lowest 12 bits with zeros.
+Form a 32-bit offset from the 20-bit immediate filling the lowest bits with zero;
+the number of places to shift is determined by the capability encoding's choice of
+the <<section_auipc_shift>> value (12, unless otherwise specified).
 Take the value of the AUIPC instruction's <<pcc>>, increment its address by the 32-bit offset using the semantics of the <<SCADDR>> instruction and write the result to `{cd}`.
+
+#End changed since last ARC review#
 +
 include::rep_range_check.adoc[]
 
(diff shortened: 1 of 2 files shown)


src/cheri/insns/auipc_32bit.adoc
```diff
@@ -15,7 +15,9 @@ include::wavedrom/rv64_lui-auipc.adoc[]
 include::base_isa_extension.adoc[]
 
 Description::
-Form a 32-bit offset from the 20-bit immediate filling the lowest 12 bits with zeros.
+Form a 32-bit offset from the 20-bit immediate filling the lowest bits with zero;
+the number of bits to fill is determined by the capability encoding's choice of
```
tariqkurd-repo:
```suggestion
the number of places to shift is determined by the capability encoding's choice of
```

src/cheri/insns/auipc_32bit.adoc
```diff
@@ -15,7 +15,9 @@ include::wavedrom/rv64_lui-auipc.adoc[]
...
+the <<section_auipc_shift>> value.
```
tariqkurd-repo: can we please have a link to the default encoding value as well as this?
so please don't have to search for the information.
Both links are essential.
One huge bugbear of mine about the risc-v spec is how hard it is to find the information.
and it's only going to get worse with the two manuals being merged and continuing to grow........

tariqkurd-repo: TBH this is such a simple thing you could just write "the shift value is 12 unless _otherwise_ specified by the encoding" and then add a link to the CHERIoT section later.

nwf: I added some words along those lines here and over in the rationale prose.

Vexingly, I cannot find a way in AsciiDoctor to have a "What Links Here" / list of in-edges in the cross-reference graph.  That'd be ideal for extensibly maintaining these kinds of hook instantiations.

tariqkurd-repo: it's quite limiting is adoc..... thanks for the updates

arichardson (commented):
I believe we had a brief note along the lines of "future CHERI extensions can redefine the AUIPC shift" that was somehow lost during the major refactoring. Sorry!

I think this text is great, my only concern is that it adds quite a lot of text to the main chapter that may confuse some readers. Maybe the AUOPC shift section should move to the encoding chapter (somewhere near the description of bounds) and we restore a short 1-line note in AUIPC instruction description? What do you think?

arichardson (approved):

PR #788, opened 2025-09-16
Hypervisor integration and other privileged spec improvements

tariqkurd-repo:
The main objective was to remove the hypervisor chapter from the priv spec, by integrating it properly.
As ever lots more things were left to discover.

### DIFF
--- a/src/cheri/hypervisor-integration.adoc
+++ b/src/cheri/hypervisor-integration.adoc
@@ -1,13 +1,12 @@
+//This file is only included in the standalone spec - not the priv spec
 [#section_hypervisor_integration]
 == Integrating {cheri_base_ext_name} and {cheri_default_ext_name} with the "H" Extension
 
-ifdef::cheri_standalone_spec[]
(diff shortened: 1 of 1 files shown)


src/cheri/hypervisor-integration.adoc
```diff
@@ -189,5 +188,7 @@ load and store capability instructions behave as described in
 xref:section_existing_riscv_insns[xrefstyle=short]. In
 {cheri_int_mode_name}, the instructions use the low XLEN bits of the base register as the effective address for the memory access and the capability authorizing the memory access is <<ddc>>.
 
+ifdef::cheri_standalone_spec[]
```
arichardson: Isn't this always true now that you are no longer including this file?

arichardson (approved):

src/cheri/hypervisor-integration.adoc
tariqkurd-repo: yes - tags removed

Issue #64, opened 2024-01-26
Increase type space by splitting on X permission


LawrenceEsswood:
This idea lifted from CheriOT.

Types for capabilities tend to only make sense for either executable capabilities or non-executable capabilities. Having the X bit be an implicit low bit of a capabilities type (when sealed) doubles the type space.

This is very useful if we then want to put more information in type space rather than permission space.

sorear:
This can't be done as written since we don't have object types in this specification.

I filed #70 which should be equivalent to this once object types are added.

LawrenceEsswood:
Although there may be no way of constructing them now, this is still relevant to decide which type the single sealed (and executable) is for the purpose of cgettype.

If the future we want the low bit of a type to be inferred from x, the type should be chosen sensibly to fit this requirement.

tariqkurd-repo:
Hi @LawrenceEsswood - is this still relevant now that https://github.com/riscv/riscv-cheri/pull/155 has been merged?

arichardson:
This is still relevant but I think more as part of a "Zcheri_types" extension that adds more kinds of objects.

nwf:
> Can we close this one now that we have multiple roots in the specification?

This is more a question of _encoding_ than of _multiple roots_, I think.  Just to say, each of these are completely possible:
- a fully-orthogonal permissions encoding to nevertheless have multiple distinct roots (consider: W^X enforcement inductive from primordial conditions) and
- a non-orthogonal scheme with a single Infinite cap used as the one root everywhere.

Indeed, the present Appendix A RV32 AP field encodings -- with and without Zylevels1 -- are examples of the latter.

But, that said: those encodings use a 2-bit "quadrant" selector with one quadrant value biconditionally implying X permission, so this kind of a split has been in place for a while.

PR #795, opened 2025-09-20
Explicitly permit CT/permission entanglement within encodings

nwf:
And permit Zys's SENTRY to use a different type as a function of permissions.
While here, remove a stale (after #768) note in Zys prose.


src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -272,6 +272,16 @@ to operate at the non-zero type;
...
 
+Capability encodings may also make the set of <<sec_cap_type>> values
+that may be used to seal a particular capability
+depend on the the permissions granted by that capability.
+For example, it can be a useful space optimization to differentiate
+the <<sec_cap_type>> values
+with which capabilities granting <<x_perm>> may be sealed from those
+with which capabilities not granting <<x_perm>> may be sealed.
+(The <<app_cap_description>> does not avail itself of this option,
+but, for example, the CHERIoT capability encoding does.)
```
arichardson, line 284: I am finding it a bit difficult to parse this sentence.

I believe the intent 1) is that the architectural capability type can be different depending on combination of granted permissions and the raw capability type field and 2) when you seal a capability the output can depend on the permissions of the input: E.g. YSENTRY on a cap with x permission gives a "code sentry" and without x permission you get a "sealed data cap"?

I guess we need a "Fixes: https://github.com/riscv/riscv-cheri/issues/64" in the commit message?

arichardson, line 284:
```suggestion
Capability encodings may also make the set of resulting <<sec_cap_type>> values
depend on the the permissions granted by the input capability.
For example, it can be a useful encoding space optimization to differentiate
the <<sec_cap_type>> values for capabilities granting <<x_perm>> from those not granting <<x_perm>> -- the <<x_perm>> in the capability encoding effectively adds an additional bit to the <<sec_cap_type>> field.
NOTE: The <<app_cap_description>> does not make use of this optimization,
but, for example, the CHERIoT capability encoding does.
```

Maybe something like this? I hope I didn't change the meaning?

nwf: That sounds right to me.  It's getting late and my ability to write English is certainly increasingly suspect. ^^;

arichardson (approved):

PR #796, opened 2025-09-20
Fix asciidoctor syntax error

nwf:


src/cheri/insns/load_exceptions.adoc
```diff
@@ -9,7 +9,6 @@ All misaligned load reservations cause a load address misaligned exception to al
 +
 endif::[]
 endif::[]
-+
```
arichardson, line 12: Oh good catch, missed this one.

It looks like the 
```
:!load_res:
:!has_cap_data:
```
below is also rendered as part of the doc, can you remove that + in the line before?

nwf: It should no longer be rendered with this `+` removed.

arichardson: perfect thank you

PR #797, opened 2025-09-20
Allow using the isa-manual makefile to build riscv-cheri.pdf

arichardson:
This avoids a lot of duplication and simplifies the CI files. The only difference is that riscv-privileged and riscv-unprivileged now have a CHERI spec version, but IMO that actually makes more sense.


arichardson:
I kept Makefile.cheri for backwards compatibility but it shouldn't be needed anymore after this.

One advantage is that we can now do `make build-pdf -j3` to build all documents at once instead of needed two makefiles. Standalone spec can still be built with `make build/riscv-cheri.pdf`.

Should I add a `make cheri-pdf` convenience target?

PR #803, opened 2025-09-23
update vector note

tariqkurd-repo:
Use the standard heading for vector: `Vector (RVY)`
Update the note about indexed loads with zero base and XLEN offsets not being useful.

### DIFF
--- a/src/cheri/vector-integration.adoc
+++ b/src/cheri/vector-integration.adoc
@@ -15,7 +15,7 @@ Under {cheri_base_ext_name}, vector loads and stores follow the standard rules f
 * If there are no _active_ elements then no CHERI exceptions will be raised.
 * CHERI exceptions are only raised on fault-only-first loads if element 0 is both _active_ and fails any exception checks.
 
-Additionally the standard rule follows that all load and stores where the base register is `x0` are reserved.
+Additionally the standard {cheri_base_ext_name} rule that all load and stores where the base register is `x0` are reserved applies to all vector memory access instructions.
 
 NOTE: The approach of using indexed loads with the base register set to the value zero and XLEN-wide offsets do not work well with CHERI as the authorizing capability must cover all of memory.
-      If the authorizing capability is actually specified as `x0` then this is reserved behavior.
+      If the authorizing capability is specified as `x0` then this instruction encoding is RESERVED.


src/cheri/vector-integration.adoc
```diff
@@ -20,4 +20,5 @@ Under {cheri_base_ext_name}, vector loads and stores follow the standard rules f
 * CHERI exceptions are only raised on fault-only-first loads if element 0 is both _active_ and fails any exception checks.
 
 NOTE: Indexed loads check the bounds of every access against the authorizing capability in `{cs1}`.
- Therefore the approach of having a zero base register and treating every element as an absolute address may not work well in this mode.
+ Therefore the approach of having base register set to zero and XLEN offsets which therefore are treated as absolute addresses implies that the capability must cover all of memory.
+ #Therefore this mode is not useful and so needs more thought#
```
arichardson: well the zero register as the base is an always trapping encoding so it should be reserved in the same way as all other loads/stores with RVY?

tariqkurd-repo: that is a possible solution to the problem - yes - and if the base is a register which just happens to be zero then you simply need to have (probably) an infinite data cap

src/cheri/vector-integration.adoc
```diff
@@ -19,5 +15,7 @@ Under {cheri_base_ext_name}, vector loads and stores follow the standard rules f
 * If there are no _active_ elements then no CHERI exceptions will be raised.
 * CHERI exceptions are only raised on fault-only-first loads if element 0 is both _active_ and fails any exception checks.
 
-NOTE: Indexed loads check the bounds of every access against the authorizing capability in `{cs1}`.
- Therefore the approach of having a zero base register and treating every element as an absolute address may not work well in this mode.
+Additionally the standard rule follows that all load and stores where the base register is `x0` are reserved.
+
+NOTE: The approach of using indexed loads with the base register set to the value zero and XLEN-wide offsets do not work well with CHERI as the authorizing capability must cover all of memory.
+      If the authorizing capability is actually specified as `x0` then this is reserved behavior.
```
arichardson:
```suggestion
      If the authorizing capability is specified as `x0` then this instruction encoding is RESERVED.
```
Maybe like this?

arichardson: Oh we already have that above. Maybe we can just drop this note?

everything is covered by `Additionally the standard rule follows that all load and stores where the base register is `x0` are reserved.`

src/cheri/vector-integration.adoc
```diff
@@ -19,5 +15,7 @@ Under {cheri_base_ext_name}, vector loads and stores follow the standard rules f
 * If there are no _active_ elements then no CHERI exceptions will be raised.
 * CHERI exceptions are only raised on fault-only-first loads if element 0 is both _active_ and fails any exception checks.
 
-NOTE: Indexed loads check the bounds of every access against the authorizing capability in `{cs1}`.
- Therefore the approach of having a zero base register and treating every element as an absolute address may not work well in this mode.
+Additionally the standard rule follows that all load and stores where the base register is `x0` are reserved.
```
arichardson:
```suggestion
Additionally the standard {cheri_base_ext_name} rule that all load and stores where the base register is `x0` are reserved applies to all vector memory access instructions.
```

src/cheri/vector-integration.adoc
arichardson: But the base authorizes the load/store, so it's always trapping? And in integer mode it's DDC so it just works fine.

tariqkurd-repo: yeah - I think this is actually ok - fortunately

PR #811, opened 2025-09-25
xycheriot: drop YBNDSWBigI

nwf:
Now that YBNDSWI in the base spec has its 8+2 immediate encoding form, there's no need for us to separately propose a large-immediate set-bounds instruction.


tariqkurd-repo:
it might be an idea to ifdef it out - in case ARC don't grant us the big immediate - and it needs to come back

tariqkurd-repo (approved):

PR #812, opened 2025-09-26
Clarify that not all CHERI instructions are available in Address Mode

Flowdalic:
The current "Hart's behavior" talbe incorrectly suggests parity between (Non-CHERI) Address Mode and (CHERI) Capability Mode regarding CHERI instruction usage. Fix this by adding a footnote to the "CHERI Instructions Allowed" column for Address Mode to note that load/store instructions revert to their original, non-CHERI encoding, preventing the use of CHERI-specific load/store operations in this mode.

### DIFF
--- a/src/cheri/riscv-priv-integration.adoc
+++ b/src/cheri/riscv-priv-integration.adoc
@@ -522,13 +522,15 @@ in connection with the effective CHERI enable and the <<cheri_execution_mode>> w
 | Authorizing capability for data memory accesses | <<ddc>>| <<ddc>> | capability in `rs1`
 | natively YLEN CSR Access Width       | ✘                 | YLEN                  | YLEN
 | Extended YLEN CSR Access Width       | XLEN              | XLEN                  | YLEN
-| CHERI Instructions Allowed           | ✘                 | ✔                     | ✔
+| CHERI Instructions Allowed           | ✘                 | ✔^3^                  | ✔
 | Summary                              | **_Fully RISC-V compatible_**^2^ | **{cheri_int_mode_name}** | **{cheri_cap_mode_name}**
 |==============================================================================
 
 ^1^ _{CRE}_ represents the effective CHERI enable for the current privilege mode.
 
 ^2^ The hart is fully compatible with standard RISC-V when {CRE}=0 provided that <<pcc>>, Xtvec, Xepc and <<ddc>> have not been changed from the default reset state (i.e., hold <<root-rx-cap>> and <<root-rw-cap>> capabilities).
+
+^3^ Capability load/store instructions are unavailable as their encoding will revert to non-CHERI (see <<section_cheri_hybrid_ext>>).
 
 ifdef::support_varxlen[]
 [#section_cheri_dyn_xlen]


src/cheri/riscv-priv-integration.adoc
```diff
@@ -522,14 +522,16 @@ in connection with the effective CHERI enable and the <<cheri_execution_mode>> w
...
-| CHERI Instructions Allowed           | ✘                 | ✔                     | ✔
+| CHERI Instructions Allowed           | ✘                 | ✔^3^                  | ✔
 | Summary                              | **_Fully RISC-V compatible_**^2^ | **{cheri_int_mode_name}** | **{cheri_cap_mode_name}**
 |==============================================================================
 
 ^1^ _{CRE}_ represents the effective CHERI enable for the current privilege mode.
 
 ^2^ The hart is fully compatible with standard RISC-V when {CRE}=0 provided that <<pcc>>, Xtvec, Xepc and <<ddc>> have not been changed from the default reset state (i.e., hold <<root-rx-cap>> and <<root-rw-cap>> capabilities).
 
+^3^ Capability load/store instructions are unavailable as their encoding will revert to non-CHERI (see <<section_cheri_hybrid_ext>>).
```
```suggestion
^3^ Capability-authorized load/store instructions are unavailable as their encoding will revert to non-CHERI (see <<section_cheri_hybrid_ext>>).
```

Maybe this wording?

Flowdalic: > The capability load stores using an integer base will still be available, this only affects the remapped instructions.

You are right. Looking a tables 48-51 it appears the remapped instructions are `C.SY`, `C.LY`, `C.SYSP`, and `C.LYSP` i.e. the *compressed* instructions for cap load/store. The uncompressed ones (`SY`, `LY`) are still available.

So how about

```suggestion
^3^ The compressed instructions to load and store capability data are unavailable as their encoding will revert to non-CHERI (see <<section_cheri_hybrid_ext>>).
```

The added footenote's text was inspired by the note in 4.1 "If RVC encodings are supported, load/store encodings will revert back to their non-CHERI encodings,". If I am correct, this note could also benefit from clarification that only the compressed instructions are affected.

On an side node, the sentence "If RVC encodings are supported, load/store encodings will revert back…" could be interpreted as "Iff RVC encodings are supported, load/store encodings will revert back…". Maybe this should be further reworded to clarify that the encodnings will always revert to their non-CHERI encodings (which I believe is the intention?).

arichardson: The new wording sounds good to me.

src/cheri/riscv-priv-integration.adoc
```diff
@@ -522,14 +522,16 @@ in connection with the effective CHERI enable and the <<cheri_execution_mode>> w
...
-| CHERI Instructions Allowed           | ✘                 | ✔                     | ✔
+| CHERI Instructions Allowed           | ✘                 | ✔^3^                  | ✔
 | Summary                              | **_Fully RISC-V compatible_**^2^ | **{cheri_int_mode_name}** | **{cheri_cap_mode_name}**
 |==============================================================================
 
 ^1^ _{CRE}_ represents the effective CHERI enable for the current privilege mode.
 
 ^2^ The hart is fully compatible with standard RISC-V when {CRE}=0 provided that <<pcc>>, Xtvec, Xepc and <<ddc>> have not been changed from the default reset state (i.e., hold <<root-rx-cap>> and <<root-rw-cap>> capabilities).
 
+^3^ The compressed instructions to load and store capability data are unavailable as their encoding will revert to non-CHERI (see <<section_cheri_hybrid_ext>>).
```
```suggestion
^3^ The compressed instructions operating on capability data are unavailable as their encoding will revert to non-CHERI (see <<section_cheri_hybrid_ext>>).
```

PR #817, opened 2025-09-30
Add isa deltas since 0.9.5

tariqkurd-repo:
I'm documenting what has changed since 0.9.5 to make it easy to explain to people.
It doesn't need to be merged, but I guess it'll be useful for others too.

### DIFF
--- a/src/cheri/app-cheri-instructions.adoc
+++ b/src/cheri/app-cheri-instructions.adoc
@@ -118,9 +118,17 @@ WARNING: {YSEAL} and {YUNSEAL} are not included in the v1.0 ratification package
 . 0.9.5 had SH[123]ADD, and the .UW forms, replaced by capability versions.
 .. This is no longer the case, so now the capability versions have new encodings.
 . There is no longer an SH4ADD instruction (i.e. the integer version).
-. Capability encodings are now in a separate extension - the default being <<app_cap_description,{rvy_default_encoding_name}>>.
-. Capability encodings are now the naming authorities for capability types (only 0/unsealed exists in the base architecture).
 . The <<{SENTRY}>> instruction is now in a separate extension {rvy_sentry_insn_ext_name}.
+
+====
+
+[NOTE]
+====
+
+The following changes are for forwards compatibility with {cheriot_ext_name}:
+
+. Capability encodings are now the naming authorities for capability types (only 0/unsealed exists in the base architecture).
+. Capability encodings are now in a separate extension - the default being <<app_cap_description,{rvy_default_encoding_name}>>.
 . <<JALR_CHERI>> has been given explicit hooks for sentry handling (especially for future forward/backward arc distinction).
 
 ====


arichardson (approved):

src/cheri/cheri_isa_tables.adoc
```diff
@@ -233,3 +233,35 @@ An {cheri_base_ext_name} core which supports {cheri_default_ext_name} adds the i
...
+include::generated/new_instructions_table_body.adoc[]
+|==============================================================================
+
+NOTE: {SCHI_BASE} and {GCHI_BASE} are actual instructions. {SCHI} and {GCHI} are pseudoinstructions.
+
+NOTE: ACPERM was replaced by {CLRPERM}. The difference being that the mask is used to clear, not retain, permission bits.
+
+NOTE: 0.9.5 had SH[123]ADD, and the .UW forms, replaced by capability versions. This is no longer the case, so now the capability versions have new encodings.
+
+NOTE: There is no longer an SH4ADD instruction (i.e. the integer version).
```
arichardson: Should we include that exception reporting was changed (even though it's priv-only?

tariqkurd-repo: Yes, I think so. Maybe in a different PR?

src/cheri/cheri_isa_tables.adoc
```diff
@@ -233,3 +233,35 @@ An {cheri_base_ext_name} core which supports {cheri_default_ext_name} adds the i
...
+.Instructions added since 0.9.5
+[#ISA_added_095]
+[width="100%",options=header,cols="2,2,2",]
+|==============================================================================
+include::generated/new_instructions_table_body.adoc[]
+|==============================================================================
+
+NOTE: {SCHI_BASE} and {GCHI_BASE} are actual instructions. {SCHI} and {GCHI} are pseudoinstructions.
+
+NOTE: ACPERM was replaced by {CLRPERM}. The difference being that the mask is used to clear, not retain, permission bits.
```
nwf: Maybe worth summarizing the rationale (the set of easily materialized constants is more useful this way around)?

src/cheri/app-cheri-instructions.adoc
```diff
@@ -84,3 +84,43 @@ include::insns/zcmt_cmjalt.adoc[]
...
+. {SCHI_BASE} and {GCHI_BASE} are actual instructions. {SCHI} and {GCHI} are pseudoinstructions.
+. ACPERM was replaced by {CLRPERM}. The difference being that the mask is used to clear, not retain, permission bits.
+.. Clearing bits makes it much simpler to form the necessary constant compared to retaining bits, and so gives better code-size.
+. 0.9.5 had SH[123]ADD, and the .UW forms, replaced by capability versions.
+.. This is no longer the case, so now the capability versions have new encodings.
+. There is no longer an SH4ADD instruction (i.e. the integer version).
+. Capability encodings are now in a separate extension - the default being <<app_cap_description,{rvy_default_encoding_name}>>.
+. Capability encodings are now the naming authorities for capability types (only 0/unsealed exists in the base architecture).
+. The <<{SENTRY}>> instruction is now in a separate extension {rvy_sentry_insn_ext_name}.
+. <<JALR_CHERI>> has been given explicit hooks for sentry handling (especially for future forward/backward arc distinction).
```
arichardson, line 132: This isn't really a semantic ISA change since the default behaviour is still the same so not sure we need to call it out?

src/cheri/app-cheri-instructions.adoc
```diff
@@ -84,3 +84,43 @@ include::insns/zcmt_cmjalt.adoc[]
...
+include::generated/old_and_new_mnemonics_table_body.adoc[]
+|==============================================================================
+
+Some instructions have been added as shown in <<ISA_added_095>>.
+
+.Instructions added since 0.9.5
+[#ISA_added_095]
+[width="100%",options=header,cols="2,2,2",]
+|==============================================================================
+include::generated/new_instructions_table_body.adoc[]
```
arichardson, line 107: Does this now include YSEAL/YUNSEAL? they aren't part of the ratified package so should probably omit it. Also this table would probably be better after the text? Otherwise it seems like there is more change than there actually was?

src/cheri/app-cheri-instructions.adoc
tariqkurd-repo: these are just useful notes on things which have changed but certainly not a complete list
the main objective was to have the old->ISA mapping using the variables so the table stays up to date with future renaming

feel free to simplify the list

src/cheri/app-cheri-instructions.adoc
tariqkurd-repo: it's an autogenerated list - I'll add a note that we're not ratifying those

src/cheri/app-cheri-instructions.adoc
tariqkurd-repo: I've broken out those three

PR #821, opened 2025-10-02
Enforce integrity checks

tariqkurd-repo:
Integrity checks can't be optional on YBLD or YSUNSEAL as it would allow creation of tagged invalid caps (e.g. reserved bits set).
So why not just enforce them in all cases.


nwf:
Because YSUNSEAL requires tagged input, any integrity invariants should already hold.  YBLD, OTOH...

tariqkurd-repo:
> Because YSUNSEAL requires tagged input, any integrity invariants should already hold. YBLD, OTOH...

ah ok - I hadn't realised - it really is only YBLD

PR #822, opened 2025-10-02
clarify mtval reporting for bounds faults

tariqkurd-repo:


src/cheri/riscv-priv-integration.adoc
```diff
@@ -227,6 +227,8 @@ endif::[]
 
 For all CHERI faults, <<mtval>> is written with the MXLEN-bit effective address which caused the fault.
 
+NOTE: In the case of the access being out of CHERI bounds, the reported address must be an out of bounds address.
```
arichardson, line 230: Hmm I'm not sure this is how it is handled by other faults? What happens for PMP checks?

tariqkurd-repo: This is all I can find in the priv spec - it's not very helpful:

Failed accesses generate an instruction, load, or store access-fault exception. Note that a single instruction may generate multiple accesses, which may not be mutually atomic. An access-fault exception is generated if at least one access generated by an instruction fails, though other accesses generated by that instruction may succeed with visible side effects. Notably, instructions that reference virtual memory are decomposed into multiple accesses.

PR #839, opened 2025-11-05
clarify base strategy

tariqkurd-repo:
also rename "default encoding" to "university encoding" as Haydn wasn't happy about the default name

all names are still provisional at this stage

I've reviewed the scheme with Haydn.

### DIFF
--- a/src/cheri/riscv-unpriv-integration.adoc
+++ b/src/cheri/riscv-unpriv-integration.adoc
@@ -34,6 +34,8 @@ endif::[]
 
 NOTE: The base + encoding format scheme is designed to simply allow more encoding formats to be added in the future, to allow the formats to be optimized for specific applications.
 
+NOTE: {rvy_uni_base_name} bases are closely based upon features from CHERI v9 cite:[cheri-v9-spec], and the bounds encoding scheme is based upon CHERI Concentrate cite:[woodruff2019cheri].
+
 Custom capability encoding formats may be implemented for vendor-specific non-standard implementations.
 
 === CHERI Overview


src/cheri/app-cap-description.adoc
```diff
@@ -1,15 +1,15 @@
-[#app_cap_description, reftext="{rvy_default_encoding_name}"]
-== {rvy_default_encoding_name} - The Default Capability Encoding in {cheri_base_ext_name}
+[#app_cap_description, reftext="{rvy_uni_base_name}"]
+== {rvy_uni_base_name} - The University of Cambridge Capability Encoding in {cheri_base_ext_name}
```
nwf, line 2: I don't like this name, since at least to my mind that's ISAv9... but if everyone else is happy with it, fine.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -5,8 +5,36 @@ ifdef::cheri_standalone_spec[]
...
 
-This chapter describes the {cheri_base32_ext_name} and {cheri_base64_ext_name} base capability instruction sets, that extend the RV32I and RV64I (or RV32E/RV64E) integer instruction sets with CHERI.
-If the hart supports RV32E/RV64E instead of RV32I/RV64I, {cheri_base_ext_name} the restrictions from <<rv32e>> apply and 16 general purpose registers instead of 32 -- but all other {cheri_base_ext_name} changes are identical so the remainder of this chapter only refers to extending RV32I/RV64I behavior.
+This chapter describes the {cheri_base32_ext_name} and {cheri_base64_ext_name} base capability instruction sets, that extend the RV32I and RV64I instruction sets with CHERI.
+Other standard options to bases are also supported such as E (16-registers), Zfinx and endianess.
+
+=== CHERI encoding formats
+
+CHERI implementations are required to choose a capability encoding format.
+The CHERI base ISA name is extended with the capability encoding format in use, and each implementation only supports one.
```
nwf: "supports only one", and that's going to be challenging for efficient VMs, I think?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -5,8 +5,36 @@ ifdef::cheri_standalone_spec[]
...
+
+The currently available base ISAs including capability encoding formats are listed below:
+
+* For {cheri_base64_ext_name}, the capability encoding format from the University of Cambridge covering the full 64-bit address space:
+** {rvy64_uni_base_name}
+** {rvy64e_uni_base}
+* For {cheri_base32_ext_name}, the capability encoding format from the University of Cambridge covering the full 32-bit address space and optional RISC-V compatibility with {cheri_default_ext_name}:
+** {rvy32_uni_base_name}
+ifndef::cheri_ratification_v1_only[]
+* For {cheri_base32_ext_name}E, the CHERIoT encoding covering 8MB of memory, optimized for smaller devices:
```
nwf: Hm?  "covering 8MB of memory" is, I think, a mischaracterization of the CHERIoT v1 encoding; it can, and its root caps do, represent all 4GiB of the 32-bit address space, with 16MiB bounds alignment.  (And then, yes, it jumps down to at most 8MiB objects at 16KiB bounds alignment.)

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -5,8 +5,36 @@ ifdef::cheri_standalone_spec[]
...
-If the hart supports RV32E/RV64E instead of RV32I/RV64I, {cheri_base_ext_name} the restrictions from <<rv32e>> apply and 16 general purpose registers instead of 32 -- but all other {cheri_base_ext_name} changes are identical so the remainder of this chapter only refers to extending RV32I/RV64I behavior.
+This chapter describes the {cheri_base32_ext_name} and {cheri_base64_ext_name} base capability instruction sets, that extend the RV32I and RV64I instruction sets with CHERI.
+Other standard options to bases are also supported such as E (16-registers), Zfinx and endianess.
+
+=== CHERI encoding formats
+
+CHERI implementations are required to choose a capability encoding format.
+The CHERI base ISA name is extended with the capability encoding format in use, and each implementation only supports one.
+
+The definition of all CHERI base ISAs include sufficient details of a capability encoding format to give enough details to allow execution of all base ISA instructions.
```
nwf, line 16: That was the intent of my parametric base proposal, but that didn't seem hugely popular.  Can you clarify what you mean by "enough details"?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -5,8 +5,36 @@ ifdef::cheri_standalone_spec[]
...
+
+As a result, the {cheri_base32_ext_name} and {cheri_base64_ext_name} names on their own do not describe complete base ISAs, as a capability encoding must be selected.
```
nwf, line 18: Can we _explicitly_ leave room for vendor-specific encodings that remain compatible with these base architectures?

src/cheri/app-cap-description.adoc
tariqkurd-repo: well - it's based on ISAv9 but is certainly not the same - I was also asked not to have references to other architectures in the spec so I think this is a reasonable compromise.

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: Supporting multiple will  need to be a future feature. Krste is not happy about supportng multiple and I don't blame him - it opens the door to nightmare scenarios like variable XLEN which RISC-V are having to back away from.

I think this is sensible for now.

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: I'm happy for you to improve the description.

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: It needs enough details to cover loads, stores, branches, capability ADD etc. so needs to have enough permissions and bounds details included to cover all of those.

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: done

src/cheri/riscv-unpriv-integration.adoc
nwf: OK, that sounds like what the parameters of my proposal were attempting to characterize.  (David later suggested that they be exposed as CSRs in the base ISA, and I think that's a great idea.)

But...

Does "enough details" preclude things like `Zylevels1` being an extension (as opposed to part of the base), since (with `x1` initially a tagged capability and `x2` pointing to memory) the instruction sequence `sy x1, 0(x2); ly x1, 0(x2)` may (ETA: clarified prose here, sorry) modify `x1` in ways beyond the base's understanding (The base's `C` and `LM` permissions already mean `x1` might change, but are other reasons for non-identity OK?)?

Does "enough details" mean that the base needs to be able to answer the question "Given an initial tagged capability in `x1` and `x2` pointing at memory, what is the value of `x1` after the instruction sequence `sy x1, 0(x2); lb x1, n(x2)` for `n` between 0 and `2*XLEN-1`?" and questions like that?

tariqkurd-repo: Zylevels being an extension is fine - allocate an extra reserved permission and then something changes.
It means a full description of loads and stores operating in the absence of Zylevels, including LM behaviour following your second point.

It must be possible for an implementation to only implement any of the defined bases with no CHERI extensions and have a working system - which won't be sensible without sentry support - but it must all be there.

Essentially every base has to be a fully defined entity which can be implemented and verified.

Does that help?

nwf: That's (ETA: mostly) the answer to the first question I expected, so I think that's good.  How about the 2nd?

src/cheri/app-cap-description.adoc
arichardson: We could discuss potential names in the next TG meeting? I agree with Wes that it is different from ISAv9, so maybe we should come up with a different name.

src/cheri/riscv-unpriv-integration.adoc
arichardson: I would say there should not be anything prohibiting supporting multiple encodings. You can support I and E bases already, so there could be some implementation defined mechanism to switch. I am not suggesting this is something that needs to be possible, but prohibiting it seems unnecessary.
```suggestion
The CHERI base ISA name is extended with the capability encoding format in use.
```

src/cheri/riscv-unpriv-integration.adoc
arichardson: I would say this is effectively the same as the parameterized based just without explicitly calling out the parameters in the specification prose.

I'm not sure I fully understand the second property, but given a base+set of extensions that add meaning to reserved bits, the precise bitwise representation should be defined.

src/cheri/riscv-unpriv-integration.adoc
```suggestion
* For {cheri_base32_ext_name}E, the CHERIoT encoding optimized for smaller devices:
```

How about just dropping that statement?

src/cheri/riscv-unpriv-integration.adoc
nwf: > I would say this is effectively the same as the parameterized based just without explicitly calling out the parameters in the specification prose.

I'm OK with that.  What do you think of David's suggestion to add CSRs holding the parameters?  That'd absolve software of understanding the `ycfg` values as far as the base understands them, so early software can do some amount of testing of its (otherwise implicit) assumptions about the base ISA.

tariqkurd-repo: Can you point me at David's proposal?

src/cheri/app-cap-description.adoc
tariqkurd-repo: as long as it's not `default` which made Haydn unhappy

src/cheri/riscv-unpriv-integration.adoc
arichardson: I think having meaningful parameters in a CSR would definitely be good. `ycfg` was just a minimal proposal, doing something more structured sounds good to me.

src/cheri/app-cap-description.adoc
tariqkurd-repo: https://github.com/riscv/riscv-cheri/issues/840

so we don't forget

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: I wonder if this can be a custom thing which we ratify later as a fast-track or do we really need it up-front?
The more stuff we add the slower the first ratification will be.

arichardson: If there is already a proposal, we could replace ycfg encoding with that? Otherwise I agree that waiting for a fast-track later makes sense. This is not strictly required since you can make the assumption in software initially.

src/cheri/app-cap-description.adoc
jrtc27, line 2: "General-purpose"?

jrtc27: No, we should resolve it *prior* to merging this.

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: I'm very much in favour of fast track later - if it's not critically important now......

src/cheri/app-cap-description.adoc
tariqkurd-repo: "General-Purpose" seems reasonable - because CHERIoT is "specialised"
The principle is to choose a specialised one if need be (e.g. small devices choose CHERIoT) and there'll be more in the future, and if not then choose general-purpose.

I'll suggest this to Haydn tonight.

src/cheri/riscv-unpriv-integration.adoc
nwf: > Can you point me at David's proposal?

I don't think there's prose so much as the idea that the scalar parameters of CHERI -- mantissa width, extra bits devoted to representable OOB range, maximum representable exponent, the AUIPC shift -- are in CSRs.  Permission compression / implication / dependencies are harder to fit into CSRs but the base has few enough of them that software can just probe `YPERMWC` behavior directly.

(I won't claim with certainty that that's _the right_ parameterization of CHERI, but nobody has so far indicated that I've gotten it wrong.  There's more detailed discussion in the parametric base proposal document.)

src/cheri/riscv-unpriv-integration.adoc
davidchisnall: The case I see being common for VMs is disabling features.

If you have two formats, A and B, where B is a superset of A, then it is trivial for hardware to support both by disabling the operations that set or check the things in B that are not in A. This allows a B-aware VM to use the features of B, but allows VMs that have been (or might be) migrated to a host that supports only A, to be restricted to A.

Which is yet another reason that the base should not be tied to a specific encoding.

jrtc27: No, that's about extensions to the base encoding chosen, not about the base encoding itself.

src/cheri/riscv-unpriv-integration.adoc
davidchisnall: > I'm very much in favour of fast track later - if it's not critically important now......

To be absolutely clear: if adding a new encoding requires ratifying a new base, even with a fast-track process, this will lead to our voting ‘no’ for ratification. I believe this is also true for Microsoft.

Solving this is the most urgent problem now.

src/cheri/app-cap-description.adoc
davidchisnall: There is no such thing as a general purpose processor (I have literally written a CACM article with this title). The phrase ‘general purpose’ in ISA design means ‘we are optimising for something but haven’t bothered to articulate what our requirements are and so they are probably wrong’. Let’s not go down that road. If this is an encoding for a specific set of use cases, please articulate them.

Issue #840, opened 2025-11-05
Choose a name for the "university" encoding format


tariqkurd-repo:
`default` was unpopular with Haydn as it directed people away from CHERIoT.
`university` is less unpopular but not well received

We need an agreeable and descriptive name for the university encoding format.

Issue #832, opened 2025-10-21
Make encoding format part of the base


tariqkurd-repo:
Krste has asked for the encoding format to be part of the base, so the base becomes parameterised.

RV32Y / RV64Y will have the university CHERI format.

We'll need a different base arch name for CHERIoT and other variants - and need a way to represent these.

So we can drop the Xycfg CSRs.

Does anyone see a need for dynamic selection, or a problem with this?

I'm worried we're going in a circle if @davidchisnall and @nwf aren't happy.

We need to define rules for the architecture string to show different encoding formats - so when the base has a Y then the encoding format is represented somehow.

@jonwoodruff  and @gameboo were present during the discussion.

jrtc27:
But this does now duplicate my #792

tariqkurd-repo:
Anything CHERIoT specific will require the CHERIoT base.

We'll have a single arch string which includes the encoding format, and list extension compatibility against each possible CHERI base.

vmurali:
Krste also suggested creating a family of "CHERI" base specs, with universityCHERI being one instantiation and CHERIoT being another. This would still require (limited) rewrites in 0.9.6 version of the spec, but at least keeps the cap encoding (and some other details) abstract, with one possible instantiation using universityCHERI presented now.

nwf:
> Anything CHERIoT specific will require the CHERIoT base.

Part of the reason for the Xycheriot draft chapter was to show _how little_ of "CHERIoT" is actually _CHERIoT-specific_ and how much of it can be refactored into a suitably extensible common core base and extensions that can, in addition to being used to build CHERIoT, also be used to reconstruct University CHERI atop the same base, for example.

[...]

Issue #791, opened 2025-09-18
Mandatory Sentries


jrtc27:
Here goes the controversial issue filing...

As I have said elsewhere, I do not believe there is a compelling reason to not make sentries (read: implicitly the ambient kind throughout unless stated otherwise) mandatory as part of the base ISA.

Sentries are the simplest way to provide a sealing mechanism for CHERI. They may not be as strong as a more tailored approach, but they are universally applicable to at least some extent. If you have any form of sealing then the additional hardware complexity of introducing sentries is small. But the benefits are simplification of the specification and the ability to have a universal, portable sealing baseline that can be used across CHERI platforms regardless of "big CHERI" or CHERIoT. If CHERIoT RTOS wishes to require the use of its own forwards/backwards edge sentries then it can do so, but that is a software policy (that presumably is then realised through a combination of code generation, ABI and CSR configuration) rather than baking it into the ISA and precluding the ability to use CHERI-aware software that has been ported to CHERI but not CHERIoT.

Sentries do not weaken security; the places they are accepted are places where unsealed capabilities are also accepted, and whilst, yes, type != 0 no longer means the capability was sealed by a party with trusted sealing authority (ignoring the whole "if you can modify a region of executable memory, including via mmap-like things not just writable regions, then you can mint whatever return sentries into that region that you like, making it much more like an ambient sentry in actual guarantees" problem), type > 1 gives you that same check.

The current strategy of sticking Zys in its own extension weakens the spec by allowing RVY to be implemented without sealing, and without providing a portable manner in which to write code that wants to seal capabilities.

arichardson:
I agree you should always have some form of sealing. And I also like ambient sentries and don't want to remove them. However, there are/will be 32-bit encodings where all the bits are used for other types (e.g. CHERIoT) and therefore Zys cannot be implemented.
Additionally, I can see RVY being used as a base ISA for an accelerator that doesn't care about permissions other than R/W and therefore wants to use all bits for bounds. If there is no/limited control-flow we should be able to reclaim this bit.

RVY is similar to RVI in that I don't think it makes much sense to implement a CPU that only supports RVY without extensions such as Zys.

IMO this is something for the RVA/RVB/RVM profile to define and I would veto any profile that does not have sealed types.

jrtc27:
[...]

PR #845, opened 2025-11-13
Draft of _L<params> for capability encodings

arichardson:
Fix https://github.com/riscv/riscv-cheri/issues/840
Fix https://github.com/riscv/riscv-cheri/issues/832
Fix https://github.com/riscv/riscv-cheri/issues/791

I've taken #839 as a start and integrated the encoding appendix into the base chapter.

This should hopefully clarify that all `RVY<XLEN>_L<params>` ISA strings are instantiations of the RVY base ISA. _L<params> was suggested  as a way to define properties that change the observable behaviour of the RISC-V system, so something like _Libigendian, or the capability encoding.

This is a draft, I may have made some mistakes while moving text around.

### DIFF
--- a/src/cheri/insns/auipc_32bit.adoc
+++ b/src/cheri/insns/auipc_32bit.adoc
@@ -72,7 +72,7 @@ and subsequent 12-bit immediate(s),
 thereby ensuring that in-bounds addresses can be reached without risk of
 the intermediate computation exceeding capability bounds.)
 Future extensions that add instructions with similar semantics
-should make use this same encoding-specified shift value or
+should make use of this same encoding-specified shift value or
 otherwise allow the capability encoding to set the shift amount.
 
 NOTE: It is possible for the compiler to generate code that is compatible with any AUIPC shift by emitting a AUIPC with a zero offset followed by a sequence of LUI and <<CADD>>/<<CADDI>>.
--- a/src/cheri/riscv-unpriv-integration.adoc
+++ b/src/cheri/riscv-unpriv-integration.adoc
@@ -6,7 +6,7 @@ WARNING: This chapter will appear in the unpriv spec after the RV32I chapter.
 endif::[]
 
 This chapter describes the {cheri_base32_ext_name} and {cheri_base64_ext_name} base capability instruction sets, that extend the RV32I and RV64I instruction sets with CHERI.
-Other standard options to bases are also supported such as E (16-registers), Zfinx and endianess.
+Other standard options to bases are also supported such as E (16-registers), Zfinx and endianness.
 
 === CHERI Overview
 
@@ -537,7 +537,7 @@ Parameters for bounds precision and representability::
 These parameters affect the behavior of <<SCBNDS>>/<<SCBNDSR>> as well as the representability check that affects the tag-clearing behavior of how far you can go out of bounds with <<SCADDR>>/<<CADD>>.
 * **Mantissa width (mw<N>)**: The mantissa width for the bounds encoding. For example, a format with {cap_rv64_mw_width} bits of mantissa, would use `mw{cap_rv64_mw_width}`.
 * **Maximum exponent (ea or e<N>)**: If _all_ meaningful values of the exponent can be encoded directly in the exponent field this uses `ea`.
-Otherwise `e<N>` describes the largest encodable exponent for a valid capability except for the on used for the whole address space.
+Otherwise `e<N>` describes the largest encodable exponent for a valid capability except for the one used for the whole address space.
 For example, the CHERIoTv1 format uses  `e14` since E=15 is used for full address space capabilities.
 The default value for this parameter is `ea` and may be omitted in that case.
 * **Representable region (rc1 or r0)**: The number of mantissa bits used to guarantee representability of out-of-bounds capability values.
--- a/src/cheri/rvy32-encoding.adoc
+++ b/src/cheri/rvy32-encoding.adoc
@@ -256,7 +256,7 @@ Capabilities in this encoding have a 1-bit field for <<sec_cap_type>> values whi
 [#section_cap_bounds32]
 ==== Bounds (EF, T, TE, B, BE, L~8~) Encoding
 
-The bounds are encoding in the same way as in <<section_cap_bounds,{rv64y_uni_base_name}>>, with the appropriate values for mantissa width and maximum exponent substituted.
+The bounds are encoded in the same way as in <<section_cap_bounds,{rv64y_uni_base_name}>>, with the appropriate values for mantissa width and maximum exponent substituted.
 Compared to <<rv64y_cap_description,{rv64y_uni_base_name}>>, this encoding uses an additional `L~8~` bit as described in xref:section_cap_bounds[xrefstyle=short].
 
 // ===== <<AUIPC_CHERI>> Representability
--- a/src/cheri/rvy64-encoding.adoc
+++ b/src/cheri/rvy64-encoding.adoc
@@ -275,7 +275,7 @@ xref:top_bound_dec[xrefstyle=short] and xref:base_bound_dec[xrefstyle=short]
 include ranges which may not be present when the bounds are decoded:
 
 * If E = 0 the lower section does not exist.
-* If E+MW=XLEN then the top section is only the least signification bit of
+* If E+MW=XLEN then the top section is only the least significant bit of
  c~t~ for the top bound, and top section doesn't exist for the bottom bound.
 * If E+MW>XLEN then neither top section exists, and so the bounds are calculated
  with no dependency on the address field _a_.
@@ -455,8 +455,8 @@ from `T` and `E` which will be identical for both the source and destination
 capabilities, thus guaranteeing that `t == t'`.  Likewise, with such values of E, the
 base bound is entirely derived from `B` and `E` and therefore `b == b'`.
 
-The calculation of the MSB of the top bound maybe inverted as specified
-xref:t_bound_msb_inversion[xrefstyle=short].
+The calculation of the MSB of the top bound may be inverted as specified
+in xref:t_bound_msb_inversion[xrefstyle=short].
 Assuming `(E < (CAP_MAX_E - 1))`, the truth-table for this inversion is as follows:
 
 .Top bound MSB inversion truth table


arichardson:
None of the awkward names should ever be used, -march=<profile> is the sensible solution.

src/cheri/rvy64-encoding.adoc
```diff
@@ -0,0 +1,593 @@
...
+Capabilities with malformed bounds:
+
+. Return both base and top bounds as zero, which affects instructions like <<GCBASE>>.
+. Cause certain manipulation instructions like <<CADDI>> to always set the {ctag} of the result to zero.
+
+===== <<AUIPC_CHERI>> Shift
+
+#Begin new since last ARC review#
+
+For the capabilities of this chapter, the <<section_auipc_shift>> value is 12.
```
tariqkurd-repo: I'd note that the value 12 is unchanged from the default, and that other encoding formats may need to change it due to retaining the representability.

src/cheri/rvy32-encoding.adoc
```diff
@@ -0,0 +1,304 @@
...
+
+[#section_cap_bounds32]
+==== Bounds (EF, T, TE, B, BE, L8) Encoding
+
+The bounds are encoding in the same way as in <<section_cap_bounds,{rvy64_uni_base_name}>>, with the appropriate values for mantissa width and maximum exponent substituted.
+Compared to {rvy64_uni_base_name}, this encoding uses an additional `L8` bit as described in xref:section_cap_bounds[xrefstyle=short].
+
+===== <<AUIPC_CHERI>> Shift
+
+For the capabilities of this encoding, the <<section_auipc_shift>> value is 12.
```
tariqkurd-repo: I'd note that the value 12 is unchanged from the default, and that other encoding formats may need to change it due to retaining the representability.

src/cheri/rvy64-encoding.adoc
```diff
@@ -0,0 +1,593 @@
...
+Each memory location or register able to hold a capability must also store the {ctag} as _out of band_ or _hidden_ information that software cannot directly set or clear.
+The capability metadata is held in the most significant bits and the address is held in the least significant bits.
+
+.Capability encoding for {rvy64_uni_base_name}
+[#cap_encoding_xlen64]
+include::img/cap-encoding-xlen64.edn[]
+
+NOTE: Reserved bits must be 0 in valid capabilities and are available for future extensions to {cheri_base_ext_name}.
+
+The encoding diagram above shows the format of capabilities in the case where all {cheri_base_ext_name} (Zy*) extension listed in this manual are supported.
```
```suggestion
```

I think that text is out of date now

src/cheri/rvy32-encoding.adoc
```diff
@@ -0,0 +1,304 @@
...
+Each memory location or register able to hold a capability must also store the {ctag} as _out of band_ or _hidden_ information that software cannot directly set or clear.
+The capability metadata is held in the most significant bits and the address is held in the least significant bits.
+
+.Capability encoding for {rvy32_uni_base_name}
+[#cap_encoding_xlen32]
+include::img/cap-encoding-xlen32.edn[]
+
+NOTE: Reserved bits must be 0 in valid capabilities and are available for future extensions to {cheri_base_ext_name}.
+
+The encoding diagram above shows the format of capabilities in the case where all {cheri_base_ext_name} (Zy*) extension listed in this manual are supported.
```
tariqkurd-repo:
```suggestion
```

I think this text is out of date now

src/cheri/attributes.adoc
```diff
@@ -67,8 +67,18 @@ endif::support_varxlen[]
 :rvy_zicbop_mod_file_name: {cheri_base_ext_name}_Zicbop_mod_insns_table_body
 // 32-bit encodings added by Zyhybrid
 :rvy_hybrid_file_name:     {cheri_default_ext_name}_insns_table_body
-// default encoding scheme
-:rvy_default_encoding_name: Zydefaultcap
+// university (default) encoding scheme
+// For some reason can't use _Lymw{cap_rv32_mw_width}e{cap_rv32_exp_max}rc1pc
+// here, probably need some macro for double expansion
+:rvy32_uni_encoding_suffix: _Lymw10e24rc1pc
```
tariqkurd-repo: and an "h" for "hybrid supported"?

src/cheri/rvy64-encoding.adoc
```diff
@@ -0,0 +1,593 @@
...
+
+The bit width of T and B are defined in terms of the mantissa width (MW) which
+is set depending on capability encoding as shown in
+xref:mantissa_bit_width[xrefstyle=short].
+
+.Mantissa width (MW) values depending on capability encoding
+[#mantissa_bit_width,options=header,align="center",width="55%"]
+|==============================================================================
+^| Encoding   ^| MW
+^| {rvy32_uni_base_name} ^| {cap_rv32_mw_width}
```
tariqkurd-repo: I think this should move into the RV32Y encoding chapter
Just keep this section about RV64Y

src/cheri/rvy64-encoding.adoc
```diff
@@ -0,0 +1,593 @@
...
+----
+
+The possible values for EW and CAP_MAX_E are shown in
+xref:exp_bit_width[xrefstyle=short].
+
+.Exponent widths and CAP_MAX_E depending on capability encoding
+[#exp_bit_width,options=header,align="center",width="55%"]
+|==============================================================================
+^| XLEN                 ^| EW ^| CAP_MAX_E
+^| {rvy32_uni_base_name} ^| {cap_rv32_exp_width}  ^| {cap_rv32_exp_max}
```
tariqkurd-repo: I think this should move into the RV32Y encoding chapter
Just keep this section about RV64Y

src/cheri/rvy32-encoding.adoc
```diff
@@ -0,0 +1,304 @@
...
+
+[#section_cap_sealed32]
+==== Capability Type (CT) Encoding
+
+Capabilities in this encoding have a 1-bit field for <<sec_cap_type>> values which behaves in the same way as the {rvy64_uni_base_name} <<section_cap_sealed>>.
+
+[#section_cap_bounds32]
+==== Bounds (EF, T, TE, B, BE, L8) Encoding
+
+The bounds are encoding in the same way as in <<section_cap_bounds,{rvy64_uni_base_name}>>, with the appropriate values for mantissa width and maximum exponent substituted.
```
tariqkurd-repo: I think all the parameters should be in a table here

tariqkurd-repo: done

src/cheri/rvy64-encoding.adoc
```diff
@@ -0,0 +1,607 @@
...
+
+* seals its return capabilities with type 1.
+
+<<JALR_CHERI>> places no constraints on the triple of input <<sec_cap_type>> value, `{cd}` selector, and `{cs1}` selector.
+That is, <<JALR_CHERI>> will, as directed, attempt to jump to any unsealed or sealed capability in any register regardless of which register comes to hold the sealed return pointer.
+
+ifdef::cheri_standalone_spec[]
+[NOTE]
+=====
+The permission encodings of <<rvy32_cap_description>>/<<rvy64_cap_description>> do not provide mappings for
```
tariqkurd-repo: forward reference to rvy32

src/cheri/rvy64-encoding.adoc
```diff
@@ -0,0 +1,607 @@
...
+
+#End changed since last ARC review#
+
+[#section_cap_bounds]
+==== Bounds (EF, T, TE, B, BE) Encoding
+
+===== Concept
+
+ifdef::cheri_v9_annotations[]
+NOTE: *CHERI v9 Note:* The bounds mantissa width is different in {rvy32_uni_base_name}.
```
tariqkurd-repo: forward reference to rvy32

src/cheri/rvy64-encoding.adoc
```diff
@@ -0,0 +1,607 @@
...
+
+The base, _b_, and top, _t_, addresses are derived from the address by
+substituting _a_[E + MW - 1:E] with B and T respectively and clearing the
+lower E bits.  The most significant bits of _a_ may be adjusted up or down by 1
+using corrections _c~b~_ and _c~t~_ to allow encoding memory regions that span
+alignment boundaries.
+
+The EF bit selects between two cases:
+
+1. EF = 1: The exponent is 0. When {rvy32_uni_base_name}, L~8~ encodes the MSB of the length,
```
tariqkurd-repo: forward reference to rvy32

src/cheri/rvy64-encoding.adoc
```diff
@@ -0,0 +1,607 @@
...
+which can be used to derive T[MW-1:MW-2], forming a full MW-wide T field.
+2. EF = 0: The exponent is _internal_ with E stored in the lower bits of T and
+B, with L~8~ used for the MSB of E when {rvy32_uni_base_name}. E is chosen so that the most
```
tariqkurd-repo: forward reference to rvy32

src/cheri/rvy64-encoding.adoc
```diff
@@ -0,0 +1,607 @@
...
+significant non-zero bit of the length of the region aligns with T[MW - 2] such
+that this bit is implied by E.
+
+The most significant two bits of T can be derived from B using
+the equality `T = B + L`, where L[MW - 2] is known from the values of EF and E
+(as well as L~8~ for {rvy32_uni_base_name}).
```
tariqkurd-repo: forward reference to rvy32

src/cheri/rvy64-encoding.adoc
```diff
@@ -0,0 +1,607 @@
...
+NOTE: A capability has infinite bounds if E=CAP_MAX_E and it is not
+malformed (see xref:section_cap_malformed[xrefstyle=short]); this check is
+equivalent to _b_=0 and _t_&#8805;2^XLEN^.
+
+[#section_cap_malformed]
+===== Malformed Capability Bounds
+
+A capability is _malformed_ if its bounds cannot be correctly decoded.
+The following check
+indicates whether a capability is malformed. `enableL8` is true when {rvy32_uni_base_name}
```
tariqkurd-repo: forward reference to rvy32

src/cheri/rvy64-encoding.adoc
```diff
@@ -0,0 +1,607 @@
...
+expanded base is 0 and top is 2^XLEN^.
+
+.Field values of the NULL capability
+[#null-cap-encoding,reftext="NULL",options=header,align=center,width="70%",cols="25,23,52"]
+|==============================================================================
+| Field    | Value  | Comment
+| {ctag_title}             | zero   | Capability is not valid
+| <<SDP-field,SDP>>        | zeros  | Grants no permissions
+| <<AP-field,AP>>          | zeros  | Grants no permissions
+| <<M-bit,M>>              | zero   | No meaning since non-executable ({rvy64_uni_base_name} and {cheri_default_ext_name} only)
```
tariqkurd-repo: rvy64_uni_base_name is redundant here

src/cheri/rvy64-encoding.adoc
```diff
@@ -0,0 +1,607 @@
...
+| CT       | zero   | Unsealed
+| EF       | zero   | Internal exponent format
+| L~8~     | zero   | Top address reconstruction bit ({rvy32_uni_base_name} only)
```
tariqkurd-repo: forward reference to rvy32

src/cheri/rvy64-encoding.adoc
```diff
@@ -0,0 +1,607 @@
...
+NOTE: This capability format has a single encoding for the _Infinite_ capability. The CHERIoT encoding format has separate roots for code and for data, and will be included in a future version of this specification.
+
+.Field values of the Infinite capability
+[#infinite-cap-encoding,reftext="Infinite"]
+[options=header,align=center,width="90%",cols="25,23,52"]
+|==============================================================================
+| Field         | Value | Comment
+| {ctag_title}  | one   | Capability is valid
+| SDP           | ones  | Grants all permissions
+| AP ({rvy32_uni_base_name}) | 0x8/0x9^1^  (see xref:cap_perms_encoding32[xrefstyle=short])
```
tariqkurd-repo: forward reference to rvy32

src/cheri/rvy64-encoding.adoc
```diff
@@ -0,0 +1,607 @@
...
+                        | Grants all permissions
+| AP ({rvy64_uni_base_name}) | 0xFF (see xref:cap_perms_encoding64[xrefstyle=short])
+                        | Grants all permissions
+| CT            | zero  | Unsealed
+| EF            | zero  | Internal exponent format
+| L~8~          | zero  | Top address reconstruction bit ({rvy32_uni_base_name} only)
```
tariqkurd-repo: forward reference to rvy32

src/cheri/rvy64-encoding.adoc
```diff
@@ -0,0 +1,607 @@
...
+.Field values of the Infinite capability
+[#infinite-cap-encoding,reftext="Infinite"]
+[options=header,align=center,width="90%",cols="25,23,52"]
+|==============================================================================
+| Field         | Value | Comment
+| {ctag_title}  | one   | Capability is valid
+| SDP           | ones  | Grants all permissions
+| AP ({rvy32_uni_base_name}) | 0x8/0x9^1^  (see xref:cap_perms_encoding32[xrefstyle=short])
+                        | Grants all permissions
+| AP ({rvy64_uni_base_name}) | 0xFF (see xref:cap_perms_encoding64[xrefstyle=short])
```
tariqkurd-repo: rvy64_uni_base_name is redundant here

src/cheri/rvy64-encoding.adoc
arichardson: Yeah I ran out of time last night to remove the final tables from this chapter

src/cheri/rvy32-encoding.adoc
```diff
@@ -218,13 +218,13 @@ NOTE: Software is completely free to define the usage of these bits.
 [#section_cap_sealed32]
 ==== Capability Type (CT) Encoding
 
-Capabilities in this encoding have a 1-bit field for <<sec_cap_type>> values which behaves in the same way as the {rvy64_uni_base_name} <<section_cap_sealed>>.
+Capabilities in this encoding have a 1-bit field for <<sec_cap_type>> values which behaves in the same way as the <<section_cap_sealed,{rvy64_uni_base_name}>>.
```
arichardson: This one was actually intentional 'as the RVY64_L... capability type (T) encoding'

src/cheri/attributes.adoc
arichardson: No this should not be needed, we are just assigning meaning to reserved bits in hybrid mode for M.

nwf: Hybrid mode does a few straightforward things, all of which seem to have a shape of things other extensions are going to want to do, too.  Specifically, it...
- adds the `ddc` CSRs,
- augments capabilities with a (X-dependent) "M-bit",
- adds getter/setter instructions for this "M-bit" (both in GPRs -- `YMODE[RW]` -- and in `pc` -- `YMODESW[IY]`), and
- modulates the semantics of load and store instructions (under the influence of `pc`'s "M-bit" and `ddc`).

Exactly one of the following things should be true:
- Hybrid does not impact capability encoding in a way that would need a new base, or
- One of the above aspects (of what is presently defined as the Zyhbrid _extension_) is, in fact, not suitable as an extension.

If the former is not true, then we have serious problems.

src/cheri/riscv-hybrid-integration.adoc
```diff
@@ -64,7 +64,7 @@ requires the CHERI RISC-V hart to support {cheri_base_ext_name} and
 The _<<cheri_execution_mode>>_ is determined by a bit in the metadata of the <<pcc>> called the <<m_bit>>.
 {cheri_default_ext_name} adds a new _<<cheri_execution_mode>>_ field (M) to the capability format.
 Although always present, it only takes effect in code capabilities, i.e., when the <<x_perm>> is set.
-The exact location of the M-bit in the capability format for XLEN=32 and XLEN=64 is described in <<app_cap_description>>.
+The exact location of the M-bit in the capability format is described in <<rvy32_cap_description>>/<<rvy64_cap_description>>.
```
nwf: I don't like the word "location" here because M is X-dependent and (in contradiction of some of the text elsewhere in this file) need not be present in X-perm-clear capabilities.  (I can raise a PR to fix said other text.)

```suggestion
{cheri_default_ext_name} is compatible only with capability encodings that specify transport of the M-bit (see <<rvy32_cap_description>> and <<rvy64_cap_description>>, for examples).
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -5,8 +5,55 @@ ifdef::cheri_standalone_spec[]
...
-If the hart supports RV32E/RV64E instead of RV32I/RV64I, {cheri_base_ext_name} the restrictions from <<rv32e>> apply and 16 general purpose registers instead of 32 -- but all other {cheri_base_ext_name} changes are identical so the remainder of this chapter only refers to extending RV32I/RV64I behavior.
+This chapter describes the {cheri_base32_ext_name} and {cheri_base64_ext_name} base capability instruction sets, that extend the RV32I and RV64I instruction sets with CHERI.
+Other standard options to bases are also supported such as E (16-registers), Zfinx and endianess.
+
+=== CHERI encoding formats
+CHERI implementations are required to choose a capability encoding format that defines certain properties of the ISA such as the bounds precision and permissions encoding.
+The CHERI base ISA name is extended with the capability encoding format in use as a suffix.
+The definition of all CHERI base ISAs includes sufficient details of a capability encoding format to give enough details to allow execution of all base ISA instructions.
```
nwf: This paragraph is perhaps at odds with the next, but the last sentence seems right to me.  So perhaps...

```suggestion
CHERI implementations make trade-offs in their "encoding formats", such as the _precision_ of bounds and the _expressible set_ of combinations of permissions, that impact certain behaviors of the ISA.
The CHERI base ISA name is extended with a compact description of these decisions,
and the definition of all CHERI base ISAs includes sufficient details of a capability encoding format to give enough details to allow execution of all base ISA instructions.
```

src/cheri/attributes.adoc
tariqkurd-repo: yes - hybrid doesn't need a new base - but any base which allows it to be implemented needs to fully describe how the M-bit is encoded

src/cheri/rvy64-encoding.adoc
tariqkurd-repo: I'm fixing this

src/cheri/rvy64-encoding.adoc
tariqkurd-repo: I'm fixing this

src/cheri/rvy64-encoding.adoc
```diff
@@ -0,0 +1,602 @@
...
+expanded base is 0 and top is 2^XLEN^.
+
+.Field values of the NULL capability
+[#null-cap-encoding,reftext="NULL",options=header,align=center,width="70%",cols="25,23,52"]
+|==============================================================================
+| Field    | Value  | Comment
+| {ctag_title}             | zero   | Capability is not valid
+| <<SDP-field,SDP>>        | zeros  | Grants no permissions
+| <<AP-field,AP>>          | zeros  | Grants no permissions
+| <<M-bit,M>>^1^             zero   | No meaning since non-executable ({cheri_default_ext_name} only)
```
tariqkurd-repo: missing a | character in the table

src/cheri/riscv-hybrid-integration.adoc
arichardson: I agree this is a much better wording, will incorporate it when I make edits later today.

src/cheri/insns/auipc_32bit.adoc
nwf, line 64: This is now a base parameter, rather than something "specified by the encoding" per se.

tariqkurd-repo: yeah - I've clarified the text - hopefully better now

src/cheri/rvy32-encoding.adoc
```diff
@@ -0,0 +1,336 @@
...
+
+Capabilities in this encoding have a 1-bit field for <<sec_cap_type>> values which behaves in the same way as the <<section_cap_sealed,{rvy64_uni_base_name}>>.
+
+[#section_cap_bounds32]
+==== Bounds (EF, T, TE, B, BE, L~8~) Encoding
+
+The bounds are encoding in the same way as in <<section_cap_bounds,{rvy64_uni_base_name}>>, with the appropriate values for mantissa width and maximum exponent substituted.
+Compared to <<rvy64_cap_description,{rvy64_uni_base_name}>>, this encoding uses an additional `L~8~` bit as described in xref:section_cap_bounds[xrefstyle=short].
+
+===== <<AUIPC_CHERI>>+<<JALR_CHERI>> Representability
```
arichardson: It's not just jalr, it's all paired 12 bit immediate functions such as ADDY and LY as well

tariqkurd-repo: yes good point - although in practice it's almost always JALR - I'll fix it

tariqkurd-repo: done - please have another look - for both rv32 and rv64

nwf: FWIW, for CHERIoT specifically, it's very often the equivalent of `AUIPC; LY`: intra-compartment calls almost always just use `JAL`, library calls fetch sentries from the calling compartment's import table (stored at PCC's base, so `AUIPC; LY; JALR`), and cross-compartment calls are done by passing a descriptor of the callee (also fetched from an import table) to the trusted switcher (whose sentry is also in the import table, so `AUIPC; LY; AUIPC; LY; JALR`).

src/cheri/rvy32-encoding.adoc
tariqkurd-repo: reverted

src/cheri/insns/auipc_32bit.adoc
```diff
@@ -19,7 +19,7 @@ include::base_isa_extension.adoc[]
 Description::
 Form a 32-bit offset from the 20-bit immediate filling the lowest bits with zero;
 the number of places to shift is determined by the capability encoding's choice of
-the <<section_auipc_shift>> value (12, unless otherwise specified).
+the <<section_auipc_shift>> value (12, unless otherwise specified by the capability encoding format).
```
nwf, line 22: As I noted in the parametric base proposal, the AUIPC shift isn't _really_ a property of the encoding format, it's just that some encoding formats struggle with too large an AUIPC shift (but software on a smaller-mantissa system _could_ always replace a potentially-bounds-violating `AUIPC n; LY` with an always-in-bounds `AUIPC (n±k); ADDI; ...; ADDI; LY` to compensate!).  The AUIPC shift really is _just_ a parameter of the base.

src/cheri/rvy32-encoding.adoc
tariqkurd-repo: interesting - do you want to enhance the text?

src/cheri/app-cheri-instructions.adoc
```diff
@@ -128,7 +128,7 @@ WARNING: {YSEAL} and {YUNSEAL} are not included in the v1.0 ratification package
 The following changes are for forwards compatibility with {cheriot_ext_name}:
 
 . Capability encodings are now the naming authorities for capability types (only 0/unsealed exists in the base architecture).
-. Capability encodings are now in a separate extension - the default being <<app_cap_description,{rvy_default_encoding_name}>>.
+. Capability encodings are now in separate base extensions.
```
nwf: Maybe...?
```suggestion
. Capability encoding formats are now in separate base parameterizations.
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -5,8 +5,55 @@ ifdef::cheri_standalone_spec[]
...
+** {rvy32_uni_base_name}
+ifndef::cheri_ratification_v1_only[]
+* For {cheri_base32_ext_name}E, the CHERIoT encoding optimized for smaller devices:
+** {rvy32e_cheriot_base_name}
+
+NOTE: RV32YE (16-register) implementations are recommended to use the {rvy32e_cheriot_base_name} format.
+
+endif::[]
+
+NOTE: The base `_L<encoding_parameters>` encoding format scheme is designed to simply more encoding formats optimized for specific applications to be added in the future.
```
nwf:
```suggestion
NOTE: The base `_L<encoding_parameters>` encoding format scheme is designed to separate the aspects of capability encoding that influence the behavior of {cheri_base_ext_name} instructions from the rest.
The intent is to simplify the future development of more (possibly domain-specific) encoding formats while also more directly conveying compatibility of architectural bases.
```

src/cheri/attributes.adoc
nwf: I _thought_, but may well have been wrong, that this would be something more like `_Lymw10e24rc1pc_Lyencuni1`, and similarly for the rest?

src/cheri/attributes.adoc
```diff
@@ -67,8 +67,18 @@ endif::support_varxlen[]
 :rvy_zicbop_mod_file_name: {cheri_base_ext_name}_Zicbop_mod_insns_table_body
 // 32-bit encodings added by Zyhybrid
 :rvy_hybrid_file_name:     {cheri_default_ext_name}_insns_table_body
-// default encoding scheme
-:rvy_default_encoding_name: Zydefaultcap
+// university (default) encoding scheme
+// For some reason can't use _Lymw{cap_rv32_mw_width}e{cap_rv32_exp_max}rc1pc
+// here, probably need some macro for double expansion
```
nwf: According to the manual, that _should_ work?  ("It also means you can reference the value of attributes which have already been defined when defining the value of an attribute." sayeth https://docs.asciidoctor.org/asciidoc/latest/attributes/attribute-entries/ .)

src/cheri/attributes.adoc
arichardson: I was thinking since _Lymw10e24rc1pc is currently the only encoding with these parameters we don't need a second `_L<encoding_variant>`. A future one with the same parameters but different other properties could use a variant/version suffix? E.g. `_Lymw10e24rc1pcv2`

src/cheri/attributes.adoc
arichardson: Yeah I am very confused why `:foo: a{some_macro}` works, but it doesn't seem to work with `:foo: a{some_macro}` when the latter contains further expansions `:some_macro: {other_macro}`... Might need some of the weird `pass:` syntax here but I ran out of patience and just inserted the raw values...

src/cheri/app-cheri-instructions.adoc
arichardson: Sounds good to me 👍

src/cheri/img/cap-encoding-xlen32.edn
```diff
@@ -14,7 +14,7 @@
 (draw-box "Reserved"    {:span 3})
 (draw-box "CT"          {:span 1})
 (draw-box "EF"          {:span 1})
-(draw-box "L8"          {:span 1})
+(draw-box "L\u2088"    {:span 1})
```
arichardson:
```suggestion
(draw-box "L\u2088"     {:span 1})
```

src/cheri/attributes.adoc
nwf: I don't love the experience of finding bugs in document authoring tools. :(

src/cheri/attributes.adoc
nwf: I worry that it wouldn't be obvious that `_Lymw10e24rc1pcv2` conforms to `_Lymw10e24rc1pc` (with all the contractual implications that might raise)?  I think...

- making the format an explicitly separate `_Lyenc{...}` field of the base name hopefully makes it clear that `...__Lymw10e24rc1pc_Lyenc1` and `_Lymw10e24rc1pc_Lyxencfoo` are both specializations of, and so conforming to, `_Lymw10e24rc1pc`, and

- if `_Lymw10e24rc1pc` implies an encoding, then `_Lymw10e24rc1pc_Lyxencfoo` has to be phrased as _replacing_ the implied encoding, rather than being of the same shape as the implicit encoding.

nwf: Also, and this is vocabulary nit picking (for which I am at least a little bit sorry), I object to calling `_Lymw10e24rc1pc` an "encoding" and viewing everything else as a "variant" of that encoding.  There's nothing about `_Lymw10e24rc1pc` that necessitates any particular field layout (though I concede that `RVY` itself, regardless of `_Ly...` suffixes, might insist that caps are a tag, an `XLEN`-bit metadata field, and a `XLEN`-bit data field, with the latter two contiguous and adjacent in memory in a particular order... I don't like it, but I can only die on so many hills at once).

arichardson: I absolutely do agree with you that strictly speaking `_Lymw10e24rc1pc` only specifies the behaviour of the instructions without necessarily mandating how those bits are laid out in memory and the concrete layout `_Lencfoo1` is a separate thing. Having it separate would make the SW compatibility more obvious.

However, the way I understand it we have been asked to combine this into one _L parameter, so a suffix for future versions seems like the best compromise to me. Also a _Lymw10e24rc1pc would also be forwards-compatible with a hypothetical `_Lymw9e24rc1pc` since it has strictly greater precision, but that is also not immediately obvious since the name is different. A future version that is compatible but allocates a few bits differently could be named `_Lymw10e24rc1pcenc2`?

I am not sure what the best approach is I am just trying to find a compromise that is somewhat acceptable to everyone

arichardson: Alternatively we could also unconditionally an `encfoo` suffix to the _L parameter and toolchains can ignore it? I was hoping to avoid having to assign a name for the encodings in the base spec.
Also that would allow using something like `_Lymw9e24r0as11pc` for CHERIoTv1 (I may have got the parameters slightly wrong) since that is currently the only encoding that exists with these parameters and a future one would need to add a suffix.

nwf: > However, the way I understand it we have been asked to combine this into one _L parameter

I thought we had been asked to combine things into one _base_, not necessarily one `_L` parameter?

> but that is also not immediately obvious since the name is different.

I agree that the _inter-base_ compatibility is not immediately obvious from the parameterization (alas) but at least the _intra-base_ compatibility is just lexical equality.

If you don't want to head down the two-parameter `_Ly[parameters]_Lyenc[name]` path, may I suggest that there _always_ be a `encN` substring in `[parameters]` and that the prose defining the parameterization schema _explicitly_ indicate that the base without a specified `enc` parameter is a thing to which implementations can (and those with vendor-specific encodings are expected to) conform.

src/cheri/attributes.adoc
arichardson: Managed to figure this out, I was referencing a macro that wasn't defined yet...

src/cheri/insns/auipc_32bit.adoc
tariqkurd-repo: You're right - and I think it's reasonable not to redefine the AUIPC shift in the base RVY chapter, which we're basing the two university formats on. Therefore it shouldn't be mentioned at all here.

I think all the text about the AUIPC shift need to go into the CHERIoT base, as that's where it's relevant. It can be redefined for all CHERIoT formats in that way.

The whole point of having different bases is to redefine the underlying RVY behaviour, which this is an example of.

We can't list everything we that change in future bases in the RVY base, so let's remove the mention from RVY and push it into CHERIoT only. @nwf what do you think?

nwf: The ambition behind the parametric base proposal, which Alex has so kindly taken on and transmogrified in this PR, was to fully define the observable behavior of the capability getters and setters, leaving only the field layout (and so, necessarily, the behavior of some load and store instructions) to the encoding.  In this spirit, `AUIPC` is a capability setter, and its behavior is fully determined by the capability mantissa width and the shift applied to the immediate.  (ETA: no, it's also dependent on the representable window, as Robert NW reminds me.)

That is, I think the AUIPC shift belongs as a parameter in the `_Ly{...}` string, and I was just noting that, in contrast to the others, it is not as overtly tied to the encoding.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -459,6 +488,79 @@ Extensions introducing new permissions may require these to be provided by root
...
+** {rvy32e_cheriot_base_name}
+
+NOTE: RV32YE (16-register) implementations are recommended to use the {rvy32e_cheriot_base_name} format.
+
+endif::[]
+
+NOTE: The base `_L<encoding_parameters>` encoding format scheme is designed to separate the aspects of capability encoding that influence the behavior of {cheri_base_ext_name} instructions from the rest.
+The intent is to simplify the future development of more (possibly domain-specific) encoding formats while also more directly conveying compatibility of architectural bases.
+
+NOTE: It is expected that the actual {cheri_base_ext_name} instantiation is chosen using a profile rather than a full architecture string and therefore the unwieldy suffix should rarely be user-visible.
```
tariqkurd-repo, line 576: ok - but what would the profile be called if it doens't include the unwieldy suffix in its name?

src/cheri/insns/auipc_32bit.adoc
tariqkurd-repo: ok - but by not including more parameters in the RVY base does not exclude them from being added by capability formats (and so) bases later - so we are still free to add more options later without changing the RVY chapter. The RVY chapter just includes things we know about _now_ which we'd like as parameters.

arichardson: I was thinking yesterday that we could actually omit the AS parameter and say that it's implied if representability isn't sufficient for +-2KB? I'll do more work on the PR later today and clarify more things such as adding the known parameters to the RVY chapter.

nwf: Robert reminds me that it's not just mantissa width that matters, but also OOB representation.  If you want to do away with the parameter in favor of deriving it from others -- which seems fine as an approach and eliminates only silly things we'd never want to do in practice -- I think the rule is:

> if `mw` and `r` imply that capabilities 1KiB in length or larger can always represent at least 3KiB out of bounds, then the AUIPC shift is 12; otherwise, it is 11.

(Since, as Peter points out, you won't use `AUIPC` with nonzero immediate unless the object is large enough that the 10-bit signed immediates of `addi` and L/S instructions won't always be able to span the whole thing from any position within (so, 1KiB or more in size) and, more specifically, the `AUIPC` is 1 KiB or more away from the location it's trying to reach.  In the worst case, that location is the upper or lower bound, and so the 12 bit shift is going to take you up to 3KiB out of bounds.)

Does that seem right?

tariqkurd-repo: I don't know if the rule is correct - I suspect it should be non-normative. Encoding formats can always override everything anyway (which is the advantage of being a base). GIven that I would say that non-normative guidance is fine in this case.

src/cheri/riscv-unpriv-integration.adoc
arichardson: `RVYB26`/`RVYA26`? And `RVYM26` for 32-bit? For application class the profile can say only the current encoding is compatible. And for embedded the profile can say the _L parameters of the following values (Current RVY32, CHERIoTv2, etc.) are legal since there is no need for binary markets there.

tariqkurd-repo: ok - so what you're saying is that the 32-bit profile(s) can be orthogonal to encoding format - I'm not sure that works though as CHERIoT wants Zyseal and not Zybld (which we still need to separate out) - so I'd expect one or more CHERIoT profiles and University RV32 profiles too.

src/cheri/insns/auipc_32bit.adoc
nwf: I remain deeply concerned about "Encoding formats can always override", because that sounds like it runs the risk of requiring ratification to claim conformance with a base, and that's the whole game here.

tariqkurd-repo: Every encoding format _is_ a base don't forget - and every one will need ratification to become part of the standard.
It's a choice about how much is changed, there's no reason to unnecessarily change the AUIPC shift for example, each one can be as simple as re-encoding the permissions.
So really we're planning to define a hierarchy - an overall CHERIoT base with AUIPC_SHIFT=11 and formats derived from that.

It's really up to you if you decide to push all the different formats through the ratification process, that's a company decision for SCI. I'm just saying what the standards process is. Talking to David last week it's inline with your plans.

tariqkurd-repo: It's also possible to put text in the base about the AUIPC shift if that helps, and not have a CHERIoT base - just base CHERIoT encoding(s) directly on the RV32Y base.
Just be aware that if it's normative text about the shift and we ratify it then it can't be changed in the base.

Although again, all encoding formats are derived bases and so have the ability to override anything they like, which is the whole principle of bases, to allow the base ISA to be altered.

nwf: > Every encoding format is a base don't forget - and every one will need ratification to become part of the standard.

Again, the whole impetus for the parametric base proposal (that is an increasingly distant progenitor to this PR; thanks for everyone's hard work) was to permit implementations with non-standard/non-ratified encodings _to nevertheless be considered conformant with a ratified base_.  If this isn't possible in this approach, we should stop and try something else.

> ust be aware that if it's normative text about the shift and we ratify it then it can't be changed in the base.
> Although again, all encoding formats are derived bases and so have the ability to override anything they like

I feel like I am missing what these two sentences are meant to convey; they seem at odds to me.

tariqkurd-repo: I think the most important thing at this point is to have a CHERIoT format in the spec which we can put forward for ARC review. We're blocked until we do that.
So I'd like to get this PR merged ASAP.

As for the AUIPC shift - you can put the definition of the calculation in the RVY base if you like. The encoding format is a derivation of the RVY base and at that point can override behaviour (which is the point of it being a base, and the source of the requirement to declare encoding formats as a base because they can change how branches, loads, stores, AUIPC etc. execute).

The CHERIoT base will be conformant with the RVY base - even if it overrides the AUIPC shift - but is that why you want the shift in the base? Maybe that's the confusion. You want enough in RVY so that CHERIoT doesn't need to define anything new for execution of the base ISA.

Are you in the meeting tomorrow?  I've missed many thursdays but will be present tomorrow.

tariqkurd-repo: if you want _all_ known parameters in the RVY base - then it can go there - but that doesn't scale as future encoding formats may want extra parameters which aren't in the base.

At that point we'll have to either extend the base somehow - which I guess will need to be an extension - or declare one or more new encoding formats which use the new parameter without altering the base.

arichardson: To clarify here, the RVY base ISA does not depend on the parameters, but any _L parameterization is allowed to specialize the behavior. So in theory a toolchain can target RVY without any _L (just less efficiently).

tariqkurd-repo: but a toolchain would need to know about the AUIPC shift..... surely?maybe I'm misunderstanding.

nwf: A _debugger_ can always disassemble `AUIPC n`, but it may not be able to hint at, or otherwise understand, its semantics without knowing the shift.  (If you are lucky enough to have relocation information attached to that `AUIPC`, it might not _need_ to understand it beyond "this is an `AUIPC` subject to relocation".)

The _compiler_ (and maybe assembler) can always (inefficiently!) articulate the 12-bit shift semantics of `AUIPC %0, n; ADDIY %0, %0, m` in a way that does not depend on the AUIPC shift and guarantees not going OOB -- that is, oblivious to representation concerns -- by using a temporary register (`%1`) and emitting (for positive `n`, without loss of generality; the negative case is similar):
```
AUIPC %0, 0

  ; if m < 0, step 2Ki shorter than AUIPC would have taken us
LUI %1, (m >= 0) ? n : (n - 1)  ; other sequences possible for small n.
ADDY %0, %0, %1                 ; this gets us "most of" the way to (n << 12) + m.

  ; if m < 0, reach upwards from 2 Ki shy of where 12-bit AUIPC would have taken us
ADDIY %0, %0, ((m >= 0) ? m : (0x1000 - m))
```
Proof sketch: Assume `n > 0` throughout.  `AUIPC n` is used only when the cap in `pc` has at least `(n - 1) * 2 KiB` of in-bounds addresses above the current offset, because the expected counterparts of `AUIPC` (`ADDI` and loads and stores) use _signed_ 12-bit displacements.  That is, `AUIPC n` may "overshoot" the intended target by up to `2 KiB`, the furthest negative reach of these displacements.  Put differently, in the integer base, `AUIPC %0, n; ADDI %0, %0, m` leaves `%0` displaced from `pc` by at minimum `n * 2 Ki - 2 Ki` or `(n - 1) * 2 Ki` (when `m = 0x800`) and at most `(n - 1) * 2 Ki + (2 Ki - 1)` or `n * 2 Ki - 1` (when `m = 0x7FF`).  To reach the same set of targets while remaining in bounds, we need to step _in the same direction_: one, `LUI (n - 1); ADDY`, that gets us `(n - 1) * 2 KiB` of the way, and the other that gets us the last bit of the way there.

In practice, of course, the toolchain may be able to fold the last `ADDYI` into the counterpart/subsequent instructions, and may be able to reuse the intermediate result (from `ADDY`) just like existing toolchains.  (CHERIoT making its `AUIPC` shift `11` means that we can always guarantee that the first step is in bounds, and our toolchain picks the correspondingly different immediates to its `AUIPC` and counterpart instructions.)

arichardson: Thanks, this is a really nice proof that it can actually work if you are willing to create less efficient code and RVY as a base is independent of the underlying encoding even with AUIPC changing.

arichardson: I'll add this to a non-normative note.

src/cheri/rvy64-encoding.adoc
```diff
@@ -0,0 +1,600 @@
...
+|==============================================================================
+
+.{rvy64_uni_base_name} extension summary
+[#rvy64_uni_base_name_ext_summary,width="100%",options=header,]
+|==============================================================================
+| Extension                    | Comment
+| <<section_cheri_hybrid_ext>> | Compatible
+| <<section_zylevels1>>        | Compatible
+ifdef::cheri_standalone_spec[]
+| <<section_zyseal>>           | Not yet compatible (need to allocate encoding)
```
tariqkurd-repo: it will be compatible - so I think I'd just list this as "compatible"
Zys should be in this list - sentries are compatible too

src/cheri/rvy64-encoding.adoc
```diff
@@ -0,0 +1,600 @@
...
+That is, <<JALR_CHERI>> will, as directed, attempt to jump to any unsealed or sealed capability in any register regardless of which register comes to hold the sealed return pointer.
+
+ifdef::cheri_standalone_spec[]
+[NOTE]
+=====
+The permission encodings of <<rvy64_cap_description>> do not provide mappings for
+the <<section_zyseal>> extension's <<zyseal_se_perm>> or <<zyseal_us_perm>>.
+Thus, without further revision, the encodings of this chapter
+are incompatible with the <<section_zyseal>> extension.
+=====
```
tariqkurd-repo, line 150: but above it says "not yet compatible" with Zyseal - so I guess it should say "incompatible" above?

src/cheri/rvy64-encoding.adoc
```diff
@@ -0,0 +1,600 @@
...
+include::img/base-bound-dec.edn[]
+
+xref:top_bound_dec[xrefstyle=short] and xref:base_bound_dec[xrefstyle=short]
+include ranges which may not be present when the bounds are decoded:
+
+* If E = 0 the lower section does not exist.
+* If E+MW=XLEN then the top section is only the least signification bit of
+ c~t~ for the top bound, and top section doesn't exist for the bottom bound.
+* If E+MW>XLEN then neither top section exists, and so the bounds are calculated
+ with no depending on the address field _a_.
```
```suggestion
 with no dependency on the address field _a_.
```

src/cheri/rvy64-encoding.adoc
```diff
@@ -0,0 +1,600 @@
...
+| T        | zeros  | Top address bits
+| T~E~     | zeros  | Exponent bits
+| B        | zeros  | Base address bits
+| B~E~     | zeros  | Exponent bits
+| Address  | zeros  | Capability address
+| Reserved | zeros  | All reserved fields
+|==============================================================================
+
+^1^ Only present if `AP_MBit=1` and {cheri_default_ext_name} is implemented.
+^2^ Only present if `enableL8=1`.
```
```suggestion

^2^ Only present if `enableL8=1`.
```

src/cheri/rvy64-encoding.adoc
```diff
@@ -0,0 +1,600 @@
...
+[#section_infinite_cap_encoding]
+==== Infinite Capability Encoding
+
+This encoding is for an _Infinite_ capability value, which
+grants all permissions while its bounds also cover the whole address space.
+It includes <<x_perm>> and so includes the <<m_bit>> if {cheri_default_ext_name} is supported.
+This infinite capability is both a <<root-rx-cap>> and a <<root-rw-cap>> capability.
+
+// Added note here for CHERIoT which enforces W xor X.
+NOTE: This capability format has a single encoding for the _Infinite_ capability. The CHERIoT encoding format has separate roots for code and for data, and will be included in a future version of this specification.
```
tariqkurd-repo: I guess we don't need the CHERIoT reference here? Can we just say that other formats are free to define separate code and data roots? this is the generalise it - other vendors will make formats in the future.

src/cheri/rvy32-encoding.adoc
```diff
@@ -0,0 +1,337 @@
...
+
+.{rvy32_uni_base_name} Feature summary
+[#rvy32_uni_base_name_feature_summary,width="100%",options=header,]
+|==============================================================================
+| Feature                      | Comment
+| Representable region         | At least 1/4 of the capability size
+| Permission encodings         | Not all combinations can be represented
+|==============================================================================
+
+[#AP-field-encoding32, reftext="AP,M-field encoding"]
```
tariqkurd-repo:
```suggestion
[#AP-field-encoding32, reftext="M-bit encoding in the AP field"]
```

src/cheri/rvy32-encoding.adoc
```diff
@@ -0,0 +1,337 @@
...
+The permissions encoding is split into four quadrants.
+The quadrant is taken from bits [4:3] of the permissions encoding.
+The meaning for bits [2:0] are shown in <<cap_perms_encoding32>> for each quadrant.
+
+Quadrants 2 and 3 are arranged to implicitly grant future permissions which may be added with the existing allocated encodings.
+Quadrant 0 does the opposite -- the encodings are allocated _not_ to implicitly add future permissions, and so granting future permissions will require new encodings.
+Quadrant 1 encodes permissions for executable capabilities.
+
+The <<m_bit>> is encoded as bit zero of the <<AP-field-encoding32>> for the executable quadrant and only assigned meaning when the implementation supports {cheri_default_ext_name} (_and_ <<x_perm>> is set).
+<<<
```
tariqkurd-repo, line 97:
```suggestion

<<<
```

src/cheri/rvy32-encoding.adoc
```diff
@@ -0,0 +1,337 @@
+[#rvy32_cap_description, reftext="{rvy32_uni_base_name}"]
+== The {rvy32_uni_base_name} Capability Base Extension for {cheri_base32_ext_name}, Version 1.0
+
+This chapter describes an in-memory format and properties of a capability encoding intended for {cheri_base32_ext_name}.
+
```
tariqkurd-repo, line 9:
```suggestion
This format is heavily based upon the {cheri_base64_ext_name} format <<rvy64_cap_description,{rvy64_uni_base_name}>> with the following changes:

* The parameters used to set the width, or presence, of fields in the encoding are defined in <<rvy32_uni_base_name_param_summary>>.
* The architectural permissions (AP) field is compressed to save encoding space, and so rules are defined for removing permissions.


```

src/cheri/rvy32-encoding.adoc
```diff
@@ -0,0 +1,337 @@
...
+[#section_cap_encoding32]
+=== Capability Encoding
+
+ifdef::cheri_v9_annotations[]
+NOTE: *CHERI v9 Note:* The encoding changes eliminate the concept of the in-memory format, and also increase precision for RV32.
+endif::[]
+
+The components of a capability, except the {ctag}, are encoded as shown in xref:cap_encoding_xlen32[xrefstyle=short].
+Each memory location or register able to hold a capability must also store the {ctag} as _out of band_ or _hidden_ information that software cannot directly set or clear.
```
tariqkurd-repo: it feels like this text should be common to all formats, along with the minimal set of permissions, and the concept of encoding memory bounds, which will define the minimum set of fields that every capability format must have. I presume, but I don't know, that CHERIoT bounds encoding is different, so the actual details can be in a specific encoding chapter.

I don't think we'd want to prescribe a specific bounds format - but to put some basic requirements on there being one.

src/cheri/insns/auipc_32bit.adoc
nwf: I've simplified the expansion and fixed a terribly embarrassing typo therein.

src/cheri/rvy32-encoding.adoc
tariqkurd-repo: I've removed this text - it's in the intro chapter

src/cheri/insns/auipc_32bit.adoc
arichardson: Add a short note to AUIPC and the text from you previous commented as commented out text (since it is probably not necessary for most readers, but still useful to have written down somewhere).

src/cheri/rvy64-encoding.adoc
```diff
@@ -0,0 +1,593 @@
...
+==== Capability Encoding Summary
+
+.{rv64y_uni_base_name} parameter summary
+[#rvy64_uni_base_name_param_summary,width="100%",options=header,]
+|==============================================================================
+| Parameter | Value                | Comment
+| MW        | {cap_rv64_mw_width}  | Mantissa width
+| EW        | {cap_rv64_exp_width} | Exponent width
+| CAP_MAX_E | {cap_rv64_exp_max}   | Maximum exponent value
+| enableL8  | 0                    | Whether the encoding format includes the L~8~ bit
```
arichardson, line 49:
```suggestion
| enableLMSB  | 0                    | Whether the encoding format includes the L~MSB~ bit
```

We should probably rename this to not hardcode the bit position? But that can be done after this PR.

src/cheri/rvy32-encoding.adoc
arichardson: Dropped this from the latest version since the AUIPC instruction listing has this information.

src/cheri/attributes.adoc
arichardson: Reminder to self: this one is still open and I will address it tomorrow.

src/cheri/riscv-unpriv-integration.adoc
arichardson: Yeah I imagine we could have RV32E and RV32I profiles? I'll mark this as resolved and we can discuss the profile plans separately.

src/cheri/attributes.adoc
arichardson: I've added some wording now that defines and `enc<N>` suffix with `enc1` being implicit and `enc0` as a way to say "don't make any assumptions". Not sure this is better than two parameters or if I'm just too tired to come up with sensible solutions. Will give this another think when I wake up tomorrow.

src/cheri/rvy32-encoding.adoc
```diff
@@ -0,0 +1,335 @@
...
+[#rvy32_uni_base_name_ext_summary,width="100%",options=header,]
+|==============================================================================
+| Extension                    | Comment
+| <<section_cheri_hybrid_ext>> | Compatible
+| <<section_zylevels1>>        | Compatible
+| <<section_rvy_sentry_insn_ext>> | Compatible
+ifdef::cheri_standalone_spec[]
+| <<section_zyseal>>           | Will be compatible once new permissions are encoded
+endif::[]
+| All RVY versions of other standard extensions | Compatible if the extension is compatible with RV32I
```
tariqkurd-repo:
```suggestion
| All RV32Y versions of other standard extensions | Compatible
```
I wonder why the extra words were there? Surely this compatible with all RV32Y versions of standard extensions by definition?

src/cheri/rvy64-encoding.adoc
```diff
@@ -0,0 +1,593 @@
...
+[#rvy64_uni_base_name_ext_summary,width="100%",options=header,]
+|==============================================================================
+| Extension                    | Comment
+| <<section_cheri_hybrid_ext>> | Compatible
+| <<section_zylevels1>>        | Compatible
+| <<section_rvy_sentry_insn_ext>> | Compatible
+ifdef::cheri_standalone_spec[]
+| <<section_zyseal>>           | Will be compatible once new permissions are encoded
+endif::[]
+| All RVY versions of other standard extensions | Compatible if the extension is compatible with RV64I
```
```suggestion
| All RV64Y versions of other standard extensions | Compatible
```
I wonder why the extra words were there? Surely this compatible with all RV64Y versions of standard extensions by definition?

tariqkurd-repo (approved):

src/cheri/rvy64-encoding.adoc
tariqkurd-repo: I tried this in  src/cheri/img/cap-encoding-xlen32.edn but could't get L-subscript-MSB-/subscript- to render.
Maybe not a big problem..... let's get this one merged and worry about it later

arichardson: Yeah let's do that after

tariqkurd-repo: I think we're ready to merge then!

arichardson:
@nwf Could you take another look at this PR? I hope I've addressed the open comments.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -459,6 +511,105 @@ Extensions introducing new permissions may require these to be provided by root
...
+
+NOTE: The base `_L<encoding_parameters>` encoding format scheme is designed to separate the aspects of capability encoding that influence the behavior of {cheri_base_ext_name} instructions from the rest.
+The intent is to simplify the future development of more (possibly domain-specific) encoding formats while also more directly conveying compatibility of architectural bases.
+
+The following parameters are currently defined to distinguish existing formats:
+
+Parameters for bounds precision and representability::
+These parameters affect the behavior of <<SCBNDS>>/<<SCBNDSR>> as well as the representability check that affects the tag-clearing behavior of how far you can go out of bounds with <<SCADDR>>/<<CADD>>.
+* **Mantissa width (mw<N>)**: The mantissa width for the bounds encoding. For example, a format with {cap_rv64_mw_width} bits of mantissa, would use `mw{cap_rv64_mw_width}`.
+* **Maximum exponent (ea or e<N>)**: If _all_ required values of the exponent can be encoded directly in the exponent field this uses `ea`, otherwise `e<N>` describes the exponent for a valid capability spanning the whole address space (e.g. `e15` for the CHERIoTv1 format).
```
rmn30: I'd express this as the maximum supported exponent except for the one required to encode the full address space, which must always be supported. For CHERIoT this would be e14.

rmn30: The reason being I think it's more useful to software to know the point beyond which you start to lose precision rather than how to encode the maximum bounds.

rmn30: Also I think you should define 'required' as otherwise it's a bit of a truism.

arichardson: Thanks that sounds good, will adjust the wording and use 14 for cheriot

PR #846, opened 2025-11-13
Zyhybrid tweaks

nwf:
- The example code for observing the execution mode was, AFAICT, backwards.
- The M field is logically not a permission but is X-permission-dependent (see the "default" RV32 encoding), so reflect that in the text.

### DIFF
--- a/src/cheri/riscv-hybrid-integration.adoc
+++ b/src/cheri/riscv-hybrid-integration.adoc
@@ -66,7 +66,7 @@ The _<<cheri_execution_mode>>_ is determined by a bit in the metadata of the <<p
 This field _need_ be present only in capabilities granting <<X-permission>>,
 as it is only ever architecturally interpreted on the capability resident in <<pcc>>.
 Capabilities not granting <<X-permission>> may or may not have a defined M field,
-and attempting to update this field may or may not clear their {ctag}s.
+and attempting to update this field may be a no-op.
 The exact location of the M-bit in the capability format for XLEN=32 and XLEN=64 is described in <<app_cap_description>>.
 
 * Mode (M)={CAP_MODE_VALUE} indicates {cheri_cap_mode_name}.


src/cheri/riscv-hybrid-integration.adoc
```diff
@@ -116,7 +118,8 @@ in their representation of architectural CHERI capabilities.
 
 The effective CHERI execution mode cannot be determined just by reading the <<m_bit>> from <<pcc>> since it also depends on the execution environment.
 The following code sequence demonstrates how a program can observe the current, effective CHERI execution mode.
-It will write {CAP_MODE_VALUE} for {cheri_cap_mode_name} and {INT_MODE_VALUE} for {cheri_int_mode_name} to `x1`:
+It will write, to `x1`, the value `1` for {cheri_cap_mode_name} (wherein `pc` has a set {ctag})
```
tariqkurd-repo, line 121: why stop using the variables for the values?

src/cheri/riscv-hybrid-integration.adoc
```diff
@@ -63,14 +63,16 @@ requires the CHERI RISC-V hart to support {cheri_base_ext_name} and
 
 The _<<cheri_execution_mode>>_ is determined by a bit in the metadata of the <<pcc>> called the <<m_bit>>.
 {cheri_default_ext_name} adds a new _<<cheri_execution_mode>>_ field (M) to the capability format.
-Although always present, it only takes effect in code capabilities, i.e., when the <<x_perm>> is set.
+This field _need_ be present only in capabilities granting <<X-permission>>,
+as it is only ever architecturally interpreted on the capability resident in <<pcc>>.
+Capabilities not granting <<X-permission>> may or may not have a defined M field,
+and attempting to update this field may or may not clear their {ctag}s.
```
```suggestion
and attempting to update this field may be a no-op.
```
I believe the only way to update it in the ISA is YMODEW and that currently says:
>If rs1 grants X-permission and rs1 's AP-field could have been produced by YPERMC, then update the M-bit of rd to ... Otherwise do not update the M-bit

Or maybe just drop the ", and..." part of this sentence?

src/cheri/riscv-hybrid-integration.adoc
```diff
@@ -63,14 +63,16 @@ requires the CHERI RISC-V hart to support {cheri_base_ext_name} and
...
 The exact location of the M-bit in the capability format for XLEN=32 and XLEN=64 is described in <<app_cap_description>>.
 
 * Mode (M)={CAP_MODE_VALUE} indicates {cheri_cap_mode_name}.
 * Mode (M)={INT_MODE_VALUE} indicates {cheri_int_mode_name}.
```
arichardson, line 73: Should I push this information into the format chapter in #845 ?

src/cheri/riscv-hybrid-integration.adoc
arichardson: Because the existing text was wrong after we changed the execution mode values. This is about the tag of the resulting capability not how the mode is encoded in YMODER or the raw in-memory format. Right now it expands to writes "0" to x1 for cap mode, but it will have a valid tag so should be "1"

Good catch! I completely missed this since I mostly looked at the ASCIIdoc source rather than the PDF.

src/cheri/riscv-hybrid-integration.adoc
nwf: I read these as being the "abstract" values of the M field as returned by the `YMODER` getter and sent to the `YMODEW` setter, so I think these lines can stay here?

src/cheri/riscv-hybrid-integration.adoc
nwf: Well, `YMODESW[IY]` also mutate the bit in `pc`... but that definitely means that it's got X-permission, and it can only be read back via `auipc` if it's `1` (hm)... Aaaanyway.  Phrasing it as "may be a no-op" is fine... that probably should say "update this field therein" or something, tho'.

BUT!  I didn't notice the "The value of rs1.tag does not affect the result" over in `YMODEW`.  I... guess it's fine to have a strange bag of bits mutator instruction, but I'd feel more comfortable with it being a no-op on untagged `rs1`?

src/cheri/riscv-hybrid-integration.adoc
arichardson: Yeah it happens to be both. I'll add a note saying something like "the architectural view of the mode" just to be clear when I get back to updating #845 .

src/cheri/riscv-hybrid-integration.adoc
arichardson: Hmm I feel like we had a discussion about this a long time ago and we decided there is no need to include the extra AND with the tag when you give it an input that you can't really meaningfully use. 
Additionally, allowing it to behave "as if valid" for untagged could potentially allow it to be used with YBLD (not that I see that as being something particularly useful).

tariqkurd-repo: I'd prefer not to change any behaviour at the point unless seriously wrong - we're trying to show that the spec is stable enough to be frozen - so would really like to keep this unchanged.

tariqkurd-repo (approved):

src/cheri/riscv-hybrid-integration.adoc
tariqkurd-repo: yes - good point - the capability encoding doesn't _have to_ encode the M-bit that way

src/cheri/riscv-hybrid-integration.adoc
arichardson: Are you okay with me applying the suggestion? Then I'll go ahead and merge?

nwf: Go for it.

PR #858, opened 2025-11-25
First draft of CHERIoT encoding chapters.

rmn30:


src/cheri/rvycheriot-encoding.adoc
```diff
@@ -0,0 +1,359 @@
...
+| p         | pc                   | Compressed permission encoding
+| as        | 11                   | <<AUIPC>> shift
+|==============================================================================
+
+.{rvy32e_cheriot_base_name} extension summary
+[#rvy32e_cheriot_base_name_ext_summary,width="100%",options=header,]
+|==============================================================================
+| Extension                    | Comment
+| <<section_cheri_hybrid_ext>> | Not supported
+| <<section_zylevels1>>        | Compatible
```
arichardson: I guess we want to note here that Zys is not compatible?

src/cheri/rvycheriot3-encoding.adoc
```diff
@@ -0,0 +1,51 @@
...
+
+.{rvy32e_cheriot2_base_name} (Alternative Version) parameter summary
+[#rvy32e_cheriot3_base_name_param_summary,%autowidth,options=header,]
+|==============================================================================
+| Parameter | Value                | Comment
+| mw        | {cap_cheriot_mw_width}  | Mantissa width
+| e         | a                    | All exponents are supported
+| rc        | 0                    | Representable region between base and top only
+| p         | pc                   | Compressed permission encoding
+| as        | 11                   | <<AUIPC>> shift
```
tariqkurd-repo:
```suggestion
| as        | 11                   | <<AUIPC_CHERI>> shift
```

src/cheri/rvycheriot2-encoding.adoc
```diff
@@ -0,0 +1,80 @@
...
+
+.{rvy32e_cheriot_base_name} parameter summary
+[#rvy32e_cheriot_base_name_param_summary,%autowidth,options=header,]
+|==============================================================================
+| Parameter | Value                | Comment
+| mw        | {cap_cheriot_mw_width}  | Mantissa width
+| e         | a                    | All exponents are supported
+| rc        | 0                    | Representable region between base and top only
+| p         | pc                   | Compressed permission encoding
+| as        | 11                   | <<AUIPC>> shift
```
tariqkurd-repo:
```suggestion
| as        | 11                   | <<AUIPC_CHERI>> shift
```

src/cheri/rvycheriot-encoding.adoc
```diff
@@ -0,0 +1,359 @@
...
+
+.{rvy32e_cheriot_base_name} parameter summary
+[#rvy32e_cheriot_base_name_param_summary,width="100%",options=header,]
+|==============================================================================
+| Parameter | Value                | Comment
+| mw        | {cap_cheriot_mw_width}  | Mantissa width
+| e         | 14                   | Exponent limit before jumping to maximum exponent
+| rc        | 0                    | Representable region between base and top only
+| p         | pc                   | Compressed permission encoding
+| as        | 11                   | <<AUIPC>> shift
```
tariqkurd-repo:
```suggestion
| as        | 11                   | <<AUIPC_SHIFT>> shift
```

src/cheri/rvycheriot-encoding.adoc
tariqkurd-repo: and something about Zybld

src/cheri/rvycheriot-encoding.adoc
```diff
@@ -0,0 +1,359 @@
...
+.{rvy32e_cheriot_base_name} extension summary
+[#rvy32e_cheriot_base_name_ext_summary,width="100%",options=header,]
+|==============================================================================
+| Extension                    | Comment
+| <<section_cheri_hybrid_ext>> | Not supported
+| <<section_zylevels1>>        | Compatible
+ifdef::cheri_standalone_spec[]
+| <<section_zyseal>>           | Compatible
+endif::[]
+| All RVY versions of other standard extensions | Compatible if the extension is compatible with RV32I
```
tariqkurd-repo:
```suggestion
| All RVY versions of other standard extensions | Compatible if the extension is compatible with RV32E
```

this format is explicitly targeting RV32E so this seems better?

src/cheri/rvy-cheriot-enc1.adoc
```diff
@@ -0,0 +1,359 @@
...
+The encoding impacts the following CHERI extensions:
+
+<<section_cheri_hybrid_ext>>::
+This encoding does not support {cheri_default_ext_name}: there is no mode bit so only purecap mode is supported.
+
+<<section_zylevels1>>::
+This encoding supports <<section_zylevels1>>.
+The <<zylevels1_gl_perm,GL>> flag is encoded in the G bit.
+
+<<section_zyseal>>::
```
tariqkurd-repo, line 66: zyseal is excluded from the unpriv spec build even though this chapter is included, so need to handle them both in the same way

src/cheri/rvycheriot-encoding.adoc
```diff
@@ -0,0 +1,359 @@
...
+
+NOTE: The reserved bit( R ) must be 0 in valid capabilities and is available for future extensions.
+
+[#cheriot_encoding_field_summary, reftext="CHERIoT capability fields"]
+[width="100%",options=header,cols="1,5"]
+|==============================================================================
+| Field      | Description
+| R          | Reserved bit, must be zero in valid capabilities and preserved in any value loaded from memory and stored.
+| G          | Global bit, as in <<zylevels1_gl_perm>>.
+| P          | Architectural Permissions, see <<cheriot-permission-encoding>>.
```
tariqkurd-repo:
```suggestion
| AP          | Architectural Permissions, see <<cheriot-permission-encoding>>.
```

for consistency with the main spec?

src/cheri/rvycheriot-encoding.adoc
```diff
@@ -0,0 +1,359 @@
...
+|===
+
+For example, if P[4:3] = 01 and P[2:0] = 101 then the decoded permissions are R, C, X, LG, and ASR.
+Note that there is no encoding for an "Infinite" capability with all permissions.
+In particular there is no overlap between the sealing permissions (SDP0, SE, US) and any other permissions and X is mutually exclusive with W.
+This means there are three <<root-cap>> capabilities from which all other capabilities can be derived using <<CLRPERM>>. These are shown in <<cheriot_root_caps>>, along with the CSRs that will contain this root on reset.
+
+[#cheriot_root_caps,width="100%",options=header,cols="1,1,2,2",align="center",reftext="CHERIoT root capabilities"]
+|===
+| Root name  | P[4:0] | Permissions          | Reset location
```
tariqkurd-repo: can't talk about reset in an unpriv spec! needs separating out

src/cheri/rvycheriot-encoding.adoc
rmn30: This is an interesting one. There's nothing preventing CHERIoT from working with RV32I we just chose E to reduce context size.

src/cheri/rvycheriot-encoding.adoc
rmn30: I think Zybld should mention sealed `rs2`. Specifically I think CHERIoT could be compatible with it provided it explicitly prevents setting the tag on a capability sealed with a non-ambient type.

rmn30: (CHERIoT doesn't define any ambient types, hence why it is incompatible  with Zys).

src/cheri/rvy-cheriot-enc1.adoc
```diff
@@ -0,0 +1,359 @@
...
+| 010      | 1 | 2   | Interrupt disabling forward sentry
+| 011      | 1 | 3   | Interrupt enabling forward sentry
+| 100      | 1 | 4   | Interrupt disabling return sentry
+| 101      | 1 | 5   | Interrupt enabling return sentry
+| 110      | 1 | 6   | Sealed with type 6
+| 111      | 1 | 7   | Sealed with type 7
+| 000      | 0 | 0   | Unsealed
+| 1 .. 7   | 0 | 9 .. 15 | Data capability sealed with given type
+|===
+
```
tariqkurd-repo, line 207: presumably it's intentional that 8 can't be encoded and 0 appears twice

src/cheri/rvycheriot-encoding.adoc
```diff
@@ -0,0 +1,359 @@
...
+[%autowidth]
+|===
+| mstatus.MIE | link register | Return capability sealing type
+
+| any         | not ra        | 1 (standard sentry)
+| 0           | ra            | 4 (interrupt disabling return sentry)
+| 1           | ra            | 5 (interrupt enabling return sentry)
+|===
+
+* <<JALR_CHERI>> sets `mstatus.MIE` to 0 when jumping to a type 2 or 4 sentry, and sets it to 1 when jumping to a type 3 or 5 sentry. Other jumps leave it unchanged.
```
tariqkurd-repo: can't talk about mstatus.MIE in unpriv, and definitely not change it
do you intend to support other user and supervisor mode with cheriot or restrict it to M-mode only?
we need a strategy here.... it really doesn't fit with the current RISC-V model

src/cheri/rvycheriot-encoding.adoc
```diff
@@ -0,0 +1,359 @@
...
+| 011      | 1 | 3   | Interrupt enabling forward sentry
+| 100      | 1 | 4   | Interrupt disabling return sentry
+| 101      | 1 | 5   | Interrupt enabling return sentry
+| 110      | 1 | 6   | Sealed with type 6
+| 111      | 1 | 7   | Sealed with type 7
+| 000      | 0 | 0   | Unsealed
+| 1 .. 7   | 0 | 9 .. 15 | Data capability sealed with given type
+|===
+
+In addition to the ordinary <<sentry_cap>> type this encoding defines four types of sentry for enforcing structured control flow and atomically changing the value of `mstatus.IE` during a <<JALR_CHERI>>.
```
tariqkurd-repo:
```suggestion
In addition to the ordinary <<sentry_cap>> type this encoding defines four types of sentry for enforcing structured control flow and atomically changing the value of `mstatus.MIE` during a <<JALR_CHERI>>.
```

src/cheri/rvycheriot-encoding.adoc
tariqkurd-repo: let's not change Zybld, so I'd recommend explictly listing both as incompatible

src/cheri/rvycheriot-encoding.adoc
tariqkurd-repo: the encoding format name is prefixed with RV32YE - so it needs to be consistent

src/cheri/rvycheriot-encoding.adoc
tariqkurd-repo: the absolute minimum would be to say "interactions with mstatus.MIE only happen if executing in M-mode" but that wil need to be in the priv spec not here ☹️ unfortunately

tariqkurd-repo: I think you'll need another version of JALR with an M-prefix to do this in a standard way, so more like MRET

src/cheri/rvycheriot-encoding.adoc
```diff
@@ -0,0 +1,359 @@
...
+// encode E
+E = (e == 24) ? 15 : e
+// truncate B and T to 9 bits
+B = b[8 : 0]
+T = t[8 : 0]
+----
+
+===== <<AUIPC_CHERI>> Representability
+
+Due to the limited representable range of this encoding there can be a problem when using <<AUIPC_CHERI>> to form addresses relative to <<PCC>>. For example, the following sequence might used to performa a pc relative relocation by an offset [imm20|imm12]:
```
tariqkurd-repo:
```suggestion
Due to the limited representable range of this encoding there can be a problem when using <<AUIPC_CHERI>> to form addresses relative to <<pcc>>. For example, the following sequence might used to perform a pc relative relocation by an offset [imm20|imm12]:
```

src/cheri/rvycheriot-encoding.adoc
nwf: https://github.com/riscv/riscv-cheri/blob/f052668635636276d738f3d2a0b44d1a812b50c8/src/cheri/insns/cbld_32bit.adoc?plain=1#L42-L46 is already present in the `YBLD` documentation, albeit in the wrong place (at least half of it should be up in the procedural description).

src/cheri/rvycheriot-encoding.adoc
nwf: Right, so: there's _absolutely nothing_ about the CHERIoT capability encoding that means it is restricted to an E machine.  (What would that even mean?  Registers are not something capabilities understand.)  I understand that a particular base is, of course, either `I` or `E`, but...

src/cheri/rvycheriot-encoding.adoc
nwf: It holds both what the other encodings call `AP` and `SDP`.  If the main spec is making reference to those fields, it should probably be corrected.

src/cheri/rvy-cheriot-enc1.adoc
nwf: Yes.  See also https://github.com/riscv/riscv-cheri/blob/f052668635636276d738f3d2a0b44d1a812b50c8/src/cheri/xycheriot.adoc?plain=1#L317-L343

src/cheri/rvycheriot-encoding.adoc
nwf: CHERIoT (the ISA, not the encoding per se) is indeed intended for M-only RISC-V machines.  It's probably fine to phrase this as `xstatus.XIE` being updated.

src/cheri/rvycheriot-encoding.adoc
davidchisnall: We are likely to ship (heterogenous) cores with *both* Y and YE bases and the same encoding on our next product, with a CHERIoT encoding.  How should we express this in the standard?

src/cheri/rvycheriot-encoding.adoc
davidchisnall: This is some nonsense in the way the upstream RISC-V specs are written, because M mode is the only *required* mode, yet the base spec doesn't let you talk about it.

Note that this is the `jalr` that is used on *every* call and return in CHERIoT, so it would be silly to make it a different instruction, even nominally.  There is no reason to restrict this behaviour to M mode, since you can always avoid handing out capabilities to type 2 or 4 sentries to unprivileged code, if you decide you do want to introduce S/U modes in a CHERIoT system.

src/cheri/rvycheriot-encoding.adoc
rmn30: It wouldn't be an actual change to the instruction for other capability encodings, just some wording that says "Clear the tag if rs2 is sealed with a non-ambient type".

src/cheri/rvycheriot-encoding.adoc
```diff
@@ -0,0 +1,359 @@
...
+
+If the sentry is not of the correct type then the {ctag} is cleared on the target <<pcc>>, leading to a trap.
+The purpose of this is to enable calls to short atomic sequences on single-core microcontrollers without granting the right to modify `mstatus.MIE` arbitrarily.
+These rules ensure that functions that are called with a new interrupt stance can return to the caller's stance as well as providing a degree of control flow integrity.
+Non-standard control flow transfers such as tail calls and calls to compiler outlined functions using a non-standard link register are still possible.
+
+[#section_cheriot_cap_bounds, reftext="CHERIoT Capability Bounds Encoding"]
+==== Bounds (E, B, T) Encoding
+
+The bounds are encoded in similar, but slightly simplified way to <<section_cap_bounds,{rv64y_uni_base_name}>>, with field B and T being substitued into the address at the offset given by the exponent to form the base and top.
```
tariqkurd-repo:
```suggestion
The bounds are encoded in similar, but slightly simplified way to <<section_cap_bounds,{rv64y_uni_base_name}>>, with field B and T being substituted into the address at the offset given by the exponent to form the base and top.
```

src/cheri/rvy-cheriot-enc1.adoc
```diff
@@ -0,0 +1,359 @@
...
+[#cheriot_base_bound_dec]
+include::img/cheriot-base-bound-dec.edn[]
+
+where the corrections, c~t~ and c~b~, are given by:
+
+[source]
+----
+A_hi = (A[e + 8 : e]) < B ? 1 : 0
+T_hi = (T < B) ? 1 : 0
+c_b = -A_hi
```
tariqkurd-repo, line 252: cb in the diagram above not c_b?

src/cheri/rvy-cheriot-enc1.adoc
```diff
@@ -0,0 +1,359 @@
...
+For a given exponent the base and top are then computed as:
+
+.Decoding of the XLEN+1 wide top (_t_) bound
+[#cheriot_top_bound_dec]
+include::img/cheriot-top-bound-dec.edn[]
+
+.Decoding of the XLEN wide base (_b_) bound
+[#cheriot_base_bound_dec]
+include::img/cheriot-base-bound-dec.edn[]
+
```
tariqkurd-repo, line 245: as e=24 gives e+8=33, maybe a statement that the top section doesn't exist in this case?

src/cheri/rvycheriot-encoding.adoc
```diff
@@ -0,0 +1,359 @@
...
+----
+auipc    rd1, imm20
+addiy    rd2, rd1, imm12
+----
+
+In this case it is possible that imm20 is negative and imm12 is positive such that the resulting address in rd2 is inside the bounds of <<pcc>> but the intermediate address in rd1 is below the base and hence unrepresentable.
+This would cause the {ctag} of rd1 to be cleared, leading to a trap when rd2 is used as a capability.
+
+To avoid this problem the the shift applied to imm20 in <<AUIPC_CHERI>> is reduced to 11 bits rather than 12, so that is possible to ensure imm12 and imm20 have the same sign, keeping the intermediate address in rd1 within the representable region.
+This limits the range of pc-relative offsets that can be formed using this sequence to ±1 GiB, but this is not a problem in practice as <pcc> will have much smaller bounds than this in real systems targeted by this encoding.
```
tariqkurd-repo:
```suggestion
This limits the range of pc-relative offsets that can be formed using this sequence to ±1 GiB, but this is not a problem in practice as <<pcc>> will have much smaller bounds than this in real systems targeted by this encoding.
```

src/cheri/rvycheriot-encoding.adoc
tariqkurd-repo: the main spec doesn't combine those two - I think it's worth adding a note saying what @nwf just said in the comment above

src/cheri/rvycheriot-encoding.adoc
tariqkurd-repo: ok - which doesn't affect the university encoding behaviour - so let's fix the ybld page?

src/cheri/rvycheriot-encoding.adoc
tariqkurd-repo: I would say that you should change the prefix to RV32Y from RV32YE and just note that this family of encoding formats is recommended for RV32YE systems, and leave it there.

src/cheri/rvycheriot-encoding.adoc
tariqkurd-repo: Extending this spec for S and U modes is out of scope for now.
Just add an assumption that JALR is executing in M-mode somewhere visible - and we'll call it good enough for now - and get some feedback.

rmn30: CHERIoT currently only supports M-mode.

tariqkurd-repo: yes - so we need to bake that into the spec - so it's unambiguous - and leave support for S and U modes as future work.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -578,11 +578,11 @@ NOTE: It is expected that the actual {cheri_base_ext_name} instantiation is chos
 For {cheri_base64_ext_name}, the number of spare bits in the capability encoding ensure sufficient future extensibility and therefore no vendor-defined formats exist.
 However, {cheri_base32_ext_name} has insufficient bits available to enable features for all target domains, so in addition to the <<rv32y_cap_description, {rv32y_uni_base_name}>> further {cheri_base32_ext_name}-extending capability formats exist:
 
-* The CHERIoT encoding optimized for smaller devices: {rvy32e_cheriot_base_name}.
+* The CHERIoT encoding optimized for smaller devices: {rvy32_cheriot_base_name}.
```
arichardson: I guess we should also link the other two formats here

src/cheri/rvycheriot-encoding.adoc
```diff
@@ -0,0 +1,367 @@
...
+[#section_cap_encoding_cheriot]
+=== Capability Encoding
+
+The components of a capability, except the {ctag}, are encoded as shown in xref:cap_encoding_cheriot[xrefstyle=short].
+
+.Capability encoding for {rvy32_cheriot_base_name}
+[#cap_encoding_cheriot]
+include::img/cap-encoding-cheriot.edn[]
+
+NOTE: The reserved bit( R ) must be 0 in valid capabilities and is available for future extensions.
```
arichardson:
```suggestion
NOTE: The reserved bit ( R) must be 0 in valid capabilities and is available for future extensions.
```

src/cheri/rvycheriot-encoding.adoc
arichardson: I think for now we should just leave this in here, and then deal with it later. There are a few priv references in the unpriv spec already.

src/cheri/rvycheriot-encoding.adoc
rmn30: I just changed it to AP but I'm running out of time to make changes today.

src/cheri/rvycheriot-encoding.adoc
```diff
@@ -0,0 +1,369 @@
+[#rv32y_cheriot_cap_description, reftext="{rvy32_cheriot_base_name} (CHERIoT V1)"]
+== The {rvy32_cheriot_base_name} Capability Base Extension for {cheri_base32_ext_name}, (CHERIoT V1)
+
+This chapter describes an in-memory format and properties of a capability encoding intended for {cheri_base32_ext_name}.
+This is the encoding used for CHERIoT, which is an RV32Y variant adapted for microcontrollers.
+Compared to <<rv32y_cap_description>> it has some additional features for building rich compartmentalization, such as extra permissions and sealing types, and some differences in the bounds encoding that trade some software compatibility for more precise bounds on small sizes and reduced microarchitectural complexity.
+
```
tariqkurd-repo, line 23:
```suggestion
NOTE: This encoding format assumes support for M-mode only. Future versions of this encoding format may support additional privilege modes.
```

src/cheri/rvycheriot-encoding.adoc
tariqkurd-repo: Yes - but this is different - I've put a suggestion above to resolve it for now saying that it assumes M-mode only.

src/cheri/rvycheriot-encoding.adoc
tariqkurd-repo: try this: https://github.com/riscv/riscv-cheri/pull/862

src/cheri/rvycheriot-encoding.adoc
nwf: I wish, again, to draw a very clear distinction between the ISA and the encoding.  The encoding simply spells out a way of storing 15 different otypes; the interpretation of those otypes is part of the ISA.  If the latter needs to be part of a "Xycheriot" extension, then so be it.

tariqkurd-repo: https://github.com/riscv/riscv-cheri/blob/5594edf4f7b9e96fcc49f04d687ac8a906acc693/src/cheri/insns/jalr_32bit.adoc?plain=1#L53

This should spell out exactly what you need then on the ISA page, to avoid redefining things for CHERIoT, so should be normative.

Xycheriot shouldn't be something we submit to ARC - if it's non-standard they won't be interested.

The encoding format can redefine things further if need be as it's a base - but ideally wouldn't need to, better to have everything in the ISA definition if we can.

davidchisnall: I would hope that in this chapter we just specify that the otypes differ based on X permission and then leave it for an extension to define what the otypes for X capabilities mean?

tariqkurd-repo: You can certainly do that - have the special sentry behaviour in an extension which we don't make visible yet.
To be honest, I think that not exposing the connection between unpriv instruction and mstatus will help us move forwards quicker, and it's something that can be discussed and then standardised later.
Divide and conquer.
Once we have the base cheriot format(s) agreed with ARC we can discuss how to layer this behaviour on top.

src/cheri/attributes.adoc
```diff
@@ -83,6 +84,17 @@ endif::support_varxlen[]
 
 :cheri_seal_ext_name: Zyseal
 
+:cheriot_unpriv_ext_name: Zyucheriot
+:cheriot_priv_ext_name: Zyscheriot
```
```suggestion
:cheriot_priv_ext_name: Smycheriot
```
I think this is the current naming scheme for m-mode exts. But not sure where the X for vendor goes for privileged ones (Xsmy or Sxmy).

tariqkurd-repo: I think it should be Smy and ratified - if not then it doesn't need to be in the doc at all (it won't end up in the ISA manual if it's not ratified)

src/cheri/attributes.adoc
```diff
@@ -126,17 +138,27 @@ endif::support_varxlen[]
...
 :rv32ye_uni_base_name: {cheri_base32_ext_name}E{rv32y_uni_encoding_suffix}
-// CHERIoT encoding scheme
-:rvy_cheriot_encoding_suffix: _Lxymw9e14r0as11pc
-:rvy32e_cheriot_base_name: {cheri_base32_ext_name}E{rvy_cheriot_encoding_suffix}
 
+:rvy_cheriot_common_base_suffix: _Lymw{cap_cheriot_mw_width}e{cap_cheriot_exp_range_max}r0as11pc
+
+:rv32y_cheriot_encoding1_name: RV32Y_LyencCheriot1
+:rv32y_cheriot_encoding2_name: RV32Y_LyencCheriot2
+:rv32y_cheriot_encoding3_name: RV32Y_LyencCheriot3
```
```suggestion
:rv32y_cheriot_encoding1_name: RV32Y_Lyenccheriot1
:rv32y_cheriot_encoding2_name: RV32Y_Lyenccheriot2
:rv32y_cheriot_encoding3_name: RV32Y_Lyenccheriot3
```

I think the arch string expects all letters for extensions (other than the first one after _) to be lower-case.

src/cheri/riscv-cheri.adoc
```diff
@@ -48,8 +48,11 @@ include::zylevels1.adoc[]
 // include::zylevelsN.adoc[]
 
 include::zyseal.adoc[]
+include::zybndsrd.adoc[]
 include::rvy64-encoding.adoc[]
 include::rvy32-encoding.adoc[]
+include::rvy-cheriot-unpriv.adoc[]
+include::rvy-cheriot-isa-priv.adoc[]
```
arichardson: This line needs to go a few lines further down into the priv chapters section.

src/cheri/introduction.adoc
```diff
@@ -142,7 +142,6 @@ The extensions in this section have not been fully prototyped and so are not con
 |<<section_cheri_vector_integration>>                       | Vector
 |<<ZCMT_CHERI>>                                             | Table Jump for {cheri_base32_ext_name}
 |<<ZCMP_CHERI>>                                             | Push/pop and double move for {cheri_base32_ext_name}
-|<<section_xycheriot>>                                      | CHERIoT capability encoding
```
arichardson, line 145: I think we should probably keep the two cheriot extensions linked in this table, but maybe change the title of the table from "experimental", to "not for initial ratification package"?

src/cheri/rvy-cheriot-enc1.adoc
```diff
@@ -0,0 +1,364 @@
+[#rv32y_cheriot_encoding1_description, reftext="{rv32y_cheriot_encoding1_name}"]
+== The {rv32y_cheriot_encoding1_name} CHERI Capability Encoding Scheme
+
+This section describes an in-memory format and properties of a capability encoding intended for {cheri_base32_ext_name}.
+It is specifically designed to be a suitable substrate for {cheri_base32_ext_name}{cheriot_unpriv_ext_name} systems.
```
tariqkurd-repo:
```suggestion
It is specifically designed to be a suitable substrate for {cheri_base32_ext_name}_{cheriot_unpriv_ext_name} systems.
```

src/cheri/rvy-cheriot-enc1.adoc
```diff
@@ -0,0 +1,364 @@
...
+
+The first of these is achieved by:
+
+- Using two extra bits for the bounds encoding
+- Giving the exponent its own field instead of embedded it in the T and B fields
+- Saving space by reducing the exponent in size by one and dropping support for exponents in the range 15 to 23.
+This limits the precision for the bounds of large capabilities (>= 8MiB).
+- dropping guaranteed out-of-bounds representability (except "one past the end")
+
+These tradeoffs are aimed at microcontroller implementaions where precise bounds on small capabilities are important to save memory on padding, large capabiliites are unlikely and microarchitectural complexity should be minimised.
```
tariqkurd-repo:
```suggestion
These tradeoffs are aimed at microcontroller implementations where precise bounds on small capabilities are important to save memory on padding, large capabiliites are unlikely and microarchitectural complexity should be minimised.
```

src/cheri/rvy-cheriot-enc1.adoc
```diff
@@ -0,0 +1,364 @@
...
+|==============================================================================
+| Feature                      | Comment
+| Representable region         | Between base and top only (inclusive)
+| Permission encodings         | Not all combinations can be represented
+|==============================================================================
+
+[#cheriot-permission-encoding, reftext="CHERIoT permission encoding"]
+==== Permissions Encoding
+
+Similarly to {rv32y_uni_base_name}, capability permisisons are encoded in a compressed form.
```
tariqkurd-repo:
```suggestion
Similarly to <<rv32y_cap_description,{rv32y_uni_base_name}>>, capability permissions are encoded in a compressed form.
```

src/cheri/rvy-cheriot-enc1.adoc
```diff
@@ -0,0 +1,364 @@
...
+
+The permission encoding space is split into quadrants using P[3:2].
+Each quadrant may include some fixed permissions (indicated with ✔) and some dependent permissions encoded using P[2:0].
+Quadrant 0 is used to encode permissions that authorise sealing (see <<section_zyseal>>) and also the single <<sec_cap_sdp,software-defined permission>>, <<zycheriot_u0_perm,U0>>. It can also encode 'no permissions'.
+Quadrant 1 encodes executable capabilities along with the dependent permission <<asr_perm,ASR>> and optional permissions <<lm_perm,LM>> and <<zylevels1_lg_perm,LG>>.
+Quadrant 2 is subdivided into octants, with octant 4 encoding combinations of <<r_perm,R>> and <<w_perm,W>> without <<c_perm,C>>
+(with the redundant not-<<r_perm,R>> and not-<<w_perm,W>> used to encode write-only with <<c_perm,C>>).
+Octant 5 encodes read-only capabilities with <<c_perm,C>> and the dependent permissions <<lm_perm,LM>> and <<zylevels1_lg_perm,LG>>.
+Quadrant 3 encodes permissions with <<r_perm,R>>, <<w_perm,W>>, <<c_perm,C>> and the dependent permissions <<zylevels1_sl_perm,SL>>, <<lm_perm,LM>> and <<zylevels1_lg_perm,LG>>.
```
tariqkurd-repo:
```suggestion

* Quadrant 0 is used to encode permissions that authorise sealing (see <<section_zyseal>>) and also the single <<sec_cap_sdp,software-defined permission>>, <<zycheriot_u0_perm,U0>>. It can also encode 'no permissions'.
* Quadrant 1 encodes executable capabilities along with the dependent permission <<asr_perm,ASR>> and optional permissions <<lm_perm,LM>> and <<zylevels1_lg_perm,LG>>.
* Quadrant 2 is subdivided into octants, with octant 4 encoding combinations of <<r_perm,R>> and <<w_perm,W>> without <<c_perm,C>> (with the redundant not-<<r_perm,R>> and not-<<w_perm,W>> used to encode write-only with <<c_perm,C>>). Octant 5 encodes read-only capabilities with <<c_perm,C>> and the dependent permissions <<lm_perm,LM>> and <<zylevels1_lg_perm,LG>>.
* Quadrant 3 encodes permissions with <<r_perm,R>>, <<w_perm,W>>, <<c_perm,C>> and the dependent permissions <<zylevels1_sl_perm,SL>>, <<lm_perm,LM>> and <<zylevels1_lg_perm,LG>>.
```

src/cheri/rvy-cheriot-enc2.adoc
```diff
@@ -0,0 +1,83 @@
+[#rv32y_cheriot_encoding2_description, reftext="{rv32y_cheriot_encoding2_name}"]
+== The {rv32y_cheriot_encoding2_name} CHERI Capability Encoding Scheme
+
+This chapter describes a variation on the <<rv32y_cheriot_encoding1_description>> encoding to support all exponents using the same number of bits.
+This change is backwards compatible for software but enables support for more precise bounds on capabilities larger than 8MiB.
+Due to increased microarchitectural complexity and the limited need for precise alignment on such large capabilities on small systems, implementations may choose to support either encoding.
+
+The changes are limited to the bounds encoding; all other aspects of the capability encoding are identical to {rv32y_cheriot_encoding1_name}.
```
tariqkurd-repo:
```suggestion
The changes are limited to the bounds encoding; all other aspects of the capability encoding are identical to <<rv32y_cheriot_encoding1_description>>.
```

src/cheri/rvy-cheriot-enc1.adoc
tariqkurd-repo: much easier to read as a list

src/cheri/rvy-cheriot-isa-priv.adoc
```diff
@@ -0,0 +1,201 @@
...
+by pairing each memory block(s)
+in which software's dynamic allocation heap(s) are to reside
+with a per-memory-block _revocation bitmap_,
+a bit-vector wherein each bit corresponds to a capability-sized memory granule.
+(That is, with each capability-sized and -aligned region of the primal memory.)
+Such memory block(s) are said to be _revokable_.
+These revocation bitmaps are also exposed (as memory) to software.
+When `{load_cap_name_lc}` transfers a valid capability (one with a set {ctag})
+from memory (revokable or not) into a hart's register file,
+it checks whether the _base_ (lower bound) of that capability is within
```
tariqkurd-repo, line 188: why the base and not the address? using the address would give a shorter logic path as no decode would be required, and cheriot doesn't have the out of range pointers like university CHERI (except by 1 byte), so you don't have the uncertainty about whether the pointer is actually in bounds or not, and so whether it should really be revoked.

arichardson: It needs to be the base for out-of-bounds pointers. I thought you could have out-of-bounds pointers in CHERIoT just no guarantee that it's possible. For small bounds you can represent anything up to 512 bytes so you could have something OOB there?

tariqkurd-repo: If we were going to formalise the specification of this (which we aren't in the short term) I guess that "any in bounds address" could be checked - or are there problems with that too?

rmn30: > I guess that "any in bounds address" could be checked

That would cause issues for the allocator implementation. It has a capability to the heap that it really doesn't want to be revoked so it knows to never set the revocation tag for the base of that.

tariqkurd-repo: you certainly wouldn't want to accidentally revoke that the capability or the system is broken

arichardson (approved):
I think we should probably go ahead and merge this.

nwf:
Rebased atop main and #870.  I don't think I lost any commits, but please check.

Issue #855, opened 2025-11-21
Changes since 0.9.6


tariqkurd-repo:
Encoding formats:

1. Rename encoding formats: https://github.com/riscv/riscv-cheri/pull/845
2. Add CHERIoT Encoding format: XXX
3. Separate YBLD into its own extension: https://github.com/riscv/riscv-cheri/pull/853

CSRs:

1. Remove xycfg pending a general RISC-V discovery mechanism: https://github.com/riscv/riscv-cheri/pull/852

PR #860, opened 2025-11-26
enable vector for v1 ratification

tariqkurd-repo:
Vector+CHERI was being held back because of indexed loads where the base address is zero, and the whole address is in XLEN elements meaning that the capability needs to cover all of memory.
However, because if the base register is x0 the encoding is reserved, then the mode is illegal anyway.
So it's ok to include now, as discussed with @aswaterman


arichardson (approved):
Thanks! I will update https://github.com/riscv/riscv-cheri/issues/855

PR #861, opened 2025-11-26
refactor .ct into YBLD rules

tariqkurd-repo:
Attempting to make YBLD compatible with CHERIoT
the ct field behaviour is unchanged, but tag clearing is new, but only affects CHERIoT

PR #862, opened 2025-11-27
Tweak CBLD prose

nwf:
Replacement for the aborted #861.

### DIFF
--- a/src/cheri/insns/cbld_32bit.adoc
+++ b/src/cheri/insns/cbld_32bit.adoc
@@ -23,6 +23,10 @@ include::wavedrom/cbld.adoc[]
 Description::
 Copy `{cs2}` to `{cd}`.
 +
+If `{cd}.ct` (that is, its <<sec_cap_type>>) is
+neither 0 nor an <<sec_cap_type_ambient,ambient>> type, then
+set `{cd}.ct` to 0.
++
 Set `{cd}.tag=1` if:
 +
 . `{cs1}.tag` is set, and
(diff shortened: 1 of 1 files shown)


src/cheri/insns/cbld_32bit.adoc
```diff
@@ -39,11 +43,7 @@ Otherwise, set `{cd}.tag=0`
 NOTE: The <<section_cap_integrity,integrity>> check on `{cs2}` is required to prevent authorising a capability with a lack of integrity.
  The <<section_cap_integrity,integrity>> check on `{cs1}` is optional.
 
-If `{cd}.ct` (that is, its <<sec_cap_type>>) is
-neither 0 nor an <<sec_cap_type_ambient,ambient>> type, then
-set `{cd}.ct` to 0.
-That is, <<CBLD>> will construct a sealed capability
-only if its type is ambiently available.
+NOTE: <<CBLD>> will construct a sealed capability only if its type is <<sec_cap_type_ambient,ambiently available>>.
```
tariqkurd-repo, line 46: ...otherwise it clears the tag....? can we state this explicitly - which is the other thing I was trying to address in the aborted PR by merging it with the set tag=1 text above.

nwf: Otherwise, *as the text says*, it builds a *tagged and unsealed* capability (if the rest of the instruction -- the superset and wellformedness checks and all that -- is happy to set the tag).  This mimics the behavior of ISAv9's CBuildCap (https://www.cl.cam.ac.uk/techreports/UCAM-CL-TR-987.pdf p178) and allows for page-in code to reconstitute non-ambient sealed caps by using `YBLD; YSEAL`.

tariqkurd-repo: ok got it - I thought it would be detagged and hadn't realised exactly where you had inserted the new text.

PR #863, opened 2025-11-27
Svucrg: Reword PTE bits extension.

qwattash:
- The CRW field is extended to 3 bits. The CD and CRG bits are logical states encoded by the CRW field instead of proper separate bits in the PTE.
 - Clarify Reserved states for PTE.U=0. This should make it easier to add extensions that define CRW states that do not imply any CD/CRG semantics, for example a kernel revocation extension can define a cap-permissive state without load-side faults.
 - Flip the polarity of the encoded "PTE.CD" bit (PTE.CRW[1]), this ensures that software unaware of the Svucrg extension can continue to use the CRW bit with the exact same semantic.

PR #867, opened 2025-11-28
Add Svucrg enable bit.

qwattash:
This replaces #863 with the following intent:

- Add sstatus.CRGE that allows to enable/disable the Svucrg extension.
    - This solves the problem of the CD polarity when software is unaware of Svucrg and assumes that PTE.CRW behaves the same. This is no longer true as the PTE.CRW=1 CD=0 CRG=0 will either cause traps on capability writes or the PTE changes to CD=1 "unexpectedly" due to hardware dirty tracking.
    - The alternative is to change the CD bit polarity and leave Svucrg always enabled. This seems less palatable also because of possible interactions between further extensions and reserved PTE.{CRW, CD, CRG} bit patterns. The disable switch may also be used to selectively disable revocation under certain privileged conditions, but this seems like a bad idea.
- Add text to clarify that the PTE.CRW changes semantic when Svucrg is enabled. In particular:
    - CRW=0 no longer implies that the page is in the "no capabiltiy" state, instead we need CRW=0 CD=0 CRG=0.
    - CRW=1 is a sufficient condition for having a page in the "capability state",  because all 4 CRW=1 bit patterns are used in that way, but it is not a necessary condition, because further extensions can define a reserved CRW=0 CD=X CRG=X state to allow load store of capabilities under certain conditions.
    
There are some questions that we may want to ensure are addressed:
   - Do we need to specify the reset value of sstatus.CRGE?
   - Is sstatus the right place ofr CRGE or is it senvcfg?
   - I stated that if CRGE=0 (i.e. Svucrg disabled) the CD, CRG and sstatus.UCRG bits remain reserved and further extensions can reuse them. I'm not sure this is exactly what we want.

### DIFF
--- a/src/cheri/cheri-pte-ext.adoc
+++ b/src/cheri/cheri-pte-ext.adoc
@@ -34,11 +34,10 @@ NOTE: Hardware-initiated memory accesses from the page-table walker are not chec
 [#cap_revocation]
 === Capability Revocation
 
+Revocation is the process of clearing the {ctag} of capabilities that point to freed memory, to prevent them from being dereferenced and thereby enforcing _temporal memory safety_.
 cite:[cornucopia-reloaded] details algorithms for preventing use-after-free in virtual memory, and non-virtual memory based systems.
-Revocation is the process of clearing the {ctag} of capabilities which point to freed memory, to prevent them from being dereferenced.
-A CHERI system which implements such a scheme is said to support _temporal safety_.
 
-{cheri_priv_crg_ext} adds temporal safety for user pages only.
+{cheri_priv_crg_ext} adds capability revocation for user pages only.
 It uses the <<sstatusreg_pte,sstatus>>.UCRG and PTE.CRG bits to control the revocation epoch, which is used to provide a load barrier to detect when loading from a page which needs sweeping.
 This is indicated by the generation in the PTE mismatching the global one in <<sstatusreg_pte,sstatus>>.UCRG.
 Stored {ctag}s to clean pages are tracked using PTE.CD to record which virtual memory pages contain dirty capabilities, and so will need to be swept.


src/cheri/cheri-pte-ext.adoc
```diff
@@ -7,14 +7,14 @@ endif::[]
 
 NOTE: _Sv32_ (for RV32) does not have any spare PTE bits, and so no features from this chapter can be implemented.
 
-The {cheri_priv_crg_ext} extension adds the ability to perform capability revocation of user mode pages (see <<cap_revocation>>) by adding the PTE.CRG, PTE.CD and <<sstatusreg_pte,sstatus>>.UCRG bits as described below.
+The {cheri_priv_crg_ext} extension is enabled when the <<sstatusreg_pte,sstatus>>.CRGE bit is set.
+When enabled, the extension adds the ability to perform capability revocation of user mode pages (see <<cap_revocation>>) by adding the PTE.CRG, PTE.CD and <<sstatusreg_pte,sstatus>>.UCRG bits as described below.
+When the extension is disabled the PTE.CD, PTE.CRG and <<sstatusreg_pte,sstatus>>.UCRG bits remain reserved and a future extension may define an alternative function; the PTE.CRW bit is interpreted as defined by {cheri_priv_vmem_ext}.
```
```suggestion
When the extension is disabled the PTE.CD, PTE.CRG and <<sstatusreg_pte,sstatus>>.UCRG bits remain reserved and a future extension may define an alternative function; the PTE.CRW bit is interpreted as defined by <<section_priv_cheri_vmem>>.
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -24,6 +24,10 @@ capabilities (i.e., private anonymous mappings) and leave
 <<sstatusreg_pte,sstatus>>.UCRG and CRG in all PTEs set to 0, which will allow
 capabilities with their tags set to be loaded and stored successfully.
```
```suggestion
capabilities with their {ctag}s set to be loaded and stored successfully.
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -24,6 +24,10 @@ capabilities (i.e., private anonymous mappings) and leave
...
 
+NOTE: When the {cheri_priv_crg_ext} is enabled (the <<sstatusreg_pte,sstatus>>.CRGE is set), the PTE.CRW bit semantic is not compatible with the definition in {cheri_priv_vmem_ext}.
```
```suggestion
NOTE: When {cheri_priv_crg_ext} is enabled (<<sstatusreg_pte,sstatus>>.CRGE is set), the semantics of PTE.CRW bit semantic may be redefined relative to the definition in  <<section_priv_cheri_vmem>>..
```

tariqkurd-repo: I think that nothing is redefined yet.... so CRGE is for the future... for when we add kernel revocation?

src/cheri/cheri-pte-ext.adoc
```diff
@@ -185,7 +189,7 @@ These cause PTE Accessed and Dirty updates to be done in software, via the excep
 [#xstatus_pte]
 === Extending the Supervisor (sstatus) and Virtual Supervisor (vsstatus) Status Registers
 
-The <<sstatusreg_pte,sstatus>> and <<vsstatusreg_pte,vsstatus>> CSRs are extended to include the new Capability Read Generation (CRG) bit as shown.
+The <<sstatusreg_pte,sstatus>> and <<vsstatusreg_pte,vsstatus>> CSRs are extended to include the new Capability Read Generation (CRG) and Capability Read Generation Enable (CRGE) bits as shown.
 
 When V=1 <<vsstatusreg_pte,vsstatus>>.UCRG is in effect.
```
```suggestion
When V=1 <<vsstatusreg_pte,vsstatus>>.UCRG and CRGE are in effect.
```

tariqkurd-repo: also the next line but I can't add a suggestion to it

tariqkurd-repo:
> * Do we need to specify the reset value of sstatus.CRGE?

I don't think so - we only specify reset values where it's really important - and in this case it is legal to hardwire CRGE=1 or CRGE=0 so we can't sepcify it

> * Is sstatus the right place ofr CRGE or is it senvcfg?

probably senvcfg - as that has enable bits in? Not 100% sure but it's probably better there

> * I stated that if CRGE=0 (i.e. Svucrg disabled) the CD, CRG and sstatus.UCRG bits remain reserved and further extensions can reuse them. I'm not sure this is exactly what we want.

that's what _reserved_ means..... can be reused for other purposes... or do you mean you want them reserved for other cheri revocation schemes?

src/cheri/cheri-pte-ext.adoc
qwattash: I think we are redefining the behaviour due to CD (as discussed elsewhere, I will repeat it here to keep track of rationale).
Consider:

- CRGE=0 then
    - CRW=0 -> strip tag on load / fault on store
    - CRW=1 -> normal operation
- CRGE=1 then
     - CRW=0 (CD=0, CRG=0) -> strip tag on load / fault on store
     - CRW=1 (CD=0, CRG=0) -> either CD=1 on store or fault on store for CD emulation.

So we actually re-define the base spec PTE.CRW=1 behaviour as PTE.CRW=1, PTE.CD=1, PTE.CRG=0. This is also spelled out at the beginning of the chapter as "minimum level of PTE support".

src/cheri/cheri-pte-ext.adoc
```diff
@@ -7,22 +7,26 @@ endif::[]
...
 NOTE: A future version of this specification may include kernel revocation which may require an <<sstatusreg_pte,sstatus>>.SCRG bit.
 
 The minimum level of PTE support is to set CRW and CD to 1 in all PTEs intended for storing
 capabilities (i.e., private anonymous mappings) and leave
 <<sstatusreg_pte,sstatus>>.UCRG and CRG in all PTEs set to 0, which will allow
-capabilities with their tags set to be loaded and stored successfully.
+capabilities with their {ctag}s set to be loaded and stored successfully.
+
+NOTE: When {cheri_priv_crg_ext} is enabled (<<sstatusreg_pte,sstatus>>.CRGE is set), the semantics of PTE.CRW bit may be redefined relative to the definition in <<section_priv_cheri_vmem>>.
+The ability to load and store capabilities in the page depends on all the PTE.CRW, PTE.CD, PTE.CRG bits.
```
```suggestion
The ability to load and store capabilities in the page, and any side-effects, depends on all of the PTE.CRW, PTE.CD, PTE.CRG bits, which effectively become a single 3-bit field controlling the behavior.
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -7,22 +7,26 @@ endif::[]
...
 
 NOTE: A future version of this specification may include kernel revocation which may require an <<sstatusreg_pte,sstatus>>.SCRG bit.
 
 The minimum level of PTE support is to set CRW and CD to 1 in all PTEs intended for storing
 capabilities (i.e., private anonymous mappings) and leave
 <<sstatusreg_pte,sstatus>>.UCRG and CRG in all PTEs set to 0, which will allow
-capabilities with their tags set to be loaded and stored successfully.
+capabilities with their {ctag}s set to be loaded and stored successfully.
+
+NOTE: When {cheri_priv_crg_ext} is enabled (<<sstatusreg_pte,sstatus>>.CRGE is set), the semantics of PTE.CRW bit may be redefined relative to the definition in <<section_priv_cheri_vmem>>.
```
```suggestion
NOTE: When {cheri_priv_crg_ext} is enabled (i.e. <<sstatusreg_pte,sstatus>>.CRGE is set), the semantics of PTE.CRW bit are redefined relative to the definition in <<section_priv_cheri_vmem>>.
```

src/cheri/cheri-pte-ext.adoc
```suggestion
The ability to load and store capabilities in the page depends on all of the PTE.CRW, PTE.CD and PTE.CRG bits which are effectively combined into a single 3-bit field.
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -7,22 +7,26 @@ endif::[]
...
+The ability to load and store capabilities in the page depends on all the PTE.CRW, PTE.CD, PTE.CRG bits.
+When PTE.CRW is clear, it is not implied that the page can not store capabilities; a future extension is allowed to define an alternative function for the reserved CRW, CD and CRG combinations that allow capability load and stores regardless of the value of the PTE.CRW bit.
```
```suggestion
Therefore, when PTE.CRW is clear, it is not required that future extensions to {cheri_priv_crg_ext} , such as for kernel revocation, cannot store capabilities to a page.```

**Comment by tariqkurd-repo** [comment]
> > I'm thinking a bit harder on this point, as it feels that if we want to use the same CRGE bit in the future (e.g. for kernel revocation) we will run in exactly the same CD polarity problem.
> 
> If it's part of the same extension then you have just the one bit that does it all. If if's part of a different extension then it gets its own enable bit. Ideally it's part of the same extension.

yes - ideally it would fit in the reserved spaces - for enabling kernel revocation - but it's not the end of the world if it doesn't

File: src/cheri/cheri-pte-ext.adoc [hunk 8]
```diff
@@ -7,22 +7,26 @@ endif::[]
 
 NOTE: _Sv32_ (for RV32) does not have any spare PTE bits, and so no features from this chapter can be implemented.
 
-The {cheri_priv_crg_ext} extension adds the ability to perform capability revocation of user mode pages (see <<cap_revocation>>) by adding the PTE.CRG, PTE.CD and <<sstatusreg_pte,sstatus>>.UCRG bits as described below.
+The {cheri_priv_crg_ext} extension is enabled when the <<sstatusreg_pte,sstatus>>.CRGE bit is set.
+When enabled, the extension adds the ability to perform capability revocation of user mode pages (see <<cap_revocation>>) by adding the PTE.CRG, PTE.CD and <<sstatusreg_pte,sstatus>>.UCRG bits as described below.
+When the extension is disabled the PTE.CD, PTE.CRG and <<sstatusreg_pte,sstatus>>.UCRG bits remain reserved and a future extension may define an alternative function; the PTE.CRW bit is interpreted as defined by <<section_priv_cheri_vmem>>.
```
arichardson, line 15: Is reserved rather than ignored really what we want here? That would not make it possible to dynamically toggle the state since reserved PTE bits must trap.

tariqkurd-repo: I think they shoudl be reserved - if svucrg is disabled then there's no defined meaning for those bits - so they could in theory be redefined by a different extesnion which uses those PTE bits

src/cheri/cheri-pte-ext.adoc
qwattash: updated text.

src/cheri/cheri-pte-ext.adoc
qwattash: I think one could use the dynamic toggling to fetch (say) the shadow bitmap capability while in the CRG fault handler in the kernel, in the presence of kernel revocation. However, upon further thinking I'm more convinced that this would be a hack and there shouldn't be a reason to turn off the ambient CRG fault handling to do that (actually there may be races and it may require some level of synchronisation which we can't do prior to disabling CRG faults). So I think there it is actually sensible to keep them reserved.

arichardson: If it's just an enable bit then shouldn't it be in senvcfg (or possibly menvcfg and require a SBI call?) and default to off. Not sure if that makes more sense but status is probably not the right place for it since it's for dynamic state?

tariqkurd-repo: sstatus might be appropriate as it affects the TLB operation like mstatus.SUM / MXR - or maybe not as that's mstatus. also satp of course - but that register is full.
it can't be senvcfg as that's for user mode config only - it would have to be menvcfg - so would you want the OS to be able to control this? if not then probably menvcfg is the correct place.

arichardson: I guess the bit allocation is up to ARC, we can put a note to them that they need to decide where this goes?

tariqkurd-repo: ```The menvcfg CSR is a 64-bit read/write register, formatted as shown in Figure 25), that controls certain characteristics of the execution environment for modes less privileged than M.```

tariqkurd-repo: ARC decide everything anyway, we dont' need to prompt them to do so!
Should this bit be under OS control? if not then let's go for menvcfg

tariqkurd-repo: discussing with alfredo we'll stick with sstatus for now - he'll add justification as a note in the PR

nwf (approved):
LGTM

src/cheri/cheri-pte-ext.adoc
```diff
@@ -196,6 +196,16 @@ When V=1 <<vsstatusreg_pte,vsstatus>>.UCRG and CRGE are in effect.
 
 <<mstatusreg_pte,mstatus>>.UCRG and CRGE also exist. Reading or writing it is equivalent to reading or writing <<sstatusreg_pte,sstatus>>.UCRG and CRGE respectively.
 
+NOTE: The CRGE bit is added to the <<mstatusreg_pte,mstatus>>, <<vsstatusreg_pte,vsstatus>> and <<sstatusreg_pte,sstatus>> registers in order to allow the S-mode software to explicitly opt in to the revocation extensions.
+The CRGE bit can not be exclusively accessible from M-mode (e.g. in <<menvcfg>>), because this would imply that M-mode can enable revocation without any guarantee that the S-mode software understands the CRW, CD and CRG semantics, and can result in breaking backward compatibility when updating the M-mode firmware.
```
nwf: I'm not sure that this last sentence is really relevant... M mode does a lot of things at the behest of less-privileged modes without acting of its own volition.  In principle, turning on `Svucrg` could be done by some non-architectural communication path through the hypervisor and to M mode (SBI/ecall/...).

In any case, the first sentence of this NOTE is fine.

jrtc27: Yeah I mean you *can* do it in M-mode, that's where ADUE lives, but it's stupid, because it requires an SBI call to do something that S-mode is in control of (it's always S-mode's choice which value to use, and there's no reason for M-mode to deny it) and has no implications for M-mode. We should not repeat that idiocy and just make it available to S-mode.

qwattash: That is a good point. Perhaps I should change the wording to say that it _could_ be done via SBI or similar interfaces by M-mode if we placed somewhere else, and this would still be allowing S-mode to opt in. You can always do something that is not necessarily sensible in M-mode and enable it blindly, but then you can't expect software to function if it doesn't know about it.

qwattash: Changed the wording now.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -185,11 +190,21 @@ These cause PTE Accessed and Dirty updates to be done in software, via the excep
...
+Alternatively, the CRGE bit could be made exclusively accessible from M-mode (e.g. in <<menvcfg>>); the S-mode software could use an extension of the SBI interface to opt in to the revocation extensions.
 
-<<mstatusreg_pte,mstatus>>.UCRG also exists. Reading or writing it is equivalent to reading or writing <<sstatusreg_pte,sstatus>>.UCRG.
+NOTE: The CRGE bit is necessary because the {cheri_priv_crg_ext} defines a different semantic for PTE.CRW=1 PTE.CD=0 PTE.CRG=0 with respect to <<section_priv_cheri_vmem>>.
+In particular, an oblivious software implementation that only sets the PTE.CRW bit would observe one of the following:
+* The PTE.CD bit is set by the hardware when a capability is stored and _Svadu_ is enabled.
+* A {cheri_excep_name_pte_st} trap when _Svade_ is enabled.
+Both these conditions are unexpected according to <<section_priv_cheri_vmem>>.
+The CRGE bit also has the benefit or maintaining matching polarity of the PTE.CD bit with respect to the data Dirty bit (PTE.D), and permit alternative uses by future extensions of the PTE.CD and PTE.CRG reserved bits (e.g. by an alternative revocation extension that re-defines the semantics of CD and CRG).
```
```suggestion
The CRGE bit also has the benefit of maintaining matching polarity of the PTE.CD bit with respect to the data Dirty bit (PTE.D), and permits alternative uses by future extensions of the PTE.CD and PTE.CRG reserved bits (e.g. by an alternative revocation extension that re-defines the semantics of CD and CRG).
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -185,11 +190,21 @@ These cause PTE Accessed and Dirty updates to be done in software, via the excep
...
-The <<sstatusreg_pte,sstatus>> and <<vsstatusreg_pte,vsstatus>> CSRs are extended to include the new Capability Read Generation (CRG) bit as shown.
+The <<sstatusreg_pte,sstatus>> and <<vsstatusreg_pte,vsstatus>> CSRs are extended to include the new Capability Read Generation (CRG) and Capability Read Generation Enable (CRGE) bits as shown.
+
+When V=1 <<vsstatusreg_pte,vsstatus>>.UCRG and CRGE are in effect.
+
+<<mstatusreg_pte,mstatus>>.UCRG and CRGE also exist. Reading or writing it is equivalent to reading or writing <<sstatusreg_pte,sstatus>>.UCRG and CRGE respectively.
 
-When V=1 <<vsstatusreg_pte,vsstatus>>.UCRG is in effect.
+NOTE: The CRGE bit is added to the <<mstatusreg_pte,mstatus>>, <<vsstatusreg_pte,vsstatus>> and <<sstatusreg_pte,sstatus>> registers in order to allow the S-mode software to explicitly opt in to the revocation extensions.
+Alternatively, the CRGE bit could be made exclusively accessible from M-mode (e.g. in <<menvcfg>>); the S-mode software could use an extension of the SBI interface to opt in to the revocation extensions.
```
tariqkurd-repo: this is suggesting a different spec so needs to go, unfortuantely

```suggestion
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -7,40 +7,38 @@ endif::[]
...
-
 NOTE: A future version of this specification may include kernel revocation which may require an <<sstatusreg_pte,sstatus>>.SCRG bit.
 
 The minimum level of PTE support is to set CRW and CD to 1 in all PTEs intended for storing
 capabilities (i.e., private anonymous mappings) and leave
 <<sstatusreg_pte,sstatus>>.UCRG and CRG in all PTEs set to 0, which will allow
-capabilities with their tags set to be loaded and stored successfully.
+capabilities with their {ctag}s set to be loaded and stored successfully.
+
+NOTE: When {cheri_priv_crg_ext} is enabled (<<sstatusreg_pte,sstatus>>.CRGE is set), the semantics of PTE.CRW bit is redefined relative to the definition in <<section_priv_cheri_vmem>>.
```
```suggestion
NOTE: When {cheri_priv_crg_ext} is enabled (<<sstatusreg_pte,sstatus>>.CRGE is set), the semantics of PTE.CRW bit are redefined relative to the definition in <<section_priv_cheri_vmem>>.
```

src/cheri/cheri-pte-ext.adoc
qwattash: That's fine.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -7,40 +7,38 @@ endif::[]
 
 NOTE: _Sv32_ (for RV32) does not have any spare PTE bits, and so no features from this chapter can be implemented.
 
-The {cheri_priv_crg_ext} extension adds the ability to perform capability revocation of user mode pages (see <<cap_revocation>>) by adding the PTE.CRG, PTE.CD and <<sstatusreg_pte,sstatus>>.UCRG bits as described below.
+The {cheri_priv_crg_ext} extension is enabled when the <<sstatusreg_pte,sstatus>>.CRGE bit is set.
```
arichardson, line 12:
```suggestion

#Begin new since last ARC review (added enable/disable bit)#
The {cheri_priv_crg_ext} extension is enabled when the <<sstatusreg_pte,sstatus>>.CRGE bit is set.
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -7,40 +7,38 @@ endif::[]
...
+When enabled, the extension adds the ability to perform capability revocation of user mode pages (see <<cap_revocation>>) by adding the PTE.CRG, PTE.CD and <<sstatusreg_pte,sstatus>>.UCRG bits as described below.
```
arichardson, line 14:
```suggestion
#End new since last ARC review#
When enabled, the extension adds the ability to perform capability revocation of user mode pages (see <<cap_revocation>>) by adding the PTE.CRG, PTE.CD and <<sstatusreg_pte,sstatus>>.UCRG bits as described below.
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -185,11 +184,20 @@ These cause PTE Accessed and Dirty updates to be done in software, via the excep
 [#xstatus_pte]
 === Extending the Supervisor (sstatus) and Virtual Supervisor (vsstatus) Status Registers
 
-The <<sstatusreg_pte,sstatus>> and <<vsstatusreg_pte,vsstatus>> CSRs are extended to include the new Capability Read Generation (CRG) bit as shown.
+The <<sstatusreg_pte,sstatus>> and <<vsstatusreg_pte,vsstatus>> CSRs are extended to include the new Capability Read Generation (CRG) and Capability Read Generation Enable (CRGE) bits as shown.
```
arichardson, line 190:
```suggestion
#Begin new since last ARC review (added enable/disable bit)#
The <<sstatusreg_pte,sstatus>> and <<vsstatusreg_pte,vsstatus>> CSRs are extended to include the new Capability Read Generation (CRG) and Capability Read Generation Enable (CRGE) bits as shown.
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -185,11 +184,20 @@ These cause PTE Accessed and Dirty updates to be done in software, via the excep
...
+NOTE: The CRGE bit is added to the <<mstatusreg_pte,mstatus>>, <<vsstatusreg_pte,vsstatus>> and <<sstatusreg_pte,sstatus>> registers in order to allow the S-mode software to explicitly opt in to the revocation extensions.
 
-<<mstatusreg_pte,mstatus>>.UCRG also exists. Reading or writing it is equivalent to reading or writing <<sstatusreg_pte,sstatus>>.UCRG.
+NOTE: The CRGE bit is necessary because the {cheri_priv_crg_ext} defines a different semantic for PTE.CRW=1 PTE.CD=0 PTE.CRG=0 with respect to <<section_priv_cheri_vmem>>.
+In particular, an oblivious software implementation that only sets the PTE.CRW bit would observe one of the following:
+* The PTE.CD bit is set by the hardware when a capability is stored and _Svadu_ is enabled.
+* A {cheri_excep_name_pte_st} trap when _Svade_ is enabled.
+Both these conditions are unexpected according to <<section_priv_cheri_vmem>>.
+The CRGE bit also has the benefit of maintaining matching polarity of the PTE.CD bit with respect to the data Dirty bit (PTE.D), and permits alternative uses by future extensions of the PTE.CD and PTE.CRG reserved bits (e.g. by an alternative revocation extension that re-defines the semantics of CD and CRG).
```
arichardson, line 203:
```suggestion
The CRGE bit also has the benefit of maintaining matching polarity of the PTE.CD bit with respect to the data Dirty bit (PTE.D), and permits alternative uses by future extensions of the PTE.CD and PTE.CRG reserved bits (e.g. by an alternative revocation extension that re-defines the semantics of CD and CRG).
#End new since last ARC review#
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -7,40 +7,38 @@ endif::[]
...
-read to a new page to sweep it clean before proceeding.
-With a page-granularity generational capability read permission, we can
-eliminate the initial permission change of all PTEs.
-In addition, a page-granularity capability write control can eliminate many pages
-from the sweep that are known to not contain capabilities.
-Page-granularity capability-dirty tracking accelerates the detection of capability-clean
-pages, which enables the revocation sweep to skip them during the visit.
+cite:[cornucopia-reloaded] details algorithms for preventing use-after-free in virtual memory, and non-virtual memory based systems.
+Revocation is the process of clearing the {ctag} of capabilities which point to freed memory, to prevent them from being dereferenced.
+A CHERI system which implements such a scheme is said to support _temporal safety_.
```
```suggestion
Revocation is the process of clearing the {ctag} of capabilities that point to freed memory, to prevent them from being dereferenced and thereby enforcing _temporal memory safety_.
cite:[cornucopia-reloaded] details algorithms for preventing use-after-free in virtual memory, and non-virtual memory based systems.
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -7,40 +7,38 @@ endif::[]
...
 
+{cheri_priv_crg_ext} adds temporal safety for user pages only.
```
```suggestion
{cheri_priv_crg_ext} adds capability revocation for user pages only.
```

PR #868, opened 2025-11-28
YTOPR

tariqkurd-repo:
CHERIoT need it, and it's so easy to include in the base ISA.
There's no hardware cost as the TOP and BASE are unpacked together.
So easier to just include it and save time later.


nwf:
I'm happy to have `YTOPR` in the base, but `YBNDSRDW` is, in a sense, the more important of the Zycheriot instructions to have in the base ISA, since its operation is significantly more difficult to replace in software (and depends heavily on the base's parameter soup).  See https://github.com/CHERIoT-Platform/cheriot-sail/pull/74/commits/8b8938982e1458e6e84ffe6269e9801ff45adc5f for our sail and https://github.com/CHERIoT-Platform/cheriot-rtos/pull/310/files for the software it replaced.

nwf (approved):

tariqkurd-repo:
>Adding a new Zytopr is fine too, I'm not sure how often this shows up in critical paths?

it won't - the top is already calculated - it just needs driving to the output - it's really no different to accessing the base.
this is one instruction which really has minimal implemetnation cost.

PR #870, opened 2025-11-28
Zybndsrd

nwf:


src/cheri/insns/scbndsrd_32bit.adoc
```diff
@@ -0,0 +1,73 @@
...
+- precisely representable given this base address.
+--
++
+Set `{cd}.tag=0` if `{cs1}.tag=0`, `{cs1}` is sealed or if `{cd}` 's bounds exceed `{cs1}` 's bounds.
++
+include::malformed_cs1_clear_tag.adoc[]
+
+[NOTE]
+=====
+If the result in `{cd}` is tagged, then its base _is_ `{cs1}.address`, as with <<SCBNDS>>.
```
tariqkurd-repo:
```suggestion
If the result in `{cd}` has its {ctag} set to zero, then its base _is_ `{cs1}.address`, as with <<SCBNDS>>.
```

we only use "tagged" for "tagged memory" in the spec, we don't refer to caps as tagged/untagged

src/cheri/insns/scbndsrd_32bit.adoc
```diff
@@ -0,0 +1,73 @@
...
+=====
+The "largest representable length value precisely representable given this base address" is
+also the largest length `l` for which a <<SCBNDS>> instruction...
+
+- whose `rs2` register holds `l` and
+
+- whose `{cs1}` register holds a capability whose bounds cover the entire address space and
+  whose address is the given base address
+
+produces a tagged result in `{cd}`.
```
tariqkurd-repo:
```suggestion
produces a result in `{cd}` with the {ctag} set to one.
```

src/cheri/insns/scbndsrd_32bit.adoc
```diff
@@ -0,0 +1,73 @@
...
+include::malformed_cs1_clear_tag.adoc[]
+
+[NOTE]
+=====
+If the result in `{cd}` is tagged, then its base _is_ `{cs1}.address`, as with <<SCBNDS>>.
+The result in `{cd}` will be zero-length if and only if `rs2[XLEN-1:0]` is zero.
+In practice, capability encodings offer byte-granularity of bounds for capabilities
+of lengths up to some threshold
+between a few hundred or a few thousand bytes, depending on the encoding;
+for requested lenghts below this threshold, no rounding is required, and
```
tariqkurd-repo:
```suggestion
for requested lengths below this threshold, no rounding is required, and
```

src/cheri/attributes.adoc
```diff
@@ -187,6 +187,11 @@ endif::support_varxlen[]
 :SCBNDSR_LC:   ybndsrw
 :SCBNDSR_DESC: Write capability bounds, rounding up if required
 
+:cheri_bndsrd_ext_name: Zybndsrd
```
tariqkurd-repo:
```suggestion
:cheri_bndsrd_ext_name: Zybndsrdw
```

it should be the whole instruction name

src/cheri/insns/scbndsrd_32bit.adoc
```diff
@@ -0,0 +1,73 @@
...
+[NOTE]
+=====
+{cheri_base_ext_name} implementations which have
+a "maximum exponent" parameter `e` whose value is some `e<N>` (and not `ea`)
+may choose to restrict their search for "largest representable length"
+to those that would use an exponent at most `N`.
+=====
+
+Included in::
+{cheri_bndsrd_ext_name}
```
tariqkurd-repo: this should be an xref

src/cheri/zybndsrd.adoc
```diff
@@ -0,0 +1,32 @@
...
+This extension adds a single instruction, <<SCBNDSRD>>,
+which writes capability bounds while
+rounding _down_ to ensure precise representability of the result.
+In contrast to <<SCBNDSR>>,
+it does not alter the requested _lower_ bound (capability base).
+
+[NOTE]
+=====
+This instruction is useful when exposing (byte) buffers across trust domains.
+Given a buffer, a cursor therein, and a length after that cursor of elements
```
tariqkurd-repo: cursor isn't defined in this spec

src/cheri/insns/scbndsrd_32bit.adoc
```diff
@@ -0,0 +1,73 @@
...
+computing the length that is a maximal mantissa shifted left by that count
+(that is, using the number of trailing zeros in the base address as the value's exponent).
+=====
+
+[NOTE]
+=====
+{cheri_base_ext_name} implementations which have
+a "maximum exponent" parameter `e` whose value is some `e<N>` (and not `ea`)
+may choose to restrict their search for "largest representable length"
+to those that would use an exponent at most `N`.
```
tariqkurd-repo: do you make use of this in the cheriot formats?

src/cheri/zybndsrd.adoc
rmn30: It's a generic English word with some specific meanings in computer science: https://en.wiktionary.org/wiki/cursor . In this case we could probably substitute it with offset or address without losing anything.

src/cheri/insns/scbndsrd_32bit.adoc
rmn30: Yes, for the V1 format which doesn't have exponents between 14 and 24. It's simpler for formats that support all exponents as you just choose the exponent that's the lesser of the one for the requested length (computed from clz(length)) and the required base alignment (ctz(base)).

src/cheri/zybndsrd.adoc
tariqkurd-repo: I've only ever seen that used in CHERI v9 - so it's probably better to introduce it or just swap it for something more commonly used

src/cheri/insns/scbndsrd_32bit.adoc
tariqkurd-repo: does this need some additional text in the encoding format chapter then? or is the note enough?

src/cheri/insns/scbndsrd_32bit.adoc
nwf: That's "set to _one_", but sure.

src/cheri/insns/scbndsrd_32bit.adoc
nwf: I think the note here is fine.  If encodings are part of the base, extensions are surely allowed to predicate on them.

nwf: Also, just to say, there's nothing _intrinsic_ about the `e<N>` (as opposed to `ea`) encodings that mean that they _must_ restrict their search like this, and for CHERIoT we could have special cased the "exponent 24" case, but in practice we don't _have_ "exponent 24" capabilities anywhere in the system after it's up and running, so it didn't seem important.  Similar considerations might apply to other implementations, too, so thus the note.

src/cheri/insns/scbndsrd_32bit.adoc
```diff
@@ -0,0 +1,73 @@
...
+`{scbndsrd_lc} {cd}, {cs1}, rs2`
+
+Encoding::
+include::wavedrom/scbndsrd_32bit.adoc[]
+
+Description::
+Copy the capability from register `{cs1}` to register `{cd}`.
+Set the base address of its bounds to the value of `{cs1}.address`.
++
+Set the length of its bounds to a the largest representable length value that is both
```
jrtc27: s/a //

But also, "representable" is redundant here given bullet point 2? Can we not just borrow the same language as YBNDSRW and say something like:

> The top is rounded down by the smallest amount needed to form a capability covering the requested base and top.

src/cheri/insns/scbndsrd_32bit.adoc
jrtc27: I don't understand this note. It reads as "if you have a maximum exponent then you only need to search up to this exponent" which is... surely what you'd expect? Why would you ever be trying to search for an exponent that exceeds the maximum exponent for the encoding? Is that not blindingly obvious?

nwf: The "maximum exponent" is a misnomer; please suggest a better name.  It's the maximum representable exponent before any omission.  That is, `e<N>` in the base parameter string means that exponents `[0,N]` are all available, as is whatever value `v` is necessary to represent the entire `XLEN`-bit address space (usually `XLEN-mw+1`, but I don't want to promise that that's always the one), but that those in the "band gap" `(N,v)` might not all be available.

jrtc27: Ok, so the set is always either `[0,N]` or `[0,N] + {v}`, and in the latter case it's only used for the singular whole-address-space bounds encoding?

src/cheri/insns/scbndsrd_32bit.adoc
nwf: It's not just "covering the requested base and top"; the provided cursor _is_ the base of the result.  Still, sure, I'll try cleaning up the wording.

src/cheri/insns/scbndsrd_32bit.adoc
jrtc27: If so, I think the sole purpose of this note is to remind the implementer that it is impossible to request whole-address-space bounds via any `YNBDS*W*` instruction, and therefore those instructions never have to consider the whole-address-space bounds case?

src/cheri/insns/scbndsrd_32bit.adoc
jrtc27: And I suppose the top isn't requested either... how about just:

> The top is rounded down by the smallest amount needed to form a representable capability.

src/cheri/insns/scbndsrd_32bit.adoc
nwf: It isn't strictly required that the set is always one of those, though I think `[0,v]` and `[0,N] + {v}` are the only examples implemented to date.  In principle you could imagine things like `[0,N] + {(N+v)/2, v}`.

But no, the note has more weight than that.  The CHERIoT v1 encoding has `N` = `14` and `v` = `24`, and so while `LUI %0, 0x2000; YBNDSRDW %2, %1, %0` _could_ yield a `%2` with length `0x2000000` (and so exponent `24`, since `14` won't reach) (assuming 24-bit alignment of `%1.address`) precisely capturing the request, the note permits implementations to return capabilities with length at most `0x1FF << 14 = 0x7FC000`.

jrtc27: If YBNDSRW can yield a given capability then YBNDSRDW should be able to as well

jrtc27: Otherwise you're getting way too specialised as "this is a bit easier for our hardware to implement and that case isn't used by our particular software so we just won't support that", which is fine, but not for standardisation, that's very much into custom extension territory.

nwf: `YBNDSRW` and `YBNDSRDW` serve pretty different purposes to software, so I'm not entirely convinced that they must have the same range of outputs.  That said, it's probably not too many gates to handle the `v` case; I'll ask people who'd know better than I.

jrtc27: Please do not repeat the mistake Morello made

arichardson: Unless this actually has a noticeable impact on implementation complexity/fmax/etc, I would much prefer it if this instruction was able to produce the same output as YBDSRW in cases with e==24. Even if this does not matter in CHERIoTv1 systems since you won't have those at runtime it is a cleaner spec if the behaviour is consistent.

arichardson (changes requested):
This looks good to me and I will merge it once we have clarified how complicated it is to handle the e==24 case. Alternatively, we could drop the note here and move it in the CHERIoTv1 chapter?

jrtc27:
Either YBNDSRDW allows such shenanigans, in which case we're back to square one, or it doesn't, in which case no amount of weasel wording in the CHERIoT v1 encoding chapter can override that.

src/cheri/insns/scbndsrd_32bit.adoc
```diff
@@ -0,0 +1,66 @@
...
+of lengths up to some threshold
+between a few hundred and a few thousand bytes, depending on the encoding;
+for requested lengths below this threshold, no rounding is required, and
+otherwise this threshold serves to guarantee a minimum length return from {SCBNDSRD}.
+=====
+
+[NOTE]
+=====
+{SCBNDSRD} finds the largest length `l` that is both
+less than the requested `rs2[XLEN-1:0]` and
```
tariqkurd-repo: is really less than, or less than or equal to?
above it says "round down by the samllest amount necessary" which presumably can include zero?

davidchisnall: Correct, this should be less than or equal to.  In the common case for the use of this in software, it doesn't do any rounding and gives you the exact bounds you asked for (which is part of why it's so important: an attacker can sometimes influence layout such that a specific point would be impossible to precisely represent and you want the happy and unhappy code paths to be the same or you'll only test the not-under-attack case).

tariqkurd-repo:
```suggestion
less than or equal to the requested `rs2[XLEN-1:0]` and
```

happy to merge with this fix

nwf: Argh, yeah, that was "at most" in past drafts and then I flubbed it in editing.

PR #873, opened 2025-12-01
restrict YBNDSWI to 8-bits (anticipating a smaller immediate following ARC review)

tariqkurd-repo:
We don't have any compelling data for the top 2 bits of the 10-bit immediate, so instead of waiting for it to be rejected by ARC we may as well trim it up-front.
The 8-bit immediate covers all the CHERIoT use cases, and almost all of the FreeBSD use cases too.

### DIFF
--- a/src/cheri/insns/scbnds_32bit.adoc
+++ b/src/cheri/insns/scbnds_32bit.adoc
@@ -45,22 +45,9 @@ Set `{cd}.tag=0` if the requested bounds cannot be encoded exactly.
 +
 include::malformed_cs1_clear_tag.adoc[]
 +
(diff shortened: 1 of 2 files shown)


nwf:
A) We'd already pretty severely restricted `YBNDSWI`: for the I-type format we'd taken, we had also required that `rs1 = rd` and `imm[11:10] = 0`.

B) If you're going to restrict to 8 bits worth of immediate, almost surely "1 through 256" is not the best choice.

C) This doesn't change that we're still expanding into a whole new I-type-shaped chunk of the encoding space (opcode=0011011, funct3=011), so... why?

Issue #875, opened 2025-12-03
YMODEW doesn't do full integrity checking


tariqkurd-repo:
https://github.com/riscv/riscv-cheri/pull/666 was intended to make integrity checks even for all instruction - i.e. check all or none
YMODEW has missed out and just says "If the AP field can not have been produced by APERM"

PR #878, opened 2025-12-03
fix YMODEW to have full integrity checks

tariqkurd-repo:
Fix https://github.com/riscv/riscv-cheri/issues/875

### DIFF
--- a/src/cheri/insns/scmode_32bit.adoc
+++ b/src/cheri/insns/scmode_32bit.adoc
@@ -22,13 +22,13 @@ include::wavedrom/scmode.adoc[]
 Description::
 Copy `{cs1}` to `{cd}`.
 +
-Clear `{cd}.tag` if `{cs1}` is sealed.
+Set `{cd}.tag=0` if `{cs1}` is sealed, or if `{cs1}` fails any <<section_cap_integrity,integrity>> check.
 +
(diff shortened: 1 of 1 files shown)


src/cheri/insns/scmode_32bit.adoc
```diff
@@ -24,7 +24,7 @@ Copy `{cs1}` to `{cd}`.
 +
 Clear `{cd}.tag` if `{cs1}` is sealed.
 +
-If `{cs1}` grants <<x_perm>> and `{cs1}` 's <<AP-field>> could have been produced by <<CLRPERM>>, then update the <<m_bit>> of `{cd}` to:
+If `{cs1}` grants <<x_perm>> and `{cs1}` passes all <<section_cap_integrity,integrity>> checks, then update the <<m_bit>> of `{cd}` to:
```
arichardson, line 27: Most other instructions have the following statement as a final step, should we do the same?

Set rd.tag=0 if rs1 fails any integrity checks

tariqkurd-repo: yes good point - I'll change it

arichardson (approved):

tariqkurd-repo:
> I guess this is technically a change but was always the intention and is just an oversight.

yes - it's a bug fix - nothing more

PR #880, opened 2025-12-03
make JALR more readable

tariqkurd-repo:
no spec changes - just text which I find easier to interpret

### DIFF
--- a/src/cheri/insns/jalr_32bit.adoc
+++ b/src/cheri/insns/jalr_32bit.adoc
@@ -16,58 +16,33 @@ include::base_isa_extension.adoc[]
 
 #Begin changed since last ARC review#
 
+
 Description::
 Indirect jump to the target capability in `{cs1}` with an address offset.
++
+NOTE: The  description below contains three hooks for extending <<JALR_CHERI>> behavior, used by extensions to give a large degree of extensibility.
+ Unless an extension, such as <<section_rvy_sentry_insn_ext>>, is implemented which explicitly reference any of the hooks, then take no action for any of them.
 +
 Operationally, copy `{cs1}` to the target <<pcc>> and then...
 +
 . Compute the target <<pcc>> address in two steps, each using the semantics of the <<SCADDR>> instruction:
 .. Increment the address of the target <<pcc>> by the sign-extended 12-bit `offset`, and
 .. set bit zero of the target <<pcc>> address to zero.
-. Clear the {ctag} of the target <<pcc>> if the triple of its <<sec_cap_type>> value, `{cd}`, and `{cs1}` is not acceptable, as defined by the capability encoding.
-+
-NOTE: All such triples are acceptable to <<rv64y_cap_description>> and <<rv32y_cap_description>>.
-. If the target <<pcc>> is a <<sentry_cap>> (for <<rv32y_cap_description>>/<<rv64y_cap_description>>: if it has <<sec_cap_type>> `1`), then unseal it.
-. If the target <<pcc>> is (still) sealed, clear its {ctag}.
-+
-NOTE: This is not possible for <<rv32y_cap_description>>/<<rv64y_cap_description>>, given its two <<sec_cap_type>> values.
+. HOOK 1: Optionally clear the {ctag} of the target <<pcc>> depending on the <<sec_cap_type>> value and the numerical values of `{cd}` and `{cs1}`
+. Unseal the target <<pcc>> if it is a <<sentry_cap>>, or clear the {ctag} if it is any other sealed type.
 . If `{cd}≠{creg}0`, compute the return capability and install it to `{cd}`:
-.. add the width of this instruction to the current <<pcc>>, and
-.. if the capability encoding defines <<sentry_cap>> <<sec_cap_type>> values for return pointers, then seal the return capability as specified.
-+
-NOTE: <<rv32y_cap_description>>/<<rv64y_cap_description>> always seals return capabilities with <<sec_cap_type>> `1`.
+.. add the width of this instruction to the current <<pcc>> using the semantics of the <<SCADDR>> instructions, and
+.. HOOK 2: Optionally seal the return capability as a <<sentry_cap>> with a <<sec_cap_type>> defined by the implemented extensions.
+. HOOK 3: Optionally make other architectural state updates.
 . Jump to the target <<pcc>>.
 +
-NOTE: It is not possible for the target <<pcc>> to be sealed by the time it is jumped to.
-. Apply any capability encoding specified architectural semantics for the type of `{cs1}`.
-+
-NOTE: There are no such additional semantics defined by the capabilities of <<rv32y_cap_description>>/<<rv64y_cap_description>>. +
-
-+
-The above procedure for <<JALR_CHERI>> defines three hooks for use by the capability encoding:
-
-1. Architectural interpretation of zero or more particular <<sec_cap_type>> values when a capability is passed as the input to <<JALR_CHERI>>.
-   These <<sec_cap_type>> values are, by definition, <<sentry_cap>> types.
-   The "architectural interpretation" for these capabilities necessarily includes unsealing as part of installation but may include other effects as well.
-+
-NOTE: For example, CHERIoT offers sentries that update a hart's `mstatus.MIE` field when invoked.
-
-2. A subset of those <<sec_cap_type>> values for use by <<JALR_CHERI>> to seal the return capability installed in the link register.
-   A sentry return capability grants the right to return exactly to the instruction after the caller's call without relative offset.
-   Moreover, a sentry return capability does not grant the callee access to the memory pointed to by the caller's <<pcc>>.
-
-3. Conditional acceptability of the triple of the two _register selectors_ (`{cd}`, `{cs1}`) used in an instance of the instruction together with the <<sec_cap_type>> value of the capability in `{cs1}`.
-+
-NOTE: For <<rv32y_cap_description>>/<<rv64y_cap_description>>, there are no conditions and all such triples are acceptable;
-again the motivation is CHERIoT, which uses register selectors to distinguish between _forward_ and _backward_ control transfers, and restricts which <<sec_cap_type>> values are acceptable in each case.
-
 [NOTE]
 =====
 When a sealed capability is passed as the input to <<JALR_CHERI>>, its address must have bit zero clear and the instruction must have a zero offset, or the target <<pcc>> will have its {ctag} set to zero, since updates to its address are interpreted with <<SCADDR>> semantics.
 =====
-
 NOTE: A future extension may raise an exception on the <<JALR_CHERI>> instruction itself if the target <<pcc>> will raise a {cheri_excep_name_pc} at the target.
-
+NOTE: A sentry defines a secure function entry point, and as such the offset in the JALR instruction must be zero.
+ This is enforced by the use of <<SCADDR>> to add the offset which clears the tag of all sealed capabilities if the offset is non-zero.
 #End changed since last ARC review#
 
 Included in::
--- a/src/cheri/riscv-unpriv-integration.adoc
+++ b/src/cheri/riscv-unpriv-integration.adoc
@@ -1151,12 +1151,18 @@ NOTE: A future extension may raise an exception on the branch instruction itself
 
 #Begin changed since last ARC review#
 
-The {rvy_sentry_insn_ext_name} extension adds the <<SENTRY>> instruction to allow sealing arbitrary capabilities
-at a <<sec_cap_type>> value specified by the capability encoding.
+The {rvy_sentry_insn_ext_name} extension:
 
-NOTE: In principle, despite the names, the type used by <<SENTRY>> might not be a <<sentry_cap>> type, though that is not true of the capabilities of <<rv32y_cap_description>>/<<rv64y_cap_description>>.
+. Defines one <<sentry_cap>> type, the unrestricted sentry type with a <<sec_cap_type>> of `1`.
+.. These _unrestricted sentry_ capabilities can be used as immutable code pointers for both forward and backward control flow edges.
+NOTE: A future extension may define more restrictive _forward-only_ and _backward-only_ sentry capabilities that can only be used by calls and returns respectively.
+. Adds the <<SENTRY>> instruction to allow sealing capabilities as sentries with <<sec_cap_type>> of `1`.
 
-NOTE: Future extensions to {cheri_base_ext_name} (such as CHERIoT) support additional <<sec_cap_type>> values, so this instruction is part of an optional extension instead of the {cheri_base_ext_name} base ISA.
+=== Interaction with <<JALR_CHERI>>
+
+{rvy_sentry_insn_ext_name} adds sealing and unsealing behavior to <<JALR_CHERI>>:
+
+. For HOOK 2: Seal the return capability as a sentry.
 
 #End changed since last ARC review#
 
--- a/src/cheri/riscv-hybrid-integration.adoc
+++ b/src/cheri/riscv-hybrid-integration.adoc
@@ -92,6 +92,8 @@ The <<m_bit>> of <<pcc>> can be updated by the instructions listed in <<tab_cher
 | <<MODESW_CAP>> |{cheri_int_mode_name}| {MODESW_CAP_DESC}
 |=======================
 
+NOTE: When <<JALR_CHERI>> copies `rs1` into <<pcc>> it includes copying the <<m_bit>> and so setting the <<cheri_execution_mode>> of the target instruction.
+
 The <<m_bit>> of a <<x_perm>>-granting capability can be read and written by the instructions listed in <<tab_hybrid_summary>>:
 
 .{cheri_default_ext_name} instructions to observe and update the mode in a capability


src/cheri/insns/jalr_32bit.adoc
```diff
@@ -16,57 +16,50 @@ include::base_isa_extension.adoc[]
...
 +
 The above procedure for <<JALR_CHERI>> defines three hooks for use by the capability encoding:
-
-1. Architectural interpretation of zero or more particular <<sec_cap_type>> values when a capability is passed as the input to <<JALR_CHERI>>.
-   These <<sec_cap_type>> values are, by definition, <<sentry_cap>> types.
-   The "architectural interpretation" for these capabilities necessary includes unsealing as part of installation but may include other effects as well.
 +
-NOTE: For example, CHERIoT offers sentries that update a hart's `mstatus.MIE` field when invoked.
-
+1. Conditional acceptability of the triple of the two _register selectors_ (`{cd}`, `{cs1}`) used in an instance of the instruction together with the <<sec_cap_type>> value of the capability in `{cs1}`.
```
arichardson: I guess we are repeating the hooks again here, maybe it would be clearer to just extend the hook text above and add the rationale as a non-normative note?

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -16,57 +16,50 @@ include::base_isa_extension.adoc[]
...
+   The "architectural interpretation" for these capabilities necessarily includes unsealing as part of installation but may include other effects as well.
+.. NOTE: For example, CHERIoT offers sentries that update a hart's `mstatus.MIE` field when invoked.
 +
-NOTE: For <<rv32y_cap_description>>/<<rv64y_cap_description>>, there are no conditions and all such triples are acceptable;
-again the motivation is CHERIoT, which uses register selectors to distinguish between _forward_ and _backward_ control transfers, and restricts which <<sec_cap_type>> values are acceptable in each case.
-
 [NOTE]
 =====
-When a sealed capability is passed as the input to <<JALR_CHERI>>, its address must have bit zero clear and the instruction must have a zero offset, or the target <<pcc>> will have its {ctag} set to zero, since updates to its address are interpreted with <<SCADDR>> semantics.
+When a sealed capability (i.e. a sentry) is passed as the input to <<JALR_CHERI>>, its address must have bit zero clear and the instruction must have a zero offset, or the target <<pcc>> will have its {ctag} set to zero, since updates to its address are interpreted with <<SCADDR>> semantics.
```
```suggestion
When a sealed capability is passed as the input to <<JALR_CHERI>>, its address must have bit zero clear and the instruction must have a zero offset, or the target <<pcc>> will have its {ctag} set to zero, since updates to its address are interpreted with <<SCADDR>> semantics.
```
Right now we only allow sentries here, but I would just drop this i.e. since it applies to all immutable capabilities.

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -16,57 +16,50 @@ include::base_isa_extension.adoc[]
 
 #Begin changed since last ARC review#
 
+
 Description::
 Indirect jump to the target capability in `{cs1}` with an address offset.
 +
+NOTE: The  description below contains three hooks for extending <<JALR_CHERI>> behavior, as defined by the specific encoding format, to give a large degree of extensibility.
```
nwf:
```suggestion
NOTE: The  description below contains three hooks for extending <<JALR_CHERI>> behavior, used by extensions and/or encoding formats to give a large degree of extensibility.
```

For CHERIoT, the hooks are used by Zycheriot and Smycheriot, not the _LyenccheriotN formats.  The rvNNy_cap_description sections don't have their semantics separated out into something analogous to Zycheriot.

That probably means phrasing updates throughout?

src/cheri/insns/jalr_32bit.adoc
tariqkurd-repo: maybe - but that was a lot of the problem with the text before - there was so much extra information that it was hard to see the actual procedure to follow
I think I've done what I can with this one - I find it clearer than before.

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -16,56 +16,48 @@ include::base_isa_extension.adoc[]
...
 Operationally, copy `{cs1}` to the target <<pcc>> and then...
 +
 . Compute the target <<pcc>> address in two steps, each using the semantics of the <<SCADDR>> instruction:
 .. Increment the address of the target <<pcc>> by the sign-extended 12-bit `offset`, and
 .. set bit zero of the target <<pcc>> address to zero.
-. Clear the {ctag} of the target <<pcc>> if the triple of its <<sec_cap_type>> value, `{cd}`, and `{cs1}` is not acceptable, as defined by the capability encoding.
-+
-NOTE: All such triples are acceptable to <<rv64y_cap_description>> and <<rv32y_cap_description>>.
-. If the target <<pcc>> is a <<sentry_cap>> (for <<rv32y_cap_description>>/<<rv64y_cap_description>>: if it has <<sec_cap_type>> `1`), then unseal it.
+. HOOK 1: Clear the {ctag} of the target <<pcc>> if the triple of its <<sec_cap_type>> value, `{cd}`, and `{cs1}` is not acceptable, as defined by the capability encoding.
```
nwf:
```suggestion
. HOOK 1: Clear the {ctag} of the target <<pcc>> if the triple of its <<sec_cap_type>> value, `{cd}`, and `{cs1}` is not acceptable, as defined by the capability encoding or an extension.
```

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -16,56 +16,48 @@ include::base_isa_extension.adoc[]
...
+. If the target <<pcc>> is a <<sentry_cap>>, then unseal it.
 . If the target <<pcc>> is (still) sealed, clear its {ctag}.
-+
-NOTE: This is not possible for <<rv32y_cap_description>>/<<rv64y_cap_description>>, given its two <<sec_cap_type>> values.
 . If `{cd}≠{creg}0`, compute the return capability and install it to `{cd}`:
 .. add the width of this instruction to the current <<pcc>>, and
-.. if the capability encoding defines <<sentry_cap>> <<sec_cap_type>> values for return pointers, then seal the return capability as specified.
-+
-NOTE: <<rv32y_cap_description>>/<<rv64y_cap_description>> always seals return capabilities with <<sec_cap_type>> `1`.
+.. HOOK 2: if the capability encoding defines <<sentry_cap>> <<sec_cap_type>> values for return pointers, then seal the return capability as specified.
```
nwf:
```suggestion
.. HOOK 2: if the capability encoding or an extension defines <<sentry_cap>> <<sec_cap_type>> values for return pointers, then seal the return capability as specified.
```

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -16,56 +16,48 @@ include::base_isa_extension.adoc[]
...
 . Jump to the target <<pcc>>.
-+
-NOTE: It is not possible for the target <<pcc>> to be sealed by the time it is jumped to.
-. Apply any capability encoding specified architectural semantics for the type of `{cs1}`.
-+
-NOTE: There are no such additional semantics defined by the capabilities of <<rv32y_cap_description>>/<<rv64y_cap_description>>. +
-
+.. NOTE: It is not possible for the target <<pcc>> to be sealed by the time it is jumped to.
+. HOOK 3: Apply any capability encoding specified architectural semantics for the type of `{cs1}`.
```
nwf:
```suggestion
. HOOK 3: Apply any capability encoding- or extension-specified architectural semantics for the type of `{cs1}`.
```

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -16,56 +16,48 @@ include::base_isa_extension.adoc[]
...
+.. NOTE: CHERIoT uses register selectors to distinguish between _forward_ and _backward_ control transfers, and restricts which <<sec_cap_type>> values are acceptable in each case.
 2. A subset of those <<sec_cap_type>> values for use by <<JALR_CHERI>> to seal the return capability installed in the link register.
    A sentry return capability grants the right to return exactly to the instruction after the caller's call without relative offset.
    Moreover, a sentry return capability does not grant the callee access to the memory pointed to by the caller's <<pcc>>.
-
-3. Conditional acceptability of the triple of the two _register selectors_ (`{cd}`, `{cs1}`) used in an instance of the instruction together with the <<sec_cap_type>> value of the capability in `{cs1}`.
+.. NOTE: For <<rv32y_cap_description>>/<<rv64y_cap_description>>, sentries have <<sec_cap_type>> set to `1` and sealed return capabilities also have type <<sec_cap_type>> set to `1`.
+ CHERIoT formats support more <<sec_cap_type>> values.
```
nwf:
```suggestion
 CHERIoT formats support more <<sec_cap_type>> values, some of which are designated as <<sentry_cap,sentry>> types by the <<sec_zycheriot_unpriv>> extension.
```

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -16,56 +16,48 @@ include::base_isa_extension.adoc[]
...
+3. Architectural interpretation of zero or more particular <<sec_cap_type>> values when a capability is passed as the input to <<JALR_CHERI>>.
+   These <<sec_cap_type>> values are, by definition, <<sentry_cap>> types.
+   The "architectural interpretation" for these capabilities necessarily includes unsealing as part of installation but may include other effects as well.
+.. NOTE: For example, CHERIoT offers sentries that update a hart's `mstatus.MIE` field when invoked.
```
nwf:
```suggestion
.. NOTE: For example, CHERIoT offers sentries that update a hart's `mstatus.MIE` field when invoked; see the the {cheriot_priv_ext_name} extension.
```

src/cheri/insns/jalr_32bit.adoc
arichardson: I wonder if this should be just be extension? The format defined which otypes can be encoded but I think this should be part of the zycheriot/zycfi extension?

src/cheri/insns/jalr_32bit.adoc
arichardson: I wonder if this should be extension again? Then it could be dependent on Zys/Zycheriot/Zycfi?

src/cheri/insns/jalr_32bit.adoc
nwf: I'd like to split out the semantics from the default encoding chapters, too.

src/cheri/insns/jalr_32bit.adoc
tariqkurd-repo: that is a good point - it should be part of Zysentry - which uses the hook - putting the hooks and extensions into practice

because it's perfectly possible (even if misguided) to have a cheri system without sentries

so this _is_ a spec bug

"If the target pc is a sentry capability, then unseal it." needs another hook

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -16,56 +16,30 @@ include::base_isa_extension.adoc[]
 
 #Begin changed since last ARC review#
 
+
 Description::
 Indirect jump to the target capability in `{cs1}` with an address offset.
 +
+NOTE: The  description below contains three hooks for extending <<JALR_CHERI>> behavior, used by extensions and/or encoding formats to give a large degree of extensibility.
+ Unless an extension and/or encoding format, such as <<section_rvy_sentry_insn_ext>>,  is implemented which explicitly reference the hooks, then take no action for any of them.
```
```suggestion
NOTE: The  description below contains three hooks for extending <<JALR_CHERI>> behavior, used by extensions to give a large degree of extensibility.
 Unless an extension, such as <<section_rvy_sentry_insn_ext>>, is implemented that explicitly reference the hooks, then take no action for any of them.
```

I thought the conclusion was to avoid mentioning encodings here, since we are only discussing semantics here. @nwf does the suggested wording sound good to you?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -1151,12 +1151,18 @@ NOTE: A future extension may raise an exception on the branch instruction itself
...
+. Defines the <<sec_cap_type>> of the sentry type to be type `1`.
+. Adds the <<SENTRY>> instruction to allow sealing capabilities as sentries with <<sec_cap_type>> of `1`.
 
-NOTE: Future extensions to {cheri_base_ext_name} (such as CHERIoT) support additional <<sec_cap_type>> values, so this instruction is part of an optional extension instead of the {cheri_base_ext_name} base ISA.
+=== Interaction with <<JALR_CHERI>>
+
+{rvy_sentry_insn_ext_name} adds sealing and unsealing behavior to <<JALR_CHERI>>:
+
+. For HOOK 1: No action.
+. For HOOK 2: If the capability in {cs1} is sealed with a sentry of type `1`, then unseal before installing in <<pcc>>.
```
```suggestion
. For HOOK 1: If the capability in {cs1} is sealed and the immediate argument is non-zero, then clear the tag of the target <<pcc>>.
. For HOOK 2: If the capability in {cs1} is sealed with a sentry of type `1`, then unseal before installing in <<pcc>>.
```
I think we lost the non-zero offset invalidates behaviour.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -1151,12 +1151,18 @@ NOTE: A future extension may raise an exception on the branch instruction itself
 
 #Begin changed since last ARC review#
 
-The {rvy_sentry_insn_ext_name} extension adds the <<SENTRY>> instruction to allow sealing arbitrary capabilities
-at a <<sec_cap_type>> value specified by the capability encoding.
+The {rvy_sentry_insn_ext_name} extension:
 
-NOTE: In principle, despite the names, the type used by <<SENTRY>> might not be a <<sentry_cap>> type, though that is not true of the capabilities of <<rv32y_cap_description>>/<<rv64y_cap_description>>.
+. Defines the <<sec_cap_type>> of the sentry type to be type `1`.
+. Adds the <<SENTRY>> instruction to allow sealing capabilities as sentries with <<sec_cap_type>> of `1`.
```
arichardson, line 1159:
```suggestion
The {rvy_sentry_insn_ext_name} extension:

. Defines the unrestricted sentry <<sec_cap_type>> with `CT=1`. 
. Adds the <<SENTRY>> instruction to allow sealing capabilities as sentries with <<sec_cap_type>> of `1`.
```

Maybe this? I'm sure @nwf can come up with some better wording than me :)

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: no - that's in "use YADDRW to do the increment" as if it's sealed then it'll clear the tag

src/cheri/insns/jalr_32bit.adoc
tariqkurd-repo: it's good for me

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: please do! and get it merge today ideally!

src/cheri/insns/jalr_32bit.adoc
nwf: Yeah, that seems somewhat OK, treating Zysentry as the extension that factors out sentry semantics from the representations in the two "default" encodings... but we should also adjust the prose in 2.3.9 analogously.

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -16,56 +16,30 @@ include::base_isa_extension.adoc[]
...
 . If `{cd}≠{creg}0`, compute the return capability and install it to `{cd}`:
 .. add the width of this instruction to the current <<pcc>>, and
-.. if the capability encoding defines <<sentry_cap>> <<sec_cap_type>> values for return pointers, then seal the return capability as specified.
-+
-NOTE: <<rv32y_cap_description>>/<<rv64y_cap_description>> always seals return capabilities with <<sec_cap_type>> `1`.
+.. HOOK 3: Optionally seal the return capability.
 . Jump to the target <<pcc>>.
 +
-NOTE: It is not possible for the target <<pcc>> to be sealed by the time it is jumped to.
-. Apply any capability encoding specified architectural semantics for the type of `{cs1}`.
```
nwf, line 42: I am _exceedingly_ wary of removing this hook.  Smycheriot is specifically an _extension_ to the _privileged_ ISA and so, absent this hook, seems doubly barred from altering JALR's semantics.

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -16,56 +16,30 @@ include::base_isa_extension.adoc[]
...
 Operationally, copy `{cs1}` to the target <<pcc>> and then...
 +
 . Compute the target <<pcc>> address in two steps, each using the semantics of the <<SCADDR>> instruction:
 .. Increment the address of the target <<pcc>> by the sign-extended 12-bit `offset`, and
 .. set bit zero of the target <<pcc>> address to zero.
-. Clear the {ctag} of the target <<pcc>> if the triple of its <<sec_cap_type>> value, `{cd}`, and `{cs1}` is not acceptable, as defined by the capability encoding.
-+
-NOTE: All such triples are acceptable to <<rv64y_cap_description>> and <<rv32y_cap_description>>.
-. If the target <<pcc>> is a <<sentry_cap>> (for <<rv32y_cap_description>>/<<rv64y_cap_description>>: if it has <<sec_cap_type>> `1`), then unseal it.
+. HOOK 1: Optionally clear the {ctag} of the target <<pcc>>.
```
nwf: That's... really quite vague, whereas the prose it replaces attempted to indicate what the "parameters" to this hook were.

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -16,56 +16,30 @@ include::base_isa_extension.adoc[]
...
+. HOOK 2: Optionally unseal the target <<pcc>> if sealed.
```
nwf: This bit of behavior should not need to be hooked; the base's definition of sentry ocap types ought to be sufficient.  This is _why_ I kept the definition of sentry semantics in the base.

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -16,56 +16,30 @@ include::base_isa_extension.adoc[]
...
 . If the target <<pcc>> is (still) sealed, clear its {ctag}.
-+
-NOTE: This is not possible for <<rv32y_cap_description>>/<<rv64y_cap_description>>, given its two <<sec_cap_type>> values.
 . If `{cd}≠{creg}0`, compute the return capability and install it to `{cd}`:
 .. add the width of this instruction to the current <<pcc>>, and
-.. if the capability encoding defines <<sentry_cap>> <<sec_cap_type>> values for return pointers, then seal the return capability as specified.
-+
-NOTE: <<rv32y_cap_description>>/<<rv64y_cap_description>> always seals return capabilities with <<sec_cap_type>> `1`.
+.. HOOK 3: Optionally seal the return capability.
```
nwf: Again, this loses the specificity of at what the hook is allowed to look.

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -16,56 +16,30 @@ include::base_isa_extension.adoc[]
...
-+
-NOTE: All such triples are acceptable to <<rv64y_cap_description>> and <<rv32y_cap_description>>.
-. If the target <<pcc>> is a <<sentry_cap>> (for <<rv32y_cap_description>>/<<rv64y_cap_description>>: if it has <<sec_cap_type>> `1`), then unseal it.
+. HOOK 1: Optionally clear the {ctag} of the target <<pcc>>.
+. HOOK 2: Optionally unseal the target <<pcc>> if sealed.
 . If the target <<pcc>> is (still) sealed, clear its {ctag}.
-+
-NOTE: This is not possible for <<rv32y_cap_description>>/<<rv64y_cap_description>>, given its two <<sec_cap_type>> values.
 . If `{cd}≠{creg}0`, compute the return capability and install it to `{cd}`:
 .. add the width of this instruction to the current <<pcc>>, and
```
nwf: While here, this should be amended to say "using the semantics of the `<<SCADDR>>` instruction" since the addition might, in principle, take us out of representable bounds and we need to make sure that we don't construct an incorrectly reinterpreted capability.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -1151,12 +1151,18 @@ NOTE: A future extension may raise an exception on the branch instruction itself
 
 #Begin changed since last ARC review#
 
-The {rvy_sentry_insn_ext_name} extension adds the <<SENTRY>> instruction to allow sealing arbitrary capabilities
-at a <<sec_cap_type>> value specified by the capability encoding.
+The {rvy_sentry_insn_ext_name} extension:
 
-NOTE: In principle, despite the names, the type used by <<SENTRY>> might not be a <<sentry_cap>> type, though that is not true of the capabilities of <<rv32y_cap_description>>/<<rv64y_cap_description>>.
+. Defines the <<sec_cap_type>> of the sentry type to be type `1`.
```
nwf: "the CT-field of the sentry type" does not type-check.

I don't particularly like it, but the phrasing we talked about in the meeting has the Zysentry extension designating `1` as _a_ sentry type (relevant solely to X-bearing caps, solely consumed by `JALR`) and as an ambiently-available type (relevant to X-granting and X-prohibiting caps, which might be subsequently used with `YSUNSEAL`; ambientness has interactions with `YBLD`).

src/cheri/rvy-cheriot-isa-unpriv.adoc
```diff
@@ -35,12 +34,8 @@ we have not yet found a compelling reason to formally specify this composition.
...
 [NOTE]
 =====
-While {cheriot_unpriv_ext_name} is nominally compatible with
-<<section_rvy_sentry_insn_ext>>,
-the operating system written for link:https://cheriot.org[CHERIoT]
-has a security model that presumes the absence of ambient sealing,
-and so this specification does not
-define any <<sec_cap_type_ambient,ambiently available>> sentry types.
+{cheriot_unpriv_ext_name} is incompatible with
+<<section_rvy_sentry_insn_ext>> as <<sec_cap_type>> of `1` is not a sentry type.
```
nwf: Zycheriot (specifically, §14.1.2 "Capability Types" of the present riscv-cheri doc) _does_ designate otype `1` as a sentry type (which, again, the base defines as "JALR unseals this type"; see §2.3.9).  In fact, it's even _almost_ "the same" sentry as in the "default" encoding's implicit semantics (specifically, it's our forward-arc interrupt-posture-inheriting sentry type).  That said, there are some significant differences: CHERIoT does not use "HOOK 2" to have JALR _produce_ capabilities with type `1` (unlike the default's, where forward- and backward-arc types aren't distinguished) and CHERIoT's encodings cannot represent type `1` capabilities that do not grant X-permission.

src/cheri/insns/jalr_32bit.adoc
tariqkurd-repo: the idea is to leave it vague - so that future extensions can clear the tag for other reasons - without being limited to the current CHERIoT requirements

src/cheri/insns/jalr_32bit.adoc
tariqkurd-repo: yes good point, the addition must be cap friendly

src/cheri/rvy-cheriot-isa-unpriv.adoc
tariqkurd-repo: The solution here may well be to stop trying to make JALR generic and simply to have it in an extension with different versions for the different encoding formats.
If we can't work out how to make this work between ourselves we're not going to get it through review.
So maybe Zysentry has one vesion of JALR, and there's another for CHERIoT.
And CHERI implementations _must_ choose a version of JALR.
Then we can all write exactly what we want JALR to do without adding the hooks.

Would that solve it?

src/cheri/riscv-unpriv-integration.adoc
arichardson: Ah missed that, but being explicit doesn't hurt

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -16,56 +16,30 @@ include::base_isa_extension.adoc[]
...
-+
-NOTE: For <<rv32y_cap_description>>/<<rv64y_cap_description>>, there are no conditions and all such triples are acceptable;
-again the motivation is CHERIoT, which uses register selectors to distinguish between _forward_ and _backward_ control transfers, and restricts which <<sec_cap_type>> values are acceptable in each case.
-
 [NOTE]
 =====
 When a sealed capability is passed as the input to <<JALR_CHERI>>, its address must have bit zero clear and the instruction must have a zero offset, or the target <<pcc>> will have its {ctag} set to zero, since updates to its address are interpreted with <<SCADDR>> semantics.
 =====
-
 NOTE: A future extension may raise an exception on the <<JALR_CHERI>> instruction itself if the target <<pcc>> will raise a {cheri_excep_name_pc} at the target.
```
```suggestion
NOTE: A sentry defines a secure function entry point, and as such the offset in the JALR instruction must be zero.
 This is enforced by the use of <<SCADDR>> to add the offset which clears the tag of all sealed capabilities if the offset is non-zero.
```

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: suggestion added

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: suggestion can't be applied..... I'll resolve this and let's try again

src/cheri/insns/jalr_32bit.adoc
tariqkurd-repo: how would you specify it?

src/cheri/insns/jalr_32bit.adoc
tariqkurd-repo: hook 4 reinstated - before jumping to the target PC

src/cheri/insns/jalr_32bit.adoc
arichardson:
```suggestion
. HOOK 1: Optionally clear the {ctag} of the target <<pcc>> depending on the <<sec_cap_type>> value and the numerical values of `{cd}` and `{cs1}`
```

arichardson: Maybe something like that

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -16,56 +16,31 @@ include::base_isa_extension.adoc[]
...
 +
 . Compute the target <<pcc>> address in two steps, each using the semantics of the <<SCADDR>> instruction:
 .. Increment the address of the target <<pcc>> by the sign-extended 12-bit `offset`, and
 .. set bit zero of the target <<pcc>> address to zero.
-. Clear the {ctag} of the target <<pcc>> if the triple of its <<sec_cap_type>> value, `{cd}`, and `{cs1}` is not acceptable, as defined by the capability encoding.
-+
-NOTE: All such triples are acceptable to <<rv64y_cap_description>> and <<rv32y_cap_description>>.
-. If the target <<pcc>> is a <<sentry_cap>> (for <<rv32y_cap_description>>/<<rv64y_cap_description>>: if it has <<sec_cap_type>> `1`), then unseal it.
+. HOOK 1: Optionally clear the {ctag} of the target <<pcc>>.
+. HOOK 2: Optionally unseal the target <<pcc>> if sealed.
```
```suggestion
. HOOK 2: Optionally unseal the target <<pcc>> if the target <<pcc>> is a <<sentry_cap>> with an acceptable <<sec_cap_type>>. 
```

arichardson: Or do we want to just say unseal all sentries? Then the tag clearing hook can deal with the cap types

arichardson: And then as Wes said it doesn't need to be a hook anymore. We can also merge the next statement into this as follows:
```suggestion
. Unseal the target <<pcc>> if it is a <<sentry_cap>>, or clear the {ctag} if it is any other sealed type.
```

src/cheri/insns/jalr_32bit.adoc
arichardson:
```suggestion
.. HOOK 3: Optionally seal the return capability as a <<sentry_cap>> with a <<sec_cap_type>> defined by the implemented extensions.
```

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -16,56 +16,31 @@ include::base_isa_extension.adoc[]
...
-+
-NOTE: This is not possible for <<rv32y_cap_description>>/<<rv64y_cap_description>>, given its two <<sec_cap_type>> values.
 . If `{cd}≠{creg}0`, compute the return capability and install it to `{cd}`:
-.. add the width of this instruction to the current <<pcc>>, and
-.. if the capability encoding defines <<sentry_cap>> <<sec_cap_type>> values for return pointers, then seal the return capability as specified.
-+
-NOTE: <<rv32y_cap_description>>/<<rv64y_cap_description>> always seals return capabilities with <<sec_cap_type>> `1`.
+.. add the width of this instruction to the current <<pcc>> using the semantics of the <<SCADDR>> instructions, and
+.. HOOK 3: Optionally seal the return capability.
+. HOOK 4: Optionally make other architectural state updates.
```
arichardson: Could add a note here that Zyhybrid makes use of this to switch between. RVY and RVI mode.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -1151,12 +1151,19 @@ NOTE: A future extension may raise an exception on the branch instruction itself
 
 #Begin changed since last ARC review#
 
-The {rvy_sentry_insn_ext_name} extension adds the <<SENTRY>> instruction to allow sealing arbitrary capabilities
-at a <<sec_cap_type>> value specified by the capability encoding.
+The {rvy_sentry_insn_ext_name} extension:
 
-NOTE: In principle, despite the names, the type used by <<SENTRY>> might not be a <<sentry_cap>> type, though that is not true of the capabilities of <<rv32y_cap_description>>/<<rv64y_cap_description>>.
+. Defines the <<sec_cap_type>> of the sentry type to be `1`.
```
```suggestion
. Defines one <<sentry_cap>> type, the unrestricted sentry type with a <<sec_cap_type>> of `1`.
.. These _unrestricted sentry_ capabilities can be used as immutable code pointers for both forward and backward control flow edges.
NOTE: A future extension may define more restrictive _forward-only_ and _backward-only_ sentry capabilities that can only be used by calls and returns respectively.
```

Something like this?

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -16,56 +16,31 @@ include::base_isa_extension.adoc[]
 
 #Begin changed since last ARC review#
 
+
 Description::
 Indirect jump to the target capability in `{cs1}` with an address offset.
 +
+NOTE: The  description below contains three hooks for extending <<JALR_CHERI>> behavior, used by extensions and/or encoding formats to give a large degree of extensibility.
+ Unless an extension and/or encoding format, such as <<section_rvy_sentry_insn_ext>>,  is implemented which explicitly reference the hooks, then take no action for any of them.
```
```suggestion
 Unless an extension, such as <<section_rvy_sentry_insn_ext>>, is implemented which explicitly reference the hooks, then take no action for any of them.
```

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -16,56 +16,31 @@ include::base_isa_extension.adoc[]
 
 #Begin changed since last ARC review#
 
+
 Description::
 Indirect jump to the target capability in `{cs1}` with an address offset.
 +
+NOTE: The  description below contains three hooks for extending <<JALR_CHERI>> behavior, used by extensions and/or encoding formats to give a large degree of extensibility.
```
```suggestion
NOTE: The  description below contains three hooks for extending <<JALR_CHERI>> behavior, used by extensions to give a large degree of extensibility.
```

src/cheri/rvy-cheriot-isa-unpriv.adoc
arichardson: I would just drop this diff from this PR. The current wording looks fine as is to me.

arichardson (approved):

src/cheri/rvy-cheriot-isa-unpriv.adoc
tariqkurd-repo:
```suggestion
While {cheriot_unpriv_ext_name} is nominally compatible with
<<section_rvy_sentry_insn_ext>>,
```

tariqkurd-repo:
```suggestion
While {cheriot_unpriv_ext_name} is nominally compatible with
<<section_rvy_sentry_insn_ext>>,
```

src/cheri/rvy-cheriot-isa-unpriv.adoc
```diff
@@ -22,8 +22,7 @@ counterparts found in the {cheriot_priv_ext_name} extension.
 {cheriot_unpriv_ext_name} assumes the presence of
 both the <<section_zylevels1>> and <<section_zyseal>> extensions.
 The present specification presumes the *absence*
-of both the <<section_cheri_hybrid_ext,{cheri_default_ext_name}>> and
-<<section_rvy_sentry_insn_ext>> extensions.
+of <<section_cheri_hybrid_ext,{cheri_default_ext_name}>>.
```
tariqkurd-repo:
```suggestion
of <<section_cheri_hybrid_ext,{cheri_default_ext_name}>> and
<<section_rvy_sentry_insn_ext>> extensions.
```

src/cheri/rvy-cheriot-isa-unpriv.adoc
```diff
@@ -22,7 +22,7 @@ counterparts found in the {cheriot_priv_ext_name} extension.
 {cheriot_unpriv_ext_name} assumes the presence of
 both the <<section_zylevels1>> and <<section_zyseal>> extensions.
 The present specification presumes the *absence*
-of both the <<section_cheri_hybrid_ext,{cheri_default_ext_name}>> and
+of <<section_cheri_hybrid_ext,{cheri_default_ext_name}>> and
```
tariqkurd-repo:
```suggestion
of  both the <<section_cheri_hybrid_ext,{cheri_default_ext_name}>> and
```

src/cheri/rvy-cheriot-isa-unpriv.adoc
```diff
@@ -22,7 +22,7 @@ counterparts found in the {cheriot_priv_ext_name} extension.
 {cheriot_unpriv_ext_name} assumes the presence of
 both the <<section_zylevels1>> and <<section_zyseal>> extensions.
 The present specification presumes the *absence*
-of both the <<section_cheri_hybrid_ext,{cheri_default_ext_name}>> and
+of  both the <<section_cheri_hybrid_ext,{cheri_default_ext_name}>> and
```
tariqkurd-repo:
```suggestion
of both the <<section_cheri_hybrid_ext,{cheri_default_ext_name}>> and
```

src/cheri/rvy-cheriot-isa-unpriv.adoc
```diff
@@ -37,10 +37,6 @@ we have not yet found a compelling reason to formally specify this composition.
 =====
 While {cheriot_unpriv_ext_name} is nominally compatible with
 <<section_rvy_sentry_insn_ext>>,
```
tariqkurd-repo, line 39:
```suggestion
<<section_rvy_sentry_insn_ext>>,
the operating system written for link:https://cheriot.org[CHERIoT]
has a security model that presumes the absence of ambient sealing,
and so this specification does not
define any <<sec_cap_type_ambient,ambiently available>> sentry types.
```

PR #890, opened 2025-12-11
fix overlapping encoding with ybndsrw

tariqkurd-repo:
fix encoding conflict


tariqkurd-repo:
I've now removed the encoding from riscv-opcodes because we're not putting this instruction forwards for ARC review yet. WHen we do I'll reinstate it.

PR #900, opened 2025-12-16
Clarify behaviour of prefetch instructions

tariqkurd-repo:
Clarify that prefetching is optional, not required in line with the main RISC-V spec

### DIFF
--- a/src/cheri/riscv-hybrid-integration.adoc
+++ b/src/cheri/riscv-hybrid-integration.adoc
@@ -35,7 +35,6 @@ All <<rvy_insn_table>> instructions, and associated CSRs, are available in addit
 +
 The authorizing capability for memory access is <<ddc>> (as opposed to `rs1`).
 That is, all memory accesses, including <<PREFETCH_W_CHERI>> and <<PREFETCH_R_CHERI>>, are implicitly authorized by <<ddc>> and only the memory address is sourced from `rs1`.
-+
 <<PREFETCH_I_CHERI>> is the exception to this rule, as <<pcc>> authorizes it.
 +
 NOTE: <<ddc>> is also used to authorize {cheri_base_ext_name} specific memory instructions such as <<LOAD_CAP>> and <<STORE_CAP>>.


jrtc27:
It’s authorised by PCC in the same way that capability mode AUIPC reads PCC. If you’re speculating on PCC then that’s what you use. Though you want to be careful about speculative execution attacks…

Also, can we please stop changing the spec in substantive ways? Maybe the branch one is a clarification and nothing more, but PREFETCH.I is not. The spec is supposed to be frozen for ARC review.

tariqkurd-repo:
> In at least some proposed uses of hybrid mode, PCC and DDC are not "infinite" capabilities and are likely to be disjoint, so I'd think anything that looks like "instruction address" should be considered PCC-authorized.

Yes - I agree - and for something like AUIPC[C] then clearly PCC needs to be used - as Jess said above.

Prefetches are hints though and for prefetch.i in integer mode only, which is quite niche, it's not worth wiring up access to the speculative PCC to every load unit in a high performance implementation.

Therefore my proposal is to change it to a recommendation that PCC is used for authorisation, but not a normative rule, so a simpler implementation can certainly keep the authorisation, but a more complex one which doesn't have PCC available in the correct place can choose not to. Even with the normative rule in place you can't prove the behaviour due to hardware prefetchers potentially fetching the same line.

davidchisnall:
I agree, an implementation that decides to deliberately introduce attack vectors for side channel attacks should be free to do so, just as it is with the cache flush instructions.

tariqkurd-repo:
> I agree, an implementation that decides to deliberately introduce attack vectors for side channel attacks should be free to do so, just as it is with the cache flush instructions.

Yes - it becomes an implementation choice - not an architectural requirement.
There are many tradeoffs between resiliance to side channel attacks and performance in high performance designs, this is another to add to the list.

src/cheri/riscv-hybrid-integration.adoc
```diff
@@ -34,8 +34,10 @@ NOTE: Instructions which are _modified_ on an {cheri_base_ext_name} architecture
 All <<rvy_insn_table>> instructions, and associated CSRs, are available in addition to RVI/RVE and all supported non-CHERI extensions.
 +
 The authorizing capability for memory access is <<ddc>> (as opposed to `rs1`).
-That is, all memory accesses, including <<PREFETCH_W_CHERI>> and <<PREFETCH_R_CHERI>>, are implicitly authorized by <<ddc>> and only the memory address is sourced from `rs1`.
-<<PREFETCH_I_CHERI>> is the exception to this rule, as <<pcc>> authorizes it.
+That is, all memory accesses, including <<PREFETCH_W_CHERI>>, <<PREFETCH_R_CHERI>> are implicitly authorized by <<ddc>> and only the memory address is sourced from `rs1`.
++
+NOTE: It is recommended to authorize the prefetched address of <<PREFETCH_I_CHERI>> against the <<pcc>>.
```
```suggestion
NOTE: It is recommended to authorize the prefetched address of <<PREFETCH_I_CHERI>> against <<pcc>> instead of <<ddc>>.
```

tariqkurd-repo:
> I think maybe we should invert the wording here: prefetch.i _should_ be authorized by PCC, but to allow easier implementation in complex microarchitectures it _may_ be authorized by ddc instead. The downside of using ddc is of course that it might not cover the code segment which then renders the instruction useless. For now ddc mostly covers the PCC range but if you use it for hybrid sandboxes I imagine it generally won't...
> 
> Not sure what the best solution here is? No checks at all? Use a stale/predicted PCC?

If the arch expects PCC and the implementation uses DDC instead it's a bit unfortunate, but then it's only a prefetch - but I don't really see why DDC should be expected to cover the instruction memory.

I'm ok with skipping the check if PCC isn't available, so just make it optional - I'll try out some wording.

src/cheri/riscv-hybrid-integration.adoc
```diff
@@ -34,8 +34,12 @@ NOTE: Instructions which are _modified_ on an {cheri_base_ext_name} architecture
...
 +
 The authorizing capability for memory access is <<ddc>> (as opposed to `rs1`).
-That is, all memory accesses, including <<PREFETCH_W_CHERI>> and <<PREFETCH_R_CHERI>>, are implicitly authorized by <<ddc>> and only the memory address is sourced from `rs1`.
-<<PREFETCH_I_CHERI>> is the exception to this rule, as <<pcc>> authorizes it.
+That is, all memory accesses, including <<PREFETCH_W_CHERI>>, <<PREFETCH_R_CHERI>> are implicitly authorized by <<ddc>> and only the memory address is sourced from `rs1`.
++
+The prefetched address of <<PREFETCH_I_CHERI>> _should_ be authorized against <<pcc>>, but the check _may_ be avoided in the implementation.
+
+NOTE: High performance implementations may not have access to the <<pcc>> to authorize the access, as <<pcc>> bounds are not otherwise required in the load/store unit.
+ Therefore this check is non-architectural only a recommendation.
```
tariqkurd-repo:
```suggestion
 Therefore this check is non-architectural and only a recommendation.
```

src/cheri/riscv-hybrid-integration.adoc
```diff
@@ -34,8 +34,12 @@ NOTE: Instructions which are _modified_ on an {cheri_base_ext_name} architecture
...
 +
 The authorizing capability for memory access is <<ddc>> (as opposed to `rs1`).
-That is, all memory accesses, including <<PREFETCH_W_CHERI>> and <<PREFETCH_R_CHERI>>, are implicitly authorized by <<ddc>> and only the memory address is sourced from `rs1`.
-<<PREFETCH_I_CHERI>> is the exception to this rule, as <<pcc>> authorizes it.
+That is, all memory accesses, including <<PREFETCH_W_CHERI>>, <<PREFETCH_R_CHERI>> are implicitly authorized by <<ddc>> and only the memory address is sourced from `rs1`.
++
+The prefetched address of <<PREFETCH_I_CHERI>> _should_ be authorized against <<pcc>>, but the check _may_ be avoided in the implementation.
+
+NOTE: High performance implementations may not have access to the <<pcc>> to authorize the access, as <<pcc>> bounds are not otherwise required in the load/store unit.
+ Therefore this check is non-architectural and only a recommendation.
```
```suggestion
NOTE: High performance implementations may not have access to the <<pcc>> to authorize the access, as <<pcc>> bounds are not otherwise required in the load/store unit.
Since instruction prefetches cannot be used to violate memory safety and only impact side-channel resistence, this check is a strong recommendation rather than mandatory.
```

Maybe something like this?

arichardson (commented):
This is probably a dumb question since I'm not a microarchitect, but could we do the authorization prior to the LSU and potentially insert stalls to ensure a non-speculative PCC is available?

If this is too awkward and introduces significant complexity for the uarch I am fine with it being optional.

Would like to see another LGTM from others (e.g. @jrtc27 @nwf @davidchisnall) that we are okay with this new approach before merging.

tariqkurd-repo:
> This is probably a dumb question since I'm not a microarchitect, but could we do the authorization prior to the LSU and potentially insert stalls to ensure a non-speculative PCC is available?

In a non-CHERI CPU the PC simply isn't available in the LSU - the cost of adding it increases with the size of CPU.
Also in larger CPUs the hardware prefetcher is more likely to have made PRFETCH.I redundant by correctly predicting the pattern.

You also really wouldn't want to introduce stalls related to an instruction which is a hint to help improve performance 😄

tariqkurd-repo:
> I meant could you do the check during decode or some other stage?

You only find out the address at execution and then it's too late

> 
> > You also really wouldn't want to introduce stalls related to an instruction which is a hint to help improve performance 😄
> 
> I guess the other option would be to ignore it completely if you can't have the PCC, but then again the hw could be doing this prefetching anyway, it's just slightly more difficult to trigger it as an attacker. 

That's the pragmatic approach in this case.

src/cheri/riscv-hybrid-integration.adoc
```diff
@@ -35,7 +35,11 @@ All <<rvy_insn_table>> instructions, and associated CSRs, are available in addit
 +
 The authorizing capability for memory access is <<ddc>> (as opposed to `rs1`).
 That is, all memory accesses, including <<PREFETCH_W_CHERI>> and <<PREFETCH_R_CHERI>>, are implicitly authorized by <<ddc>> and only the memory address is sourced from `rs1`.
-<<PREFETCH_I_CHERI>> is the exception to this rule, as <<pcc>> authorizes it.
++
+The prefetched address of <<PREFETCH_I_CHERI>> _should_ be authorized against <<pcc>>, but the check _may_ be avoided in the implementation, or the memory access associated with the prefetch _may_ not be issued if the address cannot be correctly authorized.
++
+NOTE: High performance implementations may not have access to the <<pcc>> to authorize the access, as <<pcc>> bounds are not otherwise required in the load/store unit.
```
jrtc27: This is a bit odd, because the LSU normally deals with the D$ whereas PREFETCH.I is for the I$. How does PREFETCH.I actually work today even without CHERI in such a high-performance design? I kind of suspect, as you get high-performance enough, you want a hardware prefetcher to do everything, in part because you don't want to wait for the many cycles it takes between fetching and being able to execute the PREFETCH.I, during which time you're probably hitting the instructions you were trying to prefetch?

jrtc27: So, it's true that it's not architecturally visible whether you're executing whatever unauthorised prefetches are happening or whether you're just happening to choose to prefetch matching lines, and therefore any authorisation requirements are non-normative (until you get into side-channel microarchitectural requirement specs like Zkt), but I think the justification given here is a bit odd.

tariqkurd-repo: Yes - I think what's likely to happen is that the `prefetch.i` will prime the L2 to prefetch the line, and so will probably have nothing to do with the I$ directly.

How would you fix the justification? It's not reasonable to _require_ the LSU to do the PCC bounds check.

src/cheri/riscv-hybrid-integration.adoc
```diff
@@ -35,7 +35,11 @@ All <<rvy_insn_table>> instructions, and associated CSRs, are available in addit
...
+Since instruction prefetches cannot be used to violate memory safety and only impact side-channel resistence, this check is a strong recommendation rather than mandatory.
```
tariqkurd-repo:
```suggestion
NOTE: Since instruction prefetches cannot be used to violate memory safety and only impact side-channel resistence, this check is a strong recommendation rather than mandatory.
```

src/cheri/riscv-hybrid-integration.adoc
jrtc27: So, I think the concerns about having PCC in the LSU are overblown, since you're already happy to feed PCC down the pipeline for AUIPCC in capability mode, and you're happy to feed DDC alongside the address for integer mode memory accesses, so the the logical thing would be to shove PCC where DDC normally goes and be done with it. I think the only real reason to not _require_ it is because, without any side-channel-specific extensions, microarchitecture is not architecture for RISC-V, and therefore you can prefetch anything you want at any point for any reason. So I would just drop the justification entirely, and for all the PREFETCH.x you have a strong recommendation that microarchitecture use DCC/PCC as appropriate to authorise the request, but that any such requirements belong in an extension pertaining to side-channel resistance.

src/cheri/riscv-hybrid-integration.adoc
```diff
@@ -35,7 +35,10 @@ All <<rvy_insn_table>> instructions, and associated CSRs, are available in addit
 +
 The authorizing capability for memory access is <<ddc>> (as opposed to `rs1`).
 That is, all memory accesses, including <<PREFETCH_W_CHERI>> and <<PREFETCH_R_CHERI>>, are implicitly authorized by <<ddc>> and only the memory address is sourced from `rs1`.
-<<PREFETCH_I_CHERI>> is the exception to this rule, as <<pcc>> authorizes it.
++
+The prefetched address of <<PREFETCH_I_CHERI>> _should_ be authorized against <<pcc>>, but the check _may_ be avoided in the implementation, or the memory access associated with the prefetch _may_ not be issued if the address cannot be correctly authorized.
++
+NOTE: Since instruction prefetches cannot be used to violate memory safety and only impact side-channel resistence, this check is a strong recommendation rather than mandatory.
```
jrtc27: The statement

> If the checks above pass, an implementation may opt to cache a copy of the cache block in a cache accessed by an instruction fetch in order to improve memory access latency, but this behavior is not required.

applies equally here, without all these caveats. Such statements say nothing about what happens if the checks *don't* pass. A conforming implementation can still cache that block. It just probably shouldn't.

tariqkurd-repo: yeah - fair point - I took that text from the existing CBO.* instructions.
On its own it's enough - I#'ll remove the rest.

src/cheri/riscv-hybrid-integration.adoc
```diff
@@ -35,7 +35,8 @@ All <<rvy_insn_table>> instructions, and associated CSRs, are available in addit
 +
 The authorizing capability for memory access is <<ddc>> (as opposed to `rs1`).
 That is, all memory accesses, including <<PREFETCH_W_CHERI>> and <<PREFETCH_R_CHERI>>, are implicitly authorized by <<ddc>> and only the memory address is sourced from `rs1`.
-<<PREFETCH_I_CHERI>> is the exception to this rule, as <<pcc>> authorizes it.
++
+The prefetched address of <<PREFETCH_I_CHERI>> is authorized against <<pcc>>.
```
tariqkurd-repo:
```suggestion
<<PREFETCH_I_CHERI>> is the exception to this rule, as <<pcc>> authorizes it.
```

src/cheri/riscv-hybrid-integration.adoc
```diff
@@ -35,6 +35,7 @@ All <<rvy_insn_table>> instructions, and associated CSRs, are available in addit
 +
 The authorizing capability for memory access is <<ddc>> (as opposed to `rs1`).
 That is, all memory accesses, including <<PREFETCH_W_CHERI>> and <<PREFETCH_R_CHERI>>, are implicitly authorized by <<ddc>> and only the memory address is sourced from `rs1`.
++
```
tariqkurd-repo:
```suggestion
```

arichardson (commented):
Commit message needs updating

"Clarify behaviour of prefetch instructions"

PR #901, opened 2025-12-16
fix page break formatting

tariqkurd-repo:
not sure why this is needed but it avoid <<< showing in the pdf

### DIFF
--- a/src/cheri/insns/gchi_32bit.adoc
+++ b/src/cheri/insns/gchi_32bit.adoc
@@ -26,6 +26,8 @@ Operation::
 +
 TODO
 
+<<<
+
 [#GCHI,reftext="{GCHI}"]
 ==== {GCHI}
 
--- a/src/cheri/insns/schi_32bit.adoc
+++ b/src/cheri/insns/schi_32bit.adoc
@@ -18,6 +18,8 @@ The {SCHI_BASE} instruction packs the least-significant XLEN-bits of `rs1` and `
 Included in::
 <<rvy_insn_table>>
 
+<<<
+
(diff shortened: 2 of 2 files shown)


src/cheri/insns/gchi_32bit.adoc
```diff
@@ -26,6 +26,8 @@ Operation::
 +
 TODO
 
+<<<
```
arichardson, line 29: Do we need a page break here?

src/cheri/insns/schi_32bit.adoc
```diff
@@ -18,6 +18,8 @@ The {SCHI_BASE} instruction packs the least-significant XLEN-bits of `rs1` and `
 Included in::
 <<rvy_insn_table>>
 
+<<<
```
arichardson, line 21: Similar question here

src/cheri/insns/scbndsr_32bit.adoc
```diff
@@ -1,5 +1,5 @@
-<<<
 
+<<<
```
arichardson, line 2: Ah I was wondering how to fix that

src/cheri/insns/gchi_32bit.adoc
tariqkurd-repo: it just looks neater as it avoids the page break within the pseudoinstruction  definition

src/cheri/insns/scbndsr_32bit.adoc
tariqkurd-repo: yes - this is the reason for the PR!

PR #903, opened 2025-12-17
fix mcause numbers

tariqkurd-repo:
variables were missing for mcause numbers


src/cheri/cheri-pte-ext.adoc
```diff
@@ -194,12 +194,18 @@ When V=1 <<vsstatusreg_pte,vsstatus>>.UCRG and CRGE are in effect.
 
 NOTE: The CRGE bit is added to the <<mstatusreg_pte,mstatus>>, <<vsstatusreg_pte,vsstatus>> and <<sstatusreg_pte,sstatus>> registers in order to allow the S-mode software to explicitly opt in to the revocation extensions.
 
-NOTE: The CRGE bit is necessary because the {cheri_priv_crg_ext} defines a different semantic for PTE.CRW=1 PTE.CD=0 PTE.CRG=0 with respect to <<section_priv_cheri_vmem>>.
+[NOTE]
```
arichardson, line 197: Should probably commit this formatting change separately but don't mind either way.

tariqkurd-repo: that one does render correctly, but there are still a couple of NOTE: in the pdf, which we can address with another PR

PR #913, opened 2025-12-30
Svucrg clarification

tariqkurd-repo:
Clarify and rewrite Svucrg chapter giving specific details of the load and store side barriers
Remove and merge duplicate text
Group text more logically into load-side-barrier and store-side-barrier sections

### DIFF
--- a/src/cheri/cheri-pte-ext.adoc
+++ b/src/cheri/cheri-pte-ext.adoc
@@ -9,7 +9,7 @@ NOTE: _Sv32_ (for RV32) does not have any spare PTE bits, and so no features fro
 
 
 The {cheri_priv_crg_ext} extension is enabled when the <<sstatusreg_pte,sstatus>>.CRGE bit is set.
-When enabled, the extension adds the ability to perform capability revocation of user mode pages (see <<cap_revocation>>) by adding the PTE.CRG, PTE.CD and <<sstatusreg_pte,sstatus>>.UCRG bits as described below.
+When enabled, the extension adds the ability to perform capability revocation of userspace pages (see <<cap_revocation>>) by adding the PTE.CRG, PTE.CD and <<sstatusreg_pte,sstatus>>.UCRG bits as described below.
 When the extension is disabled the PTE.CD, PTE.CRG and <<sstatusreg_pte,sstatus>>.UCRG bits remain reserved.
 
 NOTE: A future extension may define an alternative function for these reserved states; the PTE.CRW bit is interpreted as defined by <<section_priv_cheri_vmem>>.
@@ -20,9 +20,9 @@ NOTE: {cheri_priv_crg_ext} is strongly recommended but not mandatory as a future
 
 NOTE: A future version of this specification may include kernel revocation which may require an <<sstatusreg_pte,sstatus>>.SCRG bit.
 
-The minimum level of PTE support is to set CRW and CD to 1 in all PTEs intended for storing
+The minimum level of PTE support is to set PTE.CRW and PTE.CD to 1 in all PTEs intended for storing
 capabilities (i.e., private anonymous mappings) and leave
-<<sstatusreg_pte,sstatus>>.UCRG and CRG in all PTEs set to 0, which will allow
+<<sstatusreg_pte,sstatus>>.UCRG and PTE.CRG in all PTEs set to 0, which will allow
 capabilities with their {ctag}s set to be loaded and stored successfully.
 
 NOTE: When {cheri_priv_crg_ext} is enabled (<<sstatusreg_pte,sstatus>>.CRGE is set), the semantics of PTE.CRW bit are redefined relative to the definition in <<section_priv_cheri_vmem>>.
@@ -37,10 +37,12 @@ NOTE: Hardware-initiated memory accesses from the page-table walker are not chec
 Revocation is the process of clearing the {ctag} of capabilities that point to freed memory, to prevent them from being dereferenced and thereby enforcing _temporal memory safety_.
 cite:[cornucopia-reloaded] details algorithms for preventing use-after-free in virtual memory, and non-virtual memory based systems.
 
-{cheri_priv_crg_ext} adds capability revocation for user pages only.
-It uses the <<sstatusreg_pte,sstatus>>.UCRG and PTE.CRG bits to control the revocation epoch, which is used to provide a load barrier to detect when loading from a page which needs sweeping.
-This is indicated by the generation in the PTE mismatching the global one in <<sstatusreg_pte,sstatus>>.UCRG.
-Stored {ctag}s to clean pages are tracked using PTE.CD to record which virtual memory pages contain dirty capabilities, and so will need to be swept.
+{cheri_priv_crg_ext} adds capability revocation for userspace pages only.
+
+It uses the <<sstatusreg_pte,sstatus>>.UCRG and PTE.CRG bits to control the revocation epoch, which is used to provide a _load-side barrier_ to detect when loading from a page which needs sweeping (i.e. may contain dangling capabilities which point to freed memory).
+This state is indicated by the generation in PTE.CRG mismatching the global one in <<sstatusreg_pte,sstatus>>.UCRG.
+
+It also uses the PTE.CD (capability dirty) bit to provide a _store-side barrier_ to detect when stored capabilities are to a clean page, and so need to be swept in case they are dangling.
 
 [#section_extending_pte]
 === Extending the Page Table Entry Format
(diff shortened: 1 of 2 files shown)


src/cheri/cheri-pte-ext.adoc
```diff
@@ -73,104 +75,80 @@ include::img/sv57pte_crg.edn[]
...
-
-The CRW bit (defined by {cheri_priv_vmem_ext}) indicates whether reading or writing capabilities with the {ctag} set to
-the virtual page is permitted. When the CRW bit is set, capabilities are written
-as usual, and capability reads are controlled by the CRG bit.
-
-NOTE: The {ctag} bit of the stored capability is checked _after_ it is potentially
-cleared due to lack of <<c_perm>>.
-
+. A capability load or AMO cannot set the {ctag} in the target register.
+. A capability store or AMO cannot store a set {ctag} to the target memory.
```
jrtc27, line 79: This is vague. Does "cannot [store a] set tag" encompass "the data in question has tag == 0"? Because that can never do so, yet is data-dependent. The text below then gives examples, but I think being clearer up-front is best.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -73,104 +75,80 @@ include::img/sv57pte_crg.edn[]
...
+* PTE.CRW=1, and
+* PTE.CRG != <<sstatusreg_pte,sstatus>>.UCRG, and
+* PTE.U=1, and
+* The loaded {ctag} is set^1^
 
-* The same behavior as when CRW is clear, allowing software interpretation
- of this state.
-* The implementation sets the CD bit. The PTE update behaves in the same way as
- the D bit update described by the _Svadu_ extension.
+^1^ Implementations are allowed to avoid checking the {ctag} under specific conditions.
```
jrtc27, line 110: This is vague

src/cheri/cheri-pte-ext.adoc
```diff
@@ -73,104 +75,80 @@ include::img/sv57pte_crg.edn[]
...
+ A valid implementation is _never_ to check the loaded {ctag} value.
+ The choice about whether the {ctag} is _ever_ checked affects the exception priority, see <<exception-priority-cheri>>.
```
jrtc27, line 112: So if I have an implementation that sometimes checks the tag, it has to always perform the load in case it gives an access fault??

src/cheri/cheri-pte-ext.adoc
```diff
@@ -73,104 +75,80 @@ include::img/sv57pte_crg.edn[]
...
-| 0      | 1      | X                                       | X     | Reserved
-| 1      | X      | {ne} <<sstatusreg_pte,sstatus>>.UCRG | 1     | {cheri_excep_name_pte_ld}, or {cheri_excep_name_pte_ld} if tag is set for _{cheri_priv_crg_load_tag_ext}_^1^
-| 1      | X      | = <<sstatusreg_pte,sstatus>>.UCRG       | 1     | Normal operation
-| 1      | X      | X                                       | 0     | Normal operation^2^
+|PTE. CRW| PTE. CD| PTE.CRG                               | PTE.U | Load/AMO
+| 0      | 0      | 0                                     | X     | Clear loaded {ctag}
+| 0      | 0      | 1                                     | X     | Reserved
+| 0      | 1      | X                                     | X     | Reserved
+| 1      | X      | {ne} <<sstatusreg_pte,sstatus>>. UCRG | 1     | {cheri_excep_name_pte_ld} if {ctag} is set in {cd}
+| 1      | X      | = <<sstatusreg_pte,sstatus>>. UCRG    | 1     | Normal operation
```
jrtc27, line 125: Why have these got spaces before UCRG?

src/cheri/cheri-pte-ext.adoc
```diff
@@ -73,104 +75,80 @@ include::img/sv57pte_crg.edn[]
...
-| 0      | 0      | 1                                       | X     | Reserved
-| 0      | 1      | X                                       | X     | Reserved
-| 1      | X      | {ne} <<sstatusreg_pte,sstatus>>.UCRG | 1     | {cheri_excep_name_pte_ld}, or {cheri_excep_name_pte_ld} if tag is set for _{cheri_priv_crg_load_tag_ext}_^1^
-| 1      | X      | = <<sstatusreg_pte,sstatus>>.UCRG       | 1     | Normal operation
-| 1      | X      | X                                       | 0     | Normal operation^2^
+|PTE. CRW| PTE. CD| PTE.CRG                               | PTE.U | Load/AMO
+| 0      | 0      | 0                                     | X     | Clear loaded {ctag}
+| 0      | 0      | 1                                     | X     | Reserved
+| 0      | 1      | X                                     | X     | Reserved
+| 1      | X      | {ne} <<sstatusreg_pte,sstatus>>. UCRG | 1     | {cheri_excep_name_pte_ld} if {ctag} is set in {cd}
```
jrtc27, line 124: This table now mandates Svucrglct?

src/cheri/cheri-pte-ext.adoc
```diff
@@ -37,10 +37,12 @@ NOTE: Hardware-initiated memory accesses from the page-table walker are not chec
...
-{cheri_priv_crg_ext} adds capability revocation for user pages only.
-It uses the <<sstatusreg_pte,sstatus>>.UCRG and PTE.CRG bits to control the revocation epoch, which is used to provide a load barrier to detect when loading from a page which needs sweeping.
-This is indicated by the generation in the PTE mismatching the global one in <<sstatusreg_pte,sstatus>>.UCRG.
-Stored {ctag}s to clean pages are tracked using PTE.CD to record which virtual memory pages contain dirty capabilities, and so will need to be swept.
+{cheri_priv_crg_ext} adds capability revocation for userspace pages only.
+
+It uses the <<sstatusreg_pte,sstatus>>.UCRG and PTE.CRG bits to control the revocation epoch, which is used to provide a _load-side barrier_ to detect when loading from a page which needs sweeping (i.e. may contain dangling capabilities which point to freed memory).
+This state is indicated by the generation in PTE.CRG mismatching the global one in <<sstatusreg_pte,sstatus>>.UCRG.
+
+It also uses the PTE.CD (capability dirty) bit to provide a _store-side barrier_ to detect when stored capabilities are to a clean page, and so need to be swept in case they are dangling.
```
jrtc27, line 45: It's not a barrier?

src/cheri/cheri-pte-ext.adoc
```diff
@@ -73,115 +75,83 @@ include::img/sv57pte_crg.edn[]
...
 NOTE: The behavior in this section isn't relevant if:
 
-. The authorizing capability doesn't have <<c_perm>>, for loads, stores and AMO.
-. Any extension-specific mediation has already cleared the stored tag, for stores and AMOs.
-
-The CRW bit (defined by {cheri_priv_vmem_ext}) indicates whether reading or writing capabilities with the {ctag} set to
-the virtual page is permitted. When the CRW bit is set, capabilities are written
-as usual, and capability reads are controlled by the CRG bit.
+. A capability load or AMO cannot set the {ctag} in the target register.
+. A capability store or AMO cannot store a set {ctag} to the target memory.
```
arichardson, line 79: I think maybe for this one we need to split out AMO into a 3rd point? "A capability AMO annot set the {ctag} in the target register and cannot set the {ctag} in the target memory.

Also Maybe be explicit that we mean "set to one" since setting to zero is fine.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -73,115 +75,83 @@ include::img/sv57pte_crg.edn[]
...
-is controlled by CD.
+When PTE.CRW, PTE.CD, and PTE.CRG are _all_ clear, the implementation is required to clear
+loaded tags and raise {cheri_excep_name_pte_st} when the stored {ctag} is set.
 
-When CRW is set, the CRG bit indicates the current generation of the virtual
-memory page with regards to the ongoing capability revocation epoch.
-The implementation raises {cheri_excep_name_pte_ld} when a capability load
-or AMO instruction is executed and:
+NOTE: Other PTE.CD and PTE.CRG combinations when PTE.CRW=0 are reserved for future extensions.
+The reserved PTE states behave as the PTE.CRW=0, PTE.CD=0, PTE.CRG=0, unless a future extension
```
arichardson, line 93: Isn't the normal RISC-V behaviour for reserved PTE states to fault?

jrtc27: Yes

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: This is the phrase used in cornucopia reloaded and Gutstein's thesis - they talk about a load-side and store-side barrier. The point of this section is to say _how_ Svucrg is used to implement those barriers, so it's convenient to use the same name.

One thing Krste said in the first set of review feedback was that he had no idea what the Svucrg was supposed to do - so this explicitly ties it to the papers (which is also useful for my understanding)

tariqkurd-repo:
> Oh and Svucrglct needs to stay. Different architecturally-visible behaviours need to have an extension to distinguish them.

I'll leave this open until we get some ARC feedback.

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: no I don't think it is - beause they are planning to extended the physical page number into the reserved bits so I think they are ignored.

src/cheri/cheri-pte-ext.adoc
jrtc27: *If* your software foolishly chooses to implement the original store-side Cornucopia rather than the superior load-side Cornucopia *then* the software model is that it can be used to provide some kind of barrier. But (a) you shouldn't implement that, it doesn't perform well for interactive workloads due to long stop-the-world phases (b) that's *a* software model using the bit, while other uses for this bit exist, and we're writing architecture here, so we're meant to be describing what the hardware does, not one algorithm you could use it for.

src/cheri/cheri-pte-ext.adoc
jrtc27: Perhaps you should check first:

> If pte.v=0, or if pte.r=0 and pte.w=1, or if any bits or encodings that are reserved for future standard use are set within pte, stop and raise a page-fault exception corresponding to the original access type.

My response was grounded in fact not belief.

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: the intention is that it's a summary of what's above and I was trying not to duplicate text

jrtc27: But it's wrong? You've made it say the loads only trap if the tag is set. But if you only implement Svucrg and not Svucrglct that's not true.

src/cheri/cheri-pte-ext.adoc
nwf: "Barrier" is, regretfully, a remarkably overloaded term, with different meanings in (at least) the sub-disciplines of architectural memory models (ordering of load and store operations from multiple initiators), concurrent programming (synchronization of thread group execution), and garbage collection (mutator behaviors required for preserving system invariants).

The Cornucopia Reloaded paper...
- uses the word "barrier" to mean this latter sense;
- goes so far as to explicitly compare the CD-tracking feature of PTEs with traditional GC store barriers (though it points out that its machinery is somewhat unusual, being an architectural primitive and not code inserted by a compiler on behalf of a runtime system); and
- continues to rely on CD-tracking to know which pages must have their CRGs maintained, though this is a performance optimization and not required for correctness as it was in Cornucopia.

That said... surely this whole "Capability Revocation" section should be marked as informative (and not normative)?
- "{cheri_priv_crg_ext} adds capability revocation for userspace pages only"; no, it adds _a primitive that is useful for software implementation of_ capability revocation.  (CHERI+MTE and CHERIoT _might_ be able to make the claim of adding revocation in an architectural sense, but, IMHO, only sort of... in any case, surely Svucrg should not make that claim.)
- "which needs sweeping (...)" is definitely a software perspective on the matter; barring subsequent extension or platform machinery (such as CHERIoT's revocation engine or a CHERI+MTE analogue), architecture does not know what "sweeping" means or why it might be needed.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -73,104 +75,80 @@ include::img/sv57pte_crg.edn[]
...
-. Any extension-specific mediation has already cleared the stored tag, for stores and AMOs.
-
-The CRW bit (defined by {cheri_priv_vmem_ext}) indicates whether reading or writing capabilities with the {ctag} set to
-the virtual page is permitted. When the CRW bit is set, capabilities are written
-as usual, and capability reads are controlled by the CRG bit.
-
-NOTE: The {ctag} bit of the stored capability is checked _after_ it is potentially
-cleared due to lack of <<c_perm>>.
-
+. A capability load or AMO cannot set the {ctag} in the target register.
```
nwf, line 78: I think maybe this should be something like

```
The behavior in this section is not relevant for
. capability load operations which will otherwise not write a set {ctag} into the target register
. capability store operations which will otherwise not write a set {ctag} into the target register

Capability AMOs may involve one of or both of a load and store operation.
```

The quantifiers in "isn't relevant if a capability load or AMO cannot" are awkward; there's a reading in which that is discussing scenarios in which _no_ capability load can set the tag.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -73,104 +75,80 @@ include::img/sv57pte_crg.edn[]
...
-The implementation raises {cheri_excep_name_pte_ld} when a capability load
-or AMO instruction is executed and:
+NOTE: Other PTE.CD and PTE.CRG combinations when PTE.CRW=0 are reserved for future extensions.
+The reserved PTE states behave as the PTE.CRW=0, PTE.CD=0, PTE.CRG=0, unless a future extension
+defines an alternative function.
 
-* the authorizing capability grants <<c_perm>>,
-* the virtual page's CRG bit does not equal <<sstatusreg_pte,sstatus>>.UCRG, and
-* the virtual page's PTE.U is set.
+=== Load-side barrier
```
nwf, line 96: The phrase "load-side barrier" is almost surely my fault (I took to using it to emphasize the contrast between Reloaded and Cornucopia) and should probably not be propagated further if we can help it.  Just call it a load barrier, generically, or more specifically perhaps something like "Generation-based Load Traps".

src/cheri/cheri-pte-ext.adoc
```diff
@@ -73,104 +75,80 @@ include::img/sv57pte_crg.edn[]
...
 
-When the _{cheri_priv_crg_load_tag_ext}_ extension is implemented, the
-{cheri_excep_name_pte_ld} is raised when a capability load or AMO instruction
-satisfies the conditions above and the capability read from memory
-has its {ctag} set.
+The _load-side barrier_ allows the software to detect when a userspace virtual memory page need to be swept for dangling capabilities.
```
nwf, line 98: Again, that's true but remarkably software-centric...

src/cheri/cheri-pte-ext.adoc
```diff
@@ -73,104 +75,80 @@ include::img/sv57pte_crg.edn[]
...
+This is achieved by taking a trap on capability loads allowing the exception handler to sweep the page _before_ the application can load the capability.
```
nwf, line 99: (FWIW, most things in CHERI are "catch it before the bad thing happens"; this is not emphatically unusual in that sense...)

src/cheri/cheri-pte-ext.adoc
```diff
@@ -73,104 +75,80 @@ include::img/sv57pte_crg.edn[]
...
- set will introduce additional traps during revocation sweeps.
- If _{cheri_priv_crg_load_tag_ext}_ is implemented then the trap is only taken
- if the loaded {ctag} is set, to reduce software overhead from revocation sweeps.
- Checking the loaded tag affects the exception priority, see <<exception-priority>>.
+^1^ A future version of this specification may check an SCRG bit in <<sstatusreg_pte,sstatus>> for kernel revocation.
 
-NOTE: _{cheri_priv_crg_load_tag_ext}_ is an optimization for software, and as
-such implementations are allowed to conservatively fault under certain
-conditions even if the {ctag} is not set.
+=== Store-side barrier
```
nwf, line 131: Again, "-side".  Sorry.

tariqkurd-repo:
Ok - thanks for the input - the main objective of the PR is to make it clear how this relates to the revocation scheme detailed in cornucopia reloaded, which was really not clear to me before.
Also the previous text was inconsistent in places and needed restructuring and simplifying.

1. I'll reinstate Svucrglct, it's still debatable whether it should exist but that's separate a separate issue to clarifying and simplifying the text for other reasons.
2. Stop using the load/store-side barriers terminology - but I will reference the load/store-side barriers in cornucopia reloaded from the text to make the connection clear
3. Move anything software specific into non-normative notes

PR #919, opened 2026-01-20
Svucrg clarification

tariqkurd-repo:
Clean up the Svucrg text without introducing any spec changes:

1. making it is easier to read and removing duplication.
2. putting references into Cornucopia Reloaded to explain the purpose

During the process it became apparent there were were gaps in the spec:

1. the priority between CHERI load and store page faults wasn't lsited
2. the VM rules didn't include anything CHERI related

### DIFF
--- a/src/supervisor.adoc
+++ b/src/supervisor.adoc
@@ -1624,7 +1624,7 @@ __a__=__pte__.__ppn__×PAGESIZE and go to step 2.
 
 . Determine if the requested memory access is allowed by the _pte_._r_, _pte_._w_, and _pte_._x_ bits. If not, stop and raise a page-fault exception corresponding to the original access type.
 
-. If _pte_._a_=0, or if the original memory access is a store and _pte_._d_=0, or if the original memory access is a capability store and _pte_._cd_=0^1^:
+. If _pte_._a_=0, or if the original memory access is a store and _pte_._d_=0, or if the original memory access is a capability store with the to-be-stored {ctag} set and _pte_._cd_=0^1^:
 
 * If the Svade extension is implemented, stop and raise a page-fault exception corresponding to the original access type.
 * If a store to the PTE at address __a__+__va.vpn__[__i__]×PTESIZE would violate a PMA or PMP check,


src/cheri/cheri-pte-ext.adoc
```diff
@@ -31,16 +31,18 @@ Therefore, when PTE.CRW is clear, it is not required that future extensions to {
 
 NOTE: Hardware-initiated memory accesses from the page-table walker are not checked by a capability.
 
-[#cap_revocation]
+[#cap_revocation, reftext="capability revocation"]
 === Capability Revocation
 
-Revocation is the process of clearing the {ctag} of capabilities that point to freed memory, to prevent them from being dereferenced and thereby enforcing _temporal memory safety_.
-cite:[cornucopia-reloaded] details algorithms for preventing use-after-free in virtual memory, and non-virtual memory based systems.
+Revocation is the process of clearing the {ctag} of capabilities that point to freed memory, to prevent them from being dereferenced and thereby enforcing _temporal memory safety_, i.e. preventing use-after-free and use-after-reallocate vulnerabilities.
```
nwf: That's not usually what revocation means in the capability literature sense; revocation is the act of un-granting a capability and all its progeny.  It happens that there is a special case where the thing doing the (un)granting is the heap allocator and we're after heap temporal safety.

If we had a general revocation mechanism, we'd be able to do this for stack allocations as well, and we'd be able to revoke capabilities derived from particular instances of a capability without also revoking all capabilities that pointed to the same piece of memory.  Sadly, no suitable mechanism has yet emerged, and so the only implementations to date are restricted to the case of providing heap temporal safety.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -151,26 +151,17 @@ Two schemes for this are permitted, and the scheme in use is determined by wheth
...
+|PTE. CRW| PTE. CD| PTE.CRG | Store/AMO
+| 0      | 0      | 0       | {cheri_excep_name_pte_st} if stored {ctag} is set
 | 0      | 0      | 1       | Reserved
 | 0      | 1      | X       | Reserved
-| 1      | 0      | X       | {cheri_excep_name_pte_st} if stored tag is set, or <<section_hardware_pte_updates, hardware PTE.CD update>>
+| 1      | 0      | X       | {cheri_excep_name_pte_st} if stored {ctag} is set (_Svade_), or hardware PTE.CD update (_Svadu_)
 | 1      | 1      | X       | Normal operation
 |===
 
-[#section_hardware_pte_updates]
```
tariqkurd-repo, line 177: Also see lines 124 onwards from https://github.com/riscv/riscv-cheri/pull/919/commits/81dd51eec61801abb2266093a7dc092f4bf2f735

the commit history hasn't quite worked out as I intended

src/cheri/cheri-pte-ext.adoc
```diff
@@ -151,26 +151,17 @@ Two schemes for this are permitted, and the scheme in use is determined by wheth
 ^1^ A future version of this specification may check an SCRG bit in <<sstatusreg_pte,sstatus>> in this case for kernel capability revocation.
 
 [[pte_crw_cd_store_summary]]
-.Summary of Store PTE.CRW and PTE.CD behavior in the PTEs
+.Summary of capability store PTE.CRW and PTE.CD behavior in the PTEs
 [%autowidth,float="center",align="center",cols="<,<,<,<",options="header"]
 |===
-|PTE.CRW | PTE.CD | PTE.CRG | Store/AMO
-| 0      | 0      | 0       | {cheri_excep_name_pte_st} if stored tag is set
+|PTE. CRW| PTE. CD| PTE.CRG | Store/AMO
```
tariqkurd-repo: these spaces are for formatting in the in the PDF so it doesn't render as

PTE.C  PTE.C
RW     D

src/cheri/cheri-pte-ext.adoc
```diff
@@ -5,42 +5,40 @@ ifdef::cheri_standalone_spec[]
...
 
 NOTE: Hardware-initiated memory accesses from the page-table walker are not checked by a capability.
 
-[#cap_revocation]
+[#cap_revocation, reftext="capability revocation"]
 === Capability Revocation
 
-Revocation is the process of clearing the {ctag} of capabilities that point to freed memory, to prevent them from being dereferenced and thereby enforcing _temporal memory safety_.
-cite:[cornucopia-reloaded] details algorithms for preventing use-after-free in virtual memory, and non-virtual memory based systems.
+Revocation is the process of clearing the {ctag} of capabilities that point to freed memory, to prevent them from being dereferenced and thereby enforcing _temporal memory safety_, i.e. preventing use-after-free and use-after-reallocate vulnerabilities.
```
```suggestion
Revocation is the process of clearing the {ctag} of capabilities, and their progeny, that have had their permission to access memory removed.
Enforcing _temporal memory safety_ involves preventing such capabilities from being dereferenced and so prevents use-after-free and use-after-reallocate vulnerabilities.

NOTE: This extension is suitable for _heap temporal safety_ only.

```

tariqkurd-repo: is this any better @nwf ? If not please suggest a futher edit 😄

brooksdavis: There are things we can revoke with this mechanism that aren't memory capabilities and aren't from the heap as such. For example, if you have the Zyseal extension, type capabilities could be revoked. The important thing isn't  so much where it's allocated as the rate of reuse.

tariqkurd-repo: @brooksdavis how would you implement that using these hardware barriers? I'm interested.

brooksdavis: I may be off base here, but in the load barrier case, the pointer of the barrier prevent the spread of capabilities that haven't been scanned for possible revocation from an unscanned page to the register file (from where they could spread onward). It doesn't matter if those capabilities are memory capabilities or types, you just need an appropriate bitmap (or other mechanism) to determine if they should be revoked and a mechanism to do so.

In CheriBSD we don't currently revoke type capabilities (e.g., on Morello), but we do revoke both heap capabilities allocated by malloc and capabilities previously returned by mmap where the whole reservation has been unmapped. Similar ideas can be used to handle tokens, type capabilities, etc.

nwf: Svucrg is "just" a way of letting privileged software front-run capability fetches from memory.  It's a mechanism that might be useful to software.  (ETA2: It isn't a _necessary_ mechanism nor is it a _sufficient_ one.  It just seems to be a nice to have, simple enough in hardware and a performance boost to software.  Software implementing Reloaded-style global revocation has to do a lot of other work in addition to the MMU maintenance around Svucrg.)  It _happens_ that CHERIvoke/Cornucopia/Reloaded-style "global revocation" -- the special case where we can use capability monotonicity in lieu of a materialized capability provenance tree -- is particularly straightforward to implement atop this: front-run every capability load (either by Svucrg traps or by just getting to the page first in the background scan) and for each capability in the system, check some flat global state (the revocation bitmap).  (ETA: please note that "particularly straightforward" is still some 80 pages of technical report and some thousands of lines of code spanning the CheriBSD kernel and user space.)  The same approach works for sealing rights as address space, as there's nothing that ties the "flat global state" to "address space" other than software, so it's easy to have -- and in fact the CheriBSD implementation has stubs for -- a "sealing-type revocation bitmap" next to your "address-space revocation bitmap".  (By contrast, (CHERI+)MTE more tightly couples the flat global state with address space, by virtue of storing colors at (physical; please let's not talk about vMTE) memory locations.)

_In principle_ software could be maintaining more elaborate state -- something more hierarchical and less flat -- so that a capability found at a given location in memory could be associated with (a subsequence of) its provenance, enabling the front-running privileged software to discriminate between Alice's copies of (progeny of) a capability and Bob's copies of (progeny of) the same capability.  We don't know how to do that efficiently or securely, especially not in the case of adversarial (arbitrary) instructions (in contrast to, say, CHERI's monotonicity properties).

All of that to say, I think the text for Svucrg should much more stick to what it does, mention in a note that it is intended to enable privileged software to front-run capability loads, cite the Reloaded paper ("which may be useful for software wishing to provide certain forms of temporal safety [387]"), and be done with it.  IMHO, there's no need to explain revocation in this text.

tariqkurd-repo: ok - I'm simplifying the text - will make another commit a bit later

tariqkurd-repo: ok - I've simplified and resturtured the text to focus much more closely on the actual spec - please have another look.

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: text removed in my re-write in the latest commit.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,140 +50,130 @@ include::img/sv48pte_crg.edn[]
...
 [[pte_crw_crg_load_summary]]
-.Summary of Load CRW and CRG behavior in the PTEs
+.Summary of capability load PTE.CRW and PTE.CRG behavior in the PTEs
 [%autowidth,float="center",align="center",cols="<,<,<,<,<",options="header"]
 |===
-|PTE.CRW | PTE.CD | PTE.CRG                                 | PTE.U | Load/AMO
+|PTE. CRW| PTE. CD| PTE.CRG                                 | PTE. U| Load/AMO
 | 0      | 0      | 0                                       | X     | Clear loaded tag
 | 0      | 0      | 1                                       | X     | Reserved
 | 0      | 1      | X                                       | X     | Reserved
```
tariqkurd-repo: It would be a spec change - but I think this state should be defined to say "Clear loaded tag".
The rationale is to be able to control access to the capabilities on a page, and so declare CRW=0, CD=1 to mean "there are caps here but you can't see them"

The case which came up was when a whole page of memory has been freed, and CD is set then it becomes possible to prevent access to it and retain the information that it actually does include capabilities. This could be used to clear the tags via the tag controller, for example, without needing to zero the memory.

It's what the logical meaning of CRW=0, CD=1 would mean anyway.

Perhaps the main advantage of allowing this state is for Svadu hardware setting of PTE.CD.
If one core sets CRW=0, then another core cannot legally set PTE.CD on the same page without risking an invalid PTE end result, unless it doesn't an atomic CAS on the PTE bits in memory (i.e. only update if it finds that CRW=1)

see https://github.com/riscv/riscv-cheri/issues/920

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,140 +50,130 @@ include::img/sv48pte_crg.edn[]
...
-.Summary of Store CRW and CD behavior in the PTEs
+.Summary of capability store PTE.CRW and PTE.CD behavior in the PTEs
 [%autowidth,float="center",align="center",cols="<,<,<,<",options="header"]
 |===
-|PTE.CRW | PTE.CD | PTE.CRG | Store/AMO
-| 0      | 0      | 0       | {cheri_excep_name_pte_st} if stored tag is set
+|PTE. CRW| PTE. CD| PTE. CRG| Store/AMO
+| 0      | 0      | 0       | {cheri_excep_name_pte_st} if stored {ctag} is set
 | 0      | 0      | 1       | Reserved
 | 0      | 1      | X       | Reserved
```
tariqkurd-repo: I think this state should say {cheri_excep_name_pte_st} if stored {ctag} is set, for the same reason as for loads in the comment above "there are caps here but you can't see them"

see https://github.com/riscv/riscv-cheri/issues/920

src/cheri/cheri-pte-ext.adoc
```diff
@@ -5,43 +5,22 @@ ifdef::cheri_standalone_spec[]
 WARNING: This chapter will appear in the priv spec. Exact location TBD.
 endif::[]
 
+This chapter also defines _{cheri_priv_crg_load_tag_ext}_ which implies _{cheri_priv_crg_ext}_ .
```
arichardson: Not sure we need this note at the top since it's really more of an "optimization/implementation variant" of the same extension. I'd just drop it to reduce visual clutter?

arichardson: The way I see it is that _{cheri_priv_crg_load_tag_ext}_ gives a name to a specicific implememntation choice but fundamentally this is the same extension.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -5,43 +5,22 @@ ifdef::cheri_standalone_spec[]
...
-When enabled, the extension adds the ability to perform capability revocation of user mode pages (see <<cap_revocation>>) by adding the PTE.CRG, PTE.CD and <<sstatusreg_pte,sstatus>>.UCRG bits as described below.
+When enabled, the extension adds the ability to perform _capability revocation_ to prevent use-after-free and use-after-reallocation vulnerabilities of userspace pages using software algorithms such as listed in cite:[cornucopia-reloaded].
+It achieves this by adding by adding the PTE.CRG, PTE.CD and <<sstatusreg_pte,sstatus>>.UCRG bits as described below.
 When the extension is disabled the PTE.CD, PTE.CRG and <<sstatusreg_pte,sstatus>>.UCRG bits remain reserved.
 
-NOTE: A future extension may define an alternative function for these reserved states; the PTE.CRW bit is interpreted as defined by <<section_priv_cheri_vmem>>.
-
 NOTE: {cheri_priv_crg_ext} depends on Supervisor-Level ISA and {cheri_base64_ext_name} also being supported.
 
 NOTE: {cheri_priv_crg_ext} is strongly recommended but not mandatory as a future version of this specification may specify an improved method.
```
arichardson: Maybe we should rephrase this as "not a mandatory part of the RVY privileged ISA"?

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,117 +50,108 @@ include::img/sv48pte_crg.edn[]
...
 include::img/sv57pte_crg.edn[]
 
-NOTE: The behavior in this section isn't relevant if:
+=== Software Primitives
 
-. The authorizing capability doesn't have <<c_perm>>, for loads, stores and AMO.
-. Any extension-specific mediation has already cleared the stored tag, for stores and AMOs.
+The minimum level of PTE support is to set PTE.CRW and PTE.CD to 1 in all PTEs intended for storing
+capabilities (i.e., private anonymous mappings) and leave
+<<sstatusreg_pte,sstatus>>.UCRG and PTE.CRG in all PTEs set to 0, which will allow
```
arichardson: Is this true? Since we now have an enable bit, isn't it also possible to just set CRW and keep this disabled? Maybe list that as an alternative approach?

arichardson: Also I think this should probably move to the end of the section once the primitives have been described?

src/cheri/cheri-pte-ext.adoc
```diff
@@ -5,43 +5,22 @@ ifdef::cheri_standalone_spec[]
 WARNING: This chapter will appear in the priv spec. Exact location TBD.
 endif::[]
 
+This chapter also defines _{cheri_priv_crg_load_tag_ext}_ which implies _{cheri_priv_crg_ext}_ .
+
 NOTE: _Sv32_ (for RV32) does not have any spare PTE bits, and so no features from this chapter can be implemented.
 
 
 The {cheri_priv_crg_ext} extension is enabled when the <<sstatusreg_pte,sstatus>>.CRGE bit is set.
```
```suggestion
The {cheri_priv_crg_ext} extension is enabled when the new <<sstatusreg_pte,sstatus>>.CRGE bit is set.
```

Since it doesn't exist otherwise?

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,117 +50,108 @@ include::img/sv48pte_crg.edn[]
...
 
-The CRW bit (defined by {cheri_priv_vmem_ext}) indicates whether reading or writing capabilities with the {ctag} set to
-the virtual page is permitted. When the CRW bit is set, capabilities are written
-as usual, and capability reads are controlled by the CRG bit.
+The PTE.CRW bit (defined by <<section_priv_cheri_vmem>>) indicates whether loading or storing capabilities to
+the virtual page is permitted.
 
-NOTE: The {ctag} bit of the stored capability is checked _after_ it is potentially
-cleared due to lack of <<c_perm>>.
+If all the PTE.CRW, PTE.CD and PTE.CRG bits are clear, then the existing rules from <<section_priv_cheri_vmem>> are followed, regardless of the state of <<sstatusreg_pte,sstatus>>.CRGE.
```
arichardson: Since we are talking about what CRGE does, maybe easier to say something like this:
```suggestion
When <<sstatusreg_pte,sstatus>>.CRGE is clear, the PTE.CD and PTE.CRG bits are reserved and must not be set in any valid PTEs, otherwise the existing rules from <<section_priv_cheri_vmem>> are followed.
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,117 +50,108 @@ include::img/sv48pte_crg.edn[]
...
 
-If all the CRW, CD and CRG bits are clear, the "no capability state", then the existing rules from <<section_priv_cheri_vmem>> are followed:
+NOTE: These primitives may be used to implement the load-barrier and store-barrier primitives from cite:[cornucopia-reloaded].
 
-* When a capability load or AMO instruction is executed, the {ctag} bit of the
-  loaded capability is cleared before it is written to the destination register.
-* When a capability store or AMO instruction is executed and the {ctag} bit
- of the capability being written is set, the implementation raises a
- {cheri_excep_name_pte_st}.
+The primitives are enabled when all of the following are true:
```
arichardson: Doesn't this mostly duplicate the sentence above? Maybe merge them?

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,117 +50,108 @@ include::img/sv48pte_crg.edn[]
...
-* the authorizing capability grants <<c_perm>>,
-* the virtual page's CRG bit does not equal <<sstatusreg_pte,sstatus>>.UCRG, and
-* the virtual page's PTE.U is set.
+NOTE: {cheri_excep_name_pte_ld}s may be used to implement the load-barrier primitive from cite:[cornucopia-reloaded].
 
-When the _{cheri_priv_crg_load_tag_ext}_ extension is implemented, the
-{cheri_excep_name_pte_ld} is raised when a capability load or AMO instruction
-satisfies the conditions above and the capability read from memory
-has its {ctag} set.
+When PTE.CRW is set, the local PTE.CRG bit can be used to trap on capability loads or AMOs when it mismatches the global state in <<sstatusreg_pte,sstatus>>.UCRG.
```
```suggestion
When PTE.CRW is set, the local PTE.CRG bit can be used to trap on capability loads or AMOs when it does not match the value of <<sstatusreg_pte,sstatus>>.UCRG.
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,117 +50,108 @@ include::img/sv48pte_crg.edn[]
...
 
-When CRW, CD, and CRG are all clear, the implementation is required to clear
-loaded tags and raise {cheri_excep_name_pte_st} when the stored {ctag} is set.
+[start=5]
+. The loaded {ctag} is set.
 
-NOTE: Other CD and CRG combinations when CRW=0 are reserved for future extensions.
-The reserved PTE states behave as the CRW=0, CD=0, CRG=0, unless a future extension
-defines an alternative function.
+Therefore _{cheri_priv_crg_load_tag_ext}_ requires taking data dependent exceptions on load data for loads or AMOs, and also affects the exception priority, see xref:exception-priority[xrefstyle=short].
```
```suggestion
NOTE: Checking the value of the {ctag} requires taking data dependent exceptions on load data for loads or AMOs, and also affects the exception priority, see xref:exception-priority[xrefstyle=short].
Ideally all implementatations would trap precisely (taking the tag into account) rather than conservatively (on every access). However, this information may not be available in all microarchitectures, and therefore we permit both implementation choices. This can be communicated to software by showing presence of the _{cheri_priv_crg_load_tag_ext}_ extension in addition to {cheri_priv_crg_ext}.
Both choices can be handled correctly by the same software, but {cheri_priv_crg_load_tag_ext} results in fewer spurious traps. 
```
Something like this maybe?
Since this is information for the uarch I think it should be non-normative?

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,117 +50,108 @@ include::img/sv48pte_crg.edn[]
...
+
+_{cheri_priv_crg_load_tag_ext}_ is an optimization for software, and as
```
arichardson: With the suggestion above we could drop this sentence.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,117 +50,108 @@ include::img/sv48pte_crg.edn[]
...
+
+When PTE.CRW is set, the PTE.CD bit indicates that a capability was stored to the
+virtual page since the last time the PTE.CD bit was cleared.
 
-NOTE: Implementations which already take synchronous traps on loaded data, such as ECC faults, should implement _{cheri_priv_crg_load_tag_ext}_ instead of {cheri_priv_crg_ext}.
+Capability dirty tracking behavior is enabled when, in addition to the rules above, all of the following are true:
 
-_{cheri_priv_crg_load_tag_ext}_ implies _{cheri_priv_crg_ext}_.
+. A capability store or AMO instruction is executed.
+. The {ctag} bit of the capability being stored to memory is set.
```
arichardson: Do we need a non-normative note that there is no implementation choice here whether to take the tag into account since the information should be present on all uarches?

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: it's possible to just set PTE.CRW if Svucrg is disabled, but if it's enabled then PTE.CD is also required to prevent traps on stores. Also PTE.CRG needs to match status.UCRG of course.

what's the alternative approach you are suggesting?

I've moved the text down as suggested

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: ok done

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: it's normative - to allow implementations not to _always_ check the tag - removing it would be a spec change.
which might be sensible, but I'm avoiding spec changes at the moment.

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: done

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: I've simplified the text

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: no I think it does exist otherwise - it might need a "the details of CRGE is furnished by the Svucrg extension" clause somewhere - but CRGE exists regardless

arichardson: Sounds good

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,117 +43,110 @@ include::img/sv48pte_crg.edn[]
...
 
-The CRW bit (defined by {cheri_priv_vmem_ext}) indicates whether reading or writing capabilities with the {ctag} set to
-the virtual page is permitted. When the CRW bit is set, capabilities are written
-as usual, and capability reads are controlled by the CRG bit.
+The PTE.CRW bit (defined by <<section_priv_cheri_vmem>>) indicates whether loading or storing capabilities to
+the virtual page is permitted.
 
-NOTE: The {ctag} bit of the stored capability is checked _after_ it is potentially
-cleared due to lack of <<c_perm>>.
+When <<sstatusreg_pte,sstatus>>.CRGE is clear, the PTE.CD and PTE.CRG bits are reserved and must not be set in any valid PTEs, otherwise the existing rules from <<section_priv_cheri_vmem>> are followed.
```
```suggestion
When <<sstatusreg_pte,sstatus>>.CRGE is clear, the PTE.CD and PTE.CRG bits are reserved and must not be set in any valid PTEs, and the existing rules from <<section_priv_cheri_vmem>> are followed.
```
Just noticed this is ambiguous, using and seems better

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,117 +43,110 @@ include::img/sv48pte_crg.edn[]
...
 | 1      | 1      | X       | Normal operation
 |===
 
-[#section_hardware_pte_updates]
-=== Enabling Software or Hardware PTE updates
+NOTE: Because the state of PTE.CD=1 and PTE.CRW=0 is illegal, it is possible for the update of PTE.CD to fail if another thread has cleared PTE.CRW.
 
-The decision about whether to take exceptions on capability stores with the {ctag} set to a page with PTE.CRW=1 and PTE.CD=0 is determined by whether the _Svade_ and _Svadu_ extensions are enabled.
-These cause PTE Accessed and Dirty updates to be done in software, via the exception handler, or by a hardware mechanism respectively.
+NOTE: Capability dirty tracking _always_ checks the {ctag} on stored capabilities, it is not optional as it is always available.
```
```suggestion
NOTE: Capability dirty tracking _always_ checks the {ctag} on stored capabilities, it is not optional as the value of the {ctag} is always available to the microarchitecture.
```
Maybe be more explicit about 'it'

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: outsdated - will fix separately

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,47 +1,19 @@
...
 NOTE: _Sv32_ (for RV32) does not have any spare PTE bits, and so no features from this chapter can be implemented.
 
-
-The {cheri_priv_crg_ext} extension is enabled when the <<sstatusreg_pte,sstatus>>.CRGE bit is set.
-When enabled, the extension adds the ability to perform capability revocation of user mode pages (see <<cap_revocation>>) by adding the PTE.CRG, PTE.CD and <<sstatusreg_pte,sstatus>>.UCRG bits as described below.
-When the extension is disabled the PTE.CD, PTE.CRG and <<sstatusreg_pte,sstatus>>.UCRG bits remain reserved.
-
-NOTE: A future extension may define an alternative function for these reserved states; the PTE.CRW bit is interpreted as defined by <<section_priv_cheri_vmem>>.
+The {cheri_priv_crg_ext} extension is enabled when the `sstatus.CRGE` bit is set.
+When enabled, the extension adds the ability to perform _capability revocation_ to prevent use-after-free and use-after-reallocation vulnerabilities of userspace pages using software algorithms such as listed in cite:[cornucopia-reloaded].
```
nwf: I hate to continue to be a pest, but... can this _please_ be more carefully worded?

1. It's still on supervisory software to implement revocation; this extension is _just_ providing some useful architectural traps.
2. It's not the general "capability revocation" that capability-systems people mean, so some qualification of the term is ideal.
3. CHERIvoke/Cornucopia/Reloaded has never attempted to prevent UAF, _just_ UAR.  (CHERIoT's load barrier closes the UAF/UAR gap, trapping on the earlier UAF, and CHERI+MTE makes the gap significantly less meaningful but does not completely _close_ it in its mismatching-stores-fizzle mode.)

Perhaps something like...

"When enabled, the extension adds the ability for supervisory software to quickly front-run loads of capabilities from userspace pages.  Such a facility has been shown to be a useful primitive for system-level implementation of certain forms of capability revocation which, in turn, allow userspace heap allocator software to deterministically guard against use-after-reallocation cite:[cornucopia-reloaded]."

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,142 +43,135 @@ include::img/sv48pte_crg.edn[]
...
+. _pte.crw_ is set.
+. The authorizing capability has <<c_perm>>.
+. The PMA is set to (_CHERI {ctag_title}_).
 
-* When a capability load or AMO instruction is executed, the {ctag} bit of the
-  loaded capability is cleared before it is written to the destination register.
-* When a capability store or AMO instruction is executed and the {ctag} bit
- of the capability being written is set, the implementation raises a
- {cheri_excep_name_pte_st}.
+NOTE: These primitives may be used to implement the load-barrier and store-barrier primitives from cite:[cornucopia-reloaded].
```
nwf: This is probably best deferred until after what those primitives are has been spelled out a bit?  I think it risks being confusing where it is.  Or maybe just delete this one, since each of the primitives have their own informative notes below.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,142 +43,135 @@ include::img/sv48pte_crg.edn[]
...
-When the CRW bit is set, the "capability state", then the behavior of
-capability loads is controlled by CRG and the behavior of capability stores
-is controlled by CD.
+==== {cheri_excep_name_pte_ld}s
 
-When CRW is set, the CRG bit indicates the current generation of the virtual
-memory page with regards to the ongoing capability revocation cycle.
-The implementation raises {cheri_excep_name_pte_ld} when a capability load
-or AMO instruction is executed and:
+NOTE: {cheri_excep_name_pte_ld}s may be used to implement the load-barrier primitive from cite:[cornucopia-reloaded].
```
nwf: Again, probably best to be after the normative description?

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,142 +43,135 @@ include::img/sv48pte_crg.edn[]
...
+. The loaded {ctag} is set.
 
-When CRW, CD, and CRG are all clear, the implementation is required to clear
-loaded tags and raise {cheri_excep_name_pte_st} when the stored {ctag} is set.
+NOTE: Checking the value of the {ctag} requires taking data dependent exceptions on load data for loads or AMOs, and also affects the exception priority, see xref:exception-priority[xrefstyle=short].
+ Ideally all implementations would trap precisely (taking the {ctag} into account) rather than conservatively (on every capability access).
+ However, this information may not be available in all implementations, and therefore both choices are permitted.
+ This can be communicated to software by showing presence of the _{cheri_priv_crg_load_tag_ext}_ extension which implies {cheri_priv_crg_ext}.
+ Implementations which already take synchronous traps on loaded data, such as ECC faults, should implement _{cheri_priv_crg_load_tag_ext}_.
+ Both choices can be handled correctly by the same software, but {cheri_priv_crg_load_tag_ext} results in fewer spurious traps.
```
nwf:
```suggestion
 Both choices can be handled correctly by the same software, but {cheri_priv_crg_load_tag_ext} is expected to result in fewer spurious traps.
```

I don't think we have ever measured?

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,142 +43,135 @@ include::img/sv48pte_crg.edn[]
...
 
-NOTE: Other CD and CRG combinations when CRW=0 are reserved for future extensions.
-The reserved PTE states behave as the CRW=0, CD=0, CRG=0, unless a future extension
-defines an alternative function.
+If the _{cheri_priv_crg_load_tag_ext}_ extension is implemented, implementations are allowed to conservatively fault under certain conditions even if the {ctag} is not set.
```
nwf:
```suggestion
Even if the _{cheri_priv_crg_load_tag_ext}_ extension is implemented, implementations are still allowed to conservatively fault in some situations in which the {ctag} is not set.
```

Do we want to say what those situations _are_?  If we have such a carve-out, is the distinction still useful to software?

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,142 +43,135 @@ include::img/sv48pte_crg.edn[]
...
+==== Capability Dirty Tracking
 
-NOTE: _{cheri_priv_crg_load_tag_ext}_ is an optimization for software, and as
-such implementations are allowed to conservatively fault under certain
-conditions even if the {ctag} is not set.
+NOTE: Capability dirty tracking may be used to implement the store-barrier primitive from cite:[cornucopia-reloaded].
 
-NOTE: Implementations which already take synchronous traps on loaded data, such as ECC faults, should implement _{cheri_priv_crg_load_tag_ext}_ instead of {cheri_priv_crg_ext}.
+When _pte.crw_ is set, the _pte.cd_ bit indicates that a capability was stored to the
+virtual page since the last time the _pte.cd_ bit was cleared.
```
nwf, line 113: It may be worth spelling out the analogy with `W` and `D`?  "This is akin to the _pte.d_ bit indicating that a store through this PTE since the last time _pte.d_ was cleared."

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,142 +43,135 @@ include::img/sv48pte_crg.edn[]
...
+
+==== Capability Dirty Tracking
 
-NOTE: _{cheri_priv_crg_load_tag_ext}_ is an optimization for software, and as
-such implementations are allowed to conservatively fault under certain
-conditions even if the {ctag} is not set.
+NOTE: Capability dirty tracking may be used to implement the store-barrier primitive from cite:[cornucopia-reloaded].
 
-NOTE: Implementations which already take synchronous traps on loaded data, such as ECC faults, should implement _{cheri_priv_crg_load_tag_ext}_ instead of {cheri_priv_crg_ext}.
+When _pte.crw_ is set, the _pte.cd_ bit indicates that a capability was stored to the
```
nwf, line 112: It's probably worth pointing out that, despite the intended analogy, this is actually subtly different from `W` and `D`: because the latter are reserved, it's fine, if unusual, to have a `W=0 D=1` PTE.  "This PTE is (in part) responsible for having dirtied this page, but no further stores are presently permitted."  I'm not sure that state is useful, but it's there and defined, while `CRW=0 CD=1` is reserved.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,142 +43,135 @@ include::img/sv48pte_crg.edn[]
...
+virtual page since the last time the _pte.cd_ bit was cleared.
 
-_{cheri_priv_crg_load_tag_ext}_ implies _{cheri_priv_crg_ext}_.
+Capability dirty tracking behavior is enabled when, in addition to the rules above, all of the following are true:
 
-^2^ A future version of this specification may check an SCRG bit in <<sstatusreg_pte,sstatus>> for kernel revocation.
+. A capability store or AMO instruction is executed.
+. The {ctag} bit of the capability being stored to memory is set.
+. The _pte.cd_ bit is clear.
+. Any extension-specific mediation has not already cleared the stored {ctag}.
```
nwf: This should fall under "The {ctag} bit of the capability being stored to memory is set", shouldn't it?

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,142 +43,135 @@ include::img/sv48pte_crg.edn[]
...
-{cheri_excep_name_pte_ld} is raised when a capability load or AMO instruction
-satisfies the conditions above and the capability read from memory
-has its {ctag} set.
+The implementation raises a {cheri_excep_name_pte_ld} when, in addition to the rules above, all of the following are true:
 
-{cheri_priv_crg_load_tag_ext} implies {cheri_priv_crg_ext}.
+. A capability load or AMO is executed.
+. _pte.crg_ bit does not equal `sstatus.UCRG`.
+. _pte.u_ is set.
+. Any other platform specific rules have not cleared the loaded {ctag}.
```
nwf: "have not cleared" is maybe not quite the right tense: the capability hasn't yet been loaded, so its tag cannot yet have been cleared?


```suggestion
. Any other platform specific rules have not forced the loaded {ctag} to be clear.
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,142 +43,135 @@ include::img/sv48pte_crg.edn[]
...
-^2^ A future version of this specification may check an SCRG bit in <<sstatusreg_pte,sstatus>> for kernel revocation.
+. A capability store or AMO instruction is executed.
+. The {ctag} bit of the capability being stored to memory is set.
+. The _pte.cd_ bit is clear.
+. Any extension-specific mediation has not already cleared the stored {ctag}.
+
+Two schemes for this are permitted, and the scheme in use is determined by whether the _Svade_ or _Svadu_ extensions are enabled.
+
+* For _Svade_, take a {cheri_excep_name_pte_st}.
+** This allows software interpretation of this state.
```
nwf: Yyyyes, but given that `Svadu` exists, that interpretation is pretty much required, isn't it?

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,142 +43,135 @@ include::img/sv48pte_crg.edn[]
...
+* For _Svadu_, do a hardware update which sets _pte.cd=1_, following the same rules as setting _pte.a_/_pte.d_.
+** When setting _pte.cd_, the hardware update also sets _pte.a_ and _pte.d_.
```
nwf:
```suggestion
** When setting _pte.cd_, the hardware update also necessarily sets (or leaves set) _pte.a_ and _pte.d_.
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,142 +43,135 @@ include::img/sv48pte_crg.edn[]
...
+** _pte.cd_ cannot be set speculatively, following the same rule as _pte.d_.
```
nwf: "To remind" or something, since the prose just said that we're following the same rules?

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,142 +43,135 @@ include::img/sv48pte_crg.edn[]
...
+| 0       | 0      | 0       | {cheri_excep_name_pte_st} if to-be-stored {ctag} is set
+| 0       | 0      | 1       | Reserved
+| 0       | 1      | X       | Reserved
+| 1       | 0      | X       | {cheri_excep_name_pte_st} if to-be-stored {ctag} is set (_Svade_), or hardware _pte.cd_ update (_Svadu_)
+| 1       | 1      | X       | Normal operation
 |===
 
-[#section_hardware_pte_updates]
-=== Enabling Software or Hardware PTE updates
+NOTE: Because the state of _pte.cd=1_ and _pte.crw=0_ is illegal, it is possible for the update of _pte.cd_ to fail if another thread has cleared _pte.crw_.
```
nwf, line 141: Strictly, an update to _pte.d_ can fail in exactly the same way if _pte.w_ has been cleared.  That is, it's not _because_ `CRW=0 CD=1` is illegal that this fails...

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,142 +43,135 @@ include::img/sv48pte_crg.edn[]
...
 
-The decision about whether to take exceptions on capability stores with the {ctag} set to a page with PTE.CRW=1 and PTE.CD=0 is determined by whether the _Svade_ and _Svadu_ extensions are enabled.
-These cause PTE Accessed and Dirty updates to be done in software, via the exception handler, or by a hardware mechanism respectively.
+NOTE: Capability dirty tracking _always_ checks the {ctag} on stored capabilities, it is not optional as the value of the {ctag} is always available to the microarchitecture.
```
nwf: That's been a point of contention.  In speculative execution, the AGU result for a store instruction may be ready significantly before the datum to be stored is.  Historically, for at least memory-like regions of the address space, there's been no need to hold the store instruction itself in speculation after the AGU result and translation thereof have completed, but this makes translation's trapping behavior dependent on the data as well.

An open research question is whether some one-sided error here is tolerable, especially in combination with `Svadu` (cap-dirtying a cap-capdirtyable page even if it turns out that the cap-sized datum in fact had a clear tag bit).

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,142 +43,135 @@ include::img/sv48pte_crg.edn[]
...
-
-[#xstatus_pte]
-=== Extending the Supervisor (sstatus) and Virtual Supervisor (vsstatus) Status Registers
+==== Minimum support
 
-The <<sstatusreg_pte,sstatus>> and <<vsstatusreg_pte,vsstatus>> CSRs are extended to include the new Capability Read Generation (CRG) and Capability Read Generation Enable (CRGE) bits as shown.
-
-When V=1 <<vsstatusreg_pte,vsstatus>>.UCRG and CRGE are in effect.
+The minimum level of PTE support is to set _pte.crw_ and _pte.cd_ to 1 in all PTEs intended for storing
+capabilities (i.e., private anonymous mappings) and leave
```
nwf:
```suggestion
capabilities (e.g., private anonymous mappings) and leave
```

While I am not enamored of shared and/or nonymous cap-capable mappings, they are things that get built.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,142 +43,135 @@ include::img/sv48pte_crg.edn[]
...
+`sstatus.UCRG` and _pte.crg_ in all PTEs set to 0, which will allow
```
nwf:
```suggestion
`sstatus.UCRG` on all harts and _pte.crg_ in all PTEs set to 0, which will allow
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,142 +43,135 @@ include::img/sv48pte_crg.edn[]
...
-NOTE: The CRGE bit is added to the <<mstatusreg_pte,mstatus>>, <<vsstatusreg_pte,vsstatus>> and <<sstatusreg_pte,sstatus>> registers in order to allow the S-mode software to explicitly opt in to the revocation extensions.
+The `mstatus`, `sstatus` and `vsstatus` CSRs are extended to include the new Capability Read Generation (CRG) and Capability Read Generation Enable (CRGE) bits.
 
+[#sstatus_crge_note]
 [NOTE]
 ====
 
-The CRGE bit is necessary because the {cheri_priv_crg_ext} defines a different semantic for PTE.CRW=1 PTE.CD=0 PTE.CRG=0 with respect to <<section_priv_cheri_vmem>>.
-In particular, an oblivious software implementation that only sets the PTE.CRW bit would observe one of the following:
+The CRGE bit is necessary because {cheri_priv_crg_ext} defines a different semantic for the _pte.crw=1_, _pte.cd=0_ and _pte.crg=0_ state with respect to <<section_priv_cheri_vmem>>.
```
nwf, line 172: Just to check, this is necessary because CRW is defined in the Y base and CD is defined as part of `Svucrg`?

src/cheri/riscv-priv-integration.adoc
```diff
@@ -603,13 +595,9 @@ CHERI adds the concept of _CHERI page faults_. They are split into:
...
 
-The PTE.CRW bit allows {cheri_excep_name_pte_st}s to be raised.
-
-CHERI harts that implement Sv39 must also implement a revocation scheme to prevent use-after-free attacks.
-
-The current revocation scheme (<<section_cheri_priv_crg_ext,{cheri_priv_crg_ext}>>) also allows {cheri_excep_name_pte_ld}s to be raised.
+The _pte.crw_ bit allows {cheri_excep_name_pte_st}s to be raised.
 
-NOTE: A future extension may define an improved scheme.
+NOTE: {cheri_excep_name_pte_ld} faults are only raised if <<section_cheri_priv_crg_ext,{cheri_priv_crg_ext}>> and Svade are implemented.
```
nwf:
```suggestion
NOTE: {cheri_excep_name_pte_ld} faults are at present raised only if <<section_cheri_priv_crg_ext,{cheri_priv_crg_ext}>> and Svade are implemented.
```

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: done!

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: This is certainly true - however the stored tag (along with the rest of the store data) must be known before the commit point of the store - i.e. before the point where the determination is made whether to raise the exception.

This is not true for loads on any CPU I've ever worked on - the exception detection is always done before the load data is known. RISC-V does offer load data triggers, so such a thing is in the architecture, but it's optional fortuantely.

So I think the split here is sensible.

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: yes - you _always_ get CRW if you have virtual memory and CD is optional as part of Svucrg.

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: I was thinking of unusual corner cases like ECC failures or bus errors. Software should still be able to rely on the loaded tag in the normal case - so maybe we just drop this.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,142 +46,135 @@ include::img/sv48pte_crg.edn[]
...
+virtual page since the last time the _pte.cd_ bit was cleared.
 
-NOTE: Implementations which already take synchronous traps on loaded data, such as ECC faults, should implement _{cheri_priv_crg_load_tag_ext}_ instead of {cheri_priv_crg_ext}.
+Capability dirty tracking behavior is enabled when, in addition to the rules above, all of the following are true:
 
-_{cheri_priv_crg_load_tag_ext}_ implies _{cheri_priv_crg_ext}_.
+. A capability store or AMO instruction is executed.
+. The {ctag} bit of the capability being stored to memory is set.
+. The _pte.cd_ bit is clear.
+. Any extension-specific mediation has not already cleared the stored {ctag}.
```
```suggestion
```

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: yes

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo:
```suggestion
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,142 +46,134 @@ include::img/sv48pte_crg.edn[]
...
+. The _pte.cd_ bit is clear.
 
-^2^ A future version of this specification may check an SCRG bit in <<sstatusreg_pte,sstatus>> for kernel revocation.
+Two schemes for this are permitted, and the scheme in use is determined by whether the _Svade_ or _Svadu_ extensions are enabled.
+
+* For _Svade_, take a {cheri_excep_name_pte_st}.
+** This allows software interpretation of this state.
+* For _Svadu_, do a hardware update which sets _pte.cd=1_, following the same rules as setting _pte.a_/_pte.d_.
+** When setting _pte.cd_, the hardware update also sets _pte.a_ and _pte.d_.
+** _pte.cd_ cannot be set speculatively, following the same rule as _pte.d_.
```
```suggestion
```

src/cheri/cheri-pte-ext.adoc
nwf: I agree that the split is sensible, but, just to emphasize the point I was trying to make: in earlier systems, it's generally _not_ required that you know _what_ you're storing, just _where_ you're storing it, before you can be certain not to trap on that store.  That is, the store _instruction_ can speculatively retire (is that the right phrase? speculatively commit?), promising to not trap, if the store buffer is able to hold an as-yet-unknown value at a (speculatively) known address.  That is, while program-order-earlier instructions may yet still be unresolved (and may yet trap!) -- in particular, the ones computing the value to be stored -- if the AGU and TLB have signed off, the store _instruction_ is done and it might be cheaper to track its uncommitted effects in the store buffer than in the general instruction speculation state.

tariqkurd-repo: ok- try my updated version.

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: deleted.

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: done.

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: this is true - I'm just pointing out this extra condition.

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: added

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: note added

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,47 +1,22 @@
...
 NOTE: _Sv32_ (for RV32) does not have any spare PTE bits, and so no features from this chapter can be implemented.
 
+The {cheri_priv_crg_ext} extension is enabled when the `sstatus.CRGE` bit is set.
 
-The {cheri_priv_crg_ext} extension is enabled when the <<sstatusreg_pte,sstatus>>.CRGE bit is set.
-When enabled, the extension adds the ability to perform capability revocation of user mode pages (see <<cap_revocation>>) by adding the PTE.CRG, PTE.CD and <<sstatusreg_pte,sstatus>>.UCRG bits as described below.
-When the extension is disabled the PTE.CD, PTE.CRG and <<sstatusreg_pte,sstatus>>.UCRG bits remain reserved.
-
-NOTE: A future extension may define an alternative function for these reserved states; the PTE.CRW bit is interpreted as defined by <<section_priv_cheri_vmem>>.
+When enabled, the extension adds the ability for supervisor-mode software to quickly front-run loads of capabilities from userspace pages, and to track capabilities stores to pages.
```
nwf:
```suggestion
When enabled, the extension adds the ability for supervisor-mode software to quickly front-run loads of capabilities from userspace pages, and to track stores of capabilities to pages.
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,47 +1,22 @@
...
-
-The minimum level of PTE support is to set CRW and CD to 1 in all PTEs intended for storing
-capabilities (i.e., private anonymous mappings) and leave
-<<sstatusreg_pte,sstatus>>.UCRG and CRG in all PTEs set to 0, which will allow
-capabilities with their {ctag}s set to be loaded and stored successfully.
-
-NOTE: When {cheri_priv_crg_ext} is enabled (<<sstatusreg_pte,sstatus>>.CRGE is set), the semantics of PTE.CRW bit are redefined relative to the definition in <<section_priv_cheri_vmem>>.
-The ability to load and store capabilities in the page depends on all of the PTE.CRW, PTE.CD, PTE.CRG bits, which effectively become a single 3-bit field controlling the behavior.
-Therefore, when PTE.CRW is clear, it is not required that future extensions to {cheri_priv_crg_ext}, such as for kernel revocation, cannot store capabilities to a page.
+NOTE: {cheri_priv_crg_ext} is not a mandatory part of the {cheri_base64_ext_name} Privileged ISA, as a future version of this specification may specify an improved method.
```
nwf: "may specify an improved method" is awkward, but I don't immediately have anything better.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,47 +1,22 @@
...
 
+The {cheri_priv_crg_ext} extension is enabled when the `sstatus.CRGE` bit is set.
 
-The {cheri_priv_crg_ext} extension is enabled when the <<sstatusreg_pte,sstatus>>.CRGE bit is set.
-When enabled, the extension adds the ability to perform capability revocation of user mode pages (see <<cap_revocation>>) by adding the PTE.CRG, PTE.CD and <<sstatusreg_pte,sstatus>>.UCRG bits as described below.
-When the extension is disabled the PTE.CD, PTE.CRG and <<sstatusreg_pte,sstatus>>.UCRG bits remain reserved.
-
-NOTE: A future extension may define an alternative function for these reserved states; the PTE.CRW bit is interpreted as defined by <<section_priv_cheri_vmem>>.
+When enabled, the extension adds the ability for supervisor-mode software to quickly front-run loads of capabilities from userspace pages, and to track capabilities stores to pages.
+Such facilities have been shown to provide useful primitives for system-level implementation of certain forms of capability revocation which, in turn, allow userspace heap allocator software to deterministically guard against use-after-reallocation cite:[cornucopia-reloaded].
```
nwf, line 9: Thanks for accepting proposed text. :)

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,142 +46,143 @@ include::img/sv48pte_crg.edn[]
...
-NOTE: The behavior in this section isn't relevant if:
-
-. The authorizing capability doesn't have <<c_perm>>, for loads, stores and AMO.
-. Any extension-specific mediation has already cleared the stored tag, for stores and AMOs.
+=== Software Primitives
 
-The CRW bit (defined by {cheri_priv_vmem_ext}) indicates whether reading or writing capabilities with the {ctag} set to
-the virtual page is permitted. When the CRW bit is set, capabilities are written
-as usual, and capability reads are controlled by the CRG bit.
+The _pte.crw_ bit (defined by <<section_priv_cheri_vmem>>) indicates whether loading or storing capabilities to
```
nwf: This sentence feels a little out of place as the first in a "Software Primitives" section.  Separately, perhaps
```suggestion
The _pte.crw_ bit (defined by <<section_priv_cheri_vmem>>) continues to indicate whether loading or storing capabilities to
```

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo:
```suggestion
NOTE: {cheri_priv_crg_ext} is not a mandatory part of the {cheri_base64_ext_name} Privileged ISA, as a future extension may specify an alternative mechanism giving improved software efficiency.
```

any better?

arichardson: maybe mechanism instead of algorithm?

arichardson (approved):
LGTM but would like to see one more final check by @nwf / @qwattash

src/supervisor.adoc
```diff
@@ -1627,9 +1633,11 @@ type.
 * Perform the following steps atomically:
 ** Compare _pte_ to the value of the PTE at address __a__+__va.vpn__[__i__]×PTESIZE.
 ** If the values match, set _pte_._a_ to 1 and, if the
-original memory access is a store, also set _pte_._d_ to 1. Then store _pte_ to the PTE at address __a__+__va.vpn__[__i__]×PTESIZE.
+original memory access is a store, also set _pte_._d_ to 1, and if the original memory access is a capability store and _pte_._crw_ is set, also set _pte_._cd_ to 1. Then store _pte_ to the PTE at address __a__+__va.vpn__[__i__]×PTESIZE.
```
arichardson:
```suggestion
original memory access is a store, also set _pte_._d_ to 1, and if the original memory access is a capability store with the to-be-stored {ctag} set. and _pte_._crw_ is set, also set _pte_._cd_ to 1. Then store _pte_ to the PTE at address __a__+__va.vpn__[__i__]×PTESIZE.
```

src/supervisor.adoc
```diff
@@ -1614,11 +1616,15 @@ __a__=__pte__.__ppn__×PAGESIZE and go to step 2.
...
+. Determine if the requested memory access is allowed for {cheri_base64_ext_name} when {cheri_priv_crg_ext} is enabled, by the _pte_._crw_, _pte_._u_, _pte_._crg_ bits and the UCRG fields of the *status* register, given the rules in <<pte_crw_crg_load_summary>>. If not, stop and raise a {cheri_excep_name_pte_ld} exception.
+
+. Determine if the requested memory access is allowed for {cheri_base64_ext_name} by the _pte_._crw_ bit given the rules in <<pte_crw_summary>>. If not, stop and raise a {cheri_excep_name_pte_st} exception.
+
 . Determine if the requested memory access is allowed by the _pte_._r_, _pte_._w_, and _pte_._x_ bits, given the Shadow Stack Memory Protection rules. If not, stop and raise an access-fault exception.
 
 . Determine if the requested memory access is allowed by the _pte_._r_, _pte_._w_, and _pte_._x_ bits. If not, stop and raise a page-fault exception corresponding to the original access type.
 
-. If _pte_._a_=0, or if the original memory access is a store and _pte_._d_=0:
+. If _pte_._a_=0, or if the original memory access is a store and _pte_._d_=0, or if the original memory access is a capability store and _pte_._cd_=0^1^:
```
arichardson:
```suggestion
. If _pte_._a_=0, or if the original memory access is a store and _pte_._d_=0, or if the original memory access is a capability store with the to-be-stored {ctag} set and _pte_._cd_=0^1^:
```

nwf (approved):
I'm happy to land this given that there's a rework coming, as discussed today.  _If_ we can use this branch as the base of a PR in this repository (does GH allow such things?), it might make sense to keep this PR open and land both at once.

arichardson:
> I'm happy to land this given that there's a rework coming, as discussed today. _If_ we can use this branch as the base of a PR in this repository (does GH allow such things?), it might make sense to keep this PR open and land both at once.

Let's go ahead and merge this to make it easier

src/cheri/cheri-pte-ext.adoc
```diff
@@ -71,142 +46,144 @@ include::img/sv48pte_crg.edn[]
...
-
-. The authorizing capability doesn't have <<c_perm>>, for loads, stores and AMO.
-. Any extension-specific mediation has already cleared the stored tag, for stores and AMOs.
+=== Software Primitives
 
-The CRW bit (defined by {cheri_priv_vmem_ext}) indicates whether reading or writing capabilities with the {ctag} set to
-the virtual page is permitted. When the CRW bit is set, capabilities are written
-as usual, and capability reads are controlled by the CRG bit.
+The _pte.crw_ bit (defined by <<section_priv_cheri_vmem>>) continues to indicate whether loading or storing capabilities to
+the virtual page is permitted.
```
qwattash, line 52: Is this something meant to say that CRW=0 does and always will be sufficient to assert that capability load and stores are not allowed (with some different trapping semantic)? The point of the previous rework was to allow to reuse the reserved bit configuration in ways that are not compatible with CRW=0 meaning that capability load and stores are disallowed.

qwattash: This seems to be explained well below.

PR #926, opened 2026-01-29
Clarify M=1 X=0 spec

### LINKED ISSUE #925: YBLD when rs2 has X=0 M=1
The behaviour of YBLD isn't clearly specified when rs2 has X=0, M=1 which is possible for the RV64 format.
If rs2 passes all the checks against rs1, should the output have the tag set? I think not.

We have said that:

`This field needs to be present only in capabilities granting X-permission, as it is only ever architecturally interpreted on the capability resident in pc.`

This statement is also incorrect, by the way, as YMODER/YMODEW use the M-bit when not in the pc.

In the RV64 encoding chapter we say: 

`When Zyhybrid is supported, capabilities include an M-bit (bit 52). If not supported the M-bit is reserved and reads as zero.`

If we change these to:

`The M-bit is only needs to be valid in capabilities granting X-permission. Capability encoding formats which can encoding M=1 and X=0 are required to specify the behavior`

`When Zyhybrid is supported, capabilities include an M-bit (bit 52). If not supported the M-bit is reserved and reads as zero. The M-bit is _reserved_ for capabilities which don't grant X-permission `

Would that solve it?

tariqkurd-repo:
Fixes #925 

Any encoding format which allows M=1 and X=0 (only RV64Y) must specify the behaviour, which is _reserved_ in this case.

### DIFF
--- a/src/cheri/riscv-unpriv-integration.adoc
+++ b/src/cheri/riscv-unpriv-integration.adoc
@@ -582,7 +582,7 @@ CHERI enforces the following rules for all valid capabilities:
 
 . The bounds are not <<section_cap_malformed,malformed>>.
 . The capability metadata does not have any bits, or fields set to a reserved value.
-.. An example of a reserved field is permissions which cannot be legally produced by <<CLRPERM>>.
+.. An example of a reserved field value is permissions which cannot be legally produced by <<CLRPERM>>.
 
 In all cases the capability could not have been legally created, and so either:
 
--- a/src/cheri/rvy64-encoding.adoc
+++ b/src/cheri/rvy64-encoding.adoc
@@ -23,8 +23,7 @@ The encoding diagram above of the capability format includes some fields which d
 
 <<section_cheri_hybrid_ext>>::
 When {cheri_default_ext_name} is supported, capabilities include an <<m_bit>> (bit 52).
-The <<m-bit>> is _reserved_ if the capability does not grant <<x_perm>>.
-If not supported the <<m_bit>> is always _reserved_ and so must be zero in valid capabilities.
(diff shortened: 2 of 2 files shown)


src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -581,8 +581,8 @@ endif::[]
 CHERI enforces the following rules for all valid capabilities:
 
 . The bounds are not <<section_cap_malformed,malformed>>.
-. No reserved bit in the capability encoding is set.
-. The permissions can be legally produced by <<CLRPERM>>.
+. The capability metadata does not have any bits, or fields set to a reserved value.
+.. An example of a reserved field is permissions which cannot be legally produced by <<CLRPERM>>.
```
```suggestion
.. An example of a reserved field value is permissions which cannot be legally produced by <<CLRPERM>>.
```

src/cheri/rvy64-encoding.adoc
```diff
@@ -23,7 +23,8 @@ The encoding diagram above of the capability format includes some fields which d
 
 <<section_cheri_hybrid_ext>>::
 When {cheri_default_ext_name} is supported, capabilities include an <<m_bit>> (bit 52).
-If not supported the <<m_bit>> is reserved and reads as zero.
+The <<m-bit>> is _reserved_ if the capability does not grant <<x_perm>>.
+If not supported the <<m_bit>> is always _reserved_ and so must be zero in valid capabilities.
```
```suggestion
If the capability does not grant <<x_perm>> or {cheri_default_ext_name} is not supported, the <<m_bit>> is always _reserved_ and so must be zero in valid capabilities.
```
merge the two sentences?

nwf (approved):

PR #932, opened 2026-02-02
Clarify what the various Xstateen0.TID bits control access to.

### LINKED ISSUE #928: xstateen0.TID also controls access to vstidc
In section 4.2. State Enable 0 Registers

> If the base architecture is RVY then the TID bit controls access to the stidc and utidc CSRs.

I believe that vstidc should be included in that list.

(For example, just below, the description of the CSRIND bit includes vsiselect in the equivalent list of registers, and the description of the IMSIC bit includes vstopei.)

Arguably, a fuller description similar to the text for other bits would be helpful?

As a suggestion:
"If the base architecture is RVY then the TID bit in mstateen0 controls access to the stidc, vstidc, and utidc CSRs. The TID bit in hstateen0 controls access to the stidc (really vstidc) CSR. The TID bit in sstateen0 controls access to the utidc CSR."

awiseAtCodasip:
Fix #928.

In particular, that mstateen0.TID affects access to all of stidc, vstidc, and utidc.

### DIFF
--- a/src/cheri/contributors.adoc
+++ b/src/cheri/contributors.adoc
@@ -39,5 +39,6 @@ This RISC-V specification has been contributed to directly or indirectly by:
 * Robert N. M. Watson <robert.watson@cl.cam.ac.uk>
 * Toby Wenman <toby.wenman@codasip.com>
 * Jay Williams <jay.williams@codasip.com>
+* Adrian Wise <adrian.wise@codasip.com>
 * Jonathan Woodruff <jonathan.woodruff@cl.cam.ac.uk>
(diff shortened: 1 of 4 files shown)


tariqkurd-repo (approved):
LGTM - I think this is how it's supposed to work

arichardson (approved):
I haven't really looked into the hypervisor spec, but the change looks sensible to me.

tariqkurd-repo:
any objections to merging this one? it just brings the CHERI spec in line with the smstateen spec

Issue #924, opened 2026-01-27
Hypervisor exception priority table needs CHERI exceptions


tariqkurd-repo:
https://github.com/riscv/riscv-isa-manual/blob/5bf832922a583a5ec3254e59fe8fb616df51deed/src/hypervisor.adoc?plain=1#L2445

the hypervisor chapter lists CHERI exceptions in the hcause table, but they haven't been added to the exception priority table linked above.

Issue #923, opened 2026-01-26
Svucrg: CHERI load page fault should be CHERI load/AMO page fault


tariqkurd-repo:
We have added:

`CHERI load page fault`
`CHERI store/AMO page fault`

To be consistent with the main RISC-V spec.
However:

1. The `CHERI load page fault` is also relevant to `AMOSWAP.Y`, which will trap with a CRG mismatch
2. It is possible to trigger *both* faults from `AMOSWAP.Y`

So we need to rename it to `CHERI load/AMO page fault`, and we need to specify the priority if both fire.
I'd personally go for the load page fault for Svucrg, or store  page fault for Svucrglct.

(curently `AMOSWAP.Y` is the only capability AMO - but there'll be more in the future).

tariqkurd-repo:
>None of these are clear about which apply only if the tag of the capability being loaded/stored is set

We can add more details, but the "determine if the requested memory access is allowed by..." means that the reader needs to look elsewhere to find those rules - so I think it's actually correct.

> Given AMOs are RMW operations, it also seems odd to have the store exception happen before the load generation exception? For AMOSWAP it doesn't really matter because the two are concurrent, but for other AMOs they really are RMW and so there's an ordering issue there that we don't want to get the wrong way round and make it more awkward to support other AMOs in future.

Yes good point - but then we go back to square one and have "CHERI load/AMO generation page fault" and "CHERI store/AMO page fault" don't we?

Thinking about this further in the base architecture all AMOs will take a `store/AMO page fault` , even if it's the load part which has technically failed (pte.r=0) but then write only pages aren't permitted so it can be argued that the whole access fails in this case. So I think we have created something genuinely new in that the load part of an AMO can fail whereas the store part can succeed.

So maybe it should be, similar to what @jrtc27 suggested, be  `CHERI capability generation page fault` without tying it to a load or AMO and in theory it could be caused by a store in the future.

...and we *should* prioritise the generation page fault over the store page fault as the load must be succeed _first_ for the store to be issued as @jrtc27 pointed out - I'll update https://github.com/riscv/riscv-cheri/pull/919

arichardson:
> > > But if you're AMOADDing something with a 0 tag then you should never take that store fault?
> > 
> > 
> > I think you're right (CMP)SWAP is the only AMO where we could have a tag on the stored value, so maybe AMOCAS is the only one we need to worry about.
[...]

Issue #922, opened 2026-01-23
Svucrg missing additional PTE state used by the Cornucopia-style revoker.


qwattash:
# The Problem
In the presence of virtual page aliases (_X_ and _Y_) of a single physical page (_P_), the revoker can reach a state in which capabilities stored via an alias _Y_ have their tag stripped when loaded via alias _X_ when the revoker attempts to transition the page to the **cap-clean/idle** state.

## How we reach the state
Consider that the revoker has already performed a scan of both _X_ and _Y_ during the previous epoch E<sub>n-1</sub>.
Suppose that no capabilities were found in the page, therefore both _X_ and _Y_ are in the **dirtiable** state (CRW=1 CD=0 CRG=**current-generation**)

Suppose that the revoker reaches the PTE for _X_ during the current epoch E<sub>n</sub> and finds no capabilities in the page.
Then, the revoker will attempt to continue the process to detect an **idle** page, making the PTE _X_ as **cap-clean** (CRW=0 CD=0 CRG=0).

Suppose now that a capability store via _Y_ is performed before the revoker reaches _Y_. This will cause the PTE _Y_ to become **cap-dirty** (CRW=1, CD=1, CRG=**previous-generation**), but the store succeeds as expected.

Now, the same capability can not be loaded via the mapping _X_, because the **cap-clean** state has tag-stripping semantic.

## Why this is happening
This is similar to the problem previously identified with the 2-bit PTE solution. In this case the PTE differs from the semantic of the existing Morello and Toooba implementations because the cap-clean state in these can be configured to:
 1. Strip tags on load, trap on store (same as the current RVY spec) e.g. on Morello LC=0 SC=0 CDBM=0 will do this.
 2. Permit capability loads gated by generation, trap on stores. e.g. on Morello this is expressed by LC=**enabled with generation number** SC=0 CDBM=0

In practice, this is possible in other implementations because the store-side is completely orthogonal from the load-side behaviour, at the cost of even more PTE bits.

# Possible solutions
First of all, there are two observations to make. First, this issue does not violate the temporal safety invariant. Second, it is strictly limited to aliasing pages.

This does not occur on non-aliasing pages because of the following:
 1. Consider that a non-aliasing PTE mapping _X_ is transitioned from **dirtiable** to **cap-clean** (CRW=1 CD=0 CRG=X) --> (CRW=0, CD=0, CRG=0).
[...]

Issue #915, opened 2026-01-07
Svucrglct should imply Svucrg, not the reverse


tariqkurd-repo:
Currently Svucrg is an extension, and Svucrglct is offered as a refinement to it.
Svucrg conditionally traps on every loaded cap _with/without the tag being set_.
Svucrglct conditionally traps on every loaded cap _only if the tag is set_.

The "lct" is for "load capability tag"

(these traps are if PTE.CRW=1, PTE.U=1, PTE.CRG!=sstatus.UCRG)

Currently the text says that Svucrglct implies Svucrg, but it seems the reverse may be correct.
Svucrglct is the basic behaviour - trap if the tag is set.
Svucrg adds in the additional case of trapping on loaded caps where the tag is clear.

This seems more logical.

jrtc27:
By that "logic", trapping to emulate A/D would be an extension to hardware-managed A/D. I think you're too tied up on a very narrow view of what "refinement" means.

avakar:
Actually, trapping to emulate A/D _is_ an extension to hardware-managed A/D (called Svade).

jrtc27:
The way you specify it is:

Svucrg - capability loads of tagged values will trap, and of untagged values may trap, depending on the implementation
Svucrglct - capability loads of untagged values will not trap

If you want another example to compare this to, think about misaligned accesses. Definitely not trapping on them is an extension, without which you don't know.

avakar:
> Svucrglct removes trapping cases

It doesn't though, traps are allowed for invalid tags under Svucrglct. The extension would absolutely make sense if it did forbid traps, but it currently doesn't.

> The way you specify it is:

I think we're misunderstanding each other.

Current specification (unless I'm misreading the text):

Svucrg - capability loads of tagged and untagged values will trap
Svucrglct -  capability loads of tagged values will trap, and of untagged values may trap, depending on the implementation

My suggestion:

Svucrg - capability loads of tagged values will trap, and of untagged values may trap, depending on the implementation
Svucrglct - doesn't exist (edit: or always trap on tagged, never trap on untagged, that would be perfectly meaningful)

jrtc27:
"If Svucrglct is implemented then the trap is only taken if the loaded capability tag is set, to reduce software overhead from revocation sweeps."

arichardson:
Looking over the spec again, it is split across multiple sections and could be clearer, but it absolutely just removes a trapping case.

avakar:
[...]

PR #933, opened 2026-02-03
4-bit pte scheme to support all states of existing revoker software

tariqkurd-repo:
* Introduce a 4-bit RVY PTE field.
* If no revocation extension is implemented, we continue to
  have a single bit (now called Y) which enables cap loads+stores.
* If Svyrg (new name) is implemented, then implement a 4-bit
  scheme based on Morello.
* Add kernel revocation in the same way as Morello with sstatus.SYRG (new name)

It inherits the following from #929:

* Combine and rename CHERI load page faults into a single lowest priority
  fault (fixes https://github.com/riscv/riscv-cheri/issues/923)
* Remove Svucrglct - require exception when tag is set, allow exception
  when tag not set, no change in exception priority for the different
  cases (fixes https://github.com/riscv/riscv-cheri/issues/915)
* Update all exception tables to show updated priority (includes fix
  for https://github.com/riscv/riscv-cheri/issues/924)
* Add rule about allowing SC.Y to set PTE.YD even if the store fails
* Add notes about (future) AMOCAS.Y (if rs2.tag=1), AMOADD.Y (always) to
  set YD even if the store fails
* Add rule/notes about ensuring YD may be set too often, and load
  capability faults may fire too often

Fixes: https://github.com/riscv/riscv-cheri/issues/915
Fixes: https://github.com/riscv/riscv-cheri/issues/923
Fixes: https://github.com/riscv/riscv-cheri/issues/922 
Fixes: https://github.com/riscv/riscv-cheri/issues/924
Fixes: https://github.com/riscv/riscv-cheri/issues/906 

---------

Co-authored-by: Nathaniel Wesley Filardo <VP331RHQ115POU58JFRLKB7OPA0L18E3@cmx.ietfng.org>
Co-authored-by: Jessica Clarke <jrtc27@jrtc27.com>
Co-authored-by: Alexander Richardson <mail@alexrichardson.me>

### DIFF
--- a/src/cheri/riscv-priv-integration.adoc
+++ b/src/cheri/riscv-priv-integration.adoc
@@ -441,8 +441,8 @@ There are three levels of support:
 | _CHERI {ctag_title} Fault_ | Load zero {ctag} | Store/AMO Access Fault on {ctag}^1^ | No support for {ctag}s, trap on storing one
 |=========================================================================================
 
-^1^ The access fault is triggered on all capability stores or atomics such as <<STORE_CAP>> or <<AMOSWAP_CAP>> when <<c_perm>> and <<w_perm>> are granted and the {ctag} is set to one.
- For SC{LD_ST_DOT_CAP}, the access fault is raised if the {ctag} is set in {cs2} even when the store fails.
+^1^ The access fault is triggered on all capability stores or atomics such as <<STORE_CAP>> or <<AMOSWAP_CAP>> when <<c_perm>> and <<w_perm>> are granted and the to-be-stored {ctag} is set to one.
+ For SC{LD_ST_DOT_CAP}, the access fault is raised if the to-be-stored {ctag} is set in {cs2} even when the store fails.
 
 Memory regions that do not have the _CHERI {ctag_title}_ PMA do not require storage for {ctag}s.
 


src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,169 @@
...
 
-The {cheri_priv_crg_ext} extension is enabled when the `sstatus.CRGE` bit is set.
-
-When enabled, the extension adds the ability for supervisor-mode software to quickly front-run loads of capabilities from userspace pages, and to track stores of capabilities to pages.
+The extension adds the ability for supervisor-mode software to quickly front-run loads of capabilities from userspace pages, and to track stores of capabilities to pages.
 Such facilities have been shown to provide useful primitives for system-level implementation of certain forms of capability revocation which, in turn, allow userspace heap allocator software to deterministically guard against use-after-reallocation cite:[cornucopia-reloaded].
 
-It achieves this by adding by adding the _pte.crg_, _pte.cd_ and `Xstatus.UCRG` bits as described below.
-When the extension is disabled the _pte.cd_, _pte.crg_ and `Xstatus.UCRG` bits remain reserved.
+The 4-bit {cheri_base_ext_name} field is sub-divided into four 1-bit fields, two controlling capability loads (including the capability loaded during an AMO) and two controlling capability stores (again, including the capability stored as part of an AMO).
```
qwattash: As discussed, this should become a 1 + 3 bits fields with the top field remaining with the current CRW semantic. Should that be renamed YRW or something along those lines?

The CRW bit will be bit 58, if I'm not mistaken.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -220,8 +201,7 @@ ifdef::cheri_standalone_spec[]
   {bits:  1, name: 'WPRI'},
   {bits:  1, name: 'MPELP'},
   {bits:  1, name: 'MDT'},
-  {bits: 17, name: 'WPRI'},
```
qwattash, line 223: I believe these are reintroduced for the 1 + 3 bits solution.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,169 @@
...
-| 0       | 0       | 0                      | X      | Clear loaded {ctag}
-| 0       | 0       | 1                      | X      | Reserved
-| 0       | 1       | X                      | X      | Reserved
-| 1       | X       | {ne} `sstatus.` `UCRG` | 1      | {cheri_excep_name_pte_ld}, or {cheri_excep_name_pte_ld} if {ctag} is set for _{cheri_priv_crg_load_tag_ext}_
-| 1       | X       | = `sstatus.` `UCRG`    | 1      | Normal operation
-| 1       | X       | X                      | 0      | Normal operation^1^
+|_pte.lyen_| _pte.lycrg_         | _pte.u_|tag^1^| Load Capability Behavior
+| 0        | 0                   | X      |  X   | Clear loaded {ctag}
+| 0        | 1                   | X      |  X   | Normal operation
+| 1        | X                   | 0      |  X   | Normal operation^2^
```
qwattash: I suspect that taking into account kernel revocation, it becomes easier to spell out the requirement in the `Load Capability Behavior` predicated on `sstatus.xCRG` with x meaning either UCRG or SCRG:

-  `pte.lyen = 1`, `pte.lycrg == sstatus.xCRG`, `tag = 0`: implementation defined load fault or normal op
-  `pte.lyen = 1`, `pte.lycrg == sstatus.xCRG`, `tag = 1`: normal op
-  `pte.lyen = 1`, `pte.lycrg != sstatus.xCRG`, `tag = 1`: load fault

Unless we want extra reserved states, I'd take out pte.U from the table in this manner.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,169 @@
...
-|_pte.crw_|_pte.cd_|_pte.crg_| Store/AMO
-| 0       | 0      | 0       | {cheri_excep_name_pte_st} if the to-be-stored {ctag} is set
-| 0       | 0      | 1       | Reserved
-| 0       | 1      | X       | Reserved
-| 1       | 0      | X       | {cheri_excep_name_pte_st} if the to-be-stored {ctag} is set (_Svade_), or hardware _pte.cd_ update (_Svadu_)
-| 1       | 1      | X       | Normal operation
+|_pte.syen_|_pte.yd_|tag^1^| Store Capability Behavior
+| 0        | 0      |  0   | Normal operation
+| 0        | 0      |  1   | {cheri_excep_name_pte_st}
+| 0        | 1      |  X   | Normal operation
```
qwattash: This state should trap, because `syen` is clear.
ISAv9 CHERI does:

 `syen = 0, yd = X` -> trap

although ISAv9 trap is tag-dependent, as you spell out in the table.
That said, it is fairly weird to have a dirty PTE without syen in the first place and I worry a bit that people will be tempted to use this as a SW-defined flag.

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: yes - included in the latest commit

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: yes - included in the latest commit

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: yes - included in the latest commit

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: yes - included in the latest commit

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,173 @@
...
-.Sv48 page table entry
-[#sv48pte_crg]
-include::img/sv48pte_crg.edn[]
+[[svycrg_rvy_field_breakdown]]
+.Breakdown of the 4-bit _pte_._{cheri_base_ext_name_lc}_ field.
+[%autowidth,float="center",align="center",cols="<,<,<",options="header"]
+|===
+| Bit                                | Name           | Comment
+3+| *Capability load/AMO fields*
+|_pte_._{cheri_base_ext_name_lc}[0]_ | _pte_._lye_    | Load capability enable
```
jrtc27: CR or YR, no E suffix

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,173 @@
...
+|_pte_._{cheri_base_ext_name_lc}[1]_ | _pte_._lycrg_  | Load capability read generation
```
jrtc27: CRG or YRG, L+R are redundant, ditto Y+C

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,173 @@
...
+3+| *Capability store/AMO fields*
+|_pte_._{cheri_base_ext_name_lc}[2]_ | _pte_._sye_    | Store capability enable
```
jrtc27: CW or YW

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,173 @@
...
+|_pte_._{cheri_base_ext_name_lc}[3]_ | _pte_._yd_     | Capability dirty bit
```
jrtc27: (CD if using C*x* above, else leave as YD)

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,173 @@
 [#section_cheri_priv_crg_ext]
 == "{cheri_priv_crg_ext}" Extension, Version 1.0 for {cheri_base64_ext_name}
 
-NOTE: _Sv32_ (for RV32) does not have any spare PTE bits, and so no features from this chapter can be implemented.
+This extension specifies use of the {cheri_base_ext_name} field from the PTE formats (e.g., see <<sv39pte>>).
+{cheri_priv_crg_ext} implies Sv39.
 
 The {cheri_priv_crg_ext} extension is enabled when the `sstatus.CRGE` bit is set.
```
jrtc27: YRGE if using YRG below

src/cheri/attributes.adoc
```diff
@@ -79,8 +80,8 @@ endif::support_varxlen[]
 :lr_sc_bh_ext_name:     Zabhlrsc
 
 // Extension for CHERI CRG bits
-:cheri_priv_crg_ext:          Svucrg
-:cheri_priv_crg_load_tag_ext: Svucrglct
+:cheri_priv_crg_ext:          Svycrg
+:cheri_priv_crg_ext_caps:     SVYCRG
```
jrtc27: Sv[yc]rg depending on the PTE bit name, not Y+C combined

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,173 @@
...
-. _pte.crw_ is set.
 . The authorizing capability has <<c_perm>>.
 . The PMA is set to (_CHERI {ctag_title}_).
 
 ==== {cheri_excep_name_pte_ld}s
 
+_pte_._lye_ is used to enable capability loads or AMOs to write a {ctag} to {cd}.
 
-When _pte.crw_ is set, _pte.crg_ can be used to trap on capability loads or AMOs when it does not match the value of `sstatus.UCRG`.
+When _pte_._lye_ is set, _pte_._lycrg_ can be used to trap on capability loads or AMOs when it does not match the Capability Read Generation (CRG) value which is represented by the value of `sstatus.UCRG` for userspace pages (_pte_._u=1_) or `sstatus.SCRG` for kernel pages (_pte_._u=0_).
```
jrtc27: status.U[YC]RG matching the PTE bit name

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,173 @@
...
 
 The implementation raises a {cheri_excep_name_pte_ld} when, in addition to the rules above, all of the following are true:
 
 . A capability load or AMO is executed.
-.. Where an AMO can raise both faults, {cheri_excep_name_pte_ld} is prioritized above {cheri_excep_name_pte_st}.
-. _pte.crg_ does not equal `sstatus.UCRG`.
-. _pte.u_ is set.
+. _pte_._lye_ is set.
+. if _pte_._u=1_ _pte_._lycrg_ does not equal `sstatus.UCRG`.
+. if _pte_._u=0_ _pte_._lycrg_ does not equal `sstatus.SCRG`.
```
jrtc27: sstatus.S[YC]RG matching the PTE bit name

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: as in "Y Read"?

jrtc27: Yes, it's RWX in PTEs not LSX

tariqkurd-repo: ok done

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: YD it is

src/cheri/attributes.adoc
tariqkurd-repo: Svyrg

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: UYRG

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: SYRG

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: YR and YRG

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: YW

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: YRGE

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,174 @@
...
-remain reserved and must be cleared by software for forward compatibility,
-or else a page-fault exception is raised. Additionally, if the hypervisor
-extension is enabled these bits remain reserved for leaf and non-leaf PTEs
-used in guest address translation.
+The {cheri_priv_crg_ext} extension is enabled when the `sstatus.{revocation_enable}` bit is set.
 
-.Sv39 page table entry
-[#sv39pte_crg]
-include::img/sv39pte_crg.edn[]
+NOTE: The enable bit is required to redefine the usage of the _pte_._{rv64y_pte4_field_name_lc}_ field.
```
qwattash: Is it worth spelling out also future-proofing if other schemes come into existence?

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,174 @@
...
 
 . A capability load or AMO is executed.
-.. Where an AMO can raise both faults, {cheri_excep_name_pte_ld} is prioritized above {cheri_excep_name_pte_st}.
-. _pte.crg_ does not equal `sstatus.UCRG`.
-. _pte.u_ is set.
+. _pte_._{pte_cap_r_name_lc}_ is set.
+. if _pte_._u=1_, _pte_._{pte_cap_crg_pte_name_lc}_ does not equal `sstatus.{revocation_u}`.
+. if _pte_._u=0_, _pte_._{pte_cap_crg_pte_name_lc}_ does not equal `sstatus.{revocation_s}`.
 . Any other platform specific rules have not forced the loaded {ctag} to be clear.
+. The loaded {ctag} is set.
```
qwattash: I think that while this belongs in the list, it is contradicted below where the requirement is relaxed. I wonder whether we can use a less prescriptive language in this list item.
Perhaps something along the lines of the following?

> The loaded {ctag} is set or the implementation does not support data dependent exceptions on loaded capabilities.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,174 @@
...
 |===
-|_pte.crw_|_pte.cd_|_pte.crg_| Store/AMO
-| 0       | 0      | 0       | {cheri_excep_name_pte_st} if the to-be-stored {ctag} is set
-| 0       | 0      | 1       | Reserved
-| 0       | 1      | X       | Reserved
-| 1       | 0      | X       | {cheri_excep_name_pte_st} if the to-be-stored {ctag} is set (_Svade_), or hardware _pte.cd_ update (_Svadu_)
-| 1       | 1      | X       | Normal operation
+|_pte_._{pte_cap_w_name_lc}_|_pte_._{pte_cap_dirty_name_lc}_|tag^1^| Store Capability Behavior
+| 0        | 0      |  0   | Normal operation
+| 0        | 1      |  X   | {cheri_excep_name_pte_st}
```
qwattash: The YW=0 YD=1 ctag=0 configuration is trapping here, however this contradicts the conditions outlined above:

> . A capability store or AMO instruction is executed.
> . The to-be-stored {ctag} is set.                            <---------------------
> . _pte_._{pte_cap_w_name_lc}_ is set.
> . _pte_._{pte_cap_dirty_name_lc}_ is clear.

and although it explained below in relation to dirty tracking, I think it would be useful to hint at this in the list, and also mention this case directly, as this is not strictly tracking the dirtiness but instead trapping conservatively.

src/cheri/riscv-priv-integration.adoc
```diff
@@ -625,8 +650,8 @@ When the CRW bit is set, capabilities are written as usual.
 
 If the CRW bit is clear then, in priority order for AMOs:
```
qwattash: I believe we need to rename CRW


```suggestion
If the {pte_cap_rw_name} bit is clear then, in priority order for AMOs:
```

src/cheri/riscv-priv-integration.adoc
qwattash:
```suggestion
The {pte_cap_rw_name} bit indicates whether reading or writing capabilities with the {ctag} set to the virtual page is permitted.
```

qwattash:
```suggestion
When the {pte_cap_rw_name} bit is set, capabilities are written as usual.
```

qwattash: Reword mention of CRW here.

src/cheri/attributes.adoc
```diff
@@ -79,8 +80,44 @@ endif::support_varxlen[]
...
+:revocation_enable_lc:     {pte_cap_crg_pte_name}e
+
+:revocation_x:             x{pte_cap_crg_pte_name}
+:revocation_x_lc:          x{pte_cap_crg_pte_name}
+
+:revocation_u:             U{pte_cap_crg_pte_name}
+:revocation_u_lc:          u{pte_cap_crg_pte_name}
+
+:revocation_s:             S{pte_cap_crg_pte_name}
+:revocation_s_lc:          s{pte_cap_crg_pte_name}
```
```suggestion
```
I think these prefixes are stable so we probably don't need variables for them?

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: We have a comment just below the table saying exactly that, so I think it's ok

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: I've tried to improve the text whilst keeping it accurate - please have another look.

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: I think this is accurate because the first sentence above says this: 

>When pte.yw is clear, capability stores or AMOs where the to-be-stored capability tag is set will raise a CHERI Store/AMO Page Fault fault.

and only talks about dirty tracking with pte.yw=1

so I believe it covers the table correctly?

src/cheri/cheri-pte-ext.adoc
qwattash: Yeah that's ok, I think.

src/cheri/cheri-pte-ext.adoc
qwattash: Yeah I think it is much clearer now. Thanks

src/cheri/cheri-pte-ext.adoc
qwattash: Ok, yes, that makes sense upon further inspection.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,178 @@
...
-
-ifdef::cheri_v9_annotations[]
-NOTE: *CHERI v9 Note:* The current proposal is provisional and only includes
-support for load-side revocation when compared to CHERI v9.
+This extension specifies use of the {rv64y_pte4_field_name_lc} field from the PTE formats
+ifndef::cheri_standalone_spec[]
+(e.g., see <<sv39pte>>)
 endif::[]
+.
+{cheri_priv_crg_ext} implies Sv39.
```
qwattash, line 6: Is this to signify "at least" Sv39? As in, not Sv32, but Sv39, Sv48 and Sv57 are ok?

tariqkurd-repo: yes - both Sv48 and Sv57 require Sv39 - it's enough to say "implies Sv39" to mean that there is VM support for RV64.
I'll remove the ifndef/endif as it doesn't help - the xref still breaks in the standalone spec

tariqkurd-repo: done

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,178 @@
...
 . The PMA is set to (_CHERI {ctag_title}_).
 
 ==== {cheri_excep_name_pte_ld}s
 
+{cheri_priv_crg_ext} defines a new fault type:
 
-When _pte.crw_ is set, _pte.crg_ can be used to trap on capability loads or AMOs when it does not match the value of `sstatus.UCRG`.
+* {cheri_excep_name_pte_ld} (cause value {cheri_excep_cause_pte_ld})
+
+{cheri_excep_name_pte_ld} is prioritized below {cheri_excep_name_pte_st} as shown in <<exception-priority-cheri>>, as AMOs can raise both.
```
qwattash: Not sure how to read this with respect to the priorities table. I believe we say that the CHERI Load Capability Fault has lower priority than the store fault, but here it is reversed.
I think we are saying that it is legal to prioritise the load fault for AMOs specifically?

tariqkurd-repo: I've reworded it so it reads better

src/cheri/attributes.adoc
tariqkurd-repo: yes ok - eventually we'll substitute them all - after freeze!

src/cheri/cheri-pte-ext.adoc
qwattash: Re-reading once more. I think this

> When pte.yw is clear, capability stores or AMOs where the to-be-stored capability tag is set will raise a CHERI Store/AMO Page Fault fault.

applies when the ctag=1, but the table row says that we are trapping both when

- YW=0 YD=1 ctag=1
- YW=0 YD=1 ctag=0 -- in this case it can be surprising, because you are not storing a tagged value.

This suggest an alternate wording? Such as

> When pte.yw is clear, capability stores or AMOs where the to-be-stored capability tag ~is set~ may be set will raise a CHERI Store/AMO Page Fault fault.

I realise though that this is clarified below, so maybe it is fine.

src/cheri/attributes.adoc
```diff
@@ -79,8 +80,35 @@ endif::support_varxlen[]
...
+:pte_cap_r_name:           pass:quotes[YR]
+:pte_cap_r_name_lc:        pass:quotes[yr]
+
+:pte_cap_w_name:           pass:quotes[YW]
+:pte_cap_w_name_lc:        pass:quotes[yw]
+
+:pte_cap_dirty_name:       pass:quotes[YD]
+:pte_cap_dirty_name_lc:    pass:quotes[yd]
+
+:revocation_name:          pass:quotes[RG]
```
nwf: This is definitely nit-picking, but: can we call that something else, too?

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,175 @@
...
-are added to leaf PTEs in Sv39, Sv48 and Sv57 as shown in
-xref:sv39pte_crg[xrefstyle=short], xref:sv48pte_crg[xrefstyle=short] and
-xref:sv57pte_crg[xrefstyle=short] respectively. For non-leaf PTEs these bits
-remain reserved and must be cleared by software for forward compatibility,
-or else a page-fault exception is raised. Additionally, if the hypervisor
-extension is enabled these bits remain reserved for leaf and non-leaf PTEs
-used in guest address translation.
+The extension adds the ability for supervisor-mode software to quickly front-run loads of capabilities from userspace pages, and to track stores of capabilities to pages.
+Such facilities have been shown to provide useful primitives for system-level implementation of certain forms of capability revocation which, in turn, allow userspace heap allocator software to deterministically guard against use-after-reallocation cite:[cornucopia-reloaded].
+Additionally, the extension allows tracking of kernel space pages for guarding against use-after-reallocation within the OS kernel.
```
nwf: Since sstatus.SYRG is defined, it's not just userspace pages any more?

```suggestion
The extension adds the ability for supervisor-mode software to quickly front-run loads of capabilities from pages and to track stores of capabilities to pages.
Applied to userspace pages, such facilities have been shown to be useful primitives for system-level implementation of certain forms of capability revocation which, in turn, allow userspace heap allocator software to deterministically guard against use-after-reallocation cite:[cornucopia-reloaded].
Applied to kernel pages, they are analogously useful within an OS kernel.
```

src/cheri/introduction.adoc
```diff
@@ -106,12 +106,10 @@ ifdef::support_varxlen[]
 endif::support_varxlen[]
 |<<section_priv_cheri_vmem,{cheri_priv_vmem_ext}>>          | Virtual Memory
 |<<section_debug_integration_trig,{cheri_priv_debug_trig}>> | Debug triggers
-|<<section_cheri_priv_crg_ext,    {cheri_priv_crg_ext}>>^1^ | MMU-based acceleration of capability revocation for heap temporal safety
+|<<section_cheri_priv_crg_ext,    {cheri_priv_crg_ext}>>    | MMU-based acceleration of capability revocation for heap temporal safety
```
nwf:
```suggestion
|<<section_cheri_priv_crg_ext,    {cheri_priv_crg_ext}>>    | MMU-based capability data flow control and measurement
```

src/cheri/attributes.adoc
tariqkurd-repo: if we're going to rename it now is the time - but what should it be?

tariqkurd-repo: FC for Flow Control? I'm now using the `revocation_name` variable in the extension name so we can change it in one place for the CSR fields and extension name.

nwf: Sorry, I meant renaming the `revocation_name` variable itself, since, again, it's just primitives that software can use &c &c.  Perhaps `pte_cap_flow_control` or somesuch.

tariqkurd-repo: ah ok - yes I can change that - once the spec is frozen I plan to stop using the variables anyway - they're only really there becuase we keep changing our minds about things - and ARC keep asking for renaming.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,175 @@
...
-capabilities (e.g., private anonymous mappings) and leave
-`sstatus.UCRG` on all harts and _pte.crg_ in all PTEs set to 0, which will allow
-capabilities to be loaded and stored successfully.
+* For future AMOCAS{LD_ST_DOT_CAP}, it is not known whether the store will happen until the load has executed and the compare has been done. Therefore, AMOCAS{LD_ST_DOT_CAP} is likely to trigger capability dirty tracking if the {ctag} is set in {cs2}.
+* For future AMOADD{LD_ST_DOT_CAP}, the stored {ctag} depends upon the loaded {ctag} which is not known during translation and also on the execution of the {CADD}.
+ Therefore, AMOADD{LD_ST_DOT_CAP}, is likely to always trigger capability dirty tracking.
 
-[#xstatus_pte]
-=== UCRG, CRGE CSR fields
+Checking the stored {ctag} is less of a burden to the implementation than checking the loaded {ctag} for {cheri_excep_name_pte_ld}, which is why checking the loaded {ctag} is optional behavior.
```
nwf, line 149: This should probably be informative, not normative?

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,175 @@
...
-The {cheri_priv_crg_ext} extension is enabled when the `sstatus.CRGE` bit is set.
+{cheri_priv_crg_ext} implies Sv39.
 
-When enabled, the extension adds the ability for supervisor-mode software to quickly front-run loads of capabilities from userspace pages, and to track stores of capabilities to pages.
-Such facilities have been shown to provide useful primitives for system-level implementation of certain forms of capability revocation which, in turn, allow userspace heap allocator software to deterministically guard against use-after-reallocation cite:[cornucopia-reloaded].
+The {cheri_priv_crg_ext} extension is enabled when the `sstatus.{revocation_enable}` bit is set.
 
-It achieves this by adding by adding the _pte.crg_, _pte.cd_ and `Xstatus.UCRG` bits as described below.
-When the extension is disabled the _pte.cd_, _pte.crg_ and `Xstatus.UCRG` bits remain reserved.
+NOTE: The enable bit is required to redefine the usage of the _pte_._{rv64y_pte4_field_name_lc}_ field.
```
arichardson: I think we can probably drop this note. IMO it just makes the text more confusing.

src/cheri/riscv-priv-integration.adoc
```diff
@@ -560,19 +548,15 @@ endif::support_varxlen[]
...
 
-NOTE: _Sv32_ does not have any spare PTE bits, and so this bit does not exist for RV32.
+Implicit memory accesses made by the page table walker are not subject to CHERI checks.
 
-IMPORTANT: Any {cheri_base64_ext_name} hart that supports Sv39 must also implement the _pte.crw_ bit.
+NOTE: A future extension may add CHERI checks to the page table walker.
 
 [#limit_cap_prop]
-==== Limiting Capability Propagation
+=== Limiting Capability Propagation
```
nwf, line 559: The entirety of this section should be informative, not normative?

tariqkurd-repo: done

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: it's already in a big NOTE block

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: I think the note is important, if the extension was just filling in reserved states then it wouldn't need an enable but I think it's good to make the reason for the enable clear. In particular rvy[3] has a different meaning compared to not having this extension.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,176 @@
...
-When enabled, the extension adds the ability for supervisor-mode software to quickly front-run loads of capabilities from userspace pages, and to track stores of capabilities to pages.
-Such facilities have been shown to provide useful primitives for system-level implementation of certain forms of capability revocation which, in turn, allow userspace heap allocator software to deterministically guard against use-after-reallocation cite:[cornucopia-reloaded].
+The {cheri_priv_crg_ext} extension is enabled when the `sstatus.{pte_cap_fc_enable}` bit is set.
 
-It achieves this by adding by adding the _pte.crg_, _pte.cd_ and `Xstatus.UCRG` bits as described below.
-When the extension is disabled the _pte.cd_, _pte.crg_ and `Xstatus.UCRG` bits remain reserved.
+NOTE: The enable bit is required to redefine the usage of the _pte_._{rv64y_pte4_field_name_lc}_ field.
 
-NOTE: {cheri_priv_crg_ext} depends on Sv39 and {cheri_base64_ext_name} also being supported.
+The extension adds the ability for supervisor-mode software to quickly front-run loads of capabilities from pages and to track stores of capabilities to pages.
```
jrtc27: Unless you know how CHERI's load-side revocation works I'm not sure it's obvious what "quickly front-run loads of capabilities" means. Perhaps:

The extension adds the ability for supervisor-mode software to quickly enable trapping loads of capabilities from all pages of memory, incrementally allow loads of capabilities from such pages, and track stores of capabilities to all pages of memory.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,176 @@
...
+Applied to userspace pages, such facilities have been shown to be useful primitives for system-level implementation of certain forms of capability revocation which, in turn, allow userspace heap allocator software to deterministically guard against use-after-reallocation cite:[cornucopia-reloaded].
```
jrtc27:
```suggestion
Applied to userspace pages, this has been shown to allow the operating system to implement capability revocation schemes, which allow userspace memory allocators to deterministically guard against use-after-reallocation cite:[cornucopia-reloaded].
```
Simplified language (and removed notion of heap, which is a terrible term from the days of sbrk(2))

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,176 @@
...
+Applied to kernel pages, they are analogously useful within an OS kernel.
```
jrtc27:
```suggestion
Applied to kernel pages, they can similarly be used against use-after-reallocation within the kernel itself.
```
since otherwise the sentence seems a bit close to "used for the kernel you can instead use it for the kernel"

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,176 @@
...
-.Sv48 page table entry
-[#sv48pte_crg]
-include::img/sv48pte_crg.edn[]
+[[svycrg_rvy_field_breakdown]]
+.Breakdown of the 4-bit _pte_._{rv64y_pte4_field_name_lc}_ field.
+[%autowidth,float="center",align="center",cols="<,<,<",options="header"]
+|===
+| Bit                                  | Name                                | Comment
+3+| *Capability load/AMO fields*
+|_pte_._{rv64y_pte4_field_name_lc}[0]_ | _pte_._{pte_cap_r_name_lc}_         | Capability read enable
```
jrtc27: Maybe "capability-readable"? To match "The permission bits, R, W, and X, indicate whether the page is readable, writable, and executable, respectively" from the base SvNN. I definitely don't like having "enable" in here, makes it sound like a CSR enable bit.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,176 @@
...
+|_pte_._{rv64y_pte4_field_name_lc}[1]_ | _pte_._{pte_cap_crg_pte_name_lc}_   | Capability read generation
+3+| *Capability store/AMO fields*
+|_pte_._{rv64y_pte4_field_name_lc}[2]_ | _pte_._{pte_cap_w_name_lc}_         | Capability write enable
+|_pte_._{rv64y_pte4_field_name_lc}[3]_ | _pte_._{pte_cap_dirty_name_lc}_     | Capability dirty bit
```
jrtc27: I wouldn't use "bit" in the description; the bit itself can be described as "capability dirty", with "bit" only added in the context of "the capability dirty bit", like how you say "the AUIPC instruction" but don't describe it as "AUIPC instruction".

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,176 @@
...
+|===
 
-.Sv57 page table entry
-[#sv57pte_crg]
-include::img/sv57pte_crg.edn[]
+NOTE: {cheri_priv_crg_ext} is not a mandatory part of the {cheri_base64_ext_name} Privileged ISA, as a future extension may specify an alternative usage for the _pte_._{rv64y_pte4_field_name_lc}_ field allowing a different mechanism giving improved software efficiency.
```
jrtc27: Mandatory is an odd term to use here, that's more for profiles. What you really mean is that it's a separate extension rather than part of the RV64Y Privileged ISA.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,176 @@
...
 
 === Software Primitives
```
jrtc27: This is an ISA spec so we should be talking about architectural features not software primitives. I would drop this heading (it's the only one at this level within the chapter) level, and make the following paragraph something like:

When all of the following are true, Svyrg provides two related architectural features, Load Capability Faults and Capability Dirty Tracking (and either put CHERI on both or neither, don't do just the first like is currently the case).

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,176 @@
...
-. _pte.u_ is set.
+. _pte_._{pte_cap_r_name_lc}_ is set.
+. if _pte_._u=1_, _pte_._{pte_cap_crg_pte_name_lc}_ does not equal `sstatus.U{pte_cap_crg_pte_name}`.
+. if _pte_._u=0_, _pte_._{pte_cap_crg_pte_name_lc}_ does not equal `sstatus.S{pte_cap_crg_pte_name}`.
+. The loaded {ctag} is set^1^.
+.. The {cheri_excep_name_pte_ld} _must_ be taken if the loaded {ctag} is set.
+.. The {cheri_excep_name_pte_ld} _may_ be taken if the loaded {ctag} is not set.
 . Any other platform specific rules have not forced the loaded {ctag} to be clear.
 
 NOTE: An example of a platform specific rule is a hardware engine clearing {ctag}s in memory after a call to free() to offload the CPU, which may have cleared a {ctag} before the CPU loads it.
```
jrtc27: Why is this here? Firstly, free() is way too many levels up the stack in software land not architecture land. Secondly, this isn't forcing the loaded tag to be clear, it's forcing the in-memory tag to be clear, no? It's not interesting to say "if someone's stored to memory to make the tag 0 then the tag is 0".

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,176 @@
...
 . A capability load or AMO is executed.
-.. Where an AMO can raise both faults, {cheri_excep_name_pte_ld} is prioritized above {cheri_excep_name_pte_st}.
-. _pte.crg_ does not equal `sstatus.UCRG`.
-. _pte.u_ is set.
+. _pte_._{pte_cap_r_name_lc}_ is set.
+. if _pte_._u=1_, _pte_._{pte_cap_crg_pte_name_lc}_ does not equal `sstatus.U{pte_cap_crg_pte_name}`.
+. if _pte_._u=0_, _pte_._{pte_cap_crg_pte_name_lc}_ does not equal `sstatus.S{pte_cap_crg_pte_name}`.
+. The loaded {ctag} is set^1^.
+.. The {cheri_excep_name_pte_ld} _must_ be taken if the loaded {ctag} is set.
+.. The {cheri_excep_name_pte_ld} _may_ be taken if the loaded {ctag} is not set.
```
jrtc27: I don't think a sub-point of the top-level point should add cases, it should instead qualify / elaborate on what the top-level point means.

jrtc27: That is, the top-level point itself should capture the optionality of this check

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,176 @@
...
 . Any other platform specific rules have not forced the loaded {ctag} to be clear.
 
 NOTE: An example of a platform specific rule is a hardware engine clearing {ctag}s in memory after a call to free() to offload the CPU, which may have cleared a {ctag} before the CPU loads it.
 
-There is an additional rule if _{cheri_priv_crg_load_tag_ext}_ is implemented:
+^1^ Allowing flexibility about checking the loaded {ctag} gives a range of valid implementations.
```
jrtc27: Um, is that not true of every single option in the entirety of RISC-V? Otherwise there'd be no point adding the options. This seems odd to call out for this and only this case.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,176 @@
...
+| 1                         | {ne} `sstatus.x{pte_cap_crg_pte_name}`^2^ |   0   | Implementation defined choice of +
+{cheri_excep_name_pte_ld} or normal operation
+| 1                         | {ne} `sstatus.x{pte_cap_crg_pte_name}`^2^ |   1   | {cheri_excep_name_pte_ld}
+| 1                         | = `sstatus.x{pte_cap_crg_pte_name}`^2^    |   X   | Normal operation
 |===
 
-^1^ A future version of this specification may check an SCRG bit in `sstatus` in this case for trapping on kernel pages.
+^1^ The loaded {ctag}.
+
+^2^ if _pte_._u=1_, `sstatus.U{pte_cap_crg_pte_name}` else `sstatus.S{pte_cap_crg_pte_name}`.
```
jrtc27: I think you either need a comma before else or no commas at all?

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,176 @@
...
 
 NOTE: {cheri_excep_name_pte_ld}s may be used to implement the load-barrier primitive from cite:[cornucopia-reloaded].
```
jrtc27, line 88: Is it actually the primitive we care about talking about here or the revocation scheme?

src/cheri/cheri-pte-ext.adoc
jrtc27: I don't understand why this is the way it is. If the tag is 0 and yw is 0, why does it matter what yd is?

jrtc27: That is, why is the table not:

yw | yd | tag | trap?
---|----|-----|------
0 | x | 0 | no
0 | x | 1 | yes

Currently it's (rewriting to match the above more):

yw | yd | tag | trap?
---|----|-----|------
0 | 0 | 0 | no
0 | 1 | 0 | yes
0 | x | 1 | yes

That middle case does not make sense to me.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,176 @@
...
-In particular, an oblivious software implementation that only sets the _pte.crw_ bit would observe one of the following:
+NOTE: Capability dirty tracking may be used to implement the store-barrier primitive from cite:[cornucopia-reloaded].
 
-* The _pte.cd_ bit is set by the hardware when a capability is stored and _Svadu_ is enabled.
-* A {cheri_excep_name_pte_st} trap when _Svade_ is enabled.
+==== Minimum support
 
-The CRGE bit also has the benefit of maintaining matching polarity of the _pte.cd_ bit with respect to the data Dirty bit (_pte.d_).
-If the behavior of the _pte.crw=1_ _pte.cd=0_ state were defined to match the behavior when {cheri_priv_crg_ext} is not implemented, then _pte.cd_ would need to have 0 meaning dirty, which is inconsistent.
+The minimum level of PTE support is to set _pte_._{pte_cap_r_name_lc}_=1, _pte_._{pte_cap_dirty_name_lc}_=1, _pte_._{pte_cap_w_name_lc}_=1 and _pte_._{pte_cap_crg_pte_name_lc}_=0 in all PTEs intended for storing capabilities (e.g., private anonymous mappings) and set `sstatus.U{pte_cap_crg_pte_name}`=0 and `sstatus.S{pte_cap_crg_pte_name}`=0 on all harts, which will enable capabilities to be loaded and stored successfully.
```
jrtc27: This whole section is very confusing. It's talking about minimum support, in an ISA spec, which sounds a lot like what a hardware implementer needs to do, but it's not. It's about what _software_ minimally has to do if it enables the extension and wants to still be able to use capabilities. Which is all non-normative text about how you could use the thing, not how to implement it / what it does.

src/cheri/riscv-priv-integration.adoc
```diff
@@ -585,61 +572,47 @@ Such sharing through virtual memory is on the page granularity, so preventing ca
...
-NOTE: {cheri_excep_name_pte_ld} faults are at present raised only if <<section_cheri_priv_crg_ext,{cheri_priv_crg_ext}>> is enabled.
+It is prioritized against other fault types as shown in <<exception-priority-cheri>>.
 
-NOTE: {cheri_excep_name_pte_st} faults are raised under more circumstances if <<section_cheri_priv_crg_ext,{cheri_priv_crg_ext}>> and Svade are both enabled.
+=== The _pte_._{cheri_base_ext_name_lc}_ field for capability flow control
 
-==== Extending the Page Table Entry Format
+As {cheri_base32_ext_name} has no field allocated in the PTE, capability loads and stores operate as normal.
+Therefore the {cheri_base_ext_name} field is reserved in the RV64 PTE formats only (e.g. <<sv39pte>>) for use by {cheri_base64_ext_name} extensions.
+The principle is that extensions, such as <<section_cheri_priv_crg_ext,{cheri_priv_crg_ext}>> may redefine the usage of the {cheri_base_ext_name} field by adding an enable bit into `Xstatus`.
```
jrtc27, line 592: This reads more like a note to future spec authors than hardware implementors?

src/cheri/cheri-pte-ext.adoc
nwf: Sure.  Mostly I've been using "heap" as a short-hand for "not on-stack"; I agree the terminology is terrible.  I'd prefer "certain capability revocation schemes" or similar, since, as stated up thread, the global revocation implemented to date is really not what capability-systems people mean by "revocation".

src/cheri/cheri-pte-ext.adoc
nwf: I suspect this was an attempt to capture CHERIoT's capability load filter -- additional architecture that intermediates the core's view of capability tags in memory -- and accidentally got cross-threaded with the revocation state machine.  We use the term "filter" to reinforce the non-self-healing nature of the load barrier-in-the-GC-sense: it does not induce a RMW.  (Just to say: the CHERIoT revocation state machine, OTOH, does do RMWs against memory, but is not involved in the load capability path.)

In any case, I agree it's a bit of a mishmash of layers and should probably just go away.

That said, a better NOTE to have here might be to point out the tag-clearing behavior of a capability load instruction citing a `R !C` authorizing cap: because the tag in the register file _will_ be clear, regardless of the tag in memory, the exception need not take place.

src/cheri/cheri-pte-ext.adoc
nwf: I'd rather just talk about the primitives... the revocation scheme as a whole is a whole lot of software and orchestration of architecture, while this extension really is just providing the primitives that that software uses and that could be used for other purposes.

src/cheri/riscv-priv-integration.adoc
tariqkurd-repo: turned into a note

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: turned into a note

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: the primitives, of course, but the links to the revocation are useful for context

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: ok I've simplified it

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: I've simplified it

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: I've simplified it

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: ok

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: done

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: done

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: done

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: done

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: fixed

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,165 @@
...
+|_pte_._{rv64y_pte4_field_name_lc}[1]_ | _pte_._{pte_cap_crg_pte_name_lc}_   | Capability read generation
+3+| *Capability store/AMO fields*
+|_pte_._{rv64y_pte4_field_name_lc}[2]_ | _pte_._{pte_cap_w_name_lc}_         | Capability writable
+|_pte_._{rv64y_pte4_field_name_lc}[3]_ | _pte_._{pte_cap_dirty_name_lc}_     | Capability dirty
+|===
 
-.Sv57 page table entry
-[#sv57pte_crg]
-include::img/sv57pte_crg.edn[]
+NOTE: {cheri_priv_crg_ext} is a separate extension and not part of the {cheri_base64_ext_name} Privileged ISA, as a future extension may specify an alternative usage for the _pte_._{rv64y_pte4_field_name_lc}_ field allowing a different mechanism giving improved software efficiency.
```
```suggestion
NOTE: {cheri_priv_crg_ext} is a separate extension and not part of the {cheri_base64_ext_name} Privileged ISA, to allow future extensions that are incompatible with {cheri_priv_crg_ext} (but compatible with {cheri_base64_ext_name}) to re-purpose these bits.
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,165 @@
...
- The analogous page state is not permitted for capability data as _pte.crw=0, pte.cd=1_ is _reserved_.
+^1^ The to-be-stored {ctag}.
 
-Capability dirty tracking _always_ checks the {ctag} on stored capabilities when determining whether to raise the {cheri_excep_name_pte_st}.
+The capability dirty tracking is resolved during memory translation, but there are cases it is not known if there will be a {ctag} stored to memory or not at this point.
+Capability dirty tracking _must_ be triggered when _pte_._{pte_cap_w_name_lc}_=1 and _pte_._{pte_cap_dirty_name_lc}_=0 and the to-be-stored {ctag} is set, and _may_ be triggered in the following case where it is not known if there will be a stored {ctag} during translation:
 
-NOTE: Checking the stored {ctag} is less of a burden to the implementation than checking the loaded {ctag} for {cheri_priv_crg_load_tag_ext}, which is why checking the loaded {ctag} is optional behavior.
- However, a future extension may reduce the burden further by removing the check on the stored {ctag}.
+* SC{LD_ST_DOT_CAP} triggers capability dirty tracking if the {ctag} is set in {cs2}, even if the store fails.
```
arichardson, line 131: Is this the same behaviour as sc has with the normal dirty bit? If so, how about:
```suggestion
* SC{LD_ST_DOT_CAP} triggers capability dirty tracking if the {ctag} is set in {cs2}, even if the store fails.
  This matches the semantics of SC.* with regard to  _pte_._d_.
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,165 @@
...
 
-NOTE: Capability dirty tracking may be used to implement the store-barrier primitive from cite:[cornucopia-reloaded].
+[NOTE]
+====
 
-==== Minimum support
+_pte_._{pte_cap_dirty_name_lc}_ must always be set when there is a stored {ctag} to a virtual page so that software knows which pages have had capabilities stored to them.
+It may be set too often, which may cause software to examine the page to check for capabilities when none are present.
+This is a situation software is required to handle anyway, as it is always possible for all capabilities in a page to be overwritten by non-capability data.
+In this case the _pte_._{pte_cap_dirty_name_lc}_ bit will still be set.
```
```suggestion
In this case the _pte_._{pte_cap_dirty_name_lc}_ bit would still be set.
```

src/cheri/cheri-pte-ext.adoc
```diff
@@ -1,188 +1,165 @@
...
 
-The minimum level of PTE support is to set _pte.crw_ and _pte.cd_ to 1 in all PTEs intended for storing
-capabilities (e.g., private anonymous mappings) and leave
-`sstatus.UCRG` on all harts and _pte.crg_ in all PTEs set to 0, which will allow
-capabilities to be loaded and stored successfully.
+Future AMOs fall into this category:
 
-[#xstatus_pte]
-=== UCRG, CRGE CSR fields
+* For future AMOCAS{LD_ST_DOT_CAP}, it is not known whether the store will happen until the load has executed and the compare has been done. Therefore, AMOCAS{LD_ST_DOT_CAP} is likely to trigger capability dirty tracking if the {ctag} is set in {cs2}.
```
```suggestion
* For future AMOCAS{LD_ST_DOT_CAP}, it is not known whether the store will happen until the load has executed and the compare has been done.
  Therefore, AMOCAS{LD_ST_DOT_CAP} is likely to trigger capability dirty tracking if the {ctag} is set in {cs2}.
```

src/cheri/riscv-priv-integration.adoc
```diff
@@ -444,20 +442,10 @@ There are three levels of support:
 |=========================================================================================
 
 ^1^ The access fault is triggered on all capability stores or atomics such as <<STORE_CAP>> or <<AMOSWAP_CAP>> when <<c_perm>> and <<w_perm>> are granted and the {ctag} is set to one.
+ For SC{LD_ST_DOT_CAP}, the access fault is raised if the {ctag} is set in {cs2} and the store fails.
```
```suggestion
 For SC{LD_ST_DOT_CAP}, the access fault is raised if the {ctag} is set in {cs2} even when the store fails.
```
This seems easier to read to me

src/cheri/cheri-pte-ext.adoc
tariqkurd-repo: I've done this as a separate commit and combined with the comment above talking about the need for the enable bit to redefine the field.

PR #936, opened 2026-02-05
Temporary: change SRLIY encoding to avoid clash with old SCBNDSI

tariqkurd-repo:
This one is a bit strange - the 0.9.5 ISA and the 0.9.6 ISA have no conflicting encodings except for this one.
With this one change then a single simulator can support both version of the ISA.
The encodings will be replaced once ARC assigns correct encodings anyway - this one bit change helps with the migration from 0.9.5 to 0.9.6 - before everything standardises on the final approved version.

image


arichardson:
With a bit of extra code this can be handled in the simulator (e.g. an extra predicate in the decoder file for QEMU) and the values will change again anyway. I think this just makes the churn worse and we should wait for final encodings

PR #942, opened 2026-02-11
ypermc: clarify that the fixed point is not just a set of permissions

### LINKED ISSUE #941: Clarify ypermc's behavior regarding the M-bit
There is currently ambiguity in the specification regarding whether ypermc must clear metadata (specifically the M-bit) when the removal of a permission (e.g., Execute) renders that metadata invalid.

Consider the following thought experiment
1. we have read-execute capability with M=1
2. we use ypermc to zap X on that cap
3. we use our knowledge of the RVY64 encoding format to set X again (cap becomes invalid)
4. we use ybld with a sufficient authorizing cap and the cap from 3. as template cap

Has the resulting build capability set M=1?

The issue depends on whether the specification allows for "dormant" M-bits—where a valid capability lacks X permission but retains M=1 in its bit representation (unobservable via ymoder but present in the raw bits)—or if it strictly requires M to be reset to 0 when X is removed. The consensus appears to be the latter: `ypermc` must result in a "clean" state where M=0 if the resulting permissions do not support execution.

Unfortunately, ypermc's description only talks about heading to a fixed point regarding cap permissions, not metadata (like the M-bit) in general. This leaves the desired behavior vague.

I believe that this can be fixed by refining the wording slightly.

Flowdalic:
The existing wording gives the impression that the fixed point only deals with capability permissions. However, if Zyhybrid is available and ypermc is used to clear the X permission on a capablity with M=1, then ypermc also legalizes the result by setting M=0.

Clarify the wording that ypermc may not only adjust the resulting permissions, but also capability metadata, like the M-bit of Zyhybrid.

Fixes #941.

### DIFF
--- a/src/cheri/contributors.adoc
+++ b/src/cheri/contributors.adoc
@@ -42,3 +42,4 @@ This RISC-V specification has been contributed to directly or indirectly by:
 * Jonathan Woodruff <jonathan.woodruff@cl.cam.ac.uk>
 * Jason Zhijingcheng Yu <yu.zhi@comp.nus.edu.sg>
 * Robert Riglar <robert.riglar@codasip.com>
+* Florian Schmaus <florian.schmaus@codasip.com>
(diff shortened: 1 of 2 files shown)


src/cheri/insns/acperm_32bit.adoc
```diff
@@ -31,10 +31,11 @@ format shown in xref:acperm_bit_field[xrefstyle=short].
...
 NOTE: Future extensions may include hardwired permission bits, in which case they are not cleared by set bits in `rs2`.
 . Encode the resulting architectural permissions as specified by the encoding in use.
-  This may involve iterating over the rules in <<sec_permission_transitions>>,
+  This may involve iteratively applying the rules in <<sec_permission_transitions>>,
   as well as any rules added by extensions or the capability encoding,
-  to a fixed point (that is, a set of permissions not further reduced by any rule);
-  encodings may, however, specify other encoding procedures.
+  until a fixed point is reached -- specifically, capability metadata (such as
+  a set of permissions or the <<M-bit>>) that cannot be further reduced by any rule.
+  Note that encodings may specify other encoding procedures.
```
arichardson:
```suggestion
  Note that the precise procedure depends on the capability encoding.
```

Maybe like this to avoid encoding twice?

tariqkurd-repo: I don't have a chance to read this fully right now but the YPERMC page can't talk about the M-bit - that has to go in the hybrid chapter

Flowdalic: > YPERMC page can't talk about the M-bit - that has to go in the hybrid chapter

How unfortunate. Is this a technical limitation, i.e., we want to later split this in separate documents and linking between them is hard? Or is it an editorial requirement that in RISC-V specifications, a base specification can not refer to optional extensions?

tariqkurd-repo: >Or is it an editorial requirement that in RISC-V specifications, a base specification can not refer to optional extensions?

It's an editorial requirement - we must define the base without _any_ optional extensions.
This makes writing the spec very difficult.

Can you change this to adding text into the hybrid chapter instead?

nwf:
```suggestion
  until a fixed point is reached, arriving at a set of permissions not further reduced by any rule.
  Any permission-dependent capability metadata is to be updated to reflect any permissions removed by this procedure.
  Note that encodings may specify procedures for computing the new permission set; while these must be idempotent and respect the rules of <<sec_permission_transitions>>, they need not be of the form "iterate to fixed point".
```

src/cheri/insns/acperm_32bit.adoc
```diff
@@ -31,10 +31,11 @@ format shown in xref:acperm_bit_field[xrefstyle=short].
 +
 NOTE: Future extensions may include hardwired permission bits, in which case they are not cleared by set bits in `rs2`.
 . Encode the resulting architectural permissions as specified by the encoding in use.
-  This may involve iterating over the rules in <<sec_permission_transitions>>,
+  This may involve iteratively applying the rules in <<sec_permission_transitions>>,
   as well as any rules added by extensions or the capability encoding,
-  to a fixed point (that is, a set of permissions not further reduced by any rule);
-  encodings may, however, specify other encoding procedures.
+  until a fixed point is reached, arriving at a set of permissions not further reduced by any rule.
+  Any permission-dependent capability metadata is to be updated to reflect any permissions removed by this procedure.
```
arichardson:
```suggestion
  Any permission-dependent capability metadata is updated to reflect any permissions removed by this procedure.
```
This sentence is a bit confusing to me, I think we can just drop `to be`?

PR #943, opened 2026-02-11
integrity checks: clarify that this is about valid capabilities

Flowdalic:
Eventually, integrity checks apply to valid and invalid capabilities. While they are more dominant for valid capabilities, they have their use for invalid ones as well:
1. ybld uses integrity checks on invalid capabilities (as noted in the text below).
2. ymoder and ymodew use integrity checks on invalid capabilities.

So make it clear that only a *valid* capability that fails integrity checks is something abnormal.

Also refine the wording a bit.

### DIFF
--- a/src/cheri/riscv-unpriv-integration.adoc
+++ b/src/cheri/riscv-unpriv-integration.adoc
@@ -586,8 +586,8 @@ CHERI enforces the following rules for all valid capabilities:
 
 Since a capability with the {ctag} set that fails integrity checks could not have been legally created, such a failure indicates either
 
-* state corruption due to memory or logic faults, or
-* the presence of incompatible or faulty CHERI IP within the system.
+* State corruption due to memory or logic faults, or
+* The presence of incompatible or faulty CHERI IP within the system.
 
 NOTE: These checks are much less rigorous than parity or ECC protection, and are only
 used to detect simple problems with the capability metadata.


src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -584,10 +584,10 @@ CHERI enforces the following rules for all valid capabilities:
 . The capability metadata does not have any bits, or fields set to a reserved value.
 .. An example of a reserved field value is permissions which cannot be legally produced by <<CLRPERM>>.
 
-In all cases the capability could not have been legally created, and so either:
+Since a valid capability that fails integrity checks could not have been legally created, such a failure indicates either
```
```suggestion
Since a capability with the {ctag} set that fails integrity checks could not have been legally created, such a failure indicates either
```
maybe this?

Flowdalic: wfm, adjusted.

tariqkurd-repo:
Morello use "valid capability" in the way @Flowdalic suggests, but we haven't been have always say "with the capability tag set"

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -584,10 +584,10 @@ CHERI enforces the following rules for all valid capabilities:
 . The capability metadata does not have any bits, or fields set to a reserved value.
 .. An example of a reserved field value is permissions which cannot be legally produced by <<CLRPERM>>.
 
-In all cases the capability could not have been legally created, and so either:
+Since a capability with the {ctag} set that fails integrity checks could not have been legally created, such a failure indicates either
 
-* There has been a corruption of capability state due to memory or logic faults
-* There is an incompatible or faulty CHERI IP within the system
+* state corruption due to memory or logic faults, or
+* the presence of incompatible or faulty CHERI IP within the system.
```
```suggestion
* State corruption due to memory or logic faults, or
* The presence of incompatible or faulty CHERI IP within the system.
```

PR #994, opened 2026-04-17
simplify JALR/Sentry text for RV64

### LINKED ISSUE #991: Does JAL have a dependency on Zysentry?
The description of Zysentry talks about JALR and the sealing, but doesn't talk about JAL sealing the return address, maybe it should?

tariqkurd-repo:
Radically simplify the RV64Y text to get it through ARC review for sealing/unsealing/JALR and YSENTRY.
Make the JALR text more generic so it can cover CHERIoT and forward/backwards sentries for RV64 in due course, but with specualised text so that the actual behaviour can be understood with/without Zysentry.

Additionally, fix #991


src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -181,8 +181,7 @@ Software can query the bounds of a capability held in a general-purpose register
 * The _base_ is returned by the <<GCBASE>> instruction.
 * The _length_ is returned by the <<GCLEN>> instruction.
 ** <<GCLEN>> saturates the _length_ to XLEN bits
```
tariqkurd-repo: YTOPR also saturates (to be confirmed @PeterRugg ), so this comment can probably go - as it was related to how you calculate the top

src/cheri/insns/gctop_32bit.adoc
```diff
@@ -18,14 +18,14 @@ Encoding::
 include::wavedrom/gctop.adoc[]
 
 Description::
-Decode the base integer address from `{cs1}` 's bounds and write the result to `rd`.
+Decode the top integer address from `{cs1}` 's bounds, saturated to 2^MXLEN^-1, and write the result to `rd`.
```
tariqkurd-repo, line 21: @PeterRugg good to confirm that saturation is correct, as it was previously  unspecified.

nwf:
YTOPR seems fine, but YSENTRY (useful only in cases where there is at least one ambient sealing type) and YBLD (exposing encoding _much_ more directly to software) were extensions for a reason.  Thus far we've managed to do a good job of having a single "abstract base" instantiated into concrete bases by choice of encoding, with all other changes being extensions, and I'd very much like to preserve that structure (since we can't make the encodings themselves extensions or properties of platforms).

tariqkurd-repo:
> Thus far we've managed to do a good job of having a single "abstract base" instantiated into concrete bases by choice of encoding, with all other changes being extensions, and I'd very much like to preserve that structure (since we can't make the encodings themselves extensions or properties of platforms).

I'm worried that it's perceived as a bad job by ARC because of the number of extensions.

src/cheri/insns/cbld_32bit.adoc
```diff
@@ -21,10 +21,13 @@ include::wavedrom/cbld.adoc[]
 
 Description::
 Copy `{cs2}` to `{cd}`.
+ifndef::cheri_ratification_v1_only[]
 +
+//CHERIoT specific
```
arichardson:
```suggestion
// This only applies once we include non-ambient otypes
```

src/cheri/insns/gctop_32bit.adoc
arichardson: Yes that was always the intention.

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -29,10 +29,26 @@ Operationally, copy `{cs1}` to the target <<pcc>> and then...
...
++
+. Compute the target <<pcc>> address in two steps, each using the semantics of the <<SCADDR>> instruction:
+.. Increment the address of the target <<pcc>> by the sign-extended 12-bit `offset`, and
+.. set bit zero of the target <<pcc>> address to zero.
+. Unseal the target <<pcc>> if it is a <<sentry_cap>>.
+. If `{cd}≠{creg}0`, compute the return capability and install it to `{cd}`:
+.. add the width of this instruction to the current <<pcc>> using the semantics of the <<SCADDR>> instruction, and
+.. Seal the return capability as a <<sentry_cap>>.
+. Jump to the target <<pcc>>.
+endif::[]
```
```suggestion
Description::
Indirect jump to the target capability in `{cs1}` with an address offset.
+
ifndef::cheri_ratification_v1_only[]
NOTE: The  description below contains three hooks for extending <<JALR_CHERI>> behavior, used by extensions to give a large degree of extensibility.
 Unless an instruction, such as <<SENTRY>>, is implemented which explicitly reference any of the hooks, then take no action for any of them.
endif::[]
+
Operationally, copy `{cs1}` to the target <<pcc>> and then...
+
. Compute the target <<pcc>> address in two steps, each using the semantics of the <<SCADDR>> instruction:
.. Increment the address of the target <<pcc>> by the sign-extended 12-bit `offset`, and
.. set bit zero of the target <<pcc>> address to zero.
ifndef::cheri_ratification_v1_only[]
. HOOK 1: Optionally clear the {ctag} of the target <<pcc>> depending on the <<sec_cap_type>> value and the numerical values of `{cd}` and `{cs1}`
endif::[]
. Unseal the target <<pcc>> if it is a <<sentry_cap>>.
ifndef::cheri_ratification_v1_only[]
. Clear the {ctag} of the target <<pcc>> if it is any other sealed type.
endif::[]
. If `{cd}≠{creg}0`, compute the return capability and install it to `{cd}`:
.. add the width of this instruction to the current <<pcc>> using the semantics of the <<SCADDR>> instruction, and
ifndef::cheri_ratification_v1_only[]
.. HOOK 2: Optionally seal the return capability as a <<sentry_cap>> with a <<sec_cap_type>> defined by the implemented extensions.
. HOOK 3: Optionally make other architectural state updates.
endif::[]
. Jump to the target <<pcc>>.
+

NOTE: Future {cheri_base_ext_name} extensions are permitted introduce additional restrictions on the behaviour of JALR. One such example is CHERIoT differentiating between a forward and backward edge <<sentry_cap>>.
```

But I think we probably do need to keep these hooks listed since I'm not sure we can redefine behaviour later. The CHERIoT encoding can apparently redefine it but I worry about this preventing us from adding the foward-backward sentries in  later `Zysimplecfi` extension.

Maybe this suggested note is sufficient?

src/cheri/insns/sentry_32bit.adoc
```diff
@@ -20,13 +20,12 @@ Description::
 Copy `{cs1}` to `{cd}`.
 +
 Set the capability type (<<sec_cap_type>>) of `{cd}` to
-the <<sec_cap_type_ambient,ambient>> value specified by the capability encoding
-given the permissions granted by the capability in `{cs1}`.
+the value `1`, representing a sentry type.
```
arichardson:
```suggestion
ifdef::cheri_ratification_v1_only[]
the value `1`, representing a sentry type.
endif::[]
ifndef::cheri_ratification_v1_only[]
the <<sec_cap_type_ambient,ambient>> value specified by the capability encoding given the permissions granted by the capability in `{cs1}`.
endif::[]
```

src/cheri/app-cheri-instructions.adoc
```diff
@@ -116,7 +116,7 @@ include::{generated_dir}/new_instructions_table_body.adoc[]
 . 0.9.5 had SH[123]ADD, and the .UW forms, replaced by capability versions.
 .. This is no longer the case, so now the capability versions have new encodings.
 . There is no longer an SH4ADD instruction (i.e., the integer version).
-. The <<{SENTRY}>> instruction is now in a separate extension {rvy_sentry_insn_ext_name}.
+. The <<{SENTRY}>> instruction is now in a separate extension.
```
arichardson: This is wrong, no need to edit the changelog.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -9,6 +9,14 @@ This chapter introduces a new base ISA called {cheri_base64_ext_name} that exten
 In the future, 32-bit bases ({cheri_base32_ext_name}) will also be defined with near identically behavior but a different in-memory representation, so this chapter uses the term {cheri_base_ext_name} to refer to the behavior common between all bases.
 The {cheri_base_ext_name} base can be composed with other standard options to bases such as E (16-registers), Zfinx and endianness.
 
+All CHERI implementations must also specify a capability encoding format.
+
+This chapter also specifies <<rv64y_cap_description>> which is the capability encoding format for {cheri_base64_ext_name} systems.
+
+ifdef::cheri_ratification_v1_only[]
+Future versions of this specification will include capability encoding formats for {cheri_base32_ext_name}}
```
```suggestion
Future versions of this specification will include capability encoding formats for {cheri_base32_ext_name}
```

How about just keeping this as a non-normative NOTE:?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -181,8 +189,7 @@ Software can query the bounds of a capability held in a general-purpose register
 * The _base_ is returned by the <<GCBASE>> instruction.
 * The _length_ is returned by the <<GCLEN>> instruction.
 ** <<GCLEN>> saturates the _length_ to XLEN bits
-* The _top_ is returned by the <<GCTOP>> instruction (if <<section_rvy_topr_insn_ext>> is implemented).
-** Otherwise the _top_ can be calculated using a saturating addition of the <<GCBASE>> and <<GCLEN>> results.
+* The _top_ is returned by the <<GCTOP>> instruction.
```
arichardson, line 191:
```suggestion
* The _top_ is returned by the <<GCTOP>> instruction.
** <<GCTOP>> saturates the _length_ to XLEN bits.
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -298,20 +307,25 @@ All CHERI systems must support unsealed capabilities.
 Sealed capabilities::
 Capabilities with `CT≠0` are sealed against modification and cannot be dereferenced to access memory.
 Instructions that operate on capabilities will produce a result with a cleared {ctag} if the source capability is sealed and the operation would alter its address, bounds, or permissions.
-Extensions that augment capability metadata must describe their interaction(s) with sealed capabilities.
+Extensions that manipulate capability metadata must describe their interaction(s) with sealed capabilities.
```
```suggestion
Extensions that extend the capability metadata must describe their interaction(s) with sealed capabilities.
```

arichardson (commented):
The main thing missing here is a non-normative note that allows CHERIoT to opt out of having unrestricted sentries.

src/cheri/insns/jalr_32bit.adoc
arichardson: The third NOTE: is the same as the first one.

arichardson:
Overall I think we had a nice spec to deal with opting out of sentry/ybld, but if this is causing problems with review we can go down this route. As long as this doesn't make any of our future changes more difficult. We will definintely need to refine the JALR behaviour for CHERIoT and the future forward/backward sentry extension.

src/cheri/app-cheri-instructions.adoc
tariqkurd-repo:
```suggestion
. The <<{SENTRY}>> instruction is now in a separate extension Zysentry.
```

src/cheri/insns/jalr_32bit.adoc
```suggestion
```

src/cheri/app-cheri-instructions.adoc
tariqkurd-repo: fixed - I avoided the use of the variable

src/cheri/insns/sentry_32bit.adoc
```diff
@@ -20,13 +20,17 @@ Description::
...
-the <<sec_cap_type_ambient,ambient>> value specified by the capability encoding
-given the permissions granted by the capability in `{cs1}`.
+ifdef::cheri_ratification_v1_only[]
+the value `1`, representing a sentry type.
+endif::[]
+ifndef::cheri_ratification_v1_only[]
+the <<sec_cap_type_ambient,ambient>> value specified by the capability encoding given the permissions granted by the capability in `{cs1}`.
+endif::[]
 +
 Set `{cd}.tag=0` if `{cs1}` is sealed.
```
arichardson, line 27: Add a note here that if the capability encoding does not support otype 1, then this instruction is reserved?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -296,20 +306,25 @@ All CHERI systems must support unsealed capabilities.
...
-for types not defined above.
 
-Capability encodings may also make the set of <<sec_cap_type>> values
+ifndef::cheri_ratification_v1_only[]
+
+[NOTE]
+====
+In general, each of sealing and unsealing actions may require *authority*
+to operate on the non-zero type;
+capability encoding formats which define types other than in <<table_capability_types>> will specify how software expresses this authority.
```
```suggestion
capability encoding formats which define types other than the ones in <<table_capability_types>> will specify how software expresses this authority.
```

arichardson (commented):
I've got feedback that the spec is overall difficult to understand if you are not a CHERI expert. I think dropping the details on ambient Vs non ambient types from the initial version of the spec helps with this and is okay since we still have it in the full version. 

Also we need to reduce the number of extensions so this generally sounds good. 
I will review in more detail later. 

@davidchisnall @nwf let me know what you think.

src/cheri/insns/sentry_32bit.adoc
tariqkurd-repo: yes - that sounds sensible - will add that tomorrow (or feel free to do so)

tariqkurd-repo: done

davidchisnall:
I remain of the opinion that ambient sentries are a bad idea for security and that, long term, RV64Y systems will not want them, so putting them in the base seems like a bad idea.  It only costs one otype value, which is less scarce on 64-bit, but it's still a confusion of the model.  I'd be more in favour of removing them entirely than putting them in the base.

davidchisnall:
YTOPR is generally useful, I had missed that it wasn't in Y already, we've had it in CHERIoT for a long time and use it a lot.

YBLD is not a great idea and software that uses it is probably wrong unless it is the swap subsystem, which is an odd things to, but needs basically the same hardware as YSS and is not a big instruction so is just about okay to include.  It will probably cause some ecosystem issues downstream if people ship it.  It's a performance optimisation for an I/O-bound path, which harms capability-format agility if used in any other place.

YSENTRY is standardising something that @nwf threw into a research ISA as an expedient way of building some experimental things, which we removed from CHERIoT based on MSRC security review at Microsoft.  I don't mind too much having it as an extension (ideally with a 'please don't implement this' label) but it shouldn't be in the base.  

Even if ambient sentries *are* a good idea (I don't believe they are, but some people disagree), being able to differentiate forward and backward-edge control-flow arcs is important (various vendors have shipped software mitigations that do *only* this).  We shipped this in CHERIoT a while ago.  Various bits of software will need the ability to create both, so YSENTRY is inadequate as an instruction.  An MVP base should not include instructions that we know we're going to tell people to avoid ever using in a future version of base+extensions.

jrtc27:
We are aware of your opinions on ambient sentries, however, whether sentries are in the base ISA or an extension, they will be a part of the application core profiles, and therefore it makes zero difference to the compatibility story whether they're in the base ISA or not, because in practice they will have to be present.

Similarly, we are aware of your opinions on using YBLD for processing relocations, due to CHERIoT's desire to change the encoding format at will. Again, application cores will not be doing that, so this is not of concern to RV64Y, and as you and others note the hardware cost of adding it is small if you already have YSS.

nwf:
I don't love going back to a single shared taxonomy of otypes across all bases... the present spec has their definitions factored out across encodings (and that's even more defensible now than when I did it because encodings constitute separate bases).

jrtc27:
Except we don't? The CHERIoT base can do what it likes when it defines a variant that adapts "normal" Y?

nwf:
"NOTE: If a capability encoding format does not support <<sec_cap_type>> of `1` then this instruction is _reserved_." sure seems like a return to the much older compromise opt-out we had before we pushed otype semantics out of the base.

jrtc27:
But reserved is exactly what you want? It's not YSENTRY, and can be used by an extension?

tariqkurd-repo:
>before we pushed otype semantics out of the base

Pushing them out of the base has resulted in a spec which people don't understand and we can't get through review so we need a compromise to move forwards. It may be less elegant, but it's certainly more unstandable for people. And as Jessica says, you can define what you like for CHERIoT as each format is its own base, you're not strictly tied to this definition.

The big picture here is to simplify RV64Y to the point where it is understandable enough so it can pass review. Without that we can't make forward progress.

tariqkurd-repo:
I've updated this PR to only simplify Zysentry, JAL and JALR for RV64 and add the CHERIoT style forward/backward-edge sentries to greatly simplify the CHERioT integration and allow us to add the feature into RV64 in due course.

src/cheri/insns/jal_32bit.adoc
```diff
@@ -14,13 +14,14 @@ Encoding::
...
-Jump to the target <<pcc>>.
+
+Direct jump with an address offset.
 +
-Increment <<pcc>>.address by the sign-extended offset to form the target <<pcc>>.
-The <<pcc>> of the next instruction is sealed and written to `{cd}`.
+. The target address is obtained by adding the sign-extended 20-bit J-immediate to the register `{cs1}`.
+. The target <<pcc>> is obtained by applying the target address to the <<pcc>> of the JAL instruction, using the semantics of the <<SCADDR>> instruction.
+. Jump to the target <<pcc>>.
+. The <<pcc>> of the next instruction is written to `{cd}`.
```
arichardson: I think we need to keep mentioning that the return value is sealed

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -14,33 +14,31 @@ include::wavedrom/ct-unconditional-2.adoc[]
...
- Unless an extension, such as <<section_rvy_sentry_insn_ext>>, is implemented which explicitly reference any of the hooks, then take no action for any of them.
-+
-Operationally, copy `{cs1}` to the target <<pcc>> and then...
+ Unless an instruction, such as <<SENTRY>>, is implemented which explicitly reference any of the hooks, then take no action for any of them.
 +
-. Compute the target <<pcc>> address in two steps, each using the semantics of the <<SCADDR>> instruction:
-.. Increment the address of the target <<pcc>> by the sign-extended 12-bit `offset`, and
-.. set bit zero of the target <<pcc>> address to zero.
+endif::[]
+. The target address is obtained by adding the sign-extended 12-bit I-immediate to the register `{cs1}`, then setting the least-significant bit of the result to zero.
```
```suggestion
. The target address is obtained by adding the sign-extended 12-bit immediate to the register `{cs1}`, then setting the least-significant bit of the result to zero.
```

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -14,33 +14,31 @@ include::wavedrom/ct-unconditional-2.adoc[]
...
+. The target <<pcc>> is obtained by applying the target address to `{cs1}` using the semantics of the <<SCADDR>> instruction.
+ifndef::cheri_ratification_v1_only[]
 . HOOK 1: Optionally clear the {ctag} of the target <<pcc>> depending on the <<sec_cap_type>> value and the numerical values of `{cd}` and `{cs1}`
-. Unseal the target <<pcc>> if it is a <<sentry_cap>>, or clear the {ctag} if it is any other sealed type.
-. If `{cd}≠{creg}0`, compute the return capability and install it to `{cd}`:
-.. add the width of this instruction to the current <<pcc>> using the semantics of the <<SCADDR>> instructions, and
+. Unseal the target <<pcc>> if it is a <<sentry_cap>>.
```
arichardson: These rules are no longer quite correct. Scaddr would have detagged even when setting the address to the same value. I think maybe the easiest approach is to move unseal before scaddr but make it conditional on the address matching.

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -14,33 +14,31 @@ include::wavedrom/ct-unconditional-2.adoc[]
...
+endif::[]
+ifndef::cheri_ratification_v1_only[]
+. Clear the {ctag} of the target <<pcc>> if it is any other sealed type.
+endif::[]
+. The <<pcc>> of the next instruction is written to `{cd}`.
```
```suggestion
. The <<pcc>> of the next instruction is sealed and written to `{cd}`.
```
Maybe a vague sealed is good enough here to allow for it being sealed as a return sentry? Since the initial spec doesn't have then we can just say is sealed and then clarify once return sentries are defined?

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -14,33 +14,31 @@ include::wavedrom/ct-unconditional-2.adoc[]
...
+ifndef::cheri_ratification_v1_only[]
 .. HOOK 2: Optionally seal the return capability as a <<sentry_cap>> with a <<sec_cap_type>> defined by the implemented extensions.
 . HOOK 3: Optionally make other architectural state updates.
+endif::[]
 . Jump to the target <<pcc>>.
 +
-[NOTE]
-=====
-When a sealed capability is passed as the input to <<JALR_CHERI>>, its address must have bit zero clear and the instruction must have a zero offset, or the target <<pcc>> will have its {ctag} set to zero, since updates to its address are interpreted with <<SCADDR>> semantics.
```
arichardson, line 39: I think this note is actually useful since otherwise you have to look at more places to understand what happens when a sentry is passed.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -1198,9 +1226,31 @@ NOTE: A future extension may define more restrictive _forward-only_ and _backwar
 
 === Interaction with <<JALR_CHERI>>
 
-{rvy_sentry_insn_ext_name} adds sealing and unsealing behavior to <<JALR_CHERI>>:
+{rvy_sentry_insn_ext_name} adds sealing and unsealing behavior to <<JALR_CHERI>>.
```
arichardson: I'm not sure we can do this while removing the mention of hooks in the jalr. My understanding is that unprivileged extensions can't redefine behaviour of existing instructions.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -1198,9 +1226,31 @@ NOTE: A future extension may define more restrictive _forward-only_ and _backwar
...
+
+ifdef::cheri_ratification_v1_only[]
+* If `{cs1}` is _sealed_ then _unseal_ it before installing in <<pcc>> (see <<sec_cap_type>>).
```
arichardson: And target address== cs1.addr

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -1198,9 +1226,31 @@ NOTE: A future extension may define more restrictive _forward-only_ and _backwar
...
+endif::[]
 
+ifndef::cheri_ratification_v1_only[]
 . For HOOK 2: Seal the return capability as a sentry.
+endif::[]
+
+NOTE: When any sealed capability is passed as the input to <<JALR_CHERI>>, its address must have bit zero clear and the instruction must have a zero offset, or the target <<pcc>> will have its {ctag} set to zero, since updates to its address are interpreted with the semantics of the <<SCADDR>> instruction.
+ As such, a <<sentry_cap>> defines a secure function entry point, and as such the offset in the JALR instruction must be zero.
+
+NOTE: Future {cheri_base_ext_name} extensions are permitted to introduce additional restrictions on the behaviour of <<JALR_CHERI>>, by extending the number of available values in the <<sec_cap_type>> field in the capability encoding format.
```
arichardson: Why is this note in zysentry?

src/cheri/rvy32-encoding.adoc
```diff
@@ -62,7 +62,6 @@ See xref:cap_perms_encoding32[xrefstyle=short].
 | Extension                    | Comment
 | <<section_cheri_hybrid_ext>> | Compatible
 | <<section_zylevels1>>        | Compatible
-| <<section_rvy_sentry_insn_ext>> | Compatible
```
arichardson, line 65: Should revert this I think

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -318,36 +333,49 @@ from those not granting <<x_perm>>;
...
+
+They are not used by <<rv64y_cap_description>>, but are by the CHERIoT formats.
+====
 
 [#sec_cap_type_ambient]
 Ambient sealing type::
 Some capability types are said to be "ambiently available" (or just "ambient")
 if they do not require specific authority to seal a capability (with that type).
-For example, if <<section_rvy_sentry_insn_ext>> is available on a given platform,
+For example, if the <<YSENTRY>> instruction is available on a given platform,
```
```suggestion
For example, if the <<SENTRY>> instruction is available on a given platform,
```

src/cheri/insns/jalr_32bit.adoc
tariqkurd-repo: the I-immediate part is taken from the normal JALR description

src/cheri/insns/jalr_32bit.adoc
tariqkurd-repo: this is the fundamental problem - without Zysentry it _can't_ be sealed - and I don't want this page to be vague in any way. I have an idea of how to fix this.....

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: I missed this one - I've reverted to the original text.

src/cheri/rvy32-encoding.adoc
tariqkurd-repo: done

src/cheri/insns/jalr_32bit.adoc
tariqkurd-repo: I only checked the rendered version for RV64 - will check the full version

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: as a note - as adding the address increment clears the tag as part of YADDRW - so this is already covered

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: I think this is something to discuss with ARC - _how_ to describe this stuff - I'll give it another go first

src/cheri/insns/jal_32bit.adoc
tariqkurd-repo: done

src/cheri/insns/jalr_32bit.adoc
tariqkurd-repo: re-instated

arichardson (commented):
Thanks, I think this is easier to understand and should still be fully compatible with the CHERIoT behaviour. But I'd like @davidchisnall / @nwf / @vmurali to confirm.

src/cheri/insns/jal_32bit.adoc
```diff
@@ -14,13 +14,17 @@ Encoding::
 include::wavedrom/ct-unconditional.adoc[]
 
 Description::
-Jump to the target <<pcc>>.
+
+Direct jump with an address offset.
 +
-Increment <<pcc>>.address by the sign-extended offset to form the target <<pcc>>.
-The <<pcc>> of the next instruction is sealed and written to `{cd}`.
+. The target address is obtained by adding the sign-extended 20-bit J-immediate to `{cs1}.address`.
```
vmurali: Shouldn't it be pcc.address?

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -14,33 +14,30 @@ include::wavedrom/ct-unconditional-2.adoc[]
...
-=====
+NOTE: If the only supported <<sec_cap_type>> is zero, then no <<sentry_cap, sentry capabilities>> exist, and so _sealing_ and _unsealing_ have no effect.
+ This behavior requires non-zero <<sec_cap_type>> values to be implemented, such as by the <<section_rvy_sentry_insn_ext>> extension.
++
+NOTE: When any sealed capability is passed as the input to <<JALR_CHERI>>, its address must have bit zero clear and the instruction must have a zero offset, or the target <<pcc>> will have its {ctag} set to zero, since updates to its address are interpreted with the semantics of the <<SCADDR>> instruction.
```
vmurali: How is the tag cleared automatically when the offset is non zero and cs1 is sealed? I don't think it's specified anywhere.

src/cheri/insns/sentry_32bit.adoc
```diff
@@ -19,11 +19,11 @@ include::wavedrom/sentry.adoc[]
 Description::
 Copy `{cs1}` to `{cd}`.
 +
-Set the capability type (<<sec_cap_type>>) of `{cd}` to
-the <<sec_cap_type_ambient,ambient>> value specified by the capability encoding
-given the permissions granted by the capability in `{cs1}`.
+Set the capability type (<<sec_cap_type>>) of `{cd}` to the value `1`, representing a sentry type.
```
vmurali, line 22: Instead: "Set the capability type (<<sec_cap_type>>) of `{cd}` to the the default sealed type if a default sealed type exists. If default sealed type doesn't exist, this instruction is reserved."

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -271,21 +279,9 @@ architecture: normal bounds check rules should be followed.
...
-|===
-| Type Name | Integer Value | Description
-| Unsealed  | 0             | Unsealed capability granting access to memory
-| Sentry    | 1             | Immutable sealed entry point
-|===
 
-NOTE: The integer value in the table above defines the architectural capability type (i.e., the result of the <<GCTYPE>> instruction) but does not need to be encoded that way in memory.
-Other base ISAs may provide different type mappings, but the value `0` must always represent an unsealed capability.
+The capability encoding must be able to encode type `0` (_unsealed_).
+At least one non-zero type must also be implemented for control-flow integrity.
```
vmurali: Instead: "If other sealed type exists, then we can provide different levels of control-flow integrity. If a default sealed type exists, then the "SENTRY" instruction can seal automatically to that type".

src/cheri/insns/sentry_32bit.adoc
```diff
@@ -19,11 +19,11 @@ include::wavedrom/sentry.adoc[]
...
 +
 Set `{cd}.tag=0` if `{cs1}` is sealed.
++
+NOTE: If a capability encoding format does not support <<sec_cap_type>> of `1` then this instruction is _reserved_.
```
vmurali: This would become redundant if "default sealed type" is defined.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -295,21 +291,69 @@ All CHERI systems must support unsealed capabilities.
...
+[#sentry_cap,reftext="sentry capability"]
+Sentry capability type::
+Sentry capabilities provide *immutable* function pointers within a CHERI software system.
+<<sealed_cap,Sealed capabilities>> are a natural foundation, providing immutability.
+
+NOTE: Capabilities sealed with a specific <<sec_cap_type>> which can be unsealed when passed as an argument to <<JALR_CHERI>> are dubbed "sentries" (a portmanteau of "sealed entries").
+
+Sentries can establish a form of control-flow integrity between mutually distrusting code.
+
+#TODO: ARC COMMENT: "sentries" is a bad choice for a shortened form of "sealed entries" as the standard english or even CS connotation of sentry doesn't match with a sealed entry.#
```
vmurali: I mean, they do "lookout" for bad jumps.

src/cheri/insns/jal_32bit.adoc
```diff
@@ -14,13 +14,17 @@ Encoding::
...
+. The target <<pcc>> is obtained by applying the target address to the <<pcc>> of the JAL instruction, using the semantics of the <<SCADDR>> instruction.
+. The <<pcc>> of the next instruction is sealed as a _backward-edge_ <<sentry_cap>> and written to `{cd}`.
+. Jump to the target <<pcc>>.
++
+NOTE: If the only supported <<sec_cap_type>> is zero, then no <<sentry_cap, sentry capabilities>> exist, and so _sealing_ has no effect.
+ This behavior requires non-zero <<sec_cap_type>> values to be implemented, such as by the <<section_rvy_sentry_insn_ext>> extension.
```
nwf, line 26: Well, I suppose there's nothing stopping `YSENTRY` from creating backward-arc sentries (or conflating forward- and backward- types)... hrm.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -295,21 +291,69 @@ All CHERI systems must support unsealed capabilities.
...
+
+NOTE: In addition to using sealed capabilities as sentries for secure entry points, sealed capabilities can also be useful to software as secure software tokens.
+ <<YSUNSEAL>> can be used to convert such a token back to an unsealed capability.
+ A future extension may add an unseal instruction for performance.
+
+The standard capability type mapping is shown in <<table_capability_types>>.
+
+[#table_capability_types]
+.Default capability type mapping
```
vmurali: Replace "Default" with "Standard"

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -295,21 +291,69 @@ All CHERI systems must support unsealed capabilities.
...
+[width="80%",options=header,align=left,cols="1,1,5"]
+|===
+| Type Name      | Integer Value | Description
+| Unsealed       | 0             | _Unsealed_ capability granting access to memory
+| <<sentry_cap>> | 1             | Immutable _unrestricted sentry_ type for both _forward-edge_ and _backward-edge_ transitions.
```
vmurali: Add "Also serves as the default sealed type in "RV64CHERI" encoding". Certain encodings, like CHERIoT, will not have a default sealed type.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -295,21 +291,69 @@ All CHERI systems must support unsealed capabilities.
...
+|===
+
+The <<sec_cap_type>> value of 1 is enabled by the <<section_rvy_sentry_insn_ext>> extension.
+
+Any sentry type may be passed as the `{cs1}` argument to <<JALR_CHERI>>, and are _unsealed_ on deference.
+The returned `{cd}` value is also _sealed_ as a sentry.
+
+Sentry types may be restricted to _forward-edge_ only and _backward-edge_ only transitions to further improve control-flow integrity.
+
+* When <<JALR_CHERI>> is used as a function call only _forward-edge_ sentries in `{cs1}` are _unsealed_ on deference.
```
vmurali: "dereference", not "deference"

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -295,21 +291,69 @@ All CHERI systems must support unsealed capabilities.
...
+ The returned `{cd}` value is always _sealed_ as a _backward-edge_ sentry type.
+* When <<JALR_CHERI>> is used as a function return only _backward-edge_ sentries in `{cs1}` are _unsealed_ on dereference.
+
+NOTE: Only a single sentry type is defined in <<table_capability_types>>, which is used for both _forward-edge_ and _backward-edge_ transitions.
```
vmurali: Add "And this is also the "default sealing type" for the "RV64" encoding.".

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -295,21 +291,69 @@ All CHERI systems must support unsealed capabilities.
...
+
+NOTE: The integer value in the table above defines the architectural capability type (i.e., the result of the <<GCTYPE>> instruction) but does not need to be encoded that way in memory.
+
+ifndef::cheri_ratification_v1_only[]
+
+[NOTE]
+====
+In general, each of sealing and unsealing actions may require *authority*
+to operate on the non-zero type;
+capability encoding formats which define types other than the ones in <<table_capability_types>> will specify how software expresses this authority.
```
vmurali: Instead, say "If default sealing type is not defined, then each sealing and unsealing actions may require authority to operate on the non-zero type ...".?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -318,35 +362,19 @@ from those not granting <<x_perm>>;
...
-NOTE: Sentry capabilities can establish a form of control-flow integrity between mutually distrusting code.
-
-NOTE: In addition to using sealed capabilities as sentries for secure entry points, sealed capabilities can also be useful to software as secure software tokens.
- <<YSUNSEAL>> can be used to convert such a token back to an unsealed capability.
- A future extension may add an unseal instruction for performance.
-
-NOTE: The set of {cheri_base_ext_name} base ISAs do not make any use of non-zero `CT` values.
- The capability encoding format only needs to encode `CT` information if any extensions (such as {rvy_sentry_insn_ext_name}) are present which support non-zero `CT` values.
+(For the capability encoding format specified by <<rv64y_cap_description>>, that would be type `1`.)
+endif::[]
```
vmurali: Remove the word "ambiently available". Instead stick to if "default sealed type" is defined, then do this ...

vmurali (commented):
We can have a notion of whether a "default sealed type" is defined or not. If it is not defined, "SENTRY" instruction would be reserved. That way, we can simplify the text while not using the word "Ambient Sealing", etc. The comments are based on this assumption.

src/cheri/insns/jalr_32bit.adoc
arichardson: That's what yaddrw (csetaddr) does for sealed caps

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -14,33 +14,30 @@ include::wavedrom/ct-unconditional-2.adoc[]
...
-. Unseal the target <<pcc>> if it is a <<sentry_cap>>, or clear the {ctag} if it is any other sealed type.
-. If `{cd}≠{creg}0`, compute the return capability and install it to `{cd}`:
-.. add the width of this instruction to the current <<pcc>> using the semantics of the <<SCADDR>> instructions, and
-.. HOOK 2: Optionally seal the return capability as a <<sentry_cap>> with a <<sec_cap_type>> defined by the implemented extensions.
-. HOOK 3: Optionally make other architectural state updates.
+. `{cs1}` is written to the target <<pcc>>.
+. If `{cd}` is the link register, `{cs1}` is _sealed_ and `{cs1}` is not a _forward-edge_ <<sentry_cap>>, then set the {ctag} of the target <<pcc>> to zero.
+. If `{cd}=x0`, `{cs1}` is the link register, `{cs1}` is _sealed_ and `{cs1}` is not a _backward-edge_ <<sentry_cap>>, then set the {ctag} of the target <<pcc>> to zero.
+. Unseal the target <<pcc>> if it is any <<sentry_cap>>.
+. The target address is obtained by adding the sign-extended 12-bit I-immediate to `{cs1}.address`, then setting the least-significant bit of the result to zero.
```
```suggestion
. The target address is obtained by adding the sign-extended 12-bit I-immediate to `{cs1}.address`, then setting the least-significant bit of the result to zero.
. Unseal the target <<pcc>> if it is any <<sentry_cap>> and the target address is equal to `{cs1}.address`.
```

src/cheri/insns/jalr_32bit.adoc
vmurali: Hmm, but you work with a "target address", which is obtained after adding the offset. Shouldn't you explicitly say that if cs1 is sealed and there's an offset, then throw exception?

tariqkurd-repo: no exception any more - we clear the tag of the target PCC - and then take an exception at the target - this was requested by ARC - it's possible to take an exception _on_ the JAL but that will need to be a new extension

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: I added the first sentence but not the second, and I'ves topped declaring sentry type 1 in the table as CHERIoT uses that value differently

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: I fixed that by removing the integer value from the table

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: I've removed that note

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: I've removed the default sealed type and so removed that statement

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: ok - I've removed the "default sealing type" from the spec entirely and it all makes much more sense now

src/cheri/insns/sentry_32bit.adoc
tariqkurd-repo: Instead I've gone for "if the cap format doesn't encode cap_type=1 then Zysentry is incompatible"

src/cheri/insns/sentry_32bit.adoc
tariqkurd-repo: I'm not so keen on that - instead I'm bundling up cap_type=1 and YSENTRY together - so if you implement Zysentry you get both definitions together. The problem with the previous spec was the definition of cap_type=1 in the base even though CHERIoT and Zysentry use it differently, so this way there is no default sentry type.

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: true! we need to resolve this.

src/cheri/insns/jalr_32bit.adoc
vmurali: Sure, exception/clear tag. But that behavior is not explained I believe, for jalr with sentry

src/cheri/insns/jal_32bit.adoc
```diff
@@ -14,13 +14,17 @@ Encoding::
...
 +
-Increment <<pcc>>.address by the sign-extended offset to form the target <<pcc>>.
-The <<pcc>> of the next instruction is sealed and written to `{cd}`.
+. The target address is obtained by adding the sign-extended 20-bit J-immediate to `<<pcc>>.address`.
+. The target <<pcc>> is obtained by applying the target address to the <<pcc>> of the JAL instruction, using the semantics of the <<SCADDR>> instruction.
+. The <<pcc>> of the next instruction is _sealed_ as a _backward-edge_ <<sentry_cap>> and written to `{cd}`.
+. Jump to the target <<pcc>>.
++
+NOTE: If the only supported <<sec_cap_type>> is zero, then no <<sentry_cap, sentry capabilities>> exist, and so _sealing_ has no effect.
+ This behavior requires non-zero <<sec_cap_type>> values to be implemented, such as by the <<section_rvy_sentry_insn_ext>> extension.
```
nwf, line 26: I'd rather that be phrased as "if no backward-edge sentry types are supported" not "if sealing isn't supported".

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -14,33 +14,39 @@ include::wavedrom/ct-unconditional-2.adoc[]
...
 . Jump to the target <<pcc>>.
 +
+^1^ Unless redefined by an extension then the _reserved_ behavior is to set the {ctag} of the target <<pcc>> to zero.
++
 [NOTE]
-=====
-When a sealed capability is passed as the input to <<JALR_CHERI>>, its address must have bit zero clear and the instruction must have a zero offset, or the target <<pcc>> will have its {ctag} set to zero, since updates to its address are interpreted with <<SCADDR>> semantics.
-=====
+====
+If the only supported <<sec_cap_type>> is zero, then no <<sentry_cap, sentry capabilities>> exist, and so _sealing_ and _unsealing_ have no effect.
```
nwf: As above, one could support sealing without sentries.

src/cheri/insns/sentry_32bit.adoc
```diff
@@ -19,9 +19,9 @@ include::wavedrom/sentry.adoc[]
 Description::
 Copy `{cs1}` to `{cd}`.
 +
-Set the capability type (<<sec_cap_type>>) of `{cd}` to
-the <<sec_cap_type_ambient,ambient>> value specified by the capability encoding
-given the permissions granted by the capability in `{cs1}`.
+Set the capability type (<<sec_cap_type>>) of `{cd}` to the value `1`, representing a sentry type.
```
nwf, line 22: Not merely representing, but _defining_ a sentry cap, right?

src/cheri/insns/sentry_32bit.adoc
```diff
@@ -19,9 +19,9 @@ include::wavedrom/sentry.adoc[]
...
++
+NOTE: The target capability type is <<sec_cap_type_ambient,ambient>> and so the sealing action requires no specific authority or permission.
```
nwf: I think it's more correct to say that `YSENTRY` is the thing defining otype 1 as ambient (and sentry).

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -295,58 +291,87 @@ All CHERI systems must support unsealed capabilities.
...
+
+The standard capability type mapping is shown in <<table_capability_types>>, extensions and capability encoding formats may add more types.
+
+[#table_capability_types]
+.Standard capability type mapping
+[width="80%",options=header,align=left,cols="1,1,5"]
+|===
+| Type Name      | Integer Value        | Description
+| Unsealed       | 0                    | _Unsealed_ capability granting access to memory
+| <<sentry_cap>> | extension defined^1^ | Immutable _ambient sentry_ type for both _forward-edge_ and _backward-edge_ transitions.
```
nwf: "<<sentry_cap>>" was meant to be a property of otypes, not a name of an otype per se?  In CHERIoT, for example, we use type names like "backwards-edge interrupt-enabling sentry".

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -1191,16 +1216,40 @@ endif::[]
 
 The {rvy_sentry_insn_ext_name} extension:
 
-. Defines one <<sentry_cap>> type, the unrestricted sentry type with a <<sec_cap_type>> of `1`.
-.. These _unrestricted sentry_ capabilities can be used as immutable code pointers for both forward and backward control flow edges.
-NOTE: A future extension may define more restrictive _forward-only_ and _backward-only_ sentry capabilities that can only be used by calls and returns respectively.
-. Adds the <<SENTRY>> instruction to allow sealing capabilities as sentries with <<sec_cap_type>> of `1`.
+. Defines one <<sentry_cap>> type, the <<sec_cap_type_ambient,ambient>> sentry type for both _forward-edge_ and _backward-edge_ transitions, with a <<sec_cap_type>> value of `1`.
```
nwf:
```suggestion
. Defines one <<sentry_cap>> type, the <<sec_cap_type_ambient,ambient>> <<sentry_cap>> type for both _forward-edge_ and _backward-edge_ transitions, with a <<sec_cap_type>> value of `1`.
```

src/cheri/insns/sentry_32bit.adoc
tariqkurd-repo: oh - ok - so you're saying that ct-type 1 as defined by Zysentry could be interpreted as non-ambient by a new extension - I think this is ok as a new instruction can decide it whether it needs permission to seal or not.

src/cheri/insns/jalr_32bit.adoc
tariqkurd-repo: true - but not in the context of JALR? We're not trying to define JALR such that non-sentry types can be sealed.

```suggestion
If the only supported <<sec_cap_type>> is zero, then no <<sentry_cap, sentry capabilities>> exist, and so _sealing_ and _unsealing_ have no effect on JALR.
```

src/cheri/insns/sentry_32bit.adoc
tariqkurd-repo: so you're saying thet Zysentry allocates ct-type=1, and ysentry defines it to be forwards/backwards?
i don't think that works well from a definition viewpoint - it means we can't specify the JALR behavior.
JALR would need to know both the ct-type number and which instruction made it.
I'm much more confortable with ct-type=1 being defined as forwards/backwards by the extension only.

or maybe I misunderstood your point?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -295,58 +291,87 @@ All CHERI systems must support unsealed capabilities.
...
+The standard capability type mapping is shown in <<table_capability_types>>, extensions and capability encoding formats may add more types.
+
+[#table_capability_types]
+.Standard capability type mapping
+[width="80%",options=header,align=left,cols="1,1,5"]
+|===
+| Type Name      | Integer Value        | Description
+| Unsealed       | 0                    | _Unsealed_ capability granting access to memory
+| <<sentry_cap>> | extension defined^1^ | Immutable _ambient sentry_ type for both _forward-edge_ and _backward-edge_ transitions.
+| <<sentry_cap>> | extension defined    | Immutable _sentry_ type for _forward-edge_ transitions.
```
```suggestion
| forward-edge <<sentry_cap>> | extension defined    | Immutable _sentry_ type for _forward-edge_ transitions.
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -295,58 +291,87 @@ All CHERI systems must support unsealed capabilities.
...
+| <<sentry_cap>> | extension defined    | Immutable _sentry_ type for _backward-edge_ transitions.
```
```suggestion
| backward edge <<sentry_cap>> | extension defined    | Immutable _sentry_ type for _backward-edge_ transitions.
```

src/cheri/insns/jal_32bit.adoc
```diff
@@ -14,13 +14,17 @@ Encoding::
...
+Direct jump with an address offset.
 +
-Increment <<pcc>>.address by the sign-extended offset to form the target <<pcc>>.
-The <<pcc>> of the next instruction is sealed and written to `{cd}`.
+. The target address is obtained by adding the sign-extended 20-bit J-immediate to `<<pcc>>.address`.
+. The target <<pcc>> is obtained by applying the target address to the <<pcc>> of the JAL instruction, using the semantics of the <<SCADDR>> instruction.
+. The <<pcc>> of the next instruction is _sealed_ as a _backward-edge_ <<sentry_cap>> and written to `{cd}`.
+. Jump to the target <<pcc>>.
++
+NOTE: If the only supported <<sec_cap_type>> is zero, then no <<sentry_cap, sentry capabilities>> exist, and so _sealing_ has no effect.
```
tariqkurd-repo:
```suggestion
NOTE: If the only supported <<sec_cap_type>> is zero, then no  backward-edge <<sentry_cap, sentry capabilities>> exist, and so _sealing_ has no effect.
```

src/cheri/insns/jal_32bit.adoc
tariqkurd-repo: ok - I've added the "backward-edge" part

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: let's go with sentry

src/cheri/insns/jalr_32bit.adoc
nwf: I mean: a system could not satisfy "If the only supported <<sec_cap_type>> is zero" but still satisfy "no <<sentry_cap, sentry capabilities>> exist".  It is the latter, not the former, that means JALR doesn't seal.

src/cheri/riscv-unpriv-integration.adoc
```suggestion
| Unrestricted <<sentry_cap>> | extension defined^1^ | Immutable _ambient sentry_ type for both _forward-edge_ and _backward-edge_ transitions.
```
Maybe use this term?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -295,58 +291,87 @@ All CHERI systems must support unsealed capabilities.
...
+
+[#table_capability_types]
+.Standard capability type mapping
+[width="80%",options=header,align=left,cols="1,1,5"]
+|===
+| Type Name      | Integer Value        | Description
+| Unsealed       | 0                    | _Unsealed_ capability granting access to memory
+| <<sentry_cap>> | extension defined^1^ | Immutable _ambient sentry_ type for both _forward-edge_ and _backward-edge_ transitions.
+| forward-edge <<sentry_cap>> | extension defined    | Immutable _sentry_ type for _forward-edge_ transitions.
+| backward edge <<sentry_cap>> | extension defined    | Immutable _sentry_ type for _backward-edge_ transitions.
```
```suggestion
| Forward-edge <<sentry_cap>> | extension defined    | Immutable _sentry_ type for _forward-edge_ transitions.
| Backward edge <<sentry_cap>> | extension defined    | Immutable _sentry_ type for _backward-edge_ transitions.
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -1191,16 +1216,40 @@ endif::[]
 
 The {rvy_sentry_insn_ext_name} extension:
 
-. Defines one <<sentry_cap>> type, the unrestricted sentry type with a <<sec_cap_type>> of `1`.
-.. These _unrestricted sentry_ capabilities can be used as immutable code pointers for both forward and backward control flow edges.
-NOTE: A future extension may define more restrictive _forward-only_ and _backward-only_ sentry capabilities that can only be used by calls and returns respectively.
-. Adds the <<SENTRY>> instruction to allow sealing capabilities as sentries with <<sec_cap_type>> of `1`.
+. Defines one <<sentry_cap>> type, the <<sentry_cap>> type for both _forward-edge_ and _backward-edge_ transitions, with a <<sec_cap_type>> value of `1`.
```
```suggestion
. Defines one <<sentry_cap>> type, the <<sentry_cap>> type for both _forward-edge_ and _backward-edge_ control-flow transitions, with a <<sec_cap_type>> value of `1`.
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -1191,16 +1216,40 @@ endif::[]
...
+. Adds the <<SENTRY>> instruction to allow <<sec_cap_type_ambient,ambient>> sealing of capabilities as sentries with <<sec_cap_type>> value of `1`.
```
```suggestion
. Adds the <<SENTRY>> instruction to allow <<sec_cap_type_ambient,ambient sealing >> of capabilities as sentries with <<sec_cap_type>> value of `1`.
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -1191,16 +1216,40 @@ endif::[]
...
+Therefore the full operational description of <<JALR_CHERI>> is specialized to:
+
+. `{cs1}` is written to the target <<pcc>>.
+. The target address is obtained by adding the sign-extended 12-bit I-immediate to `{cs1}.address`, then setting the least-significant bit of the result to zero.
+. _Unseal_ the target <<pcc>> if it is _sealed_ as a _forward-edge_ <<sentry_cap>> (type `1` qualifies for this) and `{cs1}.address[0]` is zero and the I-immediate is zero.
+. Set the address of the target <<pcc>> to the target address using the semantics of the <<SCADDR>> instruction.
+. The <<pcc>> of the next instruction is _sealed_ as a <<sentry_cap>> of type `1` and written to `{cd}`.
+. Jump to the target <<pcc>>.
+
+NOTE: Future extensions may add other <<sentry_cap>> types, but to alter the behavior of <<JALR_CHERI>> differently to this then they will be incompatible with {rvy_sentry_insn_ext_name}.
```
arichardson: I think this note is confusing, I'd just drop it.
```suggestion
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -1191,16 +1216,40 @@ endif::[]
...
+
+{rvy_sentry_insn_ext_name} enables _sealing_ behavior in <<JAL_CHERI>> with the single <<sentry_cap>> of type `1` being available.
+
+Therefore the sealed type of `{cd}` is defined as follows:
+
+* The <<pcc>> of the next instruction is _sealed_ as a <<sentry_cap>> of type `1` and written to `{cd}`.
+
+
+NOTE: Future extensions may add other <<sentry_cap>> types, but if they alter the sealed type written by <<JAL_CHERI>> then they will be incompatible with {rvy_sentry_insn_ext_name}.
+
```
```suggestion
```

Again I would drop this to avoid confusion for readers, and just note in the Zycheriot section that it is incompatible (since JALR seals differently).

src/cheri/rvy64-encoding.adoc
```diff
@@ -125,6 +125,9 @@ The value of the <<SDP-field64>> bits of the <<GCPERM>> result maps 1:1 to the <
 [#section_cap_sealed]
 ==== Capability Type (CT) Encoding
 
+The <<sec_cap_type>> field is as specified in <<table_capability_types>>.
```
arichardson, line 128:
```suggestion
The <<sec_cap_type>> field is as specified in <<table_capability_types>>.

The capabilities of this chapter define a 1-bit field for <<sec_cap_type>> values;
this field directly encodes the values `0` and `1`:

* The encoded value `0` indicates an unsealed type.
* If {rvy_sentry_insn_ext_name} is supported, then `1` encodes an unrestricted <<sentry_cap>>.
* Otherwise, the value `1` is _reserved_.

```

arichardson (commented):
A few minor suggestions, I think we are good to go with these addressed.

src/cheri/insns/sentry_32bit.adoc
nwf: Sorry, I wasn't meaning to draw a distinction between Zysentry and the ysentry opcode.

My point was that, as far as I understand the current spec doc, it's Zysentry that allocates ct-type `1` and defines it to be all of forward-arc sentry, backward-arc sentry, and ambient.  You could, if you wanted, tease those apart -- with Zysentry just doing the first two and a separate Zysentryinsn for the latter -- but I don't think it's worth doing.

arichardson: I agree that semantically it could be cleaner to do the latter, but I think we should just conflate these in one extension to keep the spec simpler for ARC.

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -14,33 +14,39 @@ include::wavedrom/ct-unconditional-2.adoc[]
...
-.. Increment the address of the target <<pcc>> by the sign-extended 12-bit `offset`, and
-.. set bit zero of the target <<pcc>> address to zero.
-. HOOK 1: Optionally set the {ctag} of the target <<pcc>> to zero depending on the <<sec_cap_type>> value and the numerical values of `{cd}` and `{cs1}`
-. Unseal the target <<pcc>> if it is a <<sentry_cap>>, or set the {ctag} to zero if it is any other sealed type.
-. If `{cd}≠{creg}0`, compute the return capability and install it to `{cd}`:
-.. add the width of this instruction to the current <<pcc>> using the semantics of the <<SCADDR>> instructions, and
-.. HOOK 2: Optionally seal the return capability as a <<sentry_cap>> with a <<sec_cap_type>> defined by the implemented extensions.
-. HOOK 3: Optionally make other architectural state updates.
+. `{cs1}` is written to the target <<pcc>>.
+. If `{cd}` is the link register, `{cs1}` is _sealed_ and `{cs1}` is not a _forward-edge_ <<sentry_cap>>, then the behavior is _reserved_^1^.
```
```suggestion
. If `{cd}`≠x0, `{cs1}` is _sealed_ and `{cs1}` is not a _forward-edge_ <<sentry_cap>>, then the behavior is _reserved_^1^.
```

I think the reserved behaviour should apply to all "non-return" uses of jalr?

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -14,33 +14,39 @@ include::wavedrom/ct-unconditional-2.adoc[]
...
+. If `{cd}=x0`, '{cs1}' is the link register , `{cs1}` is _sealed_ and `{cs1}` is not a _backward-edge_ <<sentry_cap>>, then the behavior is _reserved_^1^.
+. The target address is obtained by adding the sign-extended 12-bit I-immediate to `{cs1}.address`, then setting the least-significant bit of the result to zero.
+. Unseal the target <<pcc>> if it is any <<sentry_cap>> and `{cs1}.address[0]`` is zero and the I-immediate is zero.
+. Set the address of the target <<pcc>> to the target address using the semantics of the <<SCADDR>> instruction.
+. The <<pcc>> of the next instruction is sealed as a _backward-edge_ <<sentry_cap>> and written to `{cd}`.
 . Jump to the target <<pcc>>.
 +
+^1^ Unless redefined by an extension then the _reserved_ behavior is to set the {ctag} of the target <<pcc>> to zero.
```
```suggestion
^1^ Extensions that introduce new sealed types must define how they interact with <<JALR_CHERI>>, otherwise the behaviour is to set the {ctag} of the target <<pcc>> to zero.
```

src/cheri/insns/jalr_32bit.adoc
tariqkurd-repo: I think that's reasonable - I guess @vmurali can confirm.

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -14,33 +14,39 @@ include::wavedrom/ct-unconditional-2.adoc[]
...
-. Unseal the target <<pcc>> if it is a <<sentry_cap>>, or set the {ctag} to zero if it is any other sealed type.
-. If `{cd}≠{creg}0`, compute the return capability and install it to `{cd}`:
-.. add the width of this instruction to the current <<pcc>> using the semantics of the <<SCADDR>> instructions, and
-.. HOOK 2: Optionally seal the return capability as a <<sentry_cap>> with a <<sec_cap_type>> defined by the implemented extensions.
-. HOOK 3: Optionally make other architectural state updates.
+. `{cs1}` is written to the target <<pcc>>.
+. If `{cd}` is the link register, `{cs1}` is _sealed_ and `{cs1}` is not a _forward-edge_ <<sentry_cap>>, then the behavior is _reserved_^1^.
+. If `{cd}=x0`, '{cs1}' is the link register , `{cs1}` is _sealed_ and `{cs1}` is not a _backward-edge_ <<sentry_cap>>, then the behavior is _reserved_^1^.
+. The target address is obtained by adding the sign-extended 12-bit I-immediate to `{cs1}.address`, then setting the least-significant bit of the result to zero.
+. Unseal the target <<pcc>> if it is any <<sentry_cap>> and `{cs1}.address[0]`` is zero and the I-immediate is zero.
```
```suggestion
. Unseal the target <<pcc>> if it is any <<sentry_cap>>, `{cs1}.address[0]` is zero, and the I-immediate is zero.
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -295,58 +291,89 @@ All CHERI systems must support unsealed capabilities.
...
+[#table_capability_types]
+.Standard capability type mapping
+[width="100%",options=header,align=left,cols="2,1,4"]
+|===
+| Type Name                     | Integer Value | Description
+| Unsealed                      | 0             | _Unsealed_ capability granting access to memory
+| Unrestricted <<sentry_cap>>^1^| 1             | Immutable _unrestricted sentry_ type for both _forward-edge_ and _backward-edge_ transitions.
+| Forward-edge  <<sentry_cap>>  | 1             | Immutable _sentry_ type for _forward-edge_ transitions.
+| Backward-edge <<sentry_cap>>  | 1             | Immutable _sentry_ type for _backward-edge_ transitions.
```
```suggestion
| Forward-edge  <<sentry_cap>>  | Defined by future extension | Immutable _sentry_ type for _forward-edge_ transitions.
| Backward-edge <<sentry_cap>>  | Defined by future extension | Immutable _sentry_ type for _backward-edge_ transitions.
```
We don't have those two types yet? They only exist in CHERIoT so far.

arichardson: Or maybe just N/A and another footnote underneath to say

^2^ These types are not available in the current set of extensions but are available on CHERIoT systems.

arichardson (commented):
One comment on the type mapping mapping table otherwise looking good.

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: true - and value `1` isn't allocated here - I'll fix that now....

src/cheri/insns/jalr_32bit.adoc
vmurali: I think this is fine

src/cheri/insns/jal_32bit.adoc
```diff
@@ -14,13 +14,17 @@ Encoding::
...
+Direct jump with an address offset.
 +
-Increment <<pcc>>.address by the sign-extended offset to form the target <<pcc>>.
-The <<pcc>> of the next instruction is sealed and written to `{cd}`.
+. The target address is obtained by adding the sign-extended 20-bit J-immediate to `<<pcc>>.address`.
+. The target <<pcc>> is obtained by applying the target address to the <<pcc>> of the JAL instruction, using the semantics of the <<SCADDR>> instruction.
+. The <<pcc>> of the next instruction is _sealed_ as a _backward-edge_ <<sentry_cap>> and written to `{cd}`.
+. Jump to the target <<pcc>>.
++
+NOTE: If the only supported <<sec_cap_type>> is zero, then no  backward-edge <<sentry_cap, sentry capabilities>> exist, and so _sealing_ has no effect.
```
arichardson:
```suggestion
NOTE: If the only supported <<sec_cap_type>> is zero, then no backward-edge <<sentry_cap, sentry capabilities>> exist, and so _sealing_ has no effect.
```

arichardson (approved):

PR #996, opened 2026-04-17
Make svyrg PTE bits the default

tariqkurd-repo:
Simplify the ARC review by making the PTE bits from Svyrg the default.
In future, the bits can be reused by adding a bit into sstatus to control the meaning.


jrtc27:
This isn't a good direction to head in. It bakes into the architecture that Svyrg is mandatory, with no way to ever undo that even if many versions of profiles have deprecated it and want to jettison it, no?

tariqkurd-repo:
> This isn't a good direction to head in. It bakes into the architecture that Svyrg is mandatory, with no way to ever undo that even if many versions of profiles have deprecated it and want to jettison it, no?

Correct - but without it what do we have? No temporal safety.

The alternative is to drop Svyrg completely and just have the simple 2-bit load-enable/store-enable scheme.
Having options and perceived indecision in the spec is really hurting us.

The hardware cost of Svyrg is very low - so I say bake it in and move on. 

We can add enable bits into sstatus in the future to repurpose some of the bits if need be.

However the SCRG bit is totally untested! It's a logical extension of UCRG but maybe we should just not do that yet as there's no software support at all for it AFAIK. That really does seem like an unnecessary risk.

arichardson (commented):
Don't we need to keep the enable bit? Otherwise I think it might be difficult to add a future extension later if there isn't already a bit to opt out of those bits being reserved?

tariqkurd-repo:
I think what we decided was revert to the previous spec but delay Svyrg for the first ratification package, but keep it for the first profile.

jrtc27:
I don't think we did?

PR #1000, opened 2026-04-22
ARC cleanup including renaming hybrid modes and M-bit->P-bit

tariqkurd-repo:
I've started from #994 to avoid merge problems later

I've reordered the CSRs in the CSV so that they are more logically grouped, as it makes the behaviour summary table much easier to read.

Most of the edits are very minor - but it does include renaming the CHERI Execution Modes to:

* Pointer Mode
* Capability Mode

As the metadata bit = 1 for Pointer Mode I've renamed it from M to P

### DIFF
--- a/src/cheri/attributes.adoc
+++ b/src/cheri/attributes.adoc
@@ -118,8 +118,8 @@ endif::support_varxlen[]
 :ctag_cap:   Capability tag
 :ctag_title: Capability Tag
 
-:cheri_int_mode_name: pass:quotes[_(Non-CHERI) Address Mode_]
-:cheri_cap_mode_name: pass:quotes[_(CHERI) Capability Mode_]
+:cheri_int_mode_name: pass:quotes[_Pointer Mode_]
+:cheri_cap_mode_name: pass:quotes[_Capability Mode_]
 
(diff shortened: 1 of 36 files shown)


src/cheri/attributes.adoc
```diff
@@ -118,8 +118,8 @@ endif::support_varxlen[]
 :ctag_cap:   Capability tag
 :ctag_title: Capability Tag
 
-:cheri_int_mode_name: pass:quotes[_(Non-CHERI) Address Mode_]
-:cheri_cap_mode_name: pass:quotes[_(CHERI) Capability Mode_]
+:cheri_int_mode_name: pass:quotes[_Pointer Mode_]
```
arichardson: I don't think we should use this name even though it was suggested since capabilities are also pointers. Can we just keep address mode?

tariqkurd-repo: The would we have to also rename the P (pointer) bit? I guess it still makes sense as we're changing the behaviour of pointers.

tariqkurd-repo: P for pointer mode - the pointers are either _addresses_ or _capabilities_ with some justifying text

PR #1001, opened 2026-04-22
Fix description and rename YLT to YSS and SYEQ to YEQ

### LINKED ISSUE #999: Should YLT be YLE?
ARC review comment - if rs1=rs2 causes YLT to return 1 then it should be YLE.
The YLT text isn't very clear - talks of subset without accurately defining the term
The YLT text refers to YBLD which does allow equality.

tariqkurd-repo:
fix #999

### DIFF
--- a/src/cheri/attributes.adoc
+++ b/src/cheri/attributes.adoc
@@ -277,9 +277,9 @@ endif::support_varxlen[]
 :SCEQ_LC:      syeq
 :SCEQ_DESC:    Capability equality comparison including {ctag}
 
-:SCSS:         YLT
-:SCSS_LC:      ylt
-:SCSS_DESC:    Capability less than comparison including {ctag}
+:SCSS:         YSS
+:SCSS_LC:      yss
+:SCSS_DESC:    Capability subset
 
 :SCMODE:       YMODEW
 :SCMODE_LC:    ymodew
--- a/src/cheri/attributes.adoc
+++ b/src/cheri/attributes.adoc
@@ -277,9 +277,9 @@ endif::support_varxlen[]
 :SCEQ_LC:      syeq
 :SCEQ_DESC:    Capability equality comparison including {ctag}
 
-:SCSS:         YSS
-:SCSS_LC:      yss
-:SCSS_DESC:    Capability subset
+:SCSS:         YTSS
+:SCSS_LC:      ytss
+:SCSS_DESC:    Capability test subset
 
 :SCMODE:       YMODEW
 :SCMODE_LC:    ymodew
--- a/src/cheri/attributes.adoc
+++ b/src/cheri/attributes.adoc
@@ -273,13 +273,13 @@ endif::support_varxlen[]
 :SCHI_LC:      yhiw
 :SCHI_DESC:    Write capability metadata and clear {ctag} (pseudo)
 
-:SCEQ:         SYEQ
-:SCEQ_LC:      syeq
+:SCEQ:         YEQ
+:SCEQ_LC:      yeq
 :SCEQ_DESC:    Capability equality comparison including {ctag}
 
-:SCSS:         YTSS
-:SCSS_LC:      ytss
-:SCSS_DESC:    Capability test subset
+:SCSS:         YSS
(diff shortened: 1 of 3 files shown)


src/cheri/attributes.adoc
```diff
@@ -277,9 +277,9 @@ endif::support_varxlen[]
 :SCEQ_LC:      syeq
 :SCEQ_DESC:    Capability equality comparison including {ctag}
 
-:SCSS:         YLT
-:SCSS_LC:      ylt
-:SCSS_DESC:    Capability less than comparison including {ctag}
+:SCSS:         YSS
+:SCSS_LC:      yss
+:SCSS_DESC:    Capability subset
```
davidchisnall, line 282: Perhaps 'capability test subset'?  The problem with subset in the name is that it's both a verb and a noun and this looks like the verb form.

tariqkurd-repo: true - it could be more specific

src/cheri/attributes.adoc
```diff
@@ -277,9 +277,9 @@ endif::support_varxlen[]
 :SCEQ_LC:      syeq
 :SCEQ_DESC:    Capability equality comparison including {ctag}
 
-:SCSS:         YLT
-:SCSS_LC:      ylt
-:SCSS_DESC:    Capability less than comparison including {ctag}
+:SCSS:         YTSS
+:SCSS_LC:      ytss
```
jrtc27: The T isn't helpful here (and goes against all the two-letter comparison suffices in RISC-V), and presumably it should be SYSS to match SYEQ?

src/cheri/attributes.adoc
```diff
@@ -277,9 +277,9 @@ endif::support_varxlen[]
...
+:SCSS_DESC:    Capability test subset
```
jrtc27: Why not "Capability subset comparison"?

src/cheri/attributes.adoc
jrtc27: Although for floating-point it's FLT not SFLT... so maybe we should be YSS and YEQ

arichardson: Krste's comments so far have been that he prefers a Y prefix for every instruction. I guess that's also why it's FLT and not SFLT.

src/cheri/insns/sceq_32bit.adoc
```diff
@@ -21,6 +21,8 @@ include::wavedrom/sceq.adoc[]
 Description::
 Set `rd` to 1 if all bits (i.e., YLEN bits and the {ctag}) of capabilities `{cs1}`
 and `{cs2}`  are equal, otherwise set `rd` to 0.
++
+NOTE: Even though this is specified as a capability operation it also functions correctly as a YLEN-wide operand equality comparison.
```
arichardson: I think this is more confusing that helpful for most people. We can keep it commented out but I don't think it should be in the PDF.

arichardson (approved):
LGTM with fixes.

src/cheri/insns/sceq_32bit.adoc
jrtc27: Yeah. A YLEN-wide equality comparison *is* a capability comparison (as ever, ambiguous as to whether YLEN encompasses the tag). And of course comparing YLEN bits can be used to compare YLEN bits.

arichardson: I know that you're trying to say is this could be reused for a Rv128inY 128-bit, but I think dropping it is better
```suggestion
```

src/cheri/insns/sceq_32bit.adoc
```diff
@@ -21,6 +21,8 @@ include::wavedrom/sceq.adoc[]
 Description::
 Set `rd` to 1 if all bits (i.e., YLEN bits and the {ctag}) of capabilities `{cs1}`
 and `{cs2}`  are equal, otherwise set `rd` to 0.
++
```
arichardson:
```suggestion
```

arichardson (approved):

PR #1002, opened 2026-04-23
Update for new instruction encodings in Custom-3

arichardson:
Using the previously reserved custom3 opcode, which becomes redefined to be the RVY major opcode with a RVY base ISA.

This includes changing the format of the immediate of YBNDSWI

Generated by a script which is included here.

Assisted-by: Gemini CLI


arichardson:
This should be now ready for review. All wavedrom updates were generated from the source python script.

arichardson:
Most of the actual logic is in src/cheri/scripts/rvy_instruction_encodings.py

src/cheri/insns/sh1234add_32bit.adoc
```diff
@@ -29,13 +29,13 @@ Mnemonics ({cheri_base64_ext_name})::
...
 [wavedrom, , svg,subs=attributes+]
 ....
-{reg:[
-    { bits:  7, name: 0x33, attr: ['OP'] },
-    { bits:  5, name: '{cd}' },
-    { bits:  3, name: 'func3', attr: ['{SH1ADD_CHERI}=011', '{SH2ADD_CHERI}=101', '{SH3ADD_CHERI}=000', 'rv64: {SH4ADD_CHERI}=111'] },
-    { bits:  5, name: 'rs1' },
-    { bits:  5, name: '{cs2}' },
-    { bits:  7, name: 0x10 },
+{reg: [
```
tariqkurd-repo, line 32: this should be inthe wavrdrom directory with the rest (I think this was my fault originally)

src/cheri/insns/sh1234add_uw_32bit.adoc
```diff
@@ -27,13 +27,13 @@ Mnemonics ({cheri_base64_ext_name})::
...
 [wavedrom, , svg,subs=attributes+]
 ....
-{reg:[
-    { bits:  7, name: 0x3b, attr: ['OP-32'] },
-    { bits:  5, name: '{cd}' },
-    { bits:  3, name: 'func3', attr: ['rv64: {SH1ADD_UW_CHERI}=011', 'rv64: {SH2ADD_UW_CHERI}=101', 'rv64: {SH3ADD_UW_CHERI}=000', 'rv64: {SH4ADD_UW_CHERI}=111'] },
-    { bits:  5, name: 'rs1' },
-    { bits:  5, name: '{cs2}' },
-    { bits:  7, name: 0x10 },
+{reg: [
```
tariqkurd-repo, line 30: and this one should move

tariqkurd-repo:
The obvious ommission is a statement saying that we're reserving CUSTOM-1/2/3 for CHERI encodings and may release some of them for custom space later.

jrtc27:
Did we decide on custom-1?

tariqkurd-repo:
We also need text about MISA.Y switching between existing custom usage and CHERI usage, I'm going to add that into #1006

tariqkurd-repo:
The current plan is to reserve all but custom-0

src/cheri/scripts/generate_custom3_adoc.py
```diff
@@ -0,0 +1,139 @@
...
+
+    # 3. R-Type 1/2-Operand
+    r2_by_f7 = {}
+    for i in rvy_insts.regular_2op_insns:
+        if isinstance(i.rs2.val, int) and hasattr(i, "f7") and isinstance(i.f7.val, int):
+            r2_by_f7.setdefault(i.f7.val, []).append(i)
+
+    table_num = 3
+    for f7_val, r2_insts in sorted(r2_by_f7.items()):
+        adoc += f"=== {table_num}. R-Type 1/2-Operand (funct3=000, funct7={format(f7_val, '07b')})\n"
```
tariqkurd-repo:
```suggestion
        adoc += f"=== {table_num}. R-Type 1-Op/2-Op (funct3=000, funct7={format(f7_val, '07b')})\n"
```

easier to read without "1/2" in the text

tariqkurd-repo:
It would be nice to make these instructions into xrefs

image

tariqkurd-repo:
I noticed that YSUNSEAL had rs2 as XLEN not YLEN - I put a comment on the gsheet

tariqkurd-repo:
This one hasn't rendered well as the string is too long 
image

tariqkurd-repo:
I've proposed renaming custom-1/2/3 to RVY-1/2/3 in #1019 - if acceptable this this PR should use RVY-3

src/cheri/insns/sh1234add_32bit.adoc
arichardson: The wavedrom is inline here instead of a seaprate file, we can just keep it that way.

src/cheri/insns/sh1234add_uw_32bit.adoc
arichardson: same here

arichardson:
I've added the requested xrefs and fixed a few minor issues. Also added a warning that these encodings are not necessarily final.

tariqkurd-repo:
this should be funct5 - not rs2 - as these formats don't have rs2

image

also here - I guess the field is defined incorrectly:

image

tariqkurd-repo:
I think that a sentence saying that these three LSBs are the types of the operands is worth having:

image

arichardson:
> this should be funct5 - not rs2 - as these formats don't have rs2
> 
> image
> also here - I guess the field is defined incorrectly:
> 
> image

Oh yeah good catch that should be easy to fix

arichardson:
Following up on this mornings meeting and further offline discussion, I am merging this now and will cut a release.

PR #1006, opened 2026-04-24
ARC review updates and includes renaming hybrid modes and M-bit->P-bit

tariqkurd-repo:
Update the hybrid mode introduction and names for the new proposed encoding scheme.

### DIFF
--- a/src/cheri/attributes.adoc
+++ b/src/cheri/attributes.adoc
@@ -124,7 +124,7 @@ endif::support_varxlen[]
 :ctag_cap:   Capability tag
 :ctag_title: Capability Tag
 
-:cheri_int_mode_name: pass:quotes[_Pointer Mode_]
+:cheri_int_mode_name: pass:quotes[_Address Mode_]
 :cheri_cap_mode_name: pass:quotes[_Capability Mode_]
 
 :c_cheri_base_ext_names:   C or Zca, {cheri_base_ext_name}
--- a/src/cheri/attributes.adoc
+++ b/src/cheri/attributes.adoc
@@ -124,8 +124,8 @@ endif::support_varxlen[]
 :ctag_cap:   Capability tag
 :ctag_title: Capability Tag
 
-:cheri_int_mode_name: pass:quotes[_(Non-CHERI) Address Mode_]
-:cheri_cap_mode_name: pass:quotes[_(CHERI) Capability Mode_]
+:cheri_int_mode_name: pass:quotes[_Pointer Mode_]
+:cheri_cap_mode_name: pass:quotes[_Capability Mode_]
 
 :c_cheri_base_ext_names:   C or Zca, {cheri_base_ext_name}
 :c_cheri_default_ext_names: C or Zca, {cheri_default_ext_name}
--- a/src/cheri/attributes.adoc
+++ b/src/cheri/attributes.adoc
@@ -121,7 +121,7 @@ endif::support_varxlen[]
 :ctag_cap:   Capability tag
 :ctag_title: Capability Tag
 
-:cheri_int_mode_name: pass:quotes[_Pointer Mode_]
+:cheri_int_mode_name: pass:quotes[_Address Mode_]
 :cheri_cap_mode_name: pass:quotes[_Capability Mode_]
 
 :c_cheri_base_ext_names:    Zca, {cheri_base_ext_name}
--- a/src/cheri/attributes.adoc
+++ b/src/cheri/attributes.adoc
@@ -17,6 +17,7 @@
 :cheri_base_ext_name_lc:   pass:quotes[rvy]
 // CHERI extension adding support for integer pointer mode (and mode bit)
 :cheri_default_ext_name:   Zyhybrid
+:cheri_mode_name:          CHERI Pointer Mode
 :cheri_priv_m_ext: Machine-Level ISA ({cheri_base_ext_name})
 :cheri_priv_m_reg_enable_ext: {cheri_default_ext_name} for Privileged Architectures
 ifdef::support_varxlen[]
@@ -121,8 +122,8 @@ endif::support_varxlen[]
 :ctag_cap:   Capability tag
 :ctag_title: Capability Tag
 
-:cheri_int_mode_name: pass:quotes[_Address Mode_]
-:cheri_cap_mode_name: pass:quotes[_Capability Mode_]
+:cheri_int_mode_name: pass:quotes[_RVI Pointer Mode_]
+:cheri_cap_mode_name: pass:quotes[_RVY Pointer Mode_]
 
 :c_cheri_base_ext_names:    Zca, {cheri_base_ext_name}
 :c_cheri_default_ext_names: Zca, {cheri_default_ext_name}
--- a/src/cheri/attributes.adoc
+++ b/src/cheri/attributes.adoc
@@ -122,8 +122,8 @@ endif::support_varxlen[]
 :ctag_cap:   Capability tag
 :ctag_title: Capability Tag
 
-:cheri_int_mode_name: pass:quotes[_RVI Pointer Mode_]
-:cheri_cap_mode_name: pass:quotes[_RVY Pointer Mode_]
+:cheri_int_mode_name: pass:quotes[_Integral Pointer Mode_]
+:cheri_cap_mode_name: pass:quotes[_Capability Pointer Mode_]
 
 :c_cheri_base_ext_names:    Zca, {cheri_base_ext_name}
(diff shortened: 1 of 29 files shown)


src/cheri/csv/CHERI_CSR.csv
```diff
@@ -1,48 +1,48 @@
-"YLEN CSR","Address","Alias","Mode","Permissions","Reset Value","Action on XLEN write","Action on YLEN write","Code Pointer","Data Pointer","Unseal On Execution","Prerequisites","Description","initial_ratification","","","","","","","","","","","","","","","","","","","",""
+"YLEN CSR","Address","Alias","Mode","Permissions","Reset Value","Action on XLEN write","Action on YLEN write","Code Pointer","Data Pointer","Unseal On Execution","Prerequisites","Description","initial_ratification","","","","","","","","","","","","","","","","","","","",""
```
arichardson, line 1: What actually changed in this file?

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -26,8 +26,8 @@ Operationally, copy `{cs1}` to the target <<pcc>> and then...
 . Compute the target <<pcc>> address in two steps, each using the semantics of the <<SCADDR>> instruction:
 .. Increment the address of the target <<pcc>> by the sign-extended 12-bit `offset`, and
 .. set bit zero of the target <<pcc>> address to zero.
-. HOOK 1: Optionally clear the {ctag} of the target <<pcc>> depending on the <<sec_cap_type>> value and the numerical values of `{cd}` and `{cs1}`
-. Unseal the target <<pcc>> if it is a <<sentry_cap>>, or clear the {ctag} if it is any other sealed type.
+. HOOK 1: Optionally set the {ctag} to zero of the target <<pcc>> depending on the <<sec_cap_type>> value and the numerical values of `{cd}` and `{cs1}`
```
```suggestion
. HOOK 1: Optionally set the {ctag} of the target <<pcc>> to zero depending on the <<sec_cap_type>> value and the numerical values of `{cd}` and `{cs1}`
```

src/cheri/csv/CHERI_CSR.csv
tariqkurd-repo: I reordered it as it makes the generated table much easier to read, as registers with the same access type are now logically grouped

src/cheri/attributes.adoc
```diff
@@ -121,8 +121,8 @@ endif::support_varxlen[]
 :ctag_cap:   Capability tag
 :ctag_title: Capability Tag
 
-:cheri_int_mode_name: pass:quotes[_(Non-CHERI) Address Mode_]
-:cheri_cap_mode_name: pass:quotes[_(CHERI) Capability Mode_]
+:cheri_int_mode_name: pass:quotes[_Address Mode_]
+:cheri_cap_mode_name: pass:quotes[_Capability Mode_]
```
tariqkurd-repo: Should these be RVY and RVI mode?
We also need text about custom space mapping according to the different modes.

src/cheri/rvy64-encoding.adoc
```diff
@@ -52,7 +52,7 @@ This capability encoding has the following properties that affect the observable
 | EW        | {cap_rv64_exp_width} | Exponent width
 | CAP_MAX_E | {cap_rv64_exp_max}   | Maximum exponent value
 | enableL8  | 0                    | Whether the encoding format includes the L~8~ bit
-| AP_MBit   | 0                    | Whether the <<m_bit>> is encoded in the AP-field
+| AP_MBit   | 0                    | Whether the <<p_bit>> is encoded in the AP-field
```
```suggestion
| AP_PBit   | 0                    | Whether the <<p_bit>> is encoded in the AP-field
```

src/cheri/attributes.adoc
jrtc27: No, because that would mislead; it's still RVY, so we can use the custom space, not RVI where it's reserved. It's just RVI-compatible in its behaviour.

tariqkurd-repo: I think that with the rename of the mode bit to the pointer bit we're saying that the pointer mode is RVI although the base arch is RVY?

jrtc27: "It's RVY, but it's RVI, but it's not actually RVI" yeah no thank you, let's stick with something more accurate and not confusing.

tariqkurd-repo: that's exactly what we discussed - RVY base but executing code compatible with RVI is it not?

image

tariqkurd-repo: this was @arichardson 's comment on the subject:

The key distinction here is the misa.Y bit as a "base ISA" selector. YMODESW does not need to be executable from the pure RVI mode (misa.Y=0). When misa.Y=0, the base ISA is strictly RVI, and Custom-3 remains completely available for vendor-specific extensions. CHERI state is completely disabled, so there is no meaningful way to use the YMODESW instruction in this state. To switch into RVY purecap mode, the software would first need to set misa.Y=1. Once misa.Y=1, the core is officially using the RVY base ISA (in RVI compatibility mode where PC.pointer_mode=RVI). In this RVY base ISA state, Custom-3 is legitimately reserved for RVY instructions, making YMODESW available to seamlessly switch PC.pointer_mode to RVY.

So I'm proposing we go for *PC.pointer_mode=RVI*

jrtc27: My point is not about what it does, but about how to describe it. We need to be very careful to not make it sound like it actually _is_ RVI, just that it is _compatible with_ RVI, and calling it "RVI pointer mode", to me, makes it sound too much like "it is RVI".

jrtc27: (Also, on top of that, RVI vs RVY is bad naming as it ignores RVE)

arichardson: Agreed, I think integer/address is the cleaner naming

tariqkurd-repo: but they've rejected that already so we're not going to converge

tariqkurd-repo: Alex please just choose a name and update the PR and merge it!

jrtc27: Have they rejected “Integer Address Pointer Mode” / “CHERI Capability Pointer Mode”?

tariqkurd-repo: Krste said "it's not an integer it's just a pointer" when we were in Cambridge. Those names are too long and cumbersome anyway

arichardson: I'll add a list of options.
How about Using `Integral Pointer Mode`/`Capability Pointer Mode` as the current version? That avoids the "it's not an integer, it's a pointer" concern since we are now using an adjective to describe the type of pointer?

jrtc27: > "it's not an integer it's just a pointer"

That's... I don't even know where to start with that one.

> Integral Pointer Mode/Capability Pointer Mode

Sure, that works for me. Keeps it down to a single word for each mode, so you can presumably stylise it as `cap.P=I(ntegral)/C(apability)`.

arichardson:
Pushed some minor updates, will rebase #1013 and #1002 on top of this.

src/cheri/attributes.adoc
arichardson: Updated now and added a list of potential naming options, hopefully nothing too silly in that list.

arichardson:
I've also update the Zicsr section now to be more readable. Hopefully this all looks good and if so I will go ahead and merge this on Monday.

arichardson (approved):

src/cheri/riscv-hybrid-integration.adoc
```diff
@@ -1,98 +1,129 @@
...
+|===
+| Compatibility Mode Name | Purecap Mode Name | Notes
+| {cheri_int_mode_name} | {cheri_cap_mode_name} | Matches compiler terminology
+| Non-CHERI Mode | CHERI Mode | Using a negative to describe a mode is awkward
+| Non-CHERI Pointer Mode | CHERI Pointer Mode | Same again just with  "Pointer"
+| Address Mode | Capability Mode |
+| Address Pointer Mode | Capability Pointer Mode | "Address Pointer" is a bit odd
+| (Non-CHERI) Address Mode | (CHERI) Capability Mode | Older rejected proposal, parens are a bit awkward
+| Integer Mode | Capability Mode | CHERIv9 naming scheme.
+| RVI Mode | RVY Mode | Omits RVE, should it be RVI/RVE mode?
```
tariqkurd-repo, line 59: RVE is a subset of RVI, so I think it can be ignored here?

tariqkurd-repo:
I think this one is good-to-go now - the only way we'll resolve the naming is to choose one from a list with input.

PR #1009, opened 2026-04-28
Reword to clarify that load LM permission clearing is based on written back tag

### LINKED ISSUE #1007: Clarify permission clearing when tag stripping
This sentence seems ambiguous on whether we check the tag on the "loaded capabilities" as they come out of memory or as they land in the register. The difference is we might strip the tag on the way because of some cleared MMU cap permissive bit. It seems like we mean only to clear the permissions bits if the actual capability written back by the instruction is tagged?

https://github.com/riscv/riscv-cheri/blob/f4cfdced3c67da30a1899639cff65d34aed15d93/src/cheri/riscv-unpriv-integration.adoc?plain=1#L400

PeterRugg:
Fixes https://github.com/riscv/riscv-cheri/issues/1007

### DIFF
--- a/src/cheri/riscv-unpriv-integration.adoc
+++ b/src/cheri/riscv-unpriv-integration.adoc
@@ -397,7 +397,7 @@ If <<c_perm>> is missing then the {ctag}s for capability loads and stores are re
 Load Mutable Permission (LM):: Allow preserving the <<w_perm>> of capabilities loaded from memory.
 If a capability grants <<r_perm>> and <<c_perm>>, but no <<lm_perm>>, then a capability loaded via this authorizing capability will have <<w_perm>> and <<lm_perm>> removed.
 +
-The permission stripping behavior _only_ applies to loaded capabilities that have their {ctag} set and are not sealed.
+The permission stripping behavior _only_ applies to loaded capabilities with their {ctag} set and that are not sealed after all other checks.
 This ensures that capability loads of non-capability data do not modify the loaded value, and that sealed capabilities are not modified.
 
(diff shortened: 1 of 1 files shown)


PeterRugg:
I checked and the levels ones are already not-confusing.

arichardson (approved):

davidchisnall:
I don't really like expressing this in terms of register writeback.  I hope RISC-V will grow AArch64-like memcpy instructions at some point in the future, and these should respect the same properties even though they never write the copied values to a register.

arichardson:
> I don't really like expressing this in terms of register writeback. I hope RISC-V will grow AArch64-like memcpy instructions at some point in the future, and these should respect the same properties even though they never write the copied values to a register.

That's a good point, do you have any other wording suggestions to resolve the ambiguity? We can't say something like "after any mmu/revocation engine clearing" here since those are defined in the priv spec.

nwf:
I think I'd like the "Determining the final value of rd" text within `lw` be reworked to be something like "Capability transforms applied on load" and maybe floated out of `lw`.  Then we could refer to that in `lw` and any other instructions that come to read caps through caps and also as the thing being extended in `Zylevels1` (which presently says "When a capability is loaded through...").

PeterRugg:
How about just "The permission stripping behavior _only_ applies to loaded capabilities that have their {ctag} set and are not sealed **after all other checks**". That covers any MMU things without calling them out specifically.

arichardson:
> How about just "The permission stripping behavior _only_ applies to loaded capabilities that have their {ctag} set and are not sealed **after all other checks**". That covers any MMU things without calling them out specifically.

Sounds good to me

PeterRugg:
I made the change.

arichardson (approved):

PR #1015, opened 2026-05-10
Use set to 0 instead of cleared consistently

arichardson:
This was another ARC comment.


nwf (approved):
Seems like a fine plan

tariqkurd-repo (approved):

PR #1019, opened 2026-05-11
Specify remapping of custom space

tariqkurd-repo:
Specify remapping of custom-n -> RVY-n
#1002 should refer to RVY-3 once this PR is merged

### DIFF
--- a/src/cheri/riscv-unpriv-integration.adoc
+++ b/src/cheri/riscv-unpriv-integration.adoc
@@ -878,19 +878,19 @@ No other exception paths are added by {cheri_base_ext_name}: in particular, capa
 |inst[4:2] .2+|000 .2+|001   .2+|010     .2+|011   .2+|100 .2+|101     .2+|110          .2+|111 (>32b)
 |inst[6:5]
 |00           |LOAD   |LOAD-FP  |_custom-0_ |MISC-MEM |OP-IMM |AUIPC      |OP-IMM-32       |_reserved_
-|01           |STORE  |STORE-FP |*RVY-1*    |AMO      |OP     |LUI        |OP-32           |_reserved_
-|10           |MADD   |MSUB     |NMSUB      |NMADD    |OP-FP  |OP-V       |*RVY-2*         |_reserved_
-|11           |BRANCH |JALR     |_reserved_ |JAL      |SYSTEM |OP-VE      |*RVY-3*         |_reserved_
+|01           |STORE  |STORE-FP |*{cheri_base_ext_name}-C*    |AMO      |OP     |LUI        |OP-32           |_reserved_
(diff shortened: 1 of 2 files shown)


src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -869,9 +869,28 @@ E.g., `lw t0, 16({abi_creg}sp)` loads a word from memory, getting the address, b
...
+|===
+
+Therefore:
+
+. _custom-0_ is still available for custom encodings
+. _custom-1/2/3_ are repurposed as *{cheri_base_ext_name}-1/2/3*
+
+NOTE: Standard {cheri_base_ext_name} encodings will _never_ be placed in _custom-0_.
+
+NOTE: Parts of *{cheri_base_ext_name}-1/2/3* may be returned to custom space in the future.
```
arichardson: I guess that will make the naming a bit confusing. Should we reverse the order? I.e. RVY-0 for custom3?

tariqkurd-repo: I thought the simplest thing was to keep the same numbers, that way the mapping is trivial

arichardson: Yeah it just makes it weird if we say there is CUSTOM-0, CUSTOM-1, and RVY-2, RVY-3. What happened to RVY-1? So maybe just allocating the RVY ones in reverse order is easiest? Since we will likely return CUSTOM-1 to custom space? Then we have CUSTOM-0, CUSTOM-1 and RVY-0, RVY-1.

tariqkurd-repo: we can do that but maybe use letters so the numbers aren't reallocated?
custom-3 -> RVY-A?
custom-2 -> RVY-B?
custom-1 -> RVY-C?

I thin it's odd if the numbers change, so maybe this is clearer?

tariqkurd-repo: ARC will give their opinion anyway

arichardson: Yeah numbers or letters are good. 

I was thinking:
custom-3 -> RVY-0?
custom-2 -> RVY-1?
custom-1 -> RVY-2?

But letters is good too. As you say it doens't matter since they will redefine it but it would be good to avoid something that will be confusing.

tariqkurd-repo: yeah - I think remapping the number is confusing.
Or we use better names RVY-MAIN, RVY-YLEN, RVY-SPARE?
But it's hard to know what YLEN and SPARE shoudl really be so maybe A, B, C is better

arichardson (approved):
Let's use A/B/C until ARC assigns final names.

arichardson (approved):

PR #1021, opened 2026-05-11
Use pointers instead of addresses in the CSR classes

arichardson:
Also avoid the awkward "for example jvt, NOTE jvt does not exist" and
use mtvec as the example CSR for extended CSRs.


marnovandermaas (approved):
This looks reasonable to me.

tariqkurd-repo (approved):

PR #1030, opened 2026-05-13
Changed the CSetBoundsImm's (YBNDSWI) immediate value encoding to simplify it

vmurali:
This should have the same hardware circuit as the encoding in the base. The difference is, it uses bit 0 instead of bit 5 to determine if it's an odd or even multiple of 8 in the third case, thus putting all multiples of 8 in order.


arichardson:
As discussed offline: this scheme covers the same output values, but is slightly more complicated to implement in HW since it requires some 3:1 MUX instead of just 2:1.

Analysis can be found here: https://gist.github.com/arichardson/56fbb8fe12b203900806341e265e26c7

PR #1032, opened 2026-05-14
Use the correct YBNDSWI encoding

### LINKED ISSUE #1031: YBNDSWI odd vs even polarity
I think the slightly different encoding has been merged into main. It should have been

Region | Decoded range (bytes) | Immediate decoding condition
-- | -- | --
Byte granular | 1, 2, …​, 255 | imm[8] == 0 and imm[7:0] > 0
8-byte granular | 256, 264, …​, 504 | imm[8] == 1 and imm[7:4] == __0000__ (even multiples of 8); imm[8] == 1 and imm[7:4] == __0001__ (odd multiples of 8)
16-byte granular | 512, 528, …​, 4080 | imm[8] == 1 and imm[7:4] >= 0010
4096 | imm[8:0] == 0

</markdown-accessiblity-table>

arichardson:
While updating the YBNDSWI encoding for the Custom-3 move, I accidentally transcribed the wrong candidate encoding (I used the candidate with inverted polarity of imm[4]). The reason we ultimately chose this scheme as the preferred one is that it has the same gate count as the one I accidentally transcribed, but a slightly simpler encoder (one fewer branch needed in the LLVM/GCC backends).

Fixes: https://github.com/riscv/riscv-cheri/issues/1031

### DIFF
--- a/src/cheri/insns/scbnds_32bit.adoc
+++ b/src/cheri/insns/scbnds_32bit.adoc
@@ -47,7 +47,7 @@ include::malformed_cs1_clear_tag.adoc[]
 <<SCBNDSI>> decodes the 9-bit immediate value `imm[8:0]` to the requested length `result` as follows:
 
 * If `imm[8:0] == 0`, `result` is `4096`.
-* If `imm[8] == 0` and `imm[8:0] > 0`, `result` is `imm[7:0]`.
+* If `imm[8] == 0` and `imm[7:0] != 0`, `result` is `imm[7:0]`.
 * If `imm[8] == 1`, `result` is decoded as follows:
 ** If `imm[7:5] != 0`, `result` is `imm[7:0] << 4`.
 ** If `imm[7:5] == 0`, `result` is `(imm[7:0] << 4) | (imm[4] ? 8 : 256)`.
--- a/src/cheri/insns/scbnds_32bit.adoc
+++ b/src/cheri/insns/scbnds_32bit.adoc
@@ -47,10 +47,10 @@ include::malformed_cs1_clear_tag.adoc[]
 <<SCBNDSI>> decodes the 9-bit immediate value `imm[8:0]` to the requested length `result` as follows:
 
 * If `imm[8:0] == 0`, `result` is `4096`.
-* If `imm[8] == 0` and `imm[7:0] > 0`, `result[7:0] = imm[7:0]` with all other bits set to `0`.
-* If `imm[8] == 1`, `result` is decoded as follows, with all other bits set to `0`:
-** If `imm[7:4] != 0000`, then `result[11:4] = imm[7:0]`.
-** If `imm[7:4] == 0000`, then `result[8] = 1`, `result[7:4] = imm[3:0]`, `result[3] = 1`.
+* If `imm[8] == 0` and `imm[8:0] > 0`, `result` is `imm[7:0]`.
+* If `imm[8] == 1`, `result` is decoded as follows:
+** If `imm[7:5] != 0`, `result` is `imm[7:0] << 4`.
+** If `imm[7:5] == 0`, `result` is `(imm[7:0] << 4) | (imm[4] ? 8 : 256)`.
 
 +
 The resulting logical encodable regions and step sizes are summarized in the table below:
@@ -63,15 +63,13 @@ The resulting logical encodable regions and step sizes are summarized in the tab
 | `1, 2, ..., 255`
 | `imm[8] == 0` and `imm[7:0] > 0`
 
-.2+<| 8-byte granular
-.2+<| `256, 264, ..., 504`
-| `imm[8] == 1` and `imm[7:4] == 0001` (even multiples of 8)
-
-| `imm[8] == 1` and `imm[7:4] == 0000` (odd multiples of 8)
+| 8-byte granular
+| `256, 264, ..., 504`
+| `imm[8:5] == 1000`
 
 .2+<| 16-byte granular
 | `512, 528, ..., 4080`
-| `imm[8] == 1` and `imm[7:4] >= 0010`
+| `imm[8] == 1` and `imm[7:5] != 000`
 
 | `4096`
 | `imm[8:0] == 0`
@@ -80,8 +78,15 @@ The resulting logical encodable regions and step sizes are summarized in the tab
 [NOTE]
 ====
 This immediate encoding scheme achieves coverage of almost all observed immediate values while decoding the 9-bit immediate field without any adders or variable shifts.
-The linear range, stride-16 range, and stride-8 region are decoded via simple multiplexers and bit wiring.
 Mapping `imm == 0` to `4096` reclaims the redundant zero encoding (already covered by `{scbnds_lc} dst, src, x0`) to support highly common page-sized objects.
+
+// The hardware implementation is a simple bit-remapping: +
+// `result[12] = (imm[8:0] == 0)`, +
+// `result[11:9] = imm[8] ? imm[7:5] : 0`, +
+// `result[8] = imm[8] ? (imm[7:5] == 0 || imm[4]) : 0`, +
+// `result[7:4] = imm[8] ? imm[3:0] : imm[7:4]`, +
+// `result[3] = imm[8] ? (imm[7:4] == 1) : imm[3]`, +
+// and `result[2:0] = imm[8] ? 0 : imm[2:0]`.
 ====
 
 Included in::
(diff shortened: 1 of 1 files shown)


src/cheri/insns/scbnds_32bit.adoc
```diff
@@ -47,10 +47,10 @@ include::malformed_cs1_clear_tag.adoc[]
...
 
 * If `imm[8:0] == 0`, `result` is `4096`.
-* If `imm[8] == 0` and `imm[7:0] > 0`, `result[7:0] = imm[7:0]` with all other bits set to `0`.
-* If `imm[8] == 1`, `result` is decoded as follows, with all other bits set to `0`:
-** If `imm[7:4] != 0000`, then `result[11:4] = imm[7:0]`.
-** If `imm[7:4] == 0000`, then `result[8] = 1`, `result[7:4] = imm[3:0]`, `result[3] = 1`.
+* If `imm[8] == 0` and `imm[8:0] > 0`, `result` is `imm[7:0]`.
+* If `imm[8] == 1`, `result` is decoded as follows:
+** If `imm[7:5] != 0`, `result` is `imm[7:0] << 4`.
+** If `imm[7:5] == 0`, `result` is `(imm[7:0] << 4) | (imm[4] ? 8 : 256)`.
```
nwf: That feels a little hinky, having `imm[4]` matter twice like that... maybe phrase this last clause as
```
** If `imm[7:5] == 0`, `result` is `256 | (imm[3:0] << 4) | (imm[4] ? 8 : 0)`.
```
which should be equivalent, since if `imm[4]` is set then the left hand side of the `|`, `r = (imm[7:0] << 4)`, has `r[8]` set and if `imm[4]` is clear then the right hand side of the `|`, `r = (imm[4] ? 8 : 256)` also has `r[8]` set?

nwf: I guess the reason for the spelling as raised is that it reuses the `imm[7:0] << 4` gates from the prior clause (which itself reuses the `imm[7:0]` slice from its prior clause)... maybe just a note that these are equivalent formulae?  Cite the reuse as to why it's bit-fiddled as `imm[7:0] << 4 | (imm[4] ? 8 : 256)` and not "just" `256 | (imm[4:0] << 3)`?

arichardson: I originally wrote it closer to your suggestion, just with `imm[4] <<3` as the last term. But then I decided to shorten it by reusing the full 7:0. Looking at it again I agree that this could make it more confusing. Will change it back to the expanded term

tariqkurd-repo: in any case we need to merge this today and make a new release with the updated version, and tell ARC about the change.

tariqkurd-repo: @rmn30 opinion?

src/cheri/insns/scbnds_32bit.adoc
```diff
@@ -47,10 +47,10 @@ include::malformed_cs1_clear_tag.adoc[]
 <<SCBNDSI>> decodes the 9-bit immediate value `imm[8:0]` to the requested length `result` as follows:
 
 * If `imm[8:0] == 0`, `result` is `4096`.
-* If `imm[8] == 0` and `imm[7:0] > 0`, `result[7:0] = imm[7:0]` with all other bits set to `0`.
-* If `imm[8] == 1`, `result` is decoded as follows, with all other bits set to `0`:
-** If `imm[7:4] != 0000`, then `result[11:4] = imm[7:0]`.
-** If `imm[7:4] == 0000`, then `result[8] = 1`, `result[7:4] = imm[3:0]`, `result[3] = 1`.
+* If `imm[8] == 0` and `imm[8:0] > 0`, `result` is `imm[7:0]`.
```
rmn30:
```suggestion
* If `imm[8] == 0` and `imm[7:0] != 0`, `result` is `imm[7:0]`.
```

rmn30: Doesn't make any functional difference but makes more sense to me.

src/cheri/insns/scbnds_32bit.adoc
rmn30: If the two encodings have equivalent coverage I don't have much of an opinion. I am struggling to get my head around the way they're written here (especially filtered through diff of asciidoc) and will have a think about it, but I certainly haven't done the gate level analysis that Alex has. It might be good if @kunyanliu would take a look. I'm a bit surprised that performance of the compiler backend is a consideration for this...

rmn30: This doesn't seem correct as written.

```
imm[7:5] == 0`, `result` is `(imm[7:0] << 4) | (imm[4] ? 8 : 256)`
```

if imm[7:5] is zero and imm[4] is 1 then this simplifies to `(imm[3:0] << 4) | 8`. Which covers values 8, 24, 40..248 which are all already covered by imm[8] == 0!

nwf: > if imm[7:5] is zero and imm[4] is 1 then this simplifies to `(imm[3:0] << 4) | 8`

Not quite: the "simplification" drops the set `imm[4]` in the left operand of the `<<`, meaning that 256 is missing from the result.  See my "which should be equivalent" above.

arichardson: Does using `256 | (imm[3:0] << 4) | (imm[4] << 3)` for that case sound good?

nwf: I think it maybe be best to have one decision-tree presentation (`if { if { if { expr } else { expr } } else { if { ...`) and another that's meant to emphasize the simplicity of the circuit, in which the reuse of `imm[4]` that felt skeezy is actually just fine?

vmurali:
I think we can change the description to make the decoding algorithm simpler if needed. Can we merge the encoding already though because, I think, the wrong spec was sent out to ARC by mistake?

vmurali (approved):
The `imm[7:4] << 4` is correct but harder to reason about. The encoding is correct, irrespective of whether we have an optimized decoder or a slightly less optimized one. We can merge this and simplify (or add a comment to) the decoding algorithm for readability later, if needed.

rmn30:
FWIW this is a table I made to aid my understanding:

image

Where rows 2 and 3 can be expressed as:
```
(imm[7:0] << 4) | (1 << 8 ) | (imm[4] << 3)
```

src/cheri/insns/scbnds_32bit.adoc
arichardson: I've uncommented the note explaining the bitwise decoding, and slightly simplified the prose. Does that look okay now?

arichardson:
> FWIW this is a table I made to aid my understanding:
> 
> image
> Where rows 2 and 3 can be expressed as:
> 
> ```
> (imm[7:0] << 4) | (1 << 8 ) | (imm[4] << 3)
> ```

Should I add this table?

rmn30:
> Should I add this table?

If you think it's helpful and correct!

src/cheri/insns/scbnds_32bit.adoc
```diff
@@ -80,8 +77,15 @@ The resulting logical encodable regions and step sizes are summarized in the tab
...
-The linear range, stride-16 range, and stride-8 region are decoded via simple multiplexers and bit wiring.
 Mapping `imm == 0` to `4096` reclaims the redundant zero encoding (already covered by `{scbnds_lc} dst, src, x0`) to support highly common page-sized objects.
+
+The hardware implementation is a simple bit-remapping: +
+`result[12] = (imm[8:0] == 0)`, +
+`result[11:9] = imm[8] ? imm[7:5] : 0`, +
+`result[8] = imm[8] ? (imm[7:5] == 0b000 || imm[4]) : 0`, +
+`result[7:4] = imm[8] ? imm[3:0] : imm[7:4]`, +
+`result[3] = imm[8] ? (imm[7:4] == 0b0001) : imm[3]`, +
+and `result[2:0] = imm[8] ? 0 : imm[2:0]`.
```
tariqkurd-repo:
```suggestion
`result[12] = (imm[8:0] == 0);` +
`result[11:9] = imm[8] ? imm[7:5] : 0;` +
`result[8] = imm[8] ? (imm[7:5] == 0b000 || imm[4]) : 0;` +
`result[7:4] = imm[8] ? imm[3:0] : imm[7:4];` +
`result[3] = imm[8] ? (imm[7:4] == 0b0001) : imm[3];` +
`result[2:0] = imm[8] ? 0 : imm[2:0];`.
```

tariqkurd-repo (approved):

rmn30 (approved):

PR #1037, opened 2026-05-26
Fix YAMASK draft opcode allocation

arichardson:
It was accidentally flagged as having a YLEN input even though it is actually only XLEN. Since we encode this in the opcode allocation it ended up in the wrong bucket. This also YMODER causes YMODER to move, but since it's just one instruction and the final opcodes are not completely frozen it probably does not make sense to add a new RESERVED instruction and keep YMODER where it was before.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ DOCS := riscv-privileged riscv-unprivileged riscv-cheri riscv-cheri-full
 
 RELEASE_TYPE ?= draft
 GIT_SHORT_HASH ?= -$(shell git rev-parse --short HEAD || true)
-NEXT_VERSION = v0.9.8.1
+NEXT_VERSION = v0.9.8.2
 ifeq ($(RELEASE_TYPE), draft)
 CHERI_SPEC_VERSION ?= $(NEXT_VERSION)-draft$(GIT_SHORT_HASH)
 else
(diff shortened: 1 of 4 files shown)


arichardson:
3 options here:
- Do what this PR does and update YAMASK+YMODER
- add a reserved encoding so we only update YAMASK and keep YMODER
- Do nothing and just add a comment that this opcode allocation does not match the scheme.

tariqkurd-repo (approved):
let's go with the PR
how did you find this? are more to come?

arichardson:
> let's go with the PR how did you find this? are more to come?

Found while updating my LLVM changes

tariqkurd-repo:
> > let's go with the PR how did you find this? are more to come?
> 
> Found while updating my LLVM changes

what's your confidence level that this is the last such quirk?
and also we need riscv-opcodes updating - can your script generate the output for that?

arichardson:
> > > let's go with the PR how did you find this? are more to come?
> > 
> > 
> > Found while updating my LLVM changes
> 
> what's your confidence level that this is the last such quirk? and also we need riscv-opcodes updating - can your script generate the output for that?

Yes it can, I'll submit a PR this week.

nwf (approved):
LGTM FWIW

PR #1038, opened 2026-05-28
DRET is not in Sdext

tariqkurd-repo:
DRET is not included in Sdext, and is an implementation recommendation only

### DIFF
--- a/src/cheri/debug-integration.adoc
+++ b/src/cheri/debug-integration.adoc
@@ -256,6 +256,8 @@ Upon entry to debug mode, <<ddc>> is saved in <<dddc>>. <<ddc>> is set to a
 
 When debug mode is exited by executing <<DRET_CHERI>>, the hart's <<ddc>> is updated to
 the capability stored in <<dddc>>. A debugger may write <<dddc>> to change the
+When debug mode is exited, the hart's <<ddc>> is updated to
+the capability stored in <<dddc>>. A debugger may write <<dddc>> to change the
 hart's context.
 
 As shown in xref:CSR_exevectors[xrefstyle=short], <<dddc>> is a data pointer,
--- a/src/cheri/debug-integration.adoc
+++ b/src/cheri/debug-integration.adoc
@@ -206,7 +206,7 @@ The <<p_bit>> of this capability is _only_ updated by executing <<MODESW_CAP>> o
 include::img/drootcreg.edn[]
 
 ==== Modified Trap-Return Instruction Behavior
-The `DRET` instruction reads the full YLEN bits of the <<mepc_y>>/<<sepc_y>> register and unseals it prior to exception return if it is a <<sentry_cap>>.
+The <<DRET_CHERI>> instruction reads the full YLEN bits of the <<mepc_y>>/<<sepc_y>> register and unseals it prior to exception return if it is a <<sentry_cap>>.
 
 include::insns/dret.adoc[leveloffset=+1]
 
--- a/src/cheri/insns/dret.adoc
+++ b/src/cheri/insns/dret.adoc
@@ -16,7 +16,7 @@ Description::
 <<DRET_CHERI>> returns from debug mode. It unseals <<dpc_y>> and writes the result into
 <<pcc>>.
 
-NOTE: The <<DRET_CHERI>> instruction is the recommended way to exit debug mode.
+NOTE: The <<DRET_CHERI>> instruction is a documented method of exiting debug mode.
 However, it is a pseudoinstruction to return that technically does not execute
 from the program buffer or memory. It currently does not require the <<pcc>> to
 grant <<asr_perm>> so it never raises an exception.
--- a/src/cheri/insns/dret.adoc
+++ b/src/cheri/insns/dret.adoc
@@ -21,8 +21,12 @@ However, it is a pseudoinstruction to return that technically does not execute
 from the program buffer or memory. It currently does not require the <<pcc>> to
 grant <<asr_perm>> so it never raises an exception.
 
+NOTE: The definition of `DRET` is not part of Sdext, and is a suggested as an implementation choice in the _RISC-V Debug Specification_.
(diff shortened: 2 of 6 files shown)


src/cheri/insns/dret.adoc
JanMatCodasip: `... is the recommended way ...`

Sorry for nit-picking: Could we clarify where this recommendation comes from?

- From this riscv-cheri spec itself?
- From the riscv-debug-spec? (I think that would be untrue statement: The riscv-debug-spec only mentions DRET as a possible implementation choice but does not make an explicit recommendation.)

JanMatCodasip (approved):
Just one optional nit-pick, otherwise LGTM. Thank you.

src/cheri/insns/dret.adoc
```suggestion
NOTE: The <<DRET_CHERI>> instruction is a documented method of exiting debug mode.
```

JanMatCodasip: Also do we actually want to make a recommendation about what the internal implementation should look like?

Instead, could we say this?

- DRET is a possible implementation choice
- The spec assumes that DRET is used, but implementations are free to use any other mechanism to trigger the debug mode exit, as long as it has the same semantics as DRET.

tariqkurd-repo: the CHERI spec doesn't assume use of DRET and doesn't recommend it either - the intention is only to specify it's beahviour if it's implemented

tariqkurd-repo: the same is true of the other standard RISC-V instructions

src/cheri/insns/dret.adoc
```diff
@@ -16,13 +16,17 @@ Description::
 <<DRET_CHERI>> returns from debug mode. It unseals <<dpc_y>> and writes the result into
 <<pcc>>.
```
JanMatCodasip, line 17: IIUC, the actions described here shall always take place on exit from the debug mode - regardless of how the exit is internally implemented and whether DRET is used or not.

Could the text be restructured/rephrased to account for that?

- First state the requirements without making reference to DRET: "When the debug mode is exit, these actions shall take place: <...>."
- Then introduce DRET as a possible implementation choice. Perhaps only as a non-normative note? "One possible documented method (implementation choice) for debug mode exit is DRET instruction, its recommended encoding is <...>, etc."

src/cheri/debug-integration.adoc
JanMatCodasip, line 261: The same situation is here:

I'd suggest to say just `when debug mode is exited`, and drop the reference to DRET because vendors may freely choose to implement the debug mode exit as they like (DRET need not be used at all).

```suggestion
When debug mode is exited, the hart's <<ddc>> is updated to
the capability stored in <<dddc>>. A debugger may write <<dddc>> to change the
hart's context.
```

arichardson (approved):
LGTM with comments addressed

src/cheri/insns/dret.adoc
```diff
@@ -16,13 +16,17 @@ Description::
...
 
-NOTE: The <<DRET_CHERI>> instruction is the recommended way to exit debug mode.
+NOTE: The <<DRET_CHERI>> instruction is a documented method of exiting debug mode.
 However, it is a pseudoinstruction to return that technically does not execute
 from the program buffer or memory. It currently does not require the <<pcc>> to
 grant <<asr_perm>> so it never raises an exception.
```
JanMatCodasip, line 22: There seem to be contradictory requirements:

- A) This note says that DRET does not require the ASR permission in pc.
- B) [In another place in the spec](https://github.com/tariqkurd-repo/riscv-cheri/blob/a5904828e943ec940fd93f3894c2027eee806e2c/src/cheri/riscv-unpriv-integration.adoc?plain=1#L447) it is said that DRET requires ASR in pc.

Which of those is correct, please?

I think that A) is right: exit from debug mode should not require ASR.

arichardson: I think it's fine to say it needs ASR since it's implicitly granted in debug mode

> As a result, the value of [pc](https://riscv.github.io/riscv-cheri/#pcc) is UNSPECIFIED in debug mode according to this specification. The [pc](https://riscv.github.io/riscv-cheri/#pcc) metadata has no architectural effect in debug mode. Therefore [ASR-permission](https://riscv.github.io/riscv-cheri/#asr_perm) is implicitly granted for access to all CSRs for instruction execution.

src/cheri/insns/dret.adoc
tariqkurd-repo: not really - this _is_ the page describing DRET so it only talks about DRET

src/cheri/insns/dret.adoc
tariqkurd-repo: I think this one is ok now with my edit.

src/cheri/insns/dret.adoc
tariqkurd-repo: yes - so that linked reference is wrong - will fix.

src/cheri/insns/dret.adoc
JanMatCodasip: @tariqkurd-repo I'm sorry but I am still not convinced (or missing a piece of information to understand this correctly).

1) I believe the "unseal dpc and write the result into pc" action should **always happen** when exit from debug mode occurs, **regardless** of how it is implemented internally in the given processor. Is that correct understanding? (Note that the use of DRET or different mechanism is just an internal implementation choice,  invisible to the external debugger.) 

2) If yes, then why is that action stated only under the DRET instruction? Shouldn't it instead be a general requirement, and moved outside of the DRET section?

tariqkurd-repo:
It's strange - I can't see this comment on the PR - I've filed https://github.com/riscv/riscv-cheri/issues/1043 to make sure this gets addressed.


JanMatCodasip:
> I've filed [#1043](https://github.com/riscv/riscv-cheri/issues/1043) to make sure this gets addressed.

Thank you.

PR #1044, opened 2026-06-01
Clean up DRET text

### LINKED ISSUE #1043: unsealing DPC on debug resume
 <<pcc>>.

@tariqkurd-repo I'm sorry but I am still not convinced (or missing a piece of information to understand this correctly).
I believe the "unseal dpc and write the result into pc" action should always happen when exit from debug mode occurs, regardless of how it is implemented internally in the given processor. Is that correct understanding? (Note that the use of DRET or different mechanism is just an internal implementation choice, invisible to the external debugger.)
If yes, then why is that action stated only under the DRET instruction? Shouldn't it instead be a general requirement, and moved outside of the DRET section?

tariqkurd-repo:
@JanMatCodasip 

Fix https://github.com/riscv/riscv-cheri/issues/1043

### DIFF
--- a/src/cheri/debug-integration.adoc
+++ b/src/cheri/debug-integration.adoc
@@ -153,10 +153,16 @@ On debug mode entry, <<dpc_y>> is updated with the
 capability in <<pcc>> whose address field is set to the address of the next
 instruction to be executed upon debug mode exit as described in the _RISC-V Debug Specification_.
 
-When leaving debug mode, an unsealed capability value is copied from the value in <<dpc_y>> and written into <<pcc>>.
-A debugger may write <<dpc_y>> to change where the hart resumes and its mode, permissions, sealing or bounds.
+When leaving debug mode, the value in <<dpc_y>> is unsealed if it is a <<sentry_cap>> and is written to <<pcc>>.
+
+NOTE: A debugger may write <<dpc_y>> to change where the hart resumes and its mode, permissions, sealing or bounds.
+ The value saved from the <<pcc>> on debug mode entry cannot be sealed.
 
 The legalization of <<dpc_y>> follows the same rules described for <<mepc_y>>.
+
+One method of resuming execution is <<DRET_CHERI>>.
+
+include::insns/dret.adoc[leveloffset=+1]
 
 [#dscratch0_y,reftext="dscratch0 ({cheri_base_ext_name})"]
 ==== Debug Scratch Register 0 (dscratch0)
@@ -218,11 +224,6 @@ The <<p_bit>> of this capability is _only_ updated by executing <<MODESW_CAP>> o
 
 .Debug root capability register
 include::img/drootcreg.edn[]
-
-==== Modified Trap-Return Instruction Behavior
-The <<DRET_CHERI>> instruction reads the full YLEN bits of the <<mepc_y>>/<<sepc_y>> register and unseals it prior to exception return if it is a <<sentry_cap>>.
-
-include::insns/dret.adoc[leveloffset=+1]
 
 [#section_hybrid_debug_integration]
 === Integrating {cheri_default_ext_name} with Sdext
--- a/src/cheri/debug-integration.adoc
+++ b/src/cheri/debug-integration.adoc
@@ -1,3 +1,4 @@
+<<<<<<< HEAD
 [#section_debug_integration_ext]
 == Integrating {cheri_base_ext_name} with Debug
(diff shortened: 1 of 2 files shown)


arichardson (approved):

src/cheri/insns/dret.adoc
JanMatCodasip: Do we please expect that the DRET's ASR requirement will change? If not, I would suggest to drop the word "currently".

```suggestion
from the program buffer or memory. It does not require the <<pcc>> to
```

src/cheri/debug-integration.adoc
JanMatCodasip: 1) Fix of punctuation (extra `,`).

2) Attempt to simplify the wording (no change in meaning).

```suggestion
When the hart is in debug mode, the _RISC-V Debug Specification_ does not specify
how the PC is updated, and says that PC-relative instructions may be illegal. This concept
```

src/cheri/debug-integration.adoc
```diff
@@ -139,11 +139,17 @@ On debug mode entry, <<dpc_y>> is updated with the
 capability in <<pcc>> whose address field is set to the address of the next
 instruction to be executed upon debug mode exit as described in the _RISC-V Debug Specification_.
```
JanMatCodasip: Wording suggestion to avoid possible confusion: `is set to` --> `equals to`.

"Is set to" could lead the reader to think that the address field gets somehow changed, which is not the case.

```suggestion
On debug mode entry, <<dpc_y>> is updated with the
capability in <<pcc>>, whose address field equals to the address of the next
instruction to be executed upon debug mode exit as described in the _RISC-V Debug Specification_.
```

src/cheri/debug-integration.adoc
```diff
@@ -139,11 +139,17 @@ On debug mode entry, <<dpc_y>> is updated with the
...
-When leaving debug mode, an unsealed capability value is copied from the value in <<dpc_y>> and written into <<pcc>>.
-A debugger may write <<dpc_y>> to change where the hart resumes and its mode, permissions, sealing or bounds.
+When leaving debug mode, the value in <<dpc_y>> is unsealed if it is a <<sentry_cap>> and is written to <<pcc>>.
+
+NOTE: A debugger may write <<dpc_y>> to change where the hart resumes and its mode, permissions, sealing or bounds.
+ The value saved from the <<pcc>> on debug mode entry cannot be sealed.
 
 The legalization of <<dpc_y>> follows the same rules described for <<mepc_y>>.
 
+One method of resuming execution is <<DRET_CHERI>>.
```
JanMatCodasip: 1) Two terms (synonyms) are used in this section: "debug mode exit" and "resuming execution". If you don't mind, I'd keep using just one of them.

2) I would like to stress also here that the use of DRET is purely an implementation choice. Something along these lines:

```suggestion
One possible implementation choice for debug mode exit is DRET.
```

or

```suggestion
Exiting debug mode can be implemented using DRET, but the implementation is free to choose any other method instead.
```

(or similar)

src/cheri/debug-integration.adoc
JanMatCodasip: There seems to be a leftover piece of text (that should have been deleted):

```suggestion
When debug mode is exited, the hart's <<ddc>> is updated to
the capability stored in <<dddc>>. A debugger may write <<dddc>> to change the
hart's context.
```

JanMatCodasip (commented):
@tariqkurd-repo Thank you for the changes, it looks good to me overall.

I have come across few more minor things and posted my suggestions for each. I do not expect to have any further comments on this PR than that.

src/cheri/debug-integration.adoc
```diff
@@ -137,13 +138,21 @@ instruction execution.
 
 On debug mode entry, <<dpc_y>> is updated with the
 capability in <<pcc>> whose address field is set to the address of the next
```
```suggestion
```

src/cheri/insns/dret.adoc
```diff
@@ -13,12 +13,13 @@ Encoding::
 include::wavedrom/dret.adoc[]
 
 Description::
-<<DRET_CHERI>> returns from debug mode. It unseals <<dpc_y>> and writes the result into
-<<pcc>>.
+<<DRET_CHERI>> returns from debug mode.
+It unseals <<dpc_y>> and writes the result into <<pcc>>.
```
```suggestion
It unseals <<dpc_y>> if it is sealed as a <<sentry_cap>> and writes the result into <<pcc>>.
```

tariqkurd-repo:
replaced by #1045

PR #1045, opened 2026-06-02
DRET clarification

### LINKED ISSUE #1043: unsealing DPC on debug resume

tariqkurd-repo:
replacement for #1044 which got messed up on a rebase
@JanMatCodasip can you review this one please instead? let's work from here.

fix #1043

### DIFF
--- a/src/cheri/debug-integration.adoc
+++ b/src/cheri/debug-integration.adoc
@@ -153,10 +153,16 @@ On debug mode entry, <<dpc_y>> is updated with the
 capability in <<pcc>> whose address field is set to the address of the next
 instruction to be executed upon debug mode exit as described in the _RISC-V Debug Specification_.
 
-When leaving debug mode, an unsealed capability value is copied from the value in <<dpc_y>> and written into <<pcc>>.
-A debugger may write <<dpc_y>> to change where the hart resumes and its mode, permissions, sealing or bounds.
+When leaving debug mode, the value in <<dpc_y>> is unsealed if it is a <<sentry_cap>> and is written to <<pcc>>.
+
+NOTE: A debugger may write <<dpc_y>> to change where the hart resumes and its mode, permissions, sealing or bounds.
+ The value saved from the <<pcc>> on debug mode entry cannot be sealed.
 
 The legalization of <<dpc_y>> follows the same rules described for <<mepc_y>>.
+
+One method of resuming execution is <<DRET_CHERI>>.
+
+include::insns/dret.adoc[leveloffset=+1]
 
 [#dscratch0_y,reftext="dscratch0 ({cheri_base_ext_name})"]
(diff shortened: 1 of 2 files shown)


src/cheri/debug-integration.adoc
```diff
@@ -151,13 +152,21 @@ instruction execution.
 
 On debug mode entry, <<dpc_y>> is updated with the
 capability in <<pcc>> whose address field is set to the address of the next
```
JanMatCodasip: There is a leftover piece of text that should be removed - probably a typo during conflict resolution.

```suggestion
```

JanMatCodasip (approved):
The changes look good to me, thank you. 

I have only found one typo.

arichardson (changes requested):
Did you somehow change the line endings for the file? There seems to be a huge diff that is hard to review.

tariqkurd-repo:
> Did you somehow change the line endings for the file? There seems to be a huge diff that is hard to review.

that final suggestion which was deleting 2 lines seems to have broken it.

nwf:
Seems sensible to me, but please do check line endings.

src/cheri/debug-integration.adoc
```diff
@@ -1,278 +1,280 @@
...
+
+The <<dpc>> register is extended to hold a capability.
+
+{TAG_RESET_DCSR}
+
+.Debug program counter capability
+include::img/dpccreg.edn[]
+
+Upon entry to debug mode, the _RISC-V Debug Specification_, does not specify how to
+When the hart is in debug mode, the _RISC-V Debug Specification_ does not specify
```
JanMatCodasip: I have just noticed one more typo - a leftover part of sentence that should be removed:

```suggestion
When the hart is in debug mode, the _RISC-V Debug Specification_ does not specify
```

src/cheri/debug-integration.adoc
```suggestion
```

tariqkurd-repo:
It seems that the diff view is broken - it's repeating lines which aren't repeated and then removing them causes all hell to break loose. Merging as it is and then let's re-read the debug chapter 😢

JanMatCodasip:
> It seems that the diff view is broken - it's repeating lines which aren't repeated and then removing them causes all hell to break loose. Merging as it is and then let's re-read the debug chapter 😢

Ok, no problem. I will re-check the text once more after the merge.

PR #1047, opened 2026-06-03
PMA clarification

tariqkurd-repo:
PMA fault didn't specify that the fault is only on stores not loads, so improved the title.
Also removed a double negative.

### DIFF
--- a/src/cheri/insns/atomic_exceptions.adoc
+++ b/src/cheri/insns/atomic_exceptions.adoc
@@ -22,7 +22,7 @@ Exceptions::
 ifdef::cap_atomic[]
 Store/AMO access fault exception when the effective address is not aligned to YLEN/8.
 +
-Store/AMO access fault if the stored {ctag} is set to one and the PMA is _CHERI {ctag_title} Fault_.
+Store/AMO access fault if the stored {ctag} is set to one and the PMA is _CHERI {ctag_title} Store Fault_.
 endif::[]
 ifndef::cap_atomic[]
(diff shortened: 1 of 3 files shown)


arichardson (commented):
Seems fine to clarify this. I think it should be implicitly obvious that faulting means there can't be any tags on load. There is no way to fault on load if there are no tags in the PMA region.

tariqkurd-repo:
> Seems fine to clarify this. I think it should be implicitly obvious that faulting means there can't be any tags on load. There is no way to fault on load if there are no tags in the PMA region.

yep that's true! there can't be any tags to be loaded - so the new name is  bit clearer

tariqkurd-repo:
One consideration is what happens if PMAs become programmable such as in Smpmpmt: https://lists.riscv.org/g/sig-soft-cpu/attachment/75/0/smpmpmt-v0p1-20240626.pdf

So it would be possible to have tag supporting memory which has access dynamically restricted.
In that case we're not going require checking loaded tags, and so _store fault_ is a better name.

arichardson:
> One consideration is what happens if PMAs become programmable such as in Smpmpmt: https://lists.riscv.org/g/sig-soft-cpu/attachment/75/0/smpmpmt-v0p1-20240626.pdf
> 
> So it would be possible to have tag supporting memory which has access dynamically restricted. In that case we're not going require checking loaded tags, and so _store fault_ is a better name.

Sure for dynamically programmable that makes more sense. I'm happy to merge this if everyone else agrees.

PR #1050, opened 2026-06-08
Move encoded AP field into RV64Y chapter

tariqkurd-repo:
Preparing the RV32Y encoding format for submission.
The main objectives are:

1. To remove the duplication with the RV64Y format chapter
2. To use the compressed AP field for RV64Y to allow more permissions to be added in future (e.g. seal/unseal) without eating up too much metadata.
3. To prevent Zyhybrid and Zylevels1 redefining the entire permissions table and all the permissions rules - it needs to be a simple addition - so that it's a scalable approach. We can't have one of these tables for every possible combination, it shouldn't be part of the specification

Essentially most of the text simply wasn't needed - and duplicated text is always rejected during ARC review.


src/cheri/riscv-hybrid-integration.adoc
```diff
@@ -123,6 +123,8 @@ Capabilities lacking <<x_perm>> may not have a defined <<p_bit>> field in their
 
 When executing <<CLRPERM>>, if <<x_perm>> is removed while the <<p_bit>> is set to one, and the capability encoding still permits representing the <<p_bit>>, then the <<p_bit>> must be set to zero.
 
+Infinite capability encodings are required to set the {cheri_mode_name} to {INT_MODE_VALUE} for compatibility.
```
arichardson, line 126: Should probably be a root-rx-cap instead?

nwf:
> To prevent Zyhybrid and Zylevels1 redefining the entire permissions table and all the permissions rules - it needs to be a simple addition

This is why I originally wrote the permissions dependency rules as a scattered definition iterated to fixed point.  That gives a compositional style that seems sufficiently expressive for all the cases we have encountered to date.

The compression is only sensibly defined once all dependency rules (and any opt-in rules like the "university" RV32 encoding added) were known.  It's not generally possible to just mix in new values to the encoding table as you go adding extensions one by one.  At the very least, they need to not numerically conflict.

> We can't have one of these tables for every possible combination

At a glance it looks like zylevels1 now _has_ such a table when it didn't before?

arichardson (changes requested):

tariqkurd-repo:
> >
> At a glance it looks like zylevels1 now _has_ such a table when it didn't before?

Zylevels1 certainly adds more permissions into the table - the question is how to document that. Previously this was all in the RV32Y uni format chapter, but as we're now considering compressed encodings for RV64Y as well it needs to be in the RV64Y chapter.

Essentially each extension which adds permissions need to say _how_ they fit. Zylevels1 does, so needs to add the AP-field encodings. Previously all such tables were only in the RV32Y chapter.

All such tables could go into the RV64Y chapter, but the way it normally works is that an extension lists everything that it needs in a self contained way. For example, I now have the P-bit (formerly M-bit) mapping in the Zyhybrid chapter only as nothing else is affected. I think this composition will be more popular with the reviewers.

tariqkurd-repo:
A big worry about this approach is that we'll disallow a set of permissions which CHERIBSD / Linux etc. rely upon, but we can't simply keep adding permission bits forever. 

We have R, W, X, C, LM, LG, GL, SL, P (mode) already as separate bits and want to implement Zyseal (eventually) giving two more permission bits meaning we've used up 11-bits.

Another option might be to exploit the reserved states better, like LM being reserved if C and R are not set.

Essentially somehow we need to reduce the combinations in a scalable way.

jrtc27:
> 2\. To use the compressed AP field for RV64Y to allow more permissions to be added in future (e.g. seal/unseal) without eating up too much metadata.

This is not the time to be making such drastic and unproven changes. If this is truly necessary for ratification then it will have to be implemented and tested at huge scale, including for things like JITs, before we can consider changing the architecture. There is massive potential for making a spec that cannot be used for its intended purpose otherwise.

tariqkurd-repo:
I agree completely. How can we fit the seal/unseal permission into RV64Y without burning two more permission bits? Is there any sensible way of doing it? I'd like to confirm how we can move to Zyseal style sentries in the future before freezing the spec.

nwf:
> I agree completely. How can we fit the seal/unseal permission into RV64Y without burning two more permission bits?

AFAICT, the (only) way that would be compatible with how we have been told to do this, with full encoding syntax given in the base, would be analogous to what the base text says now about two reserved permission bits ("LG-permission if Zylevels1 is implemented; reserved 1 otherwise.")  You'd have the base continue to "know" about Zyseal ahead of time and reserve some number of compressed/encoded values for its use.  If you are OK with multiple root capabilities (as in the ISAv9 experimental appendix and CHERIoT v1), that can be as few as three values:

- all of which would not grant any of `R`, `W`, or `X` permissions (and so cannot confer any of the other base -- `C`, `LM`, `ASR` -- or Zylevels1 -- `SL`, `LG` -- permissions)

- one to serve as granting both `SE` and `US` permissions, one to grant just the one, and one to grant just the other.

If you do still want a single root cap, you probably want to have at least four new values.  In addition to the `SE US`, `SE`, and `US` values above, you probably also want `R W X C LM ASR SL LG SE US` (all permissions / omnipotent / infinite / ...) in addition to the existing `R W X C LM ASR SL LG` (all memory permissions) value.

> Is there any sensible way of doing it?

If you consider the above sensible, I guess so?  It's not, IMHO, meaningfully _worse_ than what the current text is doing.

> I'd like to confirm how we can move to Zyseal style sentries in the future before freezing the spec.

As far as I can tell:

* Moving to a compressed permission representation will almost surely be a new base architecture.
  * I think you could phrase it as an extension that un-reserved a reserved bit in the capability encoding somewhere, set that to the non-default value, and redefined the permission field, but: why.

* If that base did not itself reserve values for Zyseal's permissions (as outlined above), adding them seems like it would require one of...
  * an extension that un-reserved some unused values from the base as outlined above, or
  * an extension that un-reserved an unused bit in the encoding and either replaced or augmented the permission field's values, or
  * another new base architecture.

Separately, theorist nit-picking: Zyseal and sentries are almost completely unrelated.

Zyseal describes capability-mediated (that is, non-ambient) (un)sealing of capabilities, that's it.  Whether or not some sealing types are interpreted as sentries -- and, in principle, whether or not those sealing types are ever made available to Zyseal's `SE`-bearing root capabilities -- is outside its wheelhouse.  The difference between a Zysentry and a Zyseal platform, as far as sentries are concerned and if sentry sealing type(s) _are_ available to Zyseal's roots, is largely one of *software* architecture: with Zyseal, all would-be sentry construction sites must hold a capability to be so.

It happens that Zysentry is more general than just "sentries", and its `YSENTRY` instruction allows ambient sealing of *non-sentry* (`X`-less) capabilities meant to be consumed by superset unsealing.  In a Zyseal-instead-of-Zysentry platform, any would-be sealer again needs to hold a capability to do so.  (Importantly for software architecture, the Zyseal case doesn't require that the would-be *unsealer* also control the *allocator* for objects to which sealed handles are made.)

tariqkurd-repo:
partly replaced by #1071

PR #1061, opened 2026-06-11
Simplify encoding name

### LINKED ISSUE #1055: Simplify RV32Y/RV64Y encoding format names
RV64LYmw14rc1ps -> RV64LYA
RV32LYmw10rc1pc -> RV32LYA

keep it simple and so that people know what they are called

tariqkurd-repo:
fix #1055 
also needs parameter chapter rewrite which is part of #1071 - but this one is the actual spec change


tariqkurd-repo:
Any opinions on this one? The feedback is that we need a _much_ simpler name, so why not go with 'A'?

PR #1062, opened 2026-06-11
improve  cap dirty tracking wording

### LINKED ISSUE #1054: is "cap dirty" the correct name for PTE.YD
The normal dirty bit says that a page is different to the backing store
The capability dirty bit does _not_ necessarily mean that a capability is present as it might have been overwritten, and so does not have the same meaning.

It's more like a "capability seen" bit

tariqkurd-repo:
fix #1054

### DIFF
--- a/src/cheri/cheri-pte-ext.adoc
+++ b/src/cheri/cheri-pte-ext.adoc
@@ -95,7 +95,7 @@ When _pte_._{pte_cap_w_name_lc}_ is clear, capability stores or AMOs where the t
 When _pte_._{pte_cap_w_name_lc}_ is set, capability stores to the virtual page are permitted.
 In addition, the _pte_._{pte_cap_dirty_name_lc}_ bit indicates that a capability was stored to the virtual page since the last time the _pte_._{pte_cap_dirty_name_lc}_ bit was cleared.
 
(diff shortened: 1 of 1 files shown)


src/cheri/cheri-pte-ext.adoc
```diff
@@ -95,7 +95,7 @@ When _pte_._{pte_cap_w_name_lc}_ is clear, capability stores or AMOs where the t
 When _pte_._{pte_cap_w_name_lc}_ is set, capability stores to the virtual page are permitted.
 In addition, the _pte_._{pte_cap_dirty_name_lc}_ bit indicates that a capability was stored to the virtual page since the last time the _pte_._{pte_cap_dirty_name_lc}_ bit was cleared.
 
-NOTE: This is akin to the _pte_._d_ bit indicating that a store occurred through this PTE since the last time _pte_._d_ was cleared.
+NOTE: This is equivalent to the behavior of the _pte_._d_ bit except that _pte_._{pte_cap_dirty_name_lc}_ tracks stores of valid capabilities instead of _all_ stores.
```
```suggestion
NOTE: This is equivalent to the behavior of the _pte_._d_ bit except that _pte_._{pte_cap_dirty_name_lc}_ only tracks stores of valid capabilities instead of _all_ stores.
```

arichardson (approved):
Doesn't really change the meaning but the language is simpler so maybe easier to parse. Seems fine to me.

PR #1064, opened 2026-06-15
Clarify that illegal AP states are reserved for RV64Y

tariqkurd-repo:
If we clarify that such states are reserved, then in future states like C without R or W can be assigned a new meaning, such as SE/US permission.

### DIFF
--- a/src/cheri/riscv-unpriv-integration.adoc
+++ b/src/cheri/riscv-unpriv-integration.adoc
@@ -477,7 +477,7 @@ endif::[]
 Capability encoding formats may impose additional constraints to reduce the number of bits necessary to represent permissions.
 
 .<<CLRPERM>> base rules for <<rv64y_cap_description,{rv64y_uni_base_name}>>
-[float="center",align="center",cols="2,2,4",options="header"]
+[#base_clrperm_rule_summary,float="center",align="center",cols="2,2,4",options="header"]
 |===
 | <<CLRPERM>> Rule                          | Permission   | Valid only if
 | [[perm_req:base:c:r-or-w,base-1]]base-1   | <<c_perm>>   | <<r_perm>> or <<w_perm>>
--- a/src/cheri/riscv-unpriv-integration.adoc
+++ b/src/cheri/riscv-unpriv-integration.adoc
@@ -477,7 +477,7 @@ endif::[]
 Capability encoding formats may impose additional constraints to reduce the number of bits necessary to represent permissions.
 
 
-.Base permission rules for <<rv64y_cap_description,{rv64y_uni_base_name}>>
+.Base permission rules Architectural Permissions
 [#base_clrperm_rule_summary,float="center",align="center",cols="2,2,4",options="header"]
 |===
 | <<CLRPERM>> Rule                          | Permission   | Valid only if
--- a/src/cheri/riscv-unpriv-integration.adoc
+++ b/src/cheri/riscv-unpriv-integration.adoc
@@ -477,7 +477,7 @@ endif::[]
 Capability encoding formats may impose additional constraints to reduce the number of bits necessary to represent permissions.
 
 
-.Base permission rules Architectural Permissions
+.<<CLRPERM>> base rules
 [#base_clrperm_rule_summary,float="center",align="center",cols="2,2,4",options="header"]
 |===
 | <<CLRPERM>> Rule                          | Permission   | Valid only if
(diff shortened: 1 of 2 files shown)


src/cheri/rvy64-encoding.adoc
```diff
@@ -77,10 +77,10 @@ endif::[]
 The permissions field is {cap_rv64_perms_width} bits wide and is encoded using one bit per architectural permission as shown in xref:cap_perms_encoding64[xrefstyle=short].
 A permission is granted if its corresponding bit, and those of any dependent permissions, are set;
 otherwise, the capability does not grant that permission.
-Certain combinations of permissions are impractical. For example, <<c_perm>> is superfluous when the capability does not grant either <<r_perm>> or <<w_perm>>.
```
arichardson, line 80: I would keep these two sentences since it explains why we have the rules.

src/cheri/rvy64-encoding.adoc
```diff
@@ -77,10 +77,10 @@ endif::[]
...
-Therefore, it is only legal to encode a subset of all combinations, but this capability encoding does not make use of this redundancy as there are sufficient reserved bits.
 
-NOTE: In the future, if there are no further remaining reserved bits, but extensions need to allocate further capability metadata bits, it will be possible to use the redundancy to create a future version of this encoding that is fully _compatible for all software_ but with observable differences in the in-memory representation of capabilities.
+The rules in <<base_clrperm_rule_summary>> are enforced and any invalid states are _reserved_ in the <<AP-field>> encoding.
```
```suggestion
Certain combinations of permissions are impractical.
For example, <<c_perm>> is superfluous when the capability does not grant either <<r_perm>> or <<w_perm>>
Therefore, it is only legal to encode a subset of all combinations and any invalid states are _reserved_ in the <<AP-field>> encoding.
```

How about keeping some of the existing text and just clarifying the reserved states?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -476,8 +476,9 @@ endif::[]
 
 Capability encoding formats may impose additional constraints to reduce the number of bits necessary to represent permissions.
 
-.<<CLRPERM>> base rules for <<rv64y_cap_description,{rv64y_uni_base_name}>>
-[float="center",align="center",cols="2,2,4",options="header"]
+
+.Base permission rules Architectural Permissions
```
arichardson: Not sure about the header, maybe just the following?
```suggestion
.<<CLRPERM>> base rules
```

arichardson (commented):
Generally looking good just a few suggestions

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: annoyingly there's a bug in asciidoctor which doen't allow me to keep \<\<CLRPERM>> in the title since adding a label on the table so it can be xref'd. I'll try again....

src/cheri/rvy64-encoding.adoc
tariqkurd-repo: suggestions merged

src/cheri/rvy64-encoding.adoc
```diff
@@ -77,10 +77,12 @@ endif::[]
 The permissions field is {cap_rv64_perms_width} bits wide and is encoded using one bit per architectural permission as shown in xref:cap_perms_encoding64[xrefstyle=short].
 A permission is granted if its corresponding bit, and those of any dependent permissions, are set;
 otherwise, the capability does not grant that permission.
-Certain combinations of permissions are impractical. For example, <<c_perm>> is superfluous when the capability does not grant either <<r_perm>> or <<w_perm>>.
-Therefore, it is only legal to encode a subset of all combinations, but this capability encoding does not make use of this redundancy as there are sufficient reserved bits.
 
-NOTE: In the future, if there are no further remaining reserved bits, but extensions need to allocate further capability metadata bits, it will be possible to use the redundancy to create a future version of this encoding that is fully _compatible for all software_ but with observable differences in the in-memory representation of capabilities.
+Certain combinations of permissions are impractical.
+For example, <<c_perm>> is superfluous when the capability does not grant either <<r_perm>> or <<w_perm>>
```
```suggestion
For example, <<c_perm>> is superfluous when the capability does not grant either <<r_perm>> or <<w_perm>>.
```

src/cheri/rvy64-encoding.adoc
```diff
@@ -77,10 +77,12 @@ endif::[]
...
+Therefore, it is only legal to encode a subset of all combinations and any invalid states are _reserved_ in the <<AP-field>> encoding.
+
+NOTE: Reserved states can be used in future to allow different formats to be encoded within the <<AP-field>>.
```
```suggestion
NOTE: Reserved states can be used in the future to allow additional permissions to be encoded within the <<AP-field>>.
```

arichardson (approved):

PR #1066, opened 2026-06-15
Move YSENTRY into the base and rename it

### LINKED ISSUE #1067: Remove the name Sentry
The name `sentry` is unpopular with ARC, so let's try `secure entry point` which is a better description.
#1066 

I've renamed YSENTRY to YUSEAL - for Unauthorised SEAL

tariqkurd-repo:
Addressing ARC feedback:

fix #1067 

We must put a sentry option into the base
The name sentry is very unpopular with ARC - so I've gone for "sealed entry point capability"

Therefore this PR:

1. Merges YSEAL and YSENTRY, calling the combined instruction YSEAL
2. YSEAL with rs1=x0 seals without authority and type 1 only if type 1 is available (Uni formats only - and will be disabled by S-mode CSR bit in future).
3. YSEAL with rs1 != x0 seals with authority as before (behaviour unchanged from CHERIoT if type 1 is not available)
4. Put the updated YSEAL into the base
5. Removes the word "sentry" except as a note. The usage in Zysentry is very simple anyway as it doesn't distinguish between any sentry types or any other sealed capabilities. I've gone for `YUSEAL` for Unauthorised SEAL. I didn't want to use `YSEAL` to avoid conflicting with Zyseal, which would have been confusing.

JALR behaviour is simplified to remove forwards/backwards sentries - the full text is still there ifdef'd out.
I confirmed that it's not a problem to add new JALR behaviour later when new sealed cap / sentry types are added later by CHERIoT.

### DIFF
--- a/src/cheri/app-hybrid-usage.adoc
+++ b/src/cheri/app-hybrid-usage.adoc
@@ -179,7 +179,7 @@ J -> J: Access runtime data using\nexplicit capabilities
 J -> J: Execute ""MODESW_INT"" (Switches to {cheri_int_mode_name_nofmt})
 
 == Calling Runtime Helpers ==
-J -> R: Call helper (Secure Entry Point Capability) (Switches to {cheri_cap_mode_name_nofmt})
+J -> R: Call helper (Sealed Entry Point Capability) (Switches to {cheri_cap_mode_name_nofmt})
 R -> R: Execute safe runtime code
 R -> J: Return to JIT code (Switches to {cheri_int_mode_name_nofmt})
 
--- a/src/cheri/app-hybrid-usage.adoc
+++ b/src/cheri/app-hybrid-usage.adoc
@@ -156,7 +156,7 @@ The main difference is that the generated code (running in {cheri_int_mode_name}
 * **Accessing C Runtime Data:** The code can access data in the surrounding C++ runtime using explicit capability-based load/store operations.
   At the same time, it uses standard integer addresses for its own heap via <<ddc>>.
   This can be achieved by executing <<MODESW_CAP>> before the memory access and <<MODESW_INT>> to return back to address interpretation.
-* **Calling Runtime Helpers:** Runtime helper functions are given to the code as sentry capabilities.
+* **Calling Runtime Helpers:** Runtime helper functions are given to the code as a <<sentry_cap>>.
    When the code calls these helpers, the processor automatically switches back to {cheri_cap_mode_name}.
   This lets the runtime run with full safety before returning to the JIT code.
 +
@@ -179,7 +179,7 @@ J -> J: Access runtime data using\nexplicit capabilities
 J -> J: Execute ""MODESW_INT"" (Switches to {cheri_int_mode_name_nofmt})
 
 == Calling Runtime Helpers ==
-J -> R: Call helper (Sentry Capability) (Switches to {cheri_cap_mode_name_nofmt})
+J -> R: Call helper (Secure Entry Point Capability) (Switches to {cheri_cap_mode_name_nofmt})
 R -> R: Execute safe runtime code
 R -> J: Return to JIT code (Switches to {cheri_int_mode_name_nofmt})
 
--- a/src/cheri/insns/dret.adoc
+++ b/src/cheri/insns/dret.adoc
@@ -14,7 +14,7 @@ include::wavedrom/dret.adoc[]
 
 Description::
 <<DRET_CHERI>> returns from debug mode.
-It unseals <<dpc_y>> if it is sealed as a <<sentry_cap>> and writes the result into <<pcc>>.
+It unseals <<dpc_y>> if it is sealed and writes the result into <<pcc>>.
 It also restores <<ddc>> from <<dddc>>.
 
 NOTE: The <<DRET_CHERI>> instruction is a documented method of exiting debug mode.
--- a/src/cheri/insns/dret.adoc
+++ b/src/cheri/insns/dret.adoc
@@ -14,7 +14,7 @@ include::wavedrom/dret.adoc[]
 
 Description::
 <<DRET_CHERI>> returns from debug mode.
-It unseals <<dpc_y>> if it is sealed and writes the result into <<pcc>>.
+It unseals <<dpc_y>> if it is a <<sentry_cap>> and writes the result into <<pcc>>.
 It also restores <<ddc>> from <<dddc>>.
 
 NOTE: The <<DRET_CHERI>> instruction is a documented method of exiting debug mode.
--- a/src/cheri/insns/jalr_32bit.adoc
+++ b/src/cheri/insns/jalr_32bit.adoc
@@ -44,6 +44,8 @@ In this case the specialized operation becomes:
  . The <<pcc>> of the next instruction is written to `{cd}`.
  . Jump to the target <<pcc>>.
 ====
++
+NOTE: The standard software calling convention uses 'x1' as the link register.
 endif::[]
 ifndef::generic_JALR[]
 // simple description with only unrestricted sentries
@@ -54,14 +56,11 @@ Indirect jump to the target capability in `{cs1}` with an address offset.
 . The target address is obtained by adding the sign-extended 12-bit I-immediate to `{cs1}.address`, then setting the least-significant bit of the result to zero.
 . Unseal the target <<pcc>> if it is a <<sentry_cap>>, `{cs1}.address[0]` is zero, and the I-immediate is zero.
 . Set the address of the target <<pcc>> to the target address using the semantics of the <<SCADDR>> instruction.
-. The <<pcc>> of the next instruction is sealed and written to `{cd}`.
+. The <<pcc>> of the next instruction is sealed as a <<sentry_cap>> and written to `{cd}`.
 . Jump to the target <<pcc>>.
-+
 endif::[]
 +
 NOTE: A future extension may raise an exception on the <<JALR_CHERI>> instruction itself if the target <<pcc>> will raise a {cheri_excep_name_pc} at the target.
-+
-NOTE: The standard software calling convention uses 'x1' as the link register.
 
 Included in::
 <<rvy_i_mod_insn_table>>
--- a/src/cheri/insns/jalr_32bit.adoc
+++ b/src/cheri/insns/jalr_32bit.adoc
@@ -14,12 +14,13 @@ include::wavedrom/ct-unconditional-2.adoc[]
 
 include::base_isa_extension.adoc[]
 
+ifdef::generic_JALR[]
+//full description including backwards/forwards-edge sentries
 Description::
 Indirect jump to the target capability in `{cs1}` with an address offset.
 +
 NOTE: The description below is generic, and implementations which need control-flow-integrity will include at least one extension which defines one or more <<sentry_cap, sentry capabilities>>.
  All such extensions are required to detail the specialized behavior of JALR to remove any ambiguity.
- For example, see <<section_rvy_sentry_insn_ext>>.
 +
 . `{cs1}` is written to the target <<pcc>>.
 . If `{cd}`≠x0, `{cs1}` is _sealed_ and `{cs1}` is not a _forward-edge_ <<sentry_cap>>, then the behavior is _reserved_^1^.
@@ -43,6 +44,20 @@ In this case the specialized operation becomes:
  . The <<pcc>> of the next instruction is written to `{cd}`.
  . Jump to the target <<pcc>>.
 ====
+endif::[]
+ifndef::generic_JALR[]
+//simple descrption with unrestricted sentries
+Description::
+Indirect jump to the target capability in `{cs1}` with an address offset.
++
+. `{cs1}` is written to the target <<pcc>>.
+. The target address is obtained by adding the sign-extended 12-bit I-immediate to `{cs1}.address`, then setting the least-significant bit of the result to zero.
+. Unseal the target <<pcc>> if it is sealed, `{cs1}.address[0]` is zero, and the I-immediate is zero.
+. Set the address of the target <<pcc>> to the target address using the semantics of the <<SCADDR>> instruction.
+. The <<pcc>> of the next instruction is sealed and written to `{cd}`.
+. Jump to the target <<pcc>>.
++
+endif::[]
 +
 NOTE: A future extension may raise an exception on the <<JALR_CHERI>> instruction itself if the target <<pcc>> will raise a {cheri_excep_name_pc} at the target.
 +
--- a/src/cheri/insns/jalr_32bit.adoc
+++ b/src/cheri/insns/jalr_32bit.adoc
@@ -46,13 +46,13 @@ In this case the specialized operation becomes:
 ====
 endif::[]
 ifndef::generic_JALR[]
-//simple descrption with unrestricted sentries
+// simple description with only unrestricted sentries
 Description::
 Indirect jump to the target capability in `{cs1}` with an address offset.
 +
 . `{cs1}` is written to the target <<pcc>>.
 . The target address is obtained by adding the sign-extended 12-bit I-immediate to `{cs1}.address`, then setting the least-significant bit of the result to zero.
-. Unseal the target <<pcc>> if it is sealed, `{cs1}.address[0]` is zero, and the I-immediate is zero.
+. Unseal the target <<pcc>> if it is a <<sentry_cap>>, `{cs1}.address[0]` is zero, and the I-immediate is zero.
 . Set the address of the target <<pcc>> to the target address using the semantics of the <<SCADDR>> instruction.
 . The <<pcc>> of the next instruction is sealed and written to `{cd}`.
 . Jump to the target <<pcc>>.
--- a/src/cheri/insns/sentry_32bit.adoc
+++ b/src/cheri/insns/sentry_32bit.adoc
@@ -16,12 +16,12 @@ Mnemonic::
 Encoding::
 include::wavedrom/sentry.adoc[]
 
-NOTE: `{cs1}{ne}x0` is currently reserved.
+NOTE: `{cs1}{ne}x0` is currently _reserved_.
 
 Description::
 Copy `{cs2}` to `{cd}`.
 +
-Seal the capability in `{cd}` by setting `{cd}.ct=1` if `{cs1}=x0`.
+Seal the capability in `{cd}` by setting `{cd}.ct=1`.
 +
 Set `{cd}.tag=0` if `{cs2}` is sealed.
 +
@@ -52,7 +52,11 @@ Encoding::
 {SENTRY} is a pseudoinstruction for <<YSEAL>> with `{cs1}=x0`.
 
 Description::
-Seal the capability in `{cs2}` as a sealed entry point capability and write the result to `{cd}`.
+Copy `{cs2}` to `{cd}`.
++
+Seal the capability in `{cd}` by setting `{cd}.ct=1`.
++
+Set `{cd}.tag=0` if `{cs2}` is sealed.
 
 Included in::
 <<rvy_insn_table>>
--- a/src/cheri/insns/sentry_32bit.adoc
+++ b/src/cheri/insns/sentry_32bit.adoc
@@ -1,5 +1,5 @@
-[#SENTRY,reftext="{SENTRY}"]
-==== {SENTRY}
+[#YSEAL,reftext="{YSEAL}"]
+==== {YSEAL}
 
 include::new_encoding_note.adoc[]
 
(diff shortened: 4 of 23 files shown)


jrtc27 (changes requested):
Marking as request changes until I have time to write a proper response

src/cheri/insns/dret.adoc
```diff
@@ -14,7 +14,7 @@ include::wavedrom/dret.adoc[]
 
 Description::
 <<DRET_CHERI>> returns from debug mode.
-It unseals <<dpc_y>> if it is sealed as a <<sentry_cap>> and writes the result into <<pcc>>.
+It unseals <<dpc_y>> if it is sealed and writes the result into <<pcc>>.
```
nwf: Why?  Unsealing a non-sentry into PCC is surprising behavior and at least some of those are going to lack `X` permission and so just fault differently anyway, but now in a way that weakens the CHERI capability security model.

src/cheri/insns/sentry_32bit.adoc
```diff
@@ -19,14 +19,16 @@ include::wavedrom/sentry.adoc[]
...
 Copy `{cs1}` to `{cd}`.
 +
-Set the capability type (<<sec_cap_type>>) of `{cd}` to the value `1`, representing a sentry type.
-+
-NOTE: {SENTRY} does not require any permission to seal, and so is considered <<sec_cap_type_ambient,ambient>>.
+Seal the capability in `{cd}` by setting `{cd}.ct=1`.
 +
 Set `{cd}.tag=0` if `{cs1}` is sealed.
++
+NOTE: This is an unauthorized sealing instruction of type 1, future extensions may require authority to be able to perform the sealing action of different <<CT-field>> values.
```
nwf: Please don't use "unauthorized" in an affirmative sense; we use that to talk about why we're throwing traps.  Something more like "ambient", "unrestricted", or something else.

src/cheri/insns/ysunseal_32bit.adoc
```diff
@@ -52,6 +52,7 @@ and so software should not assume `{cs1}==0` to be a pseudo-instruction for {cta
 // as appropriate for their intended uses.
 // =====
 
+ifndef::cheri_ratification_v1_only[]
```
nwf: Removing superset unsealing removes the sole mechanism in the existing base spec for making secure handles that doesn't require allocating executable memory?  That seems like a terrible plan.

src/cheri/insns/ysunseal_32bit.adoc
```diff
@@ -60,8 +61,7 @@ Specifically, a software component can:
 
 1. allocate the memory for these objects from a region of address space,
 
-2. render capabilities to these objects opaque by sealing
-   (with, for example, the <<SENTRY>> instruction, if present),
+2. render capabilities to these objects opaque by sealing with <<SENTRY>>,
```
nwf: Or any future extension that adds sealing mechanisms.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -374,6 +370,57 @@ Capability encoding formats will explicitly state whether these are relevant to
 They are not used by <<rv64y_cap_description>>, but are by the CHERIoT formats.
 ====
 
+endif::[]
+ifdef::cheri_ratification_v1_only[]
+[#sec_cap_type, reftext="CT-field"]
+==== Capability Type (CT)
```
nwf: Please just fold any clarifications made herein back into the earlier text, or offer this as an example of a particularly simple instance of the general thing, not as replacement text.

nwf (changes requested):
Please don't.  Please.  We spent so long getting this nicely factored and usable across our different bases and extensions and platforms...

David is out of the office, so I cannot confirm, but I suspect that this landing will mean that SCI votes against ratification and encourages others to do the same.

src/cheri/insns/sentry_32bit.adoc
tariqkurd-repo: `unrestricted` could be good

src/cheri/insns/ysunseal_32bit.adoc
tariqkurd-repo: there's no removal of superset unsealing - these are just notes - which make more sense when the cheriot chapters are included

src/cheri/insns/dret.adoc
tariqkurd-repo: This is one place where the behavior needs to be specified once multiple sealed types are defined.
The basic spec only has one sealed type - and it needs to be as simple as possible to get through review.
More types can be added later.

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: The text has to be very simple to get through review.
I don't want to remove the more general text - but there's a question about how it's used.
I think that any extension which adds more capability types will need to include the "seal with CT=x" ambient sealing etc. etc. text - that's all valid. All I've done is keep a subset for review _now_.

nwf:
I am not convinced that this is the only way forward, and I sincerely hope it isn't because it threatens to undo some of the modularity and extensibility of the base spec, which is there not just for theoretical reasons but because we have worked examples and a decade of research behind it and know that we will also want to have future standards built on it!

tariqkurd-repo:
> I am not convinced that this is the only way forward, and I sincerely hope it isn't because it threatens to undo some of the modularity and extensibility of the base spec, which is there not just for theoretical reasons but because we have worked examples and a decade of research behind it and know that we will also want to have future standards built on it!

Nothing in this PR affects CHERIoT directly - it does diverge unfortunately - but it's not trying to actually change CHERIoT at all. The longer term plan is to still move to Zyseal for RV64Y but only once there has been time to properly prototype it, and for it to have been taken fully into consideration. The big ticket item for RV64Y is the RVY version of RVA23, and we're close. Once we've achieved that then we'll be taken seriously in the apps space, I hope, and we can look seriously at improvements like implementing Zyseal. Until we have that profile complete we're still seen as experimental and we need to change that mindset by finishing it.

src/cheri/insns/sentry_32bit.adoc
tariqkurd-repo: I've merged the behaviour with YSEAL. Please have another look.

tariqkurd-repo:
> I am not convinced that this is the only way forward, and I sincerely hope it isn't because it threatens to undo some of the modularity and extensibility of the base spec, which is there not just for theoretical reasons but because we have worked examples and a decade of research behind it and know that we will also want to have future standards built on it!

None of the work wil be lost and it will all be built upon - the key thing is to stage the features going into the spec - and to get them through the ARC review. We've put too much in and need to pare it back for the first version. The end goal is the same, it's just how we get there.

src/cheri/insns/yseal_32bit.adoc
```diff
@@ -31,7 +24,12 @@ Copy `{cs2}` into `{cd}`, and then...
 
 [NOTE]
 =====
-<<YSEAL>> uses the (in-bounds) _addresss_ of the authority in `{cs1}` as the
+If `{cs1}=x0` and <<CT-field>>=1 is not supported by the capability encoding format, then `{cs1}` is interpreted as a NULL capability causing `{cd}.tag=0`.
```
nwf: `x0` is _always_ interpreted as a null cap; this isn't what you mean to be saying.

I understand the desire to make `YSENTRY` take up opcode space used by `YSEAL`, but please don't phrase it this way.  Just define Zyseal's `YSEAL` as being defined only for `cs1` not `x0`.

src/cheri/insns/ysunseal_32bit.adoc
```diff
@@ -61,7 +61,7 @@ Specifically, a software component can:
 
 1. allocate the memory for these objects from a region of address space,
 
-2. render capabilities to these objects opaque by sealing with <<SENTRY>>,
+2. render capabilities to these objects opaque by sealing with <<YSEAL>> with <<CT-field>>=1,
```
nwf: This doesn't need to specify the `CT-field` value; superset unsealing will work for any sealed subset of the unsealing authority.

src/cheri/rvy-cheriot-isa-unpriv.adoc
```diff
@@ -36,7 +36,7 @@ we have not yet found a compelling reason to formally specify this composition.
 [NOTE]
 =====
 While {cheriot_unpriv_ext_name} is nominally compatible with
-<<SENTRY>>,
+<<CT-field>>=1,
```
nwf, line 39: Previously, Xycheriot has used CT=1 for its own ends; this conflation of ambient sentries and CT=1 is at odds with that.

src/cheri/rvy-cheriot-isa-unpriv.adoc
```diff
@@ -23,7 +23,7 @@ counterparts found in the {cheriot_priv_ext_name} extension.
 both the <<section_zylevels1>> and <<section_zyseal>> extensions.
 The present specification presumes the *absence*
 of both the <<section_cheri_hybrid_ext,{cheri_default_ext_name}>> and
-the <<SENTRY>> instruction.
+no support for <<CT-field>>=1 as an argument passed to the <<YSEAL>> instruction.
```
nwf: CT is not an argument passed to YSEAL, which takes an authority cap and an input unsealed cap.  "`CT=1` as an argument" might mean an untagged 1 as authority or, as apparently done, `x0` being interpreted as such.

src/cheri/insns/yseal_32bit.adoc
nwf: Separately, would `li a0, 0; yseal a0, a0, a1` with valid unsealed cap in `a1` result in `a0` having an untagged result or a tagged CT=1 cap?

src/cheri/rvy-cheriot-isa-unpriv.adoc
tariqkurd-repo: yes - that does complicate things - maybe we can reserve CT=1 for this behaviour only?

src/cheri/insns/yseal_32bit.adoc
tariqkurd-repo: a0!=x0 so the normal (old) YSEAL rules apply - I've removed the comment about NULL cap and will just define it if rs1!=x0 as requested (commit coming soon)

src/cheri/rvy-cheriot-isa-unpriv.adoc
tariqkurd-repo: alternatively CHERIoT is a new base so is free to reallocate the CT numbers, but it would be convenient if CT=1 meant the same thing universally.

src/cheri/insns/dret.adoc
arichardson: I think we could resolve the "reviewer-perceived redundancy" by just saying 
```suggestion
It unseals <<dpc_y>> if it is a <<sentry_cap>> and writes the result into <<pcc>>.
```

That is future-proof and avoid "duplication".

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -43,6 +44,20 @@ In this case the specialized operation becomes:
  . The <<pcc>> of the next instruction is written to `{cd}`.
  . Jump to the target <<pcc>>.
 ====
+endif::[]
+ifndef::generic_JALR[]
+//simple descrption with unrestricted sentries
```
```suggestion
// simple description with only unrestricted sentries
```

src/cheri/insns/jalr_32bit.adoc
```diff
@@ -43,6 +44,20 @@ In this case the specialized operation becomes:
...
+Description::
+Indirect jump to the target capability in `{cs1}` with an address offset.
++
+. `{cs1}` is written to the target <<pcc>>.
+. The target address is obtained by adding the sign-extended 12-bit I-immediate to `{cs1}.address`, then setting the least-significant bit of the result to zero.
+. Unseal the target <<pcc>> if it is sealed, `{cs1}.address[0]` is zero, and the I-immediate is zero.
```
```suggestion
. Unseal the target <<pcc>> if it is a <<sentry_cap>>, `{cs1}.address[0]` is zero, and the I-immediate is zero.
```

src/cheri/insns/ysunseal_32bit.adoc
arichardson: I think we should keep this note. I can see the rationale for omitting the second note until after 1.0.

src/cheri/scripts/rvy_instruction_encodings.py
```diff
@@ -563,7 +563,6 @@ def next_r2type(name: str, rd: str, rs1: str, **kwargs):
         next_r2type("YTAGR", rs1="{cs1}", rd="rd"),
         next_r2type("YTYPER", rs1="{cs1}", rd="rd"),
         next_r2type("YAMASK", rs1="rs1", rd="rd"),
-        next_r2type("YSENTRY", rs1="{cs1}", rd="{cd}", ext="Zysentry"),
```
arichardson, line 566: This will affect the next instruction. But it looks like we don't have another one after (since YMODER uses different operand types) so it's ok.

src/cheri/app-hybrid-usage.adoc
```diff
@@ -179,7 +179,7 @@ J -> J: Access runtime data using\nexplicit capabilities
 J -> J: Execute ""MODESW_INT"" (Switches to {cheri_int_mode_name_nofmt})
 
 == Calling Runtime Helpers ==
-J -> R: Call helper (Sentry Capability) (Switches to {cheri_cap_mode_name_nofmt})
+J -> R: Call helper (Secure Entry Point Capability) (Switches to {cheri_cap_mode_name_nofmt})
```
arichardson:
```suggestion
J -> R: Call helper (Sealed Entry Point Capability) (Switches to {cheri_cap_mode_name_nofmt})
```

src/cheri/riscv-priv-integration.adoc
```diff
@@ -716,9 +716,9 @@ Root Executable Capability::
 Capability Type (CT)::
  See the unprivileged manual for the definition of capability type values.
 
-[#sentry_cap,reftext="sentry capability"]
-Sentry Capabilities::
- See the unprivileged manual for the definition of sentry capabilities.
+[#sentry_cap,reftext="secure entrypont capability"]
+Secure Entry Point Capabilities::
```
```suggestion
Sealed Entry Point Capabilities::
```

src/cheri/riscv-priv-integration.adoc
```diff
@@ -716,9 +716,9 @@ Root Executable Capability::
 Capability Type (CT)::
  See the unprivileged manual for the definition of capability type values.
 
-[#sentry_cap,reftext="sentry capability"]
-Sentry Capabilities::
- See the unprivileged manual for the definition of sentry capabilities.
+[#sentry_cap,reftext="secure entrypont capability"]
```
```suggestion
[#sentry_cap,reftext="sealed entrypoint capability"]
```

src/cheri/riscv-priv-integration.adoc
arichardson: Should be consistent about entrypoint vs entry point

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -330,7 +330,7 @@ Ambient sealing type::
 Some capability types are said to be _ambiently available_ (or just _ambient_)
 if they do not require specific authority to seal a capability with that type.
 Therefore sentry types can be declared _ambient_.
-For example, if <<section_rvy_sentry_insn_ext>> is available on a given platform,
+For example, if <<CT-field>>=1 is available on a given platform,
 the type with which it seals capabilities is considered _ambiently available_.
```
arichardson, line 354: I would just ifdef this for example sentence

arichardson: (the new wording does not make sense)

src/cheri/insns/sentry_32bit.adoc
```diff
@@ -8,26 +8,31 @@ NOTE: *CHERI v9 Note:* this instruction was called CSEALENTRY.
 endif::[]
 
 Synopsis::
-{SENTRY_DESC}
+{YSEAL_DESC}
 
 Mnemonic::
-`{sentry_lc} {cd}, {cs1}`
+`{yseal_lc} {cd}, {cs1}, {cs2}`
```
arichardson, line 14: Could you add {SENTRY} as an assembler alias for cs2=x0?

src/cheri/rvy64-encoding.adoc
```diff
@@ -118,14 +117,13 @@ The <<sec_cap_type>> field is as specified in <<table_capability_types>>.
 The capabilities of this chapter define a 1-bit field for <<sec_cap_type>> values;
 this field directly encodes the values `0` and `1`:
 
-* The encoded value `0` indicates an unsealed type.
-* If {rvy_sentry_insn_ext_name} is supported, then `1` encodes an unrestricted <<sentry_cap>>.
-* Otherwise, the value `1` is _reserved_.
+* The encoded value `0` indicates an _unsealed_ type.
+* The encoded value `1` indicates an _sealed_ type which can unsealed by <<JALR_CHERI>>.
```
arichardson, line 121:
```suggestion
* The encoded value `1` indicates a <<sentry_cap>> which can unsealed by <<JALR_CHERI>>.
```

arichardson:
I pushed some fixes but this still needs more discussion and review

src/cheri/insns/ysunseal_32bit.adoc
arichardson: Restored in latest version

src/cheri/riscv-unpriv-integration.adoc
arichardson: I re-merged the sections with ifdefs, hope it's better now

arichardson:
This should effectively move us close the previous proposal of always requiring an authorizing capability for sealing. Just with a special case of NULL gives an unrestricted sentry if this type is available. And to keep the spec friendly for external review the extended semantics are deferred. It's not ideal but hopefully a somewhat reasonable compromise.

jrtc27:
> This should effectively move us close the previous proposal of always requiring an authorizing capability for sealing. Just with a special case of NULL gives an unrestricted sentry if this type is available. And to keep the spec friendly for external review the extended semantics are deferred. It's not ideal but hopefully a somewhat reasonable compromise.

Why are we no longer doing the thing where the encoding with x0 is YSENTRY (under whatever name ARC won't protest), and non-x0 is reserved, such that Zysealwithotypes can use it? I don't see why we would want to abandon that and make the x0 explicit in the ISA, it's just confusing?

jrtc27:
> > This should effectively move us close the previous proposal of always requiring an authorizing capability for sealing. Just with a special case of NULL gives an unrestricted sentry if this type is available. And to keep the spec friendly for external review the extended semantics are deferred. It's not ideal but hopefully a somewhat reasonable compromise.
> 
> Why are we no longer doing the thing where the encoding with x0 is YSENTRY (under whatever name ARC won't protest), and non-x0 is reserved, such that Zysealwithotypes can use it? I don't see why we would want to abandon that and make the x0 explicit in the ISA, it's just confusing?

(Doing it based on encoding rather than NULL is explicit in the ISA so (a) you convey intentionality (b) you know in decode which case it is, rather than being value-dependent for non-x0)

arichardson:
Yes sorry by NULL I meant x0. This is what the PR does.

tariqkurd-repo:
I think this is the best we can do - I've proofread the "for v1 ratification" version it and I think it looks good.
I've made the description of YSENTRY match YSEAL (with rs1=0) so that the behaviour is explicit.
Otherwise I've just made typographical fixes.

src/cheri/insns/sentry_32bit.adoc
```diff
@@ -8,26 +8,55 @@ NOTE: *CHERI v9 Note:* this instruction was called CSEALENTRY.
...
+Mnemonic::
+`{sentry_lc} {cd}, {cs2}`
+
+Encoding::
+{SENTRY} is a pseudoinstruction for <<YSEAL>> with `{cs1}=x0`.
+
+Description::
+Copy `{cs2}` to `{cd}`.
++
+Seal the capability in `{cd}` by setting `{cd}.ct=1`.
```
tariqkurd-repo:
```suggestion
Seal the capability in `{cd}` as a <<sentry_cap>> by setting `{cd}.ct=1`.
```

tariqkurd-repo:
This is good enough for now - we need to get into the next round of review urgently.

kunyanliu:
Based on prior MSRC review findings, we believe that permitting ambient sealing could introduce security vulnerabilities, including the potential for arbitrary code execution. Given that one of Microsoft’s primary interests in adopting CHERI technology is for high-assurance systems such as roots of trust, Microsoft’s position is that ambient sentries should be excluded from both the base specification and the Linux profile.

tariqkurd-repo:
> Based on prior MSRC review findings, we believe that permitting ambient sealing could introduce security vulnerabilities, including the potential for arbitrary code execution. Given that one of Microsoft’s primary interests in adopting CHERI technology is for high-assurance systems such as roots of trust, Microsoft’s position is that ambient sentries should be excluded from both the base specification and the Linux profile.

Hi Kunyan, yes - we've discussed this extensively.
We have a plan to migrate to authorized sealing / unsealing and to disable the ambient sealing in the base by adding privileged state.
Note that this doesn't affect CHERIoT at all, which will continue as it is.

Issue #1068, opened 2026-06-15
"Access Memory" abstract commands - semantics of write operations


JanMatCodasip:
If [Access Memory abstract commands](https://docs.riscv.org/reference/debug/debug_module.html#ac-accessmemory) (defined by the debug spec) are supported on a CHERI hart, what should be the effect of the write operation on the capability tag?

Should the [semantics of the store instructions](https://riscv.github.io/riscv-cheri/snapshot/riscv-cheri-full/#section_cheri_int_load_store_insns) from the CHERI spec apply to this case? Quoted below:

> All store instructions, except for the RVY SY, always write zero to the capability tag or capability tags associated with the memory locations that are written to. Therefore, misaligned stores may clear up to two associated capability tag bits.

tariqkurd-repo:
I expect the tag to be set to zero.
Would you ever want to send a valid capability via an abtract command?
The point of drootc is to be able to get hold of a root-cap in debug mode. which other capabilities can be derived from.

JanMatCodasip:
@tariqkurd-repo Thank you for your reply. I have the same opinion but wanted to confirm this and also write down to the spec.

My proposal is under: https://github.com/riscv/riscv-cheri/pull/1069

resistor:
The hypothetical use case for writing valid capabilities via abstract command is for register save/restore to host. OpenOCD, for example, assumes that it can save device GPRs to the host, executing multiple abstract commands (note that the `dscratch` registers are not preserved across ACs), and then restoring the GPRs from the host.

Absent that ability (and I agree that writing valid capabilities via AC is undesirable otherwise), we need to consider alternative ways to allow the debugger to save and restore GPRs, such as providing additional scratch registers which are guaranteed to be preserved across ACs.

JanMatCodasip:
Hi @resistor, thank you for your reply. This thread is about memory abstract commands and there is a second one about register abstract commands. Let's please [move the discussion there](https://github.com/riscv/riscv-cheri/issues/1039#issuecomment-4797348169) to keep the conversations easier to follow for newcomers.

tariqkurd-repo:
fixed by #1069

Issue #1039, opened 2026-05-29
"Access Register" abstract commands - what shall be supported on RVY?


JanMatCodasip:
I would like to clarify here (and later add to the CHERI spec) how [Access Register abstract commands](https://docs.riscv.org/reference/debug/debug_module.html#abstractcommands) should behave on RVY harts. 

Specifically, I'd like to clarify: 

- What read/write sizes should be supported.
- The exact semantics of register writes through abstract commands.

---

In RVY, the size of GPRs (and certain CSRs) is doubled to from XLEN to YLEN.

The RISC-V debug spec says [in 3.7.1.1.](https://docs.riscv.org/reference/debug/debug_module.html#ac-accessregister):

> "If aarsize specifies a size larger than the register’s actual size, then the
> access must fail. If a register is accessible, then reads of aarsize less than
> or equal to the register’s actual size must be supported. Writing less than
> the full register may be supported, but what happens to the high bits in
> that case is UNSPECIFIED."

When I apply this requirement from the debug spec word-for-word to the YLEN-wide GPRs in RVY, my interpretation would be:

- Abstract read of YLEN bits **must** be supported (reading both address and metadata).
  - In that case the sequence called "Read the high MXLEN bits into data0-data1" ([link](https://riscv.github.io/riscv-cheri/snapshot/riscv-cheri-full/#_integrating_rvy_with_sdext)) is redundant because both address and metadata from the GPR can be simply read by an abstract command.
- Abstract read of XLEN bits **must** be supported (reading address only).

- Abstract write of YLEN bits **must** be supported (address + metadata). 
  - But what should be the semantics of such a write that touches both the address and the metadata at the same time?
- Abstract write of XLEN bits is **optional** (writing address only).
  - If it is supported, what should be the semantics? Should it use the semantics of YADDRW?
  - However, the CHERI spec seems to assume that this XLEN-wide write is mandatory (not optional) because it says: "The low MXLEN bits of a capability are read and written using normal Access Register abstract commands." (in [this note](https://riscv.github.io/riscv-cheri/snapshot/riscv-cheri-full/#_integrating_rvy_with_sdext)).
 

Do you please interpret the debug spec the same way? If not, what Access Registers abstract commands should in your opinion be mandatory, optional or forbidden? 

What should be the semantics of the Abstract register writes?

Thank you.

JanMatCodasip:
[...]

PR #1069, opened 2026-06-16
Debug integration review

JanMatCodasip:
1) <del>Specify that Access Memory abstract commands clear the capability tags, as discussed in
https://github.com/riscv/riscv-cheri/issues/1068.</del>

2) <del>Prepare a placeholder section for Access Register abstract commands, which should be filled in after we clarify this issue: https://github.com/riscv/riscv-cheri/issues/1039</del>

Overall review of the debug integration chapter.

### DIFF
--- a/src/cheri/debug-integration.adoc
+++ b/src/cheri/debug-integration.adoc
@@ -121,7 +121,7 @@ All other bits in that `data` register are UNSPECIFIED.
 When a capability is written it must be a <<cap_subset>> of the capability in <<drootc>>.
 If not then the destination {ctag} is set to zero.
 
-NOTE: For the capability subset: `cap2` is the capability being written and `cap` 1 is the value of <<drootc>>.
+NOTE: For the capability subset: `cap2` is the capability being written and `cap1` is the value of <<drootc>>.
 
 NOTE: It is valid for root capabilities not to cover the entire possible address space.
  In this case the debugger cannot claim access to more memory than is permitted by <<drootc>>.
--- a/src/cheri/debug-integration.adoc
+++ b/src/cheri/debug-integration.adoc
@@ -8,7 +8,7 @@ include::trigger-integration.adoc[]
 
 {cheri_base_ext_name} harts must provide a means for accessing capabilities in registers and in memory via the external debugger.
 
-This can be achieved by at least one of:
+This is achieved by supporting at least one of:
 
 . Implementing a program buffer configured as specified below.
 . Supporting full capability access for both Access Register and Access Memory abstract commands (`aarsize=7` and `aamsize=7`).
--- a/src/cheri/debug-integration.adoc
+++ b/src/cheri/debug-integration.adoc
@@ -93,9 +93,13 @@ no way to discover the former property.
 
 ==== Access Register Command
 
-The _Access Register_ abstract command has optional support for accessing YLEN registers.
+When an YLEN-wide register is accessible via _Access Register_ abstract command, then:
 
-If `aarsize` matches YLEN then all YLEN-bits should be accessible via the command.
+- Reads of YLEN bits or any smaller amounts of bits must be supported.
+- Writes of YLEN bits must be supported.
+- Writes of smaller amounts of bits than YLEN may optionally be supported but it is UNSPECIFIED what happens to the higher bits.
+
+NOTE: The debugger can change the low XLEN bits of a YLEN wide register by performing a read-modify-write sequence.
 This requires enough `data<n>` registers to be supported to cover the entire data width (2 for {cheri_base32_ext_name} or 4 for {cheri_base64_ext_name}).
 
 Implementations can choose to also support accessing {ctag}s as part of the access register command.
--- a/src/cheri/debug-integration.adoc
+++ b/src/cheri/debug-integration.adoc
@@ -111,6 +111,8 @@ The `data<n>` register mapping for full capability accesses is as shown in <<deb
 ^1^ arg1 represents an address so never needs to exceed DXLEN.
  Each `data<n>` register is 32-bit so a maximum of 2 are required.
 
+^2^ arg2 is not used for abstract commands, and is included for consistency with the equivalent table in the _RISC-V Debug Specification_
+
 Bit 0 of `data5` or `data10` contain the {ctag} which is either written to, or read from, the target register.
 All other bits in that `data` register are UNSPECIFIED.
 
--- a/src/cheri/debug-integration.adoc
+++ b/src/cheri/debug-integration.adoc
@@ -6,7 +6,7 @@ include::trigger-integration.adoc[]
 
 === Integrating {cheri_base_ext_name} with Sdext
 
-{cheri_base_ext_name} harts should be able to access capabilities in registers and in memory via the external debugger.
+{cheri_base_ext_name} harts must provide a means for accessing capabilities in registers and in memory via the external debugger.
 
 This can be achieved by at least one of:
 
--- a/src/cheri/debug-integration.adoc
+++ b/src/cheri/debug-integration.adoc
@@ -91,24 +91,15 @@ no way to discover the former property.
   setting `transfer` to 1 and `write` to 0).
 ====
 
-==== Access Register Command
+==== Abstract Commands
 
-When an YLEN-wide register is accessible via _Access Register_ abstract command, then:
+Abstract commands are optionally extended to access full capabilities.
 
-- Reads of YLEN bits or any smaller amounts of bits must be supported.
-- Writes of YLEN bits must be supported.
-- Writes of smaller amounts of bits than YLEN may optionally be supported but it is UNSPECIFIED what happens to the higher bits.
+`aarsize/aarmsize=7` represent full capability access.
 
-NOTE: The debugger can change the low XLEN bits of a YLEN wide register by performing a read-modify-write sequence.
-This requires enough `data<n>` registers to be supported to cover the entire data width (2 for {cheri_base32_ext_name} or 4 for {cheri_base64_ext_name}).
+The `data<n>` register mapping for full capability accesses is as shown in <<debug_data_aarsize_7>>.
 
-Implementations can choose to also support accessing {ctag}s as part of the access register command.
-
-`aarsize=7` is defined to represent full capability access.
-
-The `data<n>` register mapping is as shown in <<debug_data_aarsize_7>>.
-
-.Data register mapping when `aarsize=7`
+.Data register mapping when `aarsize/aamsize=7`
 [#debug_data_aarsize_7]
 [width="100%",options=header,cols="1,1,1,1,1"]
 |=====
@@ -123,21 +114,34 @@ The `data<n>` register mapping is as shown in <<debug_data_aarsize_7>>.
 Bit 0 of `data5` or `data10` contain the {ctag} which is either written to, or read from, the target register.
 All other bits in that `data` register are UNSPECIFIED.
 
+==== Access Register Command
+
+When a YLEN-wide register is accessible via an _Access Register_ abstract command, then:
+
+- Reads of YLEN bits or any smaller amounts of bits must be supported.
+- Writes of YLEN bits must be supported.
+- Writes of smaller amounts of bits than YLEN may optionally be supported but it is UNSPECIFIED what happens to the higher bits.
+
+NOTE: The debugger can change the low XLEN bits of a YLEN wide register by performing a read-modify-write sequence.
+This requires enough `data<n>` registers to be supported to cover the entire data width (2 for {cheri_base32_ext_name} or 4 for {cheri_base64_ext_name}).
+
+Implementations can choose to also support accessing {ctag}s as part of the access register command by supporting `aarsize=7`.
+
 When a capability is written it must be a <<cap_subset>> of the capability in <<drootc>>.
 If not then the destination {ctag} is set to zero.
+
+If `aarsize{ne}7` then {ctag}s are always read and written as zero.
 
 NOTE: For the capability subset: `cap2` is the capability being written and `cap1` is the value of <<drootc>>.
 
 NOTE: It is valid for root capabilities not to cover the entire possible address space.
- In this case the debugger cannot claim access to more memory than is permitted by <<drootc>>.
-
-NOTE: If `aarsize{ne}7` then {ctag}s are always read and written as zero.
+ In this case the debugger cannot claim access to more memory than is permitted by the current value of <<drootc>>.
 
 ==== Access Memory Command
 
-Implementations can choose to support accessing full capabilities, including the {ctag}s, via the the _Access Memory_ abstract command. Full capability access is denoted by `aamsize=7` and uses the `data<n>` register assignments specified in <<debug_data_aarsize_7>>.
+Implementations can choose to support accessing full capabilities, including the {ctag}s, via the _Access Memory_ abstract command. Full capability access is denoted by `aamsize=7` and uses the `data<n>` register assignments specified in <<debug_data_aarsize_7>>.
 
-NOTE: If `aamsize{ne}7` then {ctag}s are always read and written as zero.
+If `aamsize{ne}7` then {ctag}s are always read and written as zero.
 
 ==== Core Debug Registers
 
--- a/src/cheri/debug-integration.adoc
+++ b/src/cheri/debug-integration.adoc
@@ -85,6 +85,48 @@ no way to discover the former property.
   setting `transfer` to 1 and `write` to 0).
 ====
 
+==== Access Register Command
+
+The _Access Register_ abstract command has optional support for accessing YLEN registers.
+
+If `aarsize` matches YLEN then all YLEN-bits should be accessible via the command.
+This requires enough `data<n>` registers to be supported to cover the entire data width (2 for {cheri_base32_ext_name} or 4 for {cheri_base64_ext_name}).
+
+Implementations can choose to also support accessing {ctag}s as part of the access register command.
+
+`aarsize=7` is defined to represent full capability access.
+
+The `data<n>` register mapping is as shown in <<debug_data_aarsize_7>>.
+
+.Data register mapping when `aarsize=7`
+[#debug_data_aarsize_7]
+[width="100%",options=header,cols="1,1,1,1,1"]
+|=====
+|ISA                    | arg0/return value| arg1^1^     | arg2        | {ctag}
+|{cheri_base32_ext_name}| data0,data1      | data2       | data3,data4 | data5
+|{cheri_base64_ext_name}| data0-data3      | data4,data5 | data6-data9 | data10
+|=====
+
+^1^ arg1 represents an address so never needs to exceed DXLEN.
+ Each `data<n>` register is 32-bit so a maximum of 2 is required.
+
+Bit 0 of `data5` or `data10` contain the {ctag} which is either written to, or read from, the target register.
+All other bits in that `data` register are UNSPECIFIED.
+
+When a capability is written it must be a <<cap_subset>> of the capability in <<drootc>>.
+If not then the destination {ctag} is set to zero.
+
+NOTE: For the capability subset: `cap2` is the capability being written and `cap` 1 is the value of <<drootc>>.
+
+NOTE: It is valid for root capabilities not to cover the entire possible address space.
+ In this case the debugger cannot claim access to more memory than is permitted by <<drootc>>.
+
+==== Access Memory Command
+
+The _Access Memory_ abstract command does not support accessing {ctag}s.
+
+NOTE: A future version of this specification may supporting accessing {ctag}s in memory by permitted `aarsize=7`.
+
 ==== Core Debug Registers
 
 {cheri_base_ext_name} widens debug CSRs that are designated to hold addresses so that they can hold capabilities.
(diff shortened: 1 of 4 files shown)


tariqkurd-repo:
I have:

1. Made root capabilities _nominally_ cover all memory
2. Added aarsize=7 as an option for capability access in registers and memory via abstract commands
3. Prevented a capaiblity write from the debugger exceeding the cap in drootc

src/cheri/debug-integration.adoc
```diff
@@ -85,6 +90,50 @@ no way to discover the former property.
...
+^1^ arg1 represents an address so never needs to exceed DXLEN.
+ Each `data<n>` register is 32-bit so a maximum of 2 is required.
+
+Bit 0 of `data5` or `data10` contain the {ctag} which is either written to, or read from, the target register.
+All other bits in that `data` register are UNSPECIFIED.
+
+When a capability is written it must be a <<cap_subset>> of the capability in <<drootc>>.
+If not then the destination {ctag} is set to zero.
+
+NOTE: For the capability subset: `cap2` is the capability being written and `cap` 1 is the value of <<drootc>>.
```
```suggestion
NOTE: For the capability subset: `cap2` is the capability being written and `cap1` is the value of <<drootc>>.
```

src/cheri/debug-integration.adoc
```diff
@@ -8,6 +8,11 @@ include::trigger-integration.adoc[]
 
 {cheri_base_ext_name} harts should be able to access capabilities in registers and in memory via the external debugger.
```
JanMatCodasip: If this makes sense to you:

- `should -> must` - making it very clear that this is a requirement
- It is the external debugger who accesses the capabilities, and the harts only provide a way how to do it.

```suggestion
{cheri_base_ext_name} harts must provide a means how for accessing capabilities in registers and in memory via the external debugger.
```

src/cheri/debug-integration.adoc
```diff
@@ -8,6 +8,11 @@ include::trigger-integration.adoc[]
...
 
+This can be achieved by at least one of:
```
JanMatCodasip: How strong do we want this requirement to be?

- `can` - > `should`?
- `can` -> `shall`?
- ... ?

src/cheri/debug-integration.adoc
```diff
@@ -8,6 +8,11 @@ include::trigger-integration.adoc[]
...
+
+. Implementing a program buffer configured as specified below.
+. Supporting `aarsize=7` for abstract commands which access registers and memory, or both.
```
JanMatCodasip: - Is the second bullet point meant to say that aarsize=7/aamsize=7 must be supported for **both** Access Register and Access Memory abstract commands? If so, I would rephrase it to make it clearer.
- Does the final "or both" mean that the implementers are free to implement both the bullet points at the same time? If so, I would drop it because it is IMO redundant (we already said "at least one of") and not fully clear what it refers to.


```suggestion
This can be achieved by at least one of:

. Implementing a program buffer configured as specified below.
. Supporting full capability access for both Access Register and Access Memory abstract commands (`aarsize=7` and `aamsize=7`).
```

src/cheri/debug-integration.adoc
JanMatCodasip:
```suggestion
* `hartinfo.datasize` is at least 1 and `hartinfo.dataaccess` is 0.
```

JanMatCodasip:
```suggestion
* `abstractcs.progbufsize` is at least 4 if `dmstatus.impebreak` is 1, or at least 5 if `dmstatus.impebreak` is 0.
```

JanMatCodasip: I would suggest to now drop the sentence: "These requirements may be relaxed if ...". That's because the preceding text now clearly says that the implementation should contain at least one of: abstract access for capabilities or a sufficiently configured program buffer.

```suggestion
These requirements allow a debugger to read and write capabilities in integer
registers without disturbing other registers.
```

JanMatCodasip, line 69: Depending on the outcome of the discussion under https://github.com/riscv/riscv-cheri/issues/1039, we may drop some of the sequences because they may become redundant and get replaced by the abstract commands.

src/cheri/debug-integration.adoc
```diff
@@ -85,6 +90,50 @@ no way to discover the former property.
...
+NOTE: For the capability subset: `cap2` is the capability being written and `cap1` is the value of <<drootc>>.
+
+NOTE: It is valid for root capabilities not to cover the entire possible address space.
+ In this case the debugger cannot claim access to more memory than is permitted by <<drootc>>.
+
+NOTE: Any register write with `aarsize{ne}7` causes the {ctag} (if present) to be set to zero.
+
+==== Access Memory Command
+
+The _Access Memory_ abstract command supports accessing {ctag}s when `aarsize=7`, using the `data<n>` register format specified in <<debug_data_aarsize_7>>.
```
JanMatCodasip: 1) Could we use the formulation "the implementation can choose to support ..." also here to make it clear that the Access Memory command for a full capability is conditionally optional?

2) Typo:  The size field in Access Memory abstract commands is called `aamsize`.

```suggestion
Implementations can choose to support accessing full capabilities, including the {ctag}s, via the the _Access Memory_ abstract command. Full capability access is denoted by `aamsize=7` and uses the `data<n>` register assignments specified in <<debug_data_aarsize_7>>.
```

src/cheri/debug-integration.adoc
JanMatCodasip, line 68: I think it would be useful to add program buffer sequences for reading / writing capabilities in memory to this non-normative note. 

However, given the amount of changes in this PR is already large, I would leave that for a subsequent PR.

src/cheri/debug-integration.adoc
```diff
@@ -85,6 +90,50 @@ no way to discover the former property.
...
+
+`aarsize=7` is defined to represent full capability access.
+
+The `data<n>` register mapping is as shown in <<debug_data_aarsize_7>>.
+
+.Data register mapping when `aarsize=7`
+[#debug_data_aarsize_7]
+[width="100%",options=header,cols="1,1,1,1,1"]
+|=====
+|ISA                    | arg0/return value| arg1^1^     | arg2        | {ctag}
```
JanMatCodasip, line 103: The `arg2` column is strange - both the original debug spec and this extension contain it, but neither of the specs make any use of it.  🤔 

Should we still keep it - for consistency with the original debug spec? If yes, a note can be added, saying that the "arg2" is there for future extensions.

src/cheri/debug-integration.adoc
```diff
@@ -85,6 +90,50 @@ no way to discover the former property.
...
+
+When a capability is written it must be a <<cap_subset>> of the capability in <<drootc>>.
+If not then the destination {ctag} is set to zero.
+
+NOTE: For the capability subset: `cap2` is the capability being written and `cap1` is the value of <<drootc>>.
+
+NOTE: It is valid for root capabilities not to cover the entire possible address space.
+ In this case the debugger cannot claim access to more memory than is permitted by <<drootc>>.
+
+NOTE: Any register write with `aarsize{ne}7` causes the {ctag} (if present) to be set to zero.
```
JanMatCodasip: I would recommend to drop the "note" since this is a normative statement - requirement.

```suggestion
Any register write with `aarsize{ne}7` causes the {ctag} (if present) to be set to zero.
```

src/cheri/debug-integration.adoc
```diff
@@ -85,6 +90,50 @@ no way to discover the former property.
   setting `transfer` to 1 and `write` to 0).
 ====
 
+==== Access Register Command
+
+The _Access Register_ abstract command has optional support for accessing YLEN registers.
+
+If `aarsize` matches YLEN then all YLEN-bits should be accessible via the command.
```
JanMatCodasip: From the discussion under https://github.com/riscv/riscv-cheri/issues/1039 it so far looks to me that we should make YLEN access mandatory (not optional).

One possibility how to formulate that:

```suggestion
When an YLEN-wide register is accessible via _Access Register_ abstract command, then:

- Reads of YLEN bits or any smaller amounts of bits must be supported.
- Writes of YLEN bits must be supported.
- Writes of smaller amounts of bits than YLEN may optionally be supported but it is UNSPECIFIED what happens to the higher bits.

NOTE: The debugger can change the low XLEN bits of a YLEN wide register by performing a read-modify-write sequence.
```

Alternate possibility, which would optionally allow modify the low XLEN bits without altering the metadata in a single command:

```suggestion
When an YLEN-wide register is accessible via _Access Register_ abstract command, then:

- Reads of YLEN bits and any smaller amounts of bits must be supported.
- Writes of YLEN bits must be supported.
- Writes of XLEN bits may optionally be supported. They are supported, the capability metadata (the upper YLEN/2 bits of the register) must be left unchanged.
- Writes of less than XLEN bits may optionally be supported but it is UNSPECIFIED what happens to the higher bits.
```

src/cheri/debug-integration.adoc
```diff
@@ -85,6 +90,50 @@ no way to discover the former property.
...
+|{cheri_base64_ext_name}| data0-data3      | data4,data5 | data6-data9 | data10
+|=====
+
+^1^ arg1 represents an address so never needs to exceed DXLEN.
+ Each `data<n>` register is 32-bit so a maximum of 2 is required.
+
+Bit 0 of `data5` or `data10` contain the {ctag} which is either written to, or read from, the target register.
+All other bits in that `data` register are UNSPECIFIED.
+
+When a capability is written it must be a <<cap_subset>> of the capability in <<drootc>>.
```
JanMatCodasip, line 130: Does this mean that the present value of `drootc` is relevant - as selected by `drootsel` at the time the abstract command is issued?

src/cheri/debug-integration.adoc
```diff
@@ -85,6 +90,50 @@ no way to discover the former property.
...
+NOTE: It is valid for root capabilities not to cover the entire possible address space.
+ In this case the debugger cannot claim access to more memory than is permitted by <<drootc>>.
+
+NOTE: Any register write with `aarsize{ne}7` causes the {ctag} (if present) to be set to zero.
+
+==== Access Memory Command
+
+The _Access Memory_ abstract command supports accessing {ctag}s when `aarsize=7`, using the `data<n>` register format specified in <<debug_data_aarsize_7>>.
+
+NOTE: If `aarsize{ne}7` then {ctag}s are always read and written as zero.
```
JanMatCodasip: Similarly here - I would drop the "note" prefix.

```suggestion
If `aarsize{ne}7` then {ctag}s are always read and written as zero.
```

src/cheri/debug-integration.adoc
```diff
@@ -85,6 +90,50 @@ no way to discover the former property.
...
+
+NOTE: It is valid for root capabilities not to cover the entire possible address space.
+ In this case the debugger cannot claim access to more memory than is permitted by <<drootc>>.
+
+NOTE: Any register write with `aarsize{ne}7` causes the {ctag} (if present) to be set to zero.
+
+==== Access Memory Command
+
+The _Access Memory_ abstract command supports accessing {ctag}s when `aarsize=7`, using the `data<n>` register format specified in <<debug_data_aarsize_7>>.
+
```
JanMatCodasip, line 143: Should we state also here that the capability being written must be a subset of `drootc`, otherwise the tag is written as 0?

src/cheri/debug-integration.adoc
```diff
@@ -85,6 +90,50 @@ no way to discover the former property.
...
+If `aarsize` matches YLEN then all YLEN-bits should be accessible via the command.
+This requires enough `data<n>` registers to be supported to cover the entire data width (2 for {cheri_base32_ext_name} or 4 for {cheri_base64_ext_name}).
+
+Implementations can choose to also support accessing {ctag}s as part of the access register command.
+
+`aarsize=7` is defined to represent full capability access.
+
+The `data<n>` register mapping is as shown in <<debug_data_aarsize_7>>.
+
+.Data register mapping when `aarsize=7`
```
JanMatCodasip: To me as a reader, it was strange that this table is introduced in Access Register section but contains provisions for the Access Memory command (the arg1 column that holds memory addresses).

Could the table be moved to a generic section about abstract commands? 

Something like this:

```
==== Abstract Commands

<Common requirements go here, including the table with data registers.>

===== Access Register Command

<Specifics of access register command.>

===== Access Memory Command

<Specifics of access memory command.>
```

I think this would also make it easier to later re-insert the requirements back to the original text of the debug spec - if that is the eventual goal.

JanMatCodasip (commented):
@tariqkurd-repo Thank you for your changes. I have read through the text once more and posted all comments and suggestions that I came across.

src/cheri/debug-integration.adoc
tariqkurd-repo: not sure why that one got lost

src/cheri/debug-integration.adoc
tariqkurd-repo: it seems to be just a aa[rm]size register

src/cheri/debug-integration.adoc
tariqkurd-repo: yes! for us that's not a problem as there's only one. CHERIoT would have to choose the correct root.

src/cheri/debug-integration.adoc
tariqkurd-repo: it's a note as this is following the standard rule for memory writing, as we don't repeat normative text.

src/cheri/debug-integration.adoc
tariqkurd-repo: that's already above on line 122

src/cheri/debug-integration.adoc
tariqkurd-repo: again, this follows the standard memory access pattern of tags not being accessed if the width is not YLEN

src/cheri/debug-integration.adoc
JanMatCodasip: Ok, understood. In that case, let's keep the "NOTE" prefix.

src/cheri/debug-integration.adoc
JanMatCodasip: Ok, understood.

src/cheri/debug-integration.adoc
JanMatCodasip: The line 122 is for Access Register abstract command, so I think the same requirement should be repeated for the Access Memory command, too. (Alternatively, the requirement can be moved to some form of common section for all abstract commands, if such a section is created.)

src/cheri/debug-integration.adoc
JanMatCodasip: <del>

I also think so, but still, the `arg2` argument is not utilized anywhere else in the RISC-V debug spec, so readers may wonder why it is even part of the table.

To avoid any doubts, we may add a note along these lines:

`NOTE: The argument "arg2" is currently not used by any abstract command and is left for future extensions.`

</del>

EDIT: `arg2` is mentioned in the non-normative "B.2.11. Quick Access" chapter as a possible temporary storage where to lay off data during program buffer sequences.

src/cheri/debug-integration.adoc
JanMatCodasip: Ok, understood.

src/cheri/debug-integration.adoc
tariqkurd-repo: agreed - I won't add them but I would be happy if you do!

src/cheri/debug-integration.adoc
tariqkurd-repo: I'll add a note

JanMatCodasip: So given that, I think the table can stay as it is.

src/cheri/debug-integration.adoc
tariqkurd-repo:
```suggestion
{cheri_base_ext_name} harts must provide a means for accessing capabilities in registers and in memory via the external debugger.
```

src/cheri/debug-integration.adoc
tariqkurd-repo:
```suggestion
This is achieved by supporting at least one of:
```

src/cheri/debug-integration.adoc
```diff
@@ -6,23 +6,29 @@ include::trigger-integration.adoc[]
...
+* `abstractcs.progbufsize` is at least 4 if `dmstatus.impebreak` is 1, or at least 5 if `dmstatus.impebreak` is 0.
 
 [NOTE]
 ====
 These requirements allow a debugger to read and write capabilities in integer
 registers without disturbing other registers.  These requirements may be
 relaxed if some other means of accessing capabilities in integer registers,
-such as an extension of the Access Register abstract command, is added.
+These requirements allow a debugger to read and write capabilities in integer
+registers without disturbing other registers.
```
JanMatCodasip: This looks like a typo in an applied suggestion.

The resulting text of this paragraph should be just:

```
These requirements allow a debugger to read and write capabilities in integer
registers without disturbing other registers.
```

tariqkurd-repo: fixed

src/cheri/debug-integration.adoc
```diff
@@ -85,6 +91,60 @@ no way to discover the former property.
...
+==== Access Register Command
+
+When a YLEN-wide register is accessible via an _Access Register_ abstract command, then:
+
+- Reads of YLEN bits or any smaller amounts of bits must be supported.
+- Writes of YLEN bits must be supported.
+- Writes of smaller amounts of bits than YLEN may optionally be supported but it is UNSPECIFIED what happens to the higher bits.
+
+NOTE: The debugger can change the low XLEN bits of a YLEN wide register by performing a read-modify-write sequence.
+This requires enough `data<n>` registers to be supported to cover the entire data width (2 for {cheri_base32_ext_name} or 4 for {cheri_base64_ext_name}).
```
JanMatCodasip: Nitpicks: 

1) Blank line to visually separate the Note and the following paragraph.

2) Since the flow of the text is broken by the note, the "This requires ..." reference sounds somewhat odd to me. I tried to rephrase it below. 

3) I would say "at least 2 ... or at least 4 ..." instead of just "2 ... or 4 ...".

```suggestion
NOTE: The debugger can change the low XLEN bits of a YLEN wide register by performing a read-modify-write sequence.

The requirement to support YLEN-wide access mandates that enough `data<n>` registers are implemented to cover the entire YLEN bit-width (at least 2 for {cheri_base32_ext_name} or at least 4 for {cheri_base64_ext_name}).
```

src/cheri/debug-integration.adoc
```diff
@@ -85,6 +91,60 @@ no way to discover the former property.
...
+
+When a capability is written it must be a <<cap_subset>> of the capability in <<drootc>>.
+If not then the destination {ctag} is set to zero.
+
+If `aarsize{ne}7` then {ctag}s are always read and written as zero.
+
+NOTE: For the capability subset: `cap2` is the capability being written and `cap1` is the value of <<drootc>>.
+
+NOTE: It is valid for root capabilities not to cover the entire possible address space.
+ In this case the debugger cannot claim access to more memory than is permitted by the current value of <<drootc>>.
```
JanMatCodasip, line 138: These requirements should be repeated for the Access Memory command, too. Or, even better, moved to the common section above.

tariqkurd-repo: done

src/cheri/debug-integration.adoc
jrtc27: This doesn't make sense with otype-based sealing. You require *multiple* roots in order to derive that capability: one to authorise the initial unsealed capability, and another to authorise the sealing.

tariqkurd-repo: are you saying that a written cap needs to be a subset of at least one of the possible root caps selectable by drootcsel? But I don't understand why a single written cap wouldn't be a subset of one of them.

jrtc27: If I want to write a capability that is `0x42 [rwRW,0x40-0x50]` but _sealed with otype 7_, say, and sealing capabilities have a separate root, what singular capability would exist in drootc to authorise it? I argue none, because to construct it with code you first need to YBLD the unsealed capability using the normal root of rwRW capabilities, and then YSEAL the result with the root of sealing capabilities.

jrtc27: Noting that (a) CHERIoT's multi-root scheme has rx, rw, and sealing as three distinct roots (b) we envision any otype-based sealing extension for "normal" Y as still having two distinct roots, rwx and sealing.

tariqkurd-repo: I guess I don't understand multiple roots well enough - in that case there's no root cap which can be used as an argument to YSS against the sealed cap which would pass?
so maybe it needs to be written unsealed and then sealed in registers before being written to memory.

jrtc27: > so maybe it needs to be written unsealed and then sealed in registers before being written to memory.

That sounds a lot like "don't use abstract commands, use the program buffer"? Which we're trying to get away from?

jrtc27: > in that case there's no root cap which can be used as an argument to YSS against the sealed cap which would pass?

Well, YSS would return 1 when given the appropriate unsealed root, because subset tests ignore any sealing. But a YBLD of that bit pattern (if untagged) wouldn't give you that sealed capability.

tariqkurd-repo: so what's your suggestion for fixing it so it works with multiple roots?

jrtc27: "Is derivable from the set of available debug root capabilities"?

jrtc27: (Probably with a non-normative note explaining this is "just" a subset check against each root for the "normal" Y base but extensions, or other bases like CHERIoT, that add more roots may require multiple check)

PR #1070, opened 2026-06-16
Use "widens" instead of "extends to"

### LINKED ISSUE #1063: Capabilities register description is ambiguous
https://github.com/riscv/riscv-cheri/blob/fdf3f4f2e51e62e9e2d204a7fa666f1b97754185/src/cheri/riscv-unpriv-integration.adoc?plain=1#L66

Reading chapter "Capability Registers and Format", the sentence "RVY extends all registers that have to be able to hold addresses to 2*XLEN bits" leaves it unclear if this means:

* For each pointer holding XLEN register an additional 2 * XLEN capabilities register is created. I.e. for each x0 - x31 register 2 * XLEN extra bits in the capabilities register are needed (plus 1 bit for the V flag). - I assume this is meant.

or

* To each XLEN register XLEN capabilities bits are added. So in total for each of the original x0 - x31 registers only XLEN extra bits are needed to store the capabilities.

Please, find a description that is unambiguous.

arichardson:
This should make the spec easier to parse (especially for non-native
English speakers) where some of the more complicated sentences could be
parsed as adding new wider registers.

Fixes: https://github.com/riscv/riscv-cheri/issues/1063

### DIFF
--- a/src/cheri/riscv-unpriv-integration.adoc
+++ b/src/cheri/riscv-unpriv-integration.adoc
@@ -63,7 +63,7 @@ improve their security.
 [#sec_capability_registers]
 === Capability Registers and Format
 
-{cheri_base_ext_name} extends all registers that have to be able to hold addresses to `2*XLEN` bits (hereafter referred to as YLEN), adding metadata to protect its integrity, limit how it is manipulated, and control its use.
+{cheri_base_ext_name} widens all registers that can hold addresses from `XLEN` to `2*XLEN` bits (hereafter referred to as YLEN), adding metadata to protect its integrity, limit how it is manipulated, and control its use.
 In addition to widening to YLEN, each register also gains a one-bit {ctag} which is defined below.
 
 {cheri_base_ext_name} specifies the minimum required fields that the capability format must support, and their semantics.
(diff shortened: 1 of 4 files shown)


src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -682,8 +681,8 @@ The zero register (`x0`) is extended with zero metadata and a zero {ctag}: this
 [#pcc,reftext="{pcc}"]
 ==== The Program Counter Capability (`{pcc}`)
 
-The `pc` is extended to be a capability.
-Extending the `pc` allows the range of branches, jumps and linear execution for currently executing code to be restricted.
+The `pc` register is widened to hold a capability (referred to as the Program Counter Capability, or `{pcc}`).
```
tariqkurd-repo: We don't call it PCC anymore, just PC.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -682,8 +681,8 @@ The zero register (`x0`) is extended with zero metadata and a zero {ctag}: this
...
+Widening the `pc` register to `{pcc}` allows the range of branches, jumps and linear execution for currently executing code to be restricted.
```
tariqkurd-repo: PCC variable is just PC so this doesn't really work

arichardson:
@xypron I hope this change clarifies the text appropriately.

tariqkurd-repo (approved):

src/cheri/riscv-unpriv-integration.adoc
arichardson: That was reverted

jrtc27 (approved):
This does seem to read more clearly

PR #1091, opened 2026-06-26
encode part of AP field for RV64Y

tariqkurd-repo:
Encode R-W-C-LM-SL-LG as a single 5-bit field with easy to access reserved states
Retain _all_ valid combinations.


src/cheri/rvy64-encoding.adoc
```diff
@@ -56,39 +55,52 @@ endif::[]
...
+| 01000  5+|                  _reserved_  due to <<perm_req:base:lm:c-and-r,base-2>>
+| 01001    | R |   | C | LM |
+| 01010  5+|                  _reserved_  due to <<perm_req:base:lm:c-and-r,base-2>>
+| 01011    | R | W | C | LM | *Maximum permissions (AP_CMP_MAX)*
+6+| *Reserved states*
+| 011xx  5+|                  _reserved_ for <<section_zylevels1>>
+| 100xx  5+|                  _reserved_ for <<section_zylevels1>>
+| 10101  5+|                  _reserved_ for Zyseal sealing permission
+| 10110  5+|                  _reserved_ for Zyseal unsealing permission
+| 10111  5+|                  _reserved_ for Zyseal sealing and unsealing permission
```
nwf, line 99: It's probably worth a comment (that I don't see at a glance, but sorry if I missed it) that this implies a multi-rooted implementation of Zyseal, since `SE`/`US` will be exclusive with `R`/`W`, right?  (By contrast to the 8 states reserved Zylevels1.)

tariqkurd-repo: Yes, probably, but I'm not attempting to specify that fully right now as it hasn't been worked through for the application core.

tariqkurd-repo: I think I'll add the word "tentatively" to show intention but not guaranteeing to do it that way as we don't currently have multiple roots

arichardson (commented):
I think this actually makes the spec harder to read. Splitting the field is just confusing. Can't we just say that non useful states are reserved for future orthogonal permissions?

arichardson:
Some of these changes are good but overall I think it should be reverted.

tariqkurd-repo:
> Some of these changes are good but overall I think it should be reverted.

The real problem is that the reserved spaces are very hard to allocate.

If you draw out all of the permissions from C, R, W, LM, LG, SL and mark which are reserved (which I have in a spreadsheet) then it's a real struggle to see how anything can be encoded in those areas, so really I think we're throwing away the unused spaces.
This partially encoded field allows the reserved spaces to be cleanly allocated.
It also means that Zylevels1 only allocates 9-entries in a 64-entry table, as so few cases are valid (~0.85-bits), instead of taking up 3 whole bits.
When we add more extensions (e.g. Zyseal) we really can't afford to allocate multiple whole permission bits for them.
What's going to happen is that we'll run out of metadata bits very very quickly once people start adding their own extensions, and then the capability encoding format will start to fragment.

The RV64 format was defined with the idea of having a lot of available metadata space and so it wasn't necessary to allocate sparingly, but I think we've gone past that point (not least with the Ericcson color scheme which uses _all_ reserved bits) already before the spec has even been frozen, so this is our last chance to take action.

The encoding scheme is safe because it doesn't throw any any states, it just encodes the valid ones into fewer bits.

PR #1096, opened 2026-07-01
add MISA reset note

tariqkurd-repo:
misa normally reset to "maximum suported extensions" which is not the case for CHERI if Zyhybrid is implemented.

### DIFF
--- a/src/cheri/riscv-priv-integration.adoc
+++ b/src/cheri/riscv-priv-integration.adoc
@@ -837,6 +837,9 @@ The effective CHERI-enable for the current privilege is:
 
 NOTE: On reset CHERI is always disabled for backwards compatibility: <<misa_y>>.{CRE} resets to zero, <<ddc>>, <<pcc>>, Xtvec, Xepc are nominally <<root-cap,root capabilities>>.
 
+NOTE: Resetting `<<misa_y>>.{CRE}` to zero for compatiblity means that the reset value is _not_ the maximal set of supported extensions.
+ This aspect is key for CHERI, CHERI oblivious software runs identically on a CHERI core in {cheri_int_mode_name} as on a non-CHERI core.
+
 `<<misa_y>>.{CRE}` is a WARL field.
 M-mode software can freely set it to one, but then it can only be set back to zero if <<ddc>>, <<pcc>>, Xtvec, Xepc are root capabilities so that no CHERI configuration has taken effect.
 


src/cheri/riscv-priv-integration.adoc
```diff
@@ -837,6 +837,9 @@ The effective CHERI-enable for the current privilege is:
 
 NOTE: On reset CHERI is always disabled for backwards compatibility: <<misa_y>>.{CRE} resets to zero, <<ddc>>, <<pcc>>, Xtvec, Xepc are nominally <<root-cap,root capabilities>>.
 
+NOTE: Resetting `<<misa_y>>.{CRE}` to zero for compatiblity means that the reset value is _not_ the maximal set of supported extensions.
```
jrtc27, line 840: Wait, what's misa_y.CRE? misa has a Y, but CRE is an mfoocfg/env thing?

jrtc27: And CRE is not about whether the extension is present or not, it's how it behaves. You're still in base Y, which is the key thing.

jrtc27: (and misa.Y absolutely must be 1 on reset)

tariqkurd-repo: it's just the variable - {CRE}=Y - it's probably time to get rid of the variables.

tariqkurd-repo: misa.Y resets to 1 and is read-only-1 unless in hybrid mode when it resets to zero.
that's what's currently in the spec and has been for a long time.
I do remember discussing this before.

jrtc27: Well, that was back when Y=0 was hybrid. But after ARC feedback Y=0, now that we're in custom-N, means non-CHERI, not hybrid. That is, the CHERI instructions are not available, and may even have a different, vendor-defined, meaning. So presumably then the adaptation of the spec to that is not complete, and this is a relic of that.

jrtc27: That is, in the current world order, the reset state must be:
* misa.Y=1
* PC.P = 1 (Integer) if it exists
* xfooenv/cfg.CRE=0

tariqkurd-repo: ok

arichardson: I was under the assumption that we had decided the reset state is CHERI completely off if you support integer nice. You have to opt in in m mode first, so Y=0 sounds correct to me.

arichardson: But starting with y=1 and int mode is fine too

jrtc27: Well most systems will be sane and not support writable misa, and for those that do you don't want them to have a different reset state

arichardson: That is a really good point and it makes everything simpler.

PR #1097, opened 2026-07-01
fix misa/Xenvcfg/pc reset values

tariqkurd-repo:
replaces #1096

### DIFF
--- a/src/cheri/riscv-priv-integration.adoc
+++ b/src/cheri/riscv-priv-integration.adoc
@@ -848,8 +848,6 @@ M-mode can only set it to zero if <<ddc>>, <<pcc>>, Xtvec, Xepc are root capabil
 
 NOTE: In this way M-mode software cannot disable any CHERI checks once they have been configured.
 
-NOTE: The reset state enables the maximum number of extensions as CHERI and {cheri_int_mode_name} are both enabled.
-
 The following occurs when executing code in a privilege mode that has CHERI disabled:
 
 * Custom encoding spaces reallocated as standard encoding spaces for {cheri_base_ext_name} revert to being custom encoding spaces
--- a/src/cheri/riscv-priv-integration.adoc
+++ b/src/cheri/riscv-priv-integration.adoc
@@ -835,10 +835,13 @@ The effective CHERI-enable for the current privilege is:
 * Supervisor: `<<misa_y>>.{CRE} & <<menvcfg_y>>.{CRE}`
 * User: `<<misa_y>>.{CRE} & <<menvcfg_y>>.{CRE} & <<senvcfg_y>>.{CRE}`
 
-NOTE: On reset CHERI is always disabled for backwards compatibility: <<misa_y>>.{CRE} resets to zero, <<ddc>>, <<pcc>>, Xtvec, Xepc are nominally <<root-cap,root capabilities>>.
+The reset values are:
 
-NOTE: Resetting `<<misa_y>>.{CRE}` to zero for compatiblity means that the reset value is _not_ the maximal set of supported extensions.
- This aspect is key for CHERI, CHERI oblivious software runs identically on a CHERI core in {cheri_int_mode_name} as on a non-CHERI core.
+* `<<misa_y>>.{CRE}=1`
+* `<<menvcfg_y>>.{CRE}=0`
+* `<<senvcfg_y>>.{CRE}=0`
+* <<pcc>>.<<p_bit>>={INT_MODE_VALUE} (for {cheri_int_mode_name}).
+* <<ddc>>, <<pcc>>, Xtvec, Xepc are nominally <<root-cap,root capabilities>>.
 
 `<<misa_y>>.{CRE}` is a WARL field.
 M-mode software can freely set it to one, but then it can only be set back to zero if <<ddc>>, <<pcc>>, Xtvec, Xepc are root capabilities so that no CHERI configuration has taken effect.
--- a/src/cheri/riscv-priv-integration.adoc
+++ b/src/cheri/riscv-priv-integration.adoc
@@ -844,9 +844,11 @@ The reset values are:
 * <<ddc>>, <<pcc>>, Xtvec, Xepc are nominally <<root-cap,root capabilities>>.
 
 `<<misa_y>>.{CRE}` is a WARL field.
-M-mode software can freely set it to one, but then it can only be set back to zero if <<ddc>>, <<pcc>>, Xtvec, Xepc are root capabilities so that no CHERI configuration has taken effect.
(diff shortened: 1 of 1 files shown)


src/cheri/riscv-priv-integration.adoc
```diff
@@ -835,10 +835,13 @@ The effective CHERI-enable for the current privilege is:
...
-NOTE: Resetting `<<misa_y>>.{CRE}` to zero for compatiblity means that the reset value is _not_ the maximal set of supported extensions.
- This aspect is key for CHERI, CHERI oblivious software runs identically on a CHERI core in {cheri_int_mode_name} as on a non-CHERI core.
+* `<<misa_y>>.{CRE}=1`
+* `<<menvcfg_y>>.{CRE}=0`
+* `<<senvcfg_y>>.{CRE}=0`
+* <<pcc>>.<<p_bit>>={INT_MODE_VALUE} (for {cheri_int_mode_name}).
+* <<ddc>>, <<pcc>>, Xtvec, Xepc are nominally <<root-cap,root capabilities>>.
 
 `<<misa_y>>.{CRE}` is a WARL field.
 M-mode software can freely set it to one, but then it can only be set back to zero if <<ddc>>, <<pcc>>, Xtvec, Xepc are root capabilities so that no CHERI configuration has taken effect.
```
jrtc27: This feels a bit weirdly worded now that the reset state is 1

src/cheri/riscv-priv-integration.adoc
```diff
@@ -835,10 +835,13 @@ The effective CHERI-enable for the current privilege is:
...
 
-NOTE: On reset CHERI is always disabled for backwards compatibility: <<misa_y>>.{CRE} resets to zero, <<ddc>>, <<pcc>>, Xtvec, Xepc are nominally <<root-cap,root capabilities>>.
+The reset values are:
 
-NOTE: Resetting `<<misa_y>>.{CRE}` to zero for compatiblity means that the reset value is _not_ the maximal set of supported extensions.
- This aspect is key for CHERI, CHERI oblivious software runs identically on a CHERI core in {cheri_int_mode_name} as on a non-CHERI core.
+* `<<misa_y>>.{CRE}=1`
+* `<<menvcfg_y>>.{CRE}=0`
+* `<<senvcfg_y>>.{CRE}=0`
+* <<pcc>>.<<p_bit>>={INT_MODE_VALUE} (for {cheri_int_mode_name}).
```
jrtc27, line 843: Probably worth adding a note (inspired by the other PR) about why? Just without all the stuff about it disabling extensions (arguably, it _enables_ extensions, since Zyhybrid is an extension to the base Y :P).

tariqkurd-repo: ok done

src/cheri/riscv-priv-integration.adoc
tariqkurd-repo: ok fixed

src/cheri/riscv-priv-integration.adoc
```diff
@@ -835,16 +835,21 @@ The effective CHERI-enable for the current privilege is:
...
 * User: `<<misa_y>>.{CRE} & <<menvcfg_y>>.{CRE} & <<senvcfg_y>>.{CRE}`
 
-NOTE: On reset CHERI is always disabled for backwards compatibility: <<misa_y>>.{CRE} resets to zero, <<ddc>>, <<pcc>>, Xtvec, Xepc are nominally <<root-cap,root capabilities>>.
+The reset values are:
 
-NOTE: Resetting `<<misa_y>>.{CRE}` to zero for compatiblity means that the reset value is _not_ the maximal set of supported extensions.
- This aspect is key for CHERI, CHERI oblivious software runs identically on a CHERI core in {cheri_int_mode_name} as on a non-CHERI core.
+* `<<misa_y>>.{CRE}=1`
+* `<<menvcfg_y>>.{CRE}=0`
+* `<<senvcfg_y>>.{CRE}=0`
```
jrtc27, line 842: I do wonder if this is a sensible default for systems that lack Zyhybrid; by definition they must know about CHERI to function...

jrtc27: (To be honest, I wonder if these bits should even exist for non-Zyhybrid systems; perhaps they should just be part of that extension)

jrtc27: e.g. you can't even JAL non-x0 on such a system at any other mode by default, since that would write a capability to the link register, nor can you use CALL/TAIL since that's AUIPC+JALR and the AUIPC wants a temporary capability register

tariqkurd-repo: These bits don't exist on non hybrid systems - this is the Zyhybrid chapter, and we also only allow misa.Y to be writable if Zyhybrid is implemented.

jrtc27: Oh, ok

tariqkurd-repo: ok to merge?

arichardson (approved):

src/cheri/riscv-priv-integration.adoc
```diff
@@ -835,16 +835,21 @@ The effective CHERI-enable for the current privilege is:
...
+* <<pcc>>.<<p_bit>>={INT_MODE_VALUE} (for {cheri_int_mode_name}).
+* <<ddc>>, <<pcc>>, Xtvec, Xepc are nominally <<root-cap,root capabilities>>.
 
 `<<misa_y>>.{CRE}` is a WARL field.
-M-mode software can freely set it to one, but then it can only be set back to zero if <<ddc>>, <<pcc>>, Xtvec, Xepc are root capabilities so that no CHERI configuration has taken effect.
+M-mode can only set it to zero if <<ddc>>, <<pcc>>, Xtvec, Xepc are root capabilities so that no CHERI configuration has taken effect.
 
 NOTE: In this way M-mode software cannot disable any CHERI checks once they have been configured.
 
+NOTE: The reset state enables the maximum number of extensions as CHERI and {cheri_int_mode_name} are both enabled.
```
jrtc27: Ehhh I didn't mean to actually include this, I hoped the ":P" would have made that clear enough. It's technically true but a somewhat perverse view of the "extension" (in that adding an extension on top of the new base ISA comes almost full circle back to the original base ISA).

tariqkurd-repo: Ha ha ok I can remove it, but technically it _is_ true 😁

jrtc27: Yeah... PR LGTM if this line is deleted

PR #1099, opened 2026-07-01
User mode CSR summary was missing

tariqkurd-repo:


src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -1386,6 +1386,41 @@ include::insns/prefetch.w.adoc[leveloffset=-1]
...
+
+ifndef::cheri_standalone_spec[]
+[#section_invalid_addr_conv,reftext="invalid address conversion"]
+NOTE: See the privileged specification for the definition of invalid address handling.
+endif::[]
+
+.YLEN-wide User-Mode CSRs storing code pointers or data pointers
+[#CSR_exevectors_u]
+[width="100%",options=header,cols="1,1,1,1,1"]
+|==============================================================================
```
arichardson, line 1420: Isn't this just jvtc which is post v1?

arichardson (commented):
One question otherwise seems fine

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: Yes so should probably be fully in the ifdef, will address tomorrow

tariqkurd-repo: in fact it's also DDC which is a data pointer - JVT will also enter that table in future.

PR #1100, opened 2026-07-02
Avoid using a separate AP_CMP field 

arichardson:
This makes the spec less awkward to read.


tariqkurd-repo (approved):
I made some fixes, LGTM

PR #1107, opened 2026-07-03
revert to 1-bit per permission

tariqkurd-repo:
Simplify the spec by reverting to 1-bit per permission

### DIFF
--- a/src/cheri/rvy64-encoding.adoc
+++ b/src/cheri/rvy64-encoding.adoc
@@ -12,19 +12,20 @@ NOTE: The format is closely modeled upon features from CHERI v9 cite:[cheri-v9-s
 [#cap_encoding_xlen64]
 include::img/cap-encoding-xlen64.edn[]
 
-NOTE: AP[6:2] encodes <<r_perm,R>>/<<w_perm,W>>/<<c_perm,C>>/<<lm_perm,LM>> permissions; AP[1] is <<asr_perm,ASR>>; AP[0] is <<x_perm,X>>.
-
 NOTE: Reserved bits must be 0 in valid capabilities and are available for future extensions to {cheri_base_ext_name}.
 
 The encoding diagram above of the capability encoding format includes some fields which depend upon the presence of extensions:
 
 <<section_cheri_hybrid_ext>>::
-When <<section_cheri_hybrid_ext>> is supported, capabilities include the <<p_bit>>. Otherwise, that field is _reserved_.
+When <<section_cheri_hybrid_ext>> is implemented
++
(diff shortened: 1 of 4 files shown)


src/cheri/rvy64-encoding.adoc
```diff
@@ -55,51 +56,41 @@ endif::[]
...
-|==============================================================================
+There are 9 bits in total allocated to the <<AP-field>> and the <<p_bit>>.
+
+Once all of the permission rules in <<base_clrperm_rule_summary>> and <<zylevels1_clrperm_table>> have been applied, and also the <<p_bit>> validity depends upon <<x_perm>> (see <<sec_p_bit_hybrid_rule>>) the total valid combinations of permissions totals 180.
+
+Encoding 180 valid states in 9 bits, which allows for 512 states, is inefficient.
+
+Future extensions or capability encoding formats may address this by implementing a more compact <<AP-field>> and <<p_bit>> encoding that is fully _compatible for all software_ to increase the number of reserved bits.
+
+Such as encoding would have observable differences in the in-memory representation of capabilities.
```
arichardson, line 91: This should not be normative text, in fact I'd argue it should just be a comment that isn't rendered?

arichardson: Maybe just say "NOTE: Due to the permission dependency rules there are 90 valid states for these 9 bits. Future extensions may repurpose some of the currently invalid states."

src/cheri/rvy64-encoding.adoc
```diff
@@ -55,51 +56,41 @@ endif::[]
...
-| *100* xx 5+|  _reserved_ for <<section_zylevels1>>
-6+| *Reserved for future extensions*
-| 10101 5+|  _reserved_, possibly for future sealing permission
-| 10110 5+|  _reserved_, possibly for future unsealing permission
-| 10111 5+|  _reserved_, possibly for future sealing and unsealing permission
-| 11xxx 5+|  _reserved_ for other future use
-|==============================================================================
+There are 9 bits in total allocated to the <<AP-field>> and the <<p_bit>>.
+
+Once all of the permission rules in <<base_clrperm_rule_summary>> and <<zylevels1_clrperm_table>> have been applied, and also the <<p_bit>> validity depends upon <<x_perm>> (see <<sec_p_bit_hybrid_rule>>) the total valid combinations of permissions totals 180.
```
arichardson, line 85: it's 180 with the GL bit, 90 otherwise (include P bit).

tariqkurd-repo: ah yes ok!

PR #1108, opened 2026-07-03
Make p depends on x into a rule

tariqkurd-repo:
I think it's simpler to express P depends on X as a rule, so it's regular with the other rules and makes the spec easier to follow.


arichardson (approved):

arichardson:
I don't recall why we didn't do this in the first piece. Maybe @nwf remembers?

tariqkurd-repo:
> I don't recall why we didn't do this in the first piece. Maybe Nathaniel Wesley Filardo (@nwf) remembers?

it was @nwf 's text I've replaced certainly.

PR #1109, opened 2026-07-03
Clarify the misa.Y text and move the write behaviour to that section

### LINKED ISSUE #734: should we allow MISA.Y to be writeable even if hybrid mode isn't supported?

arichardson:
Also drop the requirement for Zyhybrid for writable misa.Y, Zyhybrid adds support for toggling the pointer mode but the base ISA always remains RVY, so this should not be dependent on Zyhybrid (even though I imagine most/all implementation that support setting misa.Y would also support Zyhyrid)

this is a spec change as misa.Y is WARL without implementing Zyhybrid

Fix #734

### DIFF
--- a/src/cheri/riscv-priv-integration.adoc
+++ b/src/cheri/riscv-priv-integration.adoc
@@ -152,7 +152,8 @@ When the base ISA is {cheri_base_ext_name}, the Y bit is one.
 
 `<<misa_y,misa>>.{CRE}` is a WARL field.
 If the implementation supports a writable <<misa_y,misa>>.{CRE}, it can only be set to zero if all registers used for implicit CHERI checks are currently configured as root capabilities.
-These implicit check registers are <<pcc>>, `__x__tvec`, `__x__epc`, and <<ddc>> (if {cheri_default_ext_name} is implemented).
+
+These implicit check registers are <<pcc>>, `__x__tvec`, `__x__epc`.
 
 NOTE: This restriction ensures M-mode software cannot drop CHERI checks by modifying <<misa_y,misa>> once they have been configured.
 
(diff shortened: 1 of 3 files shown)


tariqkurd-repo (approved):

src/cheri/riscv-priv-integration.adoc
```diff
@@ -152,7 +152,8 @@ When the base ISA is {cheri_base_ext_name}, the Y bit is one.
 
 `<<misa_y,misa>>.{CRE}` is a WARL field.
 If the implementation supports a writable <<misa_y,misa>>.{CRE}, it can only be set to zero if all registers used for implicit CHERI checks are currently configured as root capabilities.
-These implicit check registers are <<pcc>>, `__x__tvec`, `__x__epc`, and <<ddc>> (if {cheri_default_ext_name} is implemented).
+
+These implicit check registers are <<pcc>>, `__x__tvec`, `__x__epc`.
```
arichardson, line 156: I would have kept ddc listed here it's always annoying having to look in lots of other places to find out what the exact rules are.

tariqkurd-repo: I'll put an xref in, I agree it's annoying

arichardson (commented):
Minor comment otherwise your improvement LGTM

arichardson (commented):
Oh I see you added it thanks.

PR #1111, opened 2026-07-03
Fix AP note

tariqkurd-repo:


src/cheri/rvy64-encoding.adoc
```diff
@@ -82,11 +82,13 @@ NOTE: Future extensions which define new permissions must augment <<cap_perms_en
 
 There are 9 bits in total allocated to the <<AP-field>> and the <<p_bit>>.
 
-Once all of the permission rules in <<base_clrperm_rule_summary>> and <<zylevels1_clrperm_table>> have been applied, and also the <<p_bit>> validity depends upon <<x_perm>> (see <<sec_p_bit_hybrid_rule>>) the total valid combinations of permissions totals 180.
+* The total valid combinations of C, W, R, LM, LG and SL are 18.
```
arichardson, line 85: Can we put all of this in a [NOTE] block?

arichardson: This is non normative

arichardson (commented):
Thanks this wording is better but we should still make it a NOTE IMO

arichardson (approved):
Please ignore I didn't see it was already in a note block since GitHub truncated it.

PR #1113, opened 2026-07-06
Remove redundant "Included in::" field from instruction listings

arichardson:
The extension summary tables already list every instruction, so this field just repeats that information. It also adds noise and sometimes pushes an instruction's listing over a page boundary, which is annoying for PDF formatting. Easy to bring back by reverting this commit if needed.


tariqkurd-repo:
the unpriv manual does have the "included in" section, but there's no consistency
I'm ok with ifdef'ing these to shorten the doc maybe?

PR #1119, opened 2026-07-06
Fixes for the CMO chapter

arichardson:
CBO.CLEAN/FLUSH fault only when no byte of the cache block is in
bounds, while CBO.INVAL/ZERO require the whole block in bounds.  This
is intentional (clean/flush are non-destructive) but looked like a
copy-paste error without a rationale note.

---

Specify CBO.INVAL ASR check applies even when executed as a flush

The ASR-permission requirement is on executing the instruction, not on
the operation it ends up performing.  Also note that the illegal
instruction exception matches MRET/SRET and ASR-gated CSR accesses.

---

Clarify PREFETCH.I/R/W authorization checks

- State what happens when a check fails: the instruction executes as a
  nop and no prefetch is performed (previously unstated).
- Fix the garbled bounds bullet ("Any bytes of the requested cache
  line requested") and make it explicit that only one byte of the
  cache block needs to be in bounds, with a rationale note (a hardware
  prefetcher could prefetch the same block in the background anyway).
- Use "cache block" (Zicbop terminology) and "raise" instead of
  "throw".

### DIFF
--- a/src/cheri/insns/prefetch.i.adoc
+++ b/src/cheri/insns/prefetch.i.adoc
@@ -31,15 +31,19 @@ sign-extended offset encoded in imm[11:0], where imm[4:0] equals 0b00000, is
 likely to be accessed by an instruction fetch in the near future.
 +
 The authorizing capability is `{cs1}`.
-This instruction does not throw any exceptions.
+This instruction does not raise any exceptions.
 +
 The following authorization checks are performed:
 
 * `{cs1}.tag=1`
 * `{cs1}` is not sealed
-* Any bytes of the requested cache line requested are in `{cs1}` bounds
+* At least one byte of the requested cache block is within `{cs1}` 's bounds
 * `{cs1}` grants <<x_perm>>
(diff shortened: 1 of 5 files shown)


src/cheri/insns/prefetch.i.adoc
```diff
@@ -31,16 +31,20 @@ sign-extended offset encoded in imm[11:0], where imm[4:0] equals 0b00000, is
...
 The following authorization checks are performed:
 
 * `{cs1}.tag=1`
 * `{cs1}` is not sealed
-* Any bytes of the requested cache line requested are in `{cs1}` bounds
+* At least one byte of the requested cache block is within `{cs1}` 's bounds
 * `{cs1}` grants <<x_perm>>
 * `{cs1}` passes all <<section_cap_integrity, integrity>> checks
 
+If any of these checks fail, the instruction executes as a `nop` and no prefetch is performed.
```
arichardson, line 44: This might be redundant but maybe best to be explicit

tariqkurd-repo (approved):

PR #1125, opened 2026-07-06
Spell out that BEQ/BNE with rs1=rs2 is reserved

arichardson:
The rs1<=rs2 reservation includes the equal-register case, which the assembler operand-swapping note cannot rescue.  State this explicitly and point at j/jal for unconditional branches.

### DIFF
--- a/src/cheri/riscv-unpriv-integration.adoc
+++ b/src/cheri/riscv-unpriv-integration.adoc
@@ -1194,11 +1194,13 @@ include::insns/jalr_32bit.adoc[]
 ==== BEQ, BNE ({cheri_base_ext_name})
 
 For `beq` and `bne` only, if `rs1≤rs2` then the encoding is _reserved_.
+This includes all encodings with `rs1=rs2`.
 
 NOTE: The reserved encodings are redundant and may be used by future extensions, suggestions include branching on {ctag} values only, or YLEN-bit compares.
 
 NOTE: For `beq rs1, rs2, offset` and `bne rs1, rs2, offset`, assemblers must emit the operand order whose register-number ordering is not reserved.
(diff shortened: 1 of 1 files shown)


tariqkurd-repo (approved):

jrtc27:
I can’t immediately see text saying this, but do we limit messing with BEQ/BNE’s encoding to capability pointer mode?

arichardson:
> I can’t immediately see text saying this, but do we limit messing with BEQ/BNE’s encoding to capability pointer mode?

Yes, it's only in capability mode.

jrtc27:
Hm, where do we say that? I see:
> Instructions which are modified on an RVY architecture (see [Table 24](https://riscv.github.io/riscv-cheri/snapshot/unprivileged/#RVY_I_mod_insns_table_body)) revert to their standard RV32[IE]/RV64[IE] behavior.

But Table 24 is just AUIPC, JAL and JALR, no BEQ/BE?

jrtc27:
Also [7.9.2. Changes to PC](https://riscv.github.io/riscv-cheri/snapshot/unprivileged/#_changes_to_pc) should probably include the BEQ/BNE change, in both its description and its specific list of PC-related instructions changed?

arichardson:
Good catch, right now its just implicit. We should fix that

tariqkurd-repo:
that is a fair comment - but also all branches also change as the target PC could be tag cleared if unrepresentable - but I think that's explicit in the PC update part of the spec.
I'll add BEQ/BNE into the table

jrtc27:
Eh, that's a bit different, and only matters if PC's bounds don't cover the whole address space, at which point you're already knowingly opting into new Y behaviour. Whereas the reserving of the opcode (and potentially allocating it to something else) is unconditionally visible.

tariqkurd-repo:
ok - see #1130

PR #1126, opened 2026-07-06
Another round of minor specification review fixes

arichardson:
Fixes feedback from various human+LLM reviews.
All more substantial changes have been split into separate PRs, this one
is just an accumulation of various wording clarifications, grammar and rendering fixes.

Also added a note linking the "longer base name" proposal, and avoid using reset value
in unprivileged (talk about EEI instead).

PR #1133, opened 2026-07-07
Convert leftover editorial notes into ARC-NOTEs

arichardson:
The speculative-execution section and the system-implications appendix
still contained informal author notes.  Turn them into explicit
ARC-NOTE questions (matching the existing ARC-NOTE style) so ARC can
answer them during review.  Also fixes the "paced" typo.


arichardson:
Extracted from #1126

PR #1142, opened 2026-07-07
Fix inverted EF values in the malformedLSB note

arichardson:
The predicate is !EF && (E == 0 && enableL8), so it is the EF=0, E=0
encodings that are declared malformed; EF=1 is the canonical
survivor.  The note said the opposite.


PeterRugg:
Oops, yes, EF is still the "internal exponent" bit in my head, so getting the sense right is hard...

PeterRugg (approved):

arichardson:
Thanks for confirming!

PR #1143, opened 2026-07-07
Tidy the cap bounds decoding section

tariqkurd-repo:
#1142 made me realise that the way we present the bounds decoding really isn't clear.
This PR lists all metadata fields in a table, so the bit positions are easier to access, and shows which variables are from parameters and are from the metadata.

Also reword most of this section to make it easier to understand

### DIFF
--- a/src/cheri/rvy64-encoding.adoc
+++ b/src/cheri/rvy64-encoding.adoc
@@ -205,7 +205,8 @@ EF=1: The exponent is zero; TE, BE and L8 form part of the address bounds.
 
 ^1^ Only valid if `enableL8=1`.
 
-The variables in <<rvy64_uni_base_name_bndsenc_param_summary>> are either taken directly from the capability metadata (see <<cap_encoding_xlen64_field_table>>) or calculated from it, and are used as shown in the <<bounds_variable_usage,pseudo-code>> below.
+The variables in <<rvy64_uni_base_name_bndsenc_param_summary>> are either taken directly from the capability metadata (see <<cap_encoding_xlen64_field_table>>) or, in the case of the upper and lower bits of `T` and `B`, are calculated from it.
+They are used as shown in the <<bounds_variable_usage,pseudo-code>> below.
 
 The bit width of T and B are defined in terms of the mantissa width (MW) which
 is set depending on capability encoding as shown in
--- a/src/cheri/rvy64-encoding.adoc
+++ b/src/cheri/rvy64-encoding.adoc
@@ -27,6 +27,25 @@ When <<section_zylevels1>> is implemented:
 * The <<zylevels1_gl_flag,GL>> flag is implemented, otherwise _reserved zero_.
 * The <<AP-field>> includes <<zylevels1_lg_perm>> and <<zylevels1_sl_perm>>. Otherwise both are _reserved one_.
 
+.{rv64y_uni_base_name} capability encoding format metadata fields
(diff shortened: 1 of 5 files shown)


src/cheri/rvy64-encoding.adoc
```diff
@@ -163,20 +182,28 @@ The capability can be used to access any memory location A in the range base
...
+.{rv64y_uni_base_name} capability bounds encoding fields
+[#rvy64_uni_base_name_bndsenc_param_summary,width="100%",options=header,cols="1,4"]
+|==============================================================================
+| Parameter | Description
+| T[MW-1:0] | Value substituted into the capability's address to decode the top address.
+| TE        | Value either substituted into the exponent or the top address depending on EF.
+| B         | Value substituted into the capability's address to decode the base address.
+| BE        | Value either substituted into the exponent or the base address depending on EF.
+| E         | Exponent that determines the position at which B and T are substituted into the capability's address.
+| L8        | Value used to form either the MSB of the exponent or the MSB of the top bound.
```
PeterRugg:
```suggestion
| L8        | (enableL8=1 only) Value used to form either the MSB of the exponent or the MSB of the top bound.
```

src/cheri/rvy64-encoding.adoc
```diff
@@ -163,20 +182,28 @@ The capability can be used to access any memory location A in the range base
...
+| EF        | Exponent format flag indicating the encoding for T, B and E: +
+1. The exponent is calculated from TE and BE, and conditionally L8, if EF=0, so it is 'internal'. +
```
PeterRugg: The 1 and 2 labels confused me. Can't they be 0 and 1 to match the values of EF they're talking about?

PeterRugg (approved):
Looks good: I think this helps a lot. Just a couple of very minor tweaks that might prevent some confusion.

src/cheri/rvy64-encoding.adoc
PeterRugg: (And also in the RV32 table if you think it makes sense)

src/cheri/rvy64-encoding.adoc
```diff
@@ -198,19 +200,19 @@ The bounds field has the following components:
...
 | EF        | Exponent format flag indicating the encoding for T, B and E: +
 EF=0: The exponent is calculated from TE and BE, and conditionally L8, so it is 'internal'. +
-EF=1: The exponent is zero, TE, BE and L8 form part of the address bounds.
+EF=1: The exponent is zero; TE, BE and L8 form part of the address bounds.
 |==============================================================================
 
 ^1^ Only valid if `enableL8=1`.
 
-The variables in <<rvy64_uni_base_name_bndsenc_param_summary>> are taken directly from the capability metadata (see <<cap_encoding_xlen64_field_table>>) and used as shown in the <<bounds_variable_usage,pseudo-code>> below.
+The variables in <<rvy64_uni_base_name_bndsenc_param_summary>> are either taken directly from the capability metadata (see <<cap_encoding_xlen64_field_table>>) or calculated from it, and are used as shown in the <<bounds_variable_usage,pseudo-code>> below.
```
tariqkurd-repo: the idea was just to show what's in the metadata - but the upper and lower bits of T and B muddy the water, so I guess this is ok

tariqkurd-repo: I made a slight adjustment

arichardson (approved):
Much better to read.

PR #1144, opened 2026-07-07
Make misa.y=0 condition implementation defined

tariqkurd-repo:
This is to make the misa.y=0 detection simpler in the hardware.

### DIFF
--- a/src/cheri/riscv-priv-integration.adoc
+++ b/src/cheri/riscv-priv-integration.adoc
@@ -159,6 +159,7 @@ These implicit check registers are <<pcc>>, `__x__tvec`, `__x__epc`.
 NOTE: The list includes <<ddc>> if <<section_cheri_hybrid_ext>> is implemented.
 
 NOTE: This restriction ensures M-mode software cannot drop CHERI checks by modifying <<misa_y,misa>> once they have been configured.
+ The exact behavior of the restriction is implementation defined, and can be as simple as detecting whether any CSR writes or <<pcc>> metadata updates have been executed in {cheri_cap_mode_name}.
 
 Either I or E is _also_ set to show how many registers are present.
 
--- a/src/cheri/riscv-priv-integration.adoc
+++ b/src/cheri/riscv-priv-integration.adoc
@@ -159,7 +159,7 @@ These implicit check registers are <<pcc>>, `__x__tvec`, `__x__epc`.
 NOTE: The list includes <<ddc>> if <<section_cheri_hybrid_ext>> is implemented.
 
 NOTE: This restriction ensures M-mode software cannot drop CHERI checks by modifying <<misa_y,misa>> once they have been configured.
- The exact behavior of the restriction is implementation defined, and can be as simple as detecting whether any  metadata updates to the relevant CSRs, or the <<pcc>>, have been executed at all in {cheri_cap_mode_name}.
+ The exact behavior of the restriction is implementation defined, and does not need to, for example, check that capability bounds have been narrowed and then reprogrammed _back_ to infinity. Detecting any bounds reduction is sufficient.
 
 Either I or E is _also_ set to show how many registers are present.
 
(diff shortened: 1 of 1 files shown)


src/cheri/riscv-priv-integration.adoc
```diff
@@ -159,6 +159,7 @@ These implicit check registers are <<pcc>>, `__x__tvec`, `__x__epc`.
 NOTE: The list includes <<ddc>> if <<section_cheri_hybrid_ext>> is implemented.
 
 NOTE: This restriction ensures M-mode software cannot drop CHERI checks by modifying <<misa_y,misa>> once they have been configured.
+ The exact behavior of the restriction is implementation defined, and can be as simple as detecting whether any CSR writes or <<pcc>> metadata updates have been executed in {cheri_cap_mode_name}.
```
```suggestion
 The exact behavior of the restriction is implementation defined, and can be as simple as detecting whether any  metadata updates to the relevant CSRs, or the <<pcc>>, have been executed at all in {cheri_cap_mode_name}.
```

src/cheri/riscv-priv-integration.adoc
```diff
@@ -159,6 +159,7 @@ These implicit check registers are <<pcc>>, `__x__tvec`, `__x__epc`.
 NOTE: The list includes <<ddc>> if <<section_cheri_hybrid_ext>> is implemented.
 
 NOTE: This restriction ensures M-mode software cannot drop CHERI checks by modifying <<misa_y,misa>> once they have been configured.
+ The exact behavior of the restriction is implementation defined, and can be as simple as detecting whether any  metadata updates to the relevant CSRs, or the <<pcc>>, have been executed at all in {cheri_cap_mode_name}.
```
jrtc27: That's not the same thing as in the normative text?

tariqkurd-repo: That's true - the point is that it covers the case that the normative text needs, and allows more flexibility.
Realistically, software has either set misa.y=0 immediately on boot or isn't going to at all.
Dynamically checking that the PC, DDC etc. are all infinite is annoying, especially for the PC with multiple versions of the metadata in flight in an OoO machine.

it's to allow for simpler implementations which still meet the basic criteria.

jrtc27: I mean, if you're allowing writable misa, you're already accepting that you have to stop your entire pipeline and flush, there's no sensible way that can be done OoO. And the infinite check is easy, you just check against hard-coded bit patterns.

arichardson (commented):
I agree with @jrtc27 I'm not sure this "conservatively prevents updates if any write to one of the CSRs has occurred" is actually worth it.

You have to stall anyway and then checking a few bit patterns shouldn't be too bad (you could always track one more bit per register whether it is infinite)?

tariqkurd-repo:
> I agree with Jessica Clarke (@jrtc27) I'm not sure this "conservatively prevents updates if any write to one of the CSRs has occurred" is actually worth it.
> 
> You have to stall anyway and then checking a few bit patterns shouldn't be too bad (you could always track one more bit per register whether it is infinite)?

you're not thinking about big implementations, I already have a complaint about this one as it's currently worded.

The point is that the design can _choose_ not to allow misa.y to be writeable under more conditions as it's WARL.
The minimum is the normative text.
The maximum is always-1
There's a middle ground where the implementation can choose when to make it writeable.
Simple implementations will do it exactly but more complex ones won't because the PC checking is much more complex.

src/cheri/riscv-priv-integration.adoc
tariqkurd-repo:
```suggestion
 The exact behavior of the restriction is implementation defined, and does not need to, for example, check that capability bounds have been narrowed and then reprogrammed _back_ to infinity. Detecting any bounds reduction is sufficient.
```

arichardson:
If we phrase it as the implementation may be more conservative, but MUST deny it if non-infinite, then this seems fine to me. I can't think of a reason to disable Y for M-mode dynamically? @jrtc27 what do you think?

jrtc27:
Why was this merged without approval when both Alex and I were in agreement that it needed different wording?

jrtc27:
I think you want to phrase the normative text as something like:

> If the implementation supports a writable <<misa_y,misa>>.{CRE}, the implementation must not allow setting it to zero if any register used for implicit CHERI checks is not currently configured as a root capability.

and the non-normative text could be something like:

> A simpler conforming implementation could prevent setting <<misa_y,misa>>.{CRE} to zero if the capability metadata of any register used for implicit CHERI checks has ever been written to, rather than dynamically checking the current metadata.

PR #1147, opened 2026-07-07
Clarify Load Capability Fault conditions

arichardson:
Rewrite the description of pte.yr/pte.yrg for loads and replace the confusing "Optionally, the loaded tag is set" list item with a plain statement of the conservative-trapping allowance.

Also note in the base definition that the reserved bits are allocated by Svyrg, which redefines the whole field when enabled, and state in the Svyrg chapter that with YRGE=0 only pte.rvy[3] (pte.y) has meaning.

Extracted from #1126

### DIFF
--- a/src/cheri/cheri-pte-ext.adoc
+++ b/src/cheri/cheri-pte-ext.adoc
@@ -18,9 +18,10 @@ endif::[]
 
 {cheri_priv_crg_ext} implies Sv39.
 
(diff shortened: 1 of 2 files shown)


arichardson (commented):
@tariqkurd-repo since this is just what you already reviewed in #1126, I assume you are happy with it, but just checking again.

tariqkurd-repo (approved):

PR #1148, opened 2026-07-07
Explain why YBNDSRW sail checks rs1 instead of rounded bounds

arichardson:
The Sail only checks the requested region against rs1's bounds; state
the bounds-rounding containment guarantee that makes this sufficient
so the tag-clearing bullet is not misread as a separate dynamic
check.

Also add a separating blank line before ysunseal page break. The <<< on
the first line was glued onto the trailing paragraph of the preceding
YBNDSRW page and rendered literally.

### DIFF
--- a/src/cheri/insns/scbndsr_32bit.adoc
+++ b/src/cheri/insns/scbndsr_32bit.adoc
@@ -38,3 +38,5 @@ include::malformed_cs1_clear_tag.adoc[]
 Operation::
 +
 sail::execute[clause="SCBNDSR(_, _, _)",part=body,unindent]
+
+NOTE: Checking only the _requested_ region with `inCapBoundsNoWrap` is sufficient, since the rounding performed by `setCapBounds` can never grow the bounds beyond those of `{cs1}`.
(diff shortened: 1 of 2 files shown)


arichardson:
@PeterRugg not sure if this is a good explanation or if we need something more verbose.

src/cheri/insns/scbndsr_32bit.adoc
```diff
@@ -38,3 +38,5 @@ include::malformed_cs1_clear_tag.adoc[]
 Operation::
 +
 sail::execute[clause="SCBNDSR(_, _, _)",part=body,unindent]
+
+NOTE: Checking only the _requested_ region with `inCapBoundsNoWrap` is sufficient, since the rounding performed by `setCapBounds` can never grow the bounds beyond those of `{cs1}`.
```
tariqkurd-repo, line 42: this is just to control Claude really
it almost wants to be a "Claude note!"

tariqkurd-repo (approved):

src/cheri/insns/scbndsr_32bit.adoc
arichardson: Well I think readers could have the same thought so I think it's worth landing.

PR #1149, opened 2026-07-07
Note why the YTYPER Sail reads the sealed boolean

arichardson:
The model stores the 1-bit CT-field as the boolean sealed; wider
CT-field encodings must return the full field value.


arichardson:
Not sure if this is worth adding as a NOTE, but it was pointed out to me that this could be read as the former "CGetSealed" code.

tariqkurd-repo (approved):

PR #1158, opened 2026-07-13
Improve wording for misa.Y writable check

arichardson:
Using Jessica's suggested wording from PR #1144.
No functional change intended, just fix the wording to be correct.

Co-authored-by: Jessica Clarke <jrtc27@jrtc27.com>

### DIFF
--- a/src/cheri/riscv-priv-integration.adoc
+++ b/src/cheri/riscv-priv-integration.adoc
@@ -151,15 +151,12 @@ endif::[]
 When the base ISA is {cheri_base_ext_name}, the Y bit is one.
 
 `<<misa_y,misa>>.{CRE}` is a WARL field.
-If the implementation supports a writable <<misa_y,misa>>.{CRE}, the implementation must not allow setting it to zero if any register used for implicit CHERI checks is not currently configured as a root capability.
-Otherwise, the write to <<misa_y,misa>>.{CRE} is ignored.
+If any of <<pcc>>, `__x__tvec`, or `__x__epc` is not currently configured as a root capability then zero is not a legal value for <<misa_y,misa>>.{CRE}.
 
-These implicit check registers are <<pcc>>, `__x__tvec`, `__x__epc`.
-
-NOTE: The list includes <<ddc>> if <<section_cheri_hybrid_ext>> is implemented.
+NOTE: The list also includes <<ddc>> if <<section_cheri_hybrid_ext>> is implemented.
 
 NOTE: This restriction ensures M-mode software cannot drop CHERI checks by modifying <<misa_y,misa>> once they have been configured.
- A simpler conforming implementation could prevent setting <<misa_y,misa>>.{CRE} to zero if the capability metadata of any register used for implicit CHERI checks has ever been written to, rather than dynamically checking the current metadata.
+ A simpler conforming implementation could prevent setting <<misa_y,misa>>.{CRE} to zero if the capability metadata of any of these registers has ever been written to, rather than dynamically checking the current metadata.
 
 Either misa.I or misa.E is _also_ set to indicate whether 32 or 16 general purpose registers are present.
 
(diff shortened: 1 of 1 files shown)


src/cheri/riscv-priv-integration.adoc
```diff
@@ -151,15 +151,15 @@ endif::[]
 When the base ISA is {cheri_base_ext_name}, the Y bit is one.
 
 `<<misa_y,misa>>.{CRE}` is a WARL field.
-If the implementation supports a writable <<misa_y,misa>>.{CRE}, it can only be set to zero if all registers used for implicit CHERI checks are currently configured as root capabilities.
+If the implementation supports a writable <<misa_y,misa>>.{CRE}, the implementation must not allow setting it to zero if any register used for implicit CHERI checks is not currently configured as a root capability.
 Otherwise, the write to <<misa_y,misa>>.{CRE} is ignored.
```
jrtc27: I think this sentence now doesn't follow, and that's implied by the only different value not being legal. so it should just disappear

src/cheri/riscv-priv-integration.adoc
```diff
@@ -151,15 +151,15 @@ endif::[]
 When the base ISA is {cheri_base_ext_name}, the Y bit is one.
 
 `<<misa_y,misa>>.{CRE}` is a WARL field.
-If the implementation supports a writable <<misa_y,misa>>.{CRE}, it can only be set to zero if all registers used for implicit CHERI checks are currently configured as root capabilities.
+If the implementation supports a writable <<misa_y,misa>>.{CRE}, the implementation must not allow setting it to zero if any register used for implicit CHERI checks is not currently configured as a root capability.
```
jrtc27: Technically the 'If" is a bit redundant, because the second half trivially holds if misa.Y isn't writable. I wonder if the better, RISC-V standard-ese, way to write it would be:


```suggestion
If any register used for implicit CHERI checks is not currently configured as a root capability then zero is not a legal value for <<misa_y,misa>>.{CRE}.
```

src/cheri/riscv-priv-integration.adoc
```diff
@@ -151,15 +151,15 @@ endif::[]
...
 Otherwise, the write to <<misa_y,misa>>.{CRE} is ignored.
 
 These implicit check registers are <<pcc>>, `__x__tvec`, `__x__epc`.
 
 NOTE: The list includes <<ddc>> if <<section_cheri_hybrid_ext>> is implemented.
```
jrtc27: Uh, how can you set misa.Y=0 without hybrid? There's no DDC to authorise loads/stores. If you have an encoding that allows RWX capabilities then you could handwave and say that PCC is authorising all data accesses, but that's really quite horrendous, and I don't see a world in which it's useful to support misa.Y=[01] but not hybrid, at that point hybrid support is basically free to add.

src/cheri/riscv-priv-integration.adoc
tariqkurd-repo: we don't use the phrase "implicit cheri checks" elsewhere - I wonder if it's worth spelling them out more clearly?

jrtc27 (commented):
Apparently I didn't submit my review yesterday, but it looks like I finished it...

src/cheri/riscv-priv-integration.adoc
tariqkurd-repo: it's a good point about not being able to authorise the loads and stores without DDC, which is part of Zyhybrid

arichardson: That would only be possible on a machine that either has "pure RVI/RVE" or RVY purecap-only. You could only switch over once on boot. Not sure if anyone is planning to build such a system, but I think it could make sense to support. You can then have one tapeout and have a fuse and let the bootrom decide whether it's a plain RISC-V or always RVY. However, thinking more about this, such a system would need to start with Y=0 since the there is no P bit that can be set on startup to be running in integer mode.

So in conclusion I think we do need either hybrid, or a system where misa.Y resets to 0. Maybe cleanest to just keep the "you need hybrid to make this writable"?

jrtc27: Yeah, I think we should leave all that kind of stuff up to vendors. After all, if it's hidden in boot ROMs, then nobody's the wiser what crimes are being committed against the architecture.

tariqkurd-repo (approved):

PR #1166, opened 2026-07-16
In int mode cap load/stores can have rs1=x0

### LINKED ISSUE #1165: LY, SY, AMOSWAP.Y, LR.Y, SC.Y inconsistent use of rs1=x0 between modes
We say that in cap mode all load/store/AMOs are reserved if rs1=x0

This is already problematic as I don't know if we can define this correctly in riscv-opcodes as it involves changing _all_ existing and future load/stores.

Additionally though the capability accesses are still reserved in int mode, whereas the other accesses aren't.

To be consistent LY, SY, AMOSWAP.Y, LR.Y, SC.Y should all permit rs1=x0 in int mode.
...or we declare the inconsistency in a note.

Also see https://github.com/riscv/riscv-opcodes/pull/420#issuecomment-4991654306 from https://github.com/riscv/riscv-opcodes/pull/420

tariqkurd-repo:
fix #1165

### DIFF
--- a/src/cheri/riscv-hybrid-integration.adoc
+++ b/src/cheri/riscv-hybrid-integration.adoc
@@ -94,6 +94,10 @@ That is, all data memory accesses are implicitly authorized by <<ddc>> with only
 +
 NOTE: <<ddc>> is also used to authorize memory instructions added by {cheri_base_ext_name} (such as <<LOAD_CAP>> and <<STORE_CAP>>).
 +
+The restriction of not allowing `rs1=x0` is removed and _all_ data memory accesses are permitted to use `x0` as the base address.
++
+NOTE: Capability memory access instructions are also permitted to have `rs1=x0`.
++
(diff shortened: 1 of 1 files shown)


jrtc27 (approved):
(but probably best to wait for Alex to also check)

tariqkurd-repo:
@aswaterman FYI - we've found a spec bug

arichardson:
> Andrew Waterman (@aswaterman) FYI - we've found a spec bug

I would not necessarily call this a spec bug. We say that in the RVY base the rs1=x0 encodings are illegal. The intention of Zyhybrid has always been that we revert back to exactly the same behaviour RVI, so I'd say this is a clarification.
It was always the expected behaviour and is what I implemented in QEMU.

arichardson (approved):

tariqkurd-repo:
It certainly is the intended behaviour, but the spec didn't say it, but it does now 😄 
The spec now reads as intended.

PR #1169, opened 2026-07-20
Add figures to illustrate the bounds decoding correction

PeterRugg:
As requested by @tariqkurd-repo, who felt the correction is currently under-explained.


tariqkurd-repo (approved):

src/cheri/rvy64-encoding.adoc
```diff
@@ -324,6 +324,15 @@ xref:cap_encoding_cb[xrefstyle=short].
 `R` marks the lower boundary of the <<section_cap_representable_check,representable range>> (see <<cap_bounds_map>>).
 If `A`, `T` or `B` is less than `R`, then it lies in the 2^E+MW^ aligned region above `R`, and the corrections _c~t~_ and _c~b~_ adjust the upper bits of the decoded bounds accordingly.
 
+As shown in xref:section_cap_bounds_format[xrefstyle=short] below, the upper bits of the capability address are used to form the upper bits of the bounds.
+Sometimes the bounds need to be adjusted up or down relative to the capability address, because the memory bound and the capability address cross alignment boundaries (address 0x20000 in <<capformat_correction_illustration>>)
```
```suggestion
Sometimes the bounds need to be adjusted up or down relative to the capability address, because the memory bound and the capability address cross alignment boundaries (address 0x20000 in <<capformat_correction_illustration>>).
```

nwf (approved):
Oh nice!  I'm glad to see the one figure I drew ages ago still being of utility, and `capformat-correction-cases` is very helpful.  (That should probably go in the next CHERI ISA doc, if it isn't already there and there is to be such a thing?)

src/cheri/img/capformat-correction-illustration.png
arichardson, line 1: Isn't this almost exactly the same as the existing cap-bounds-map.png just with addr added?

Looking at both of them side by side: I always assumed that the line in cap-bounds-map.png was the current address, but I realize now it's just the aligned region.

I find having the legend on the right is quite helpful, can we add it to the new image? It might also make sense to have the address match?
Can't we check in the source files as well to make it easier to re-generate?

Image

src/cheri/img/capformat-correction-cases.png
arichardson, line 1: Should we drop the "multiple of s = 2^E+MW" from the bottom of this diagram? I find it adds more confusion that clarification?

src/cheri/rvy64-encoding.adoc
```diff
@@ -324,6 +324,15 @@ xref:cap_encoding_cb[xrefstyle=short].
 `R` marks the lower boundary of the <<section_cap_representable_check,representable range>> (see <<cap_bounds_map>>).
 If `A`, `T` or `B` is less than `R`, then it lies in the 2^E+MW^ aligned region above `R`, and the corrections _c~t~_ and _c~b~_ adjust the upper bits of the decoded bounds accordingly.
 
+As shown in xref:section_cap_bounds_format[xrefstyle=short] below, the upper bits of the capability address are used to form the upper bits of the bounds.
+Sometimes the bounds need to be adjusted up or down relative to the capability address, because the memory bound and the capability address cross alignment boundaries (address 0x20000 in <<capformat_correction_illustration>>).
+
+[#capformat_correction_illustration]
+.Example capability bounds that use the correction. The base address bound is an s-aligned region below the address, meaning the upper bits (above the mantissa) of the base are smaller than the upper bits of the address. Therefore c~b~=-1. The address and top address bound are in the same region, so c~t~=0.
```
arichardson: I don't think we define `s` anywhere?

arichardson: Oh it's in the image, never mind.

```suggestion
.Example capability bounds that use the correction. The base is an s-aligned region below the address, meaning the upper bits (above the mantissa) of the base are smaller than the upper bits of the address. Therefore c~b~=-1. The address and top address bound are in the same region, so c~t~=0.
```

src/cheri/img/capformat-correction-illustration.png
arichardson: Thinking about this a bit more, I believe the cap-bounds-map diagram is probably a bit too complicated for where it currently is. I think all we need in this "intro section" is the diagram of the representable space and the derefenceable region without all the other lines. If you point me at the source of that diagram I can try to create a simpler one.

This would also avoid the need for these forward references now that we have the full diagram in the encoding chapter.

```
[#cap_bounds_map]
.Memory address bounds and representable range encoded within a capability
image::../cheri/img/cap-bounds-map.png[width=80%,align=center]

E, MW and R in the figure are all introduced in xref:section_cap_bounds_decoding[xrefstyle=short] along with the bounds decoding.
```

src/cheri/rvy64-encoding.adoc
```diff
@@ -324,6 +324,15 @@ xref:cap_encoding_cb[xrefstyle=short].
 `R` marks the lower boundary of the <<section_cap_representable_check,representable range>> (see <<cap_bounds_map>>).
 If `A`, `T` or `B` is less than `R`, then it lies in the 2^E+MW^ aligned region above `R`, and the corrections _c~t~_ and _c~b~_ adjust the upper bits of the decoded bounds accordingly.
 
+As shown in xref:section_cap_bounds_format[xrefstyle=short] below, the upper bits of the capability address are used to form the upper bits of the bounds.
+Sometimes the bounds need to be adjusted up or down relative to the capability address, because the memory bound and the capability address cross alignment boundaries (address 0x20000 in <<capformat_correction_illustration>>).
```
```suggestion
As shown in xref:section_cap_bounds_format[xrefstyle=short] below, the upper bits of the bound are computed by substituting B or T into the upper bits of the capability address.                          
This substitution is only correct if the memory bound and the capability address lie within the same 2^E+MW^-aligned region.                                                                               
When they do not (address 0x20000 in <<capformat_correction_illustration>>), the correction factors c~t~ and c~b~ adjust the computed upper bits to select the correct region. 
```

How about this wording to avoid using "sometimes" (which I don't think we should have in a specification) and to also avoid "bounds being adjusted". This could be interpreted as adjusting bounds of existing capabilities (e.g. YBNDSW), this makes it more clear that we are changing the way they are being computed.

src/cheri/img/capformat-correction-illustration.png
PeterRugg: Okay, it was deliberate that it matched the existing figure. The idea was to highlight with respect to it exactly what the corrections were doing. I can easily restore the braces.

src/cheri/img/capformat-correction-cases.png
PeterRugg: Maybe... Although I'm not sure how you're supposed to know what s is without it?

src/cheri/rvy64-encoding.adoc
PeterRugg: Ah, I think we used that terminology more heavily in the past, I guess to distinguish from the encoded base and top. I'll remove the references.

arichardson:
There is some trailing whitespace. I have a few ideas on how to simplify this section but let's go merge this first!

arichardson (approved):
Could you fix the trailing whitespace to make ci happy?

arichardson (commented):
Oh and could you add the tikz sources?

PeterRugg:
Okay, I should add the tikz sources, but it will need slight tidy-up (I did some terrible things to specialise the addresses in the cases diagram). I should also check @nwf is happy with the source being shared?

nwf:
> I should also check Nathaniel Wesley Filardo (@nwf) is happy with the source being shared?

I am, but it's probably simplest to treat that source as being under the license of the CHERI ISA rather than something to which I retain copyright, since it was work for hire during my postdoc.  In any case, I think that license is sufficiently permissive.

PeterRugg:
Great, thanks. I'll tidy things up and add them to this PR.

wmat:
Are these images source generated, i.e. LaTeX/TikZ?  Just thinking with respect to maintenance.

PeterRugg:
Okay, done!

arichardson (approved):

arichardson:
> Are these images source generated, i.e. LaTeX/TikZ? Just thinking with respect to maintenance.

Yes the sources have been included now.

PR #1171, opened 2026-07-27
Add G-stage translation note

tariqkurd-repo:
The note saying that G-stage translation tables don't have an RVY field seems to have been forgotten.

### DIFF
--- a/src/cheri/hypervisor-integration.adoc
+++ b/src/cheri/hypervisor-integration.adoc
@@ -10,7 +10,9 @@ IMPORTANT: {not_v1_ratification_package_experimental}
 * The <<HLV_CAP>> and <<HSV_CAP>> encodings are added.
 
 [#hstatus_y,reftext="hstatus ({cheri_base_ext_name})"]
-=== Hypervisor Status Register (hstatus)
+
+NOTE: The {cheri_base_ext_name} field is not currently defined in G-stage PTEs and so remains reserved.
+ A future extension may define this field.
 
 VSXL and VSBE follow the same restrictions as <<mstatus_y>>.SXL and <<mstatus_y>>.SBE respectively.
 
--- a/src/cheri/hypervisor-integration.adoc
+++ b/src/cheri/hypervisor-integration.adoc
@@ -12,7 +12,7 @@ IMPORTANT: {not_v1_ratification_package_experimental}
 [#hstatus_y,reftext="hstatus ({cheri_base_ext_name})"]
 
 NOTE: The {cheri_base_ext_name} field is not currently defined in G-stage PTEs and so remains reserved.
- A future extension may define this field.
+ A future extension may define this field to allow the hypervisor to control access to capabilities for guest OS translations.
 
 VSXL and VSBE follow the same restrictions as <<mstatus_y>>.SXL and <<mstatus_y>>.SBE respectively.
 


jrtc27:
But that's not true? G-stage translations do need to be able to distinguish between memory regions that can and can't hold capabilities (e.g. an emulated MMIO region backed by normal memory that can bear capabilities according to the real PMA still should not be able to do so from the perspective of the guest no matter how they configure their PTEs). What they don't do is include any generation information for revocation, at least on Morello.

tariqkurd-repo:
> But that's not true? G-stage translations do need to be able to distinguish between memory regions that can and can't hold capabilities (e.g. an emulated MMIO region backed by normal memory that can bear capabilities according to the real PMA still should not be able to do so from the perspective of the guest no matter how they configure their PTEs). What they don't do is include any generation information for revocation, at least on Morello.

None of that is specified though. Are you saying that G-stage PTEs should support the Y bit only, so that the G-stage translation can specify tag-strip on loads/fault on stores if and tag is set - if Y=0 else normal operation if Y=1?

Or should they support YW/YD? Half of Svyrg?

jrtc27:
Morello gives you its equivalents of YR, YW, and YD, but not YRG. I can see all of those three having sensible uses, and require nothing special to support, so assumed that's what we had specified. YRG doesn't make sense in the same way, and would require some kind of separate hypervisor generation CSR bit if you were to give it any possibly-reasonable meaning at all. I don't know whether we currently end up relying on YD for Morello, and probably each of YR and YW are set if and only if the other is (though note the bits are backwards from how they're named on Morello due to the base architecture's definition of DBM), but it's not hard to see what you would use them for, and it would be unnecessarily annoying to have to special-case whether to set just Y or both YR and YW on a PTE based on which stage it is.

tariqkurd-repo:
@jrtc27 do you know how they are actually used on Morello?
Certainly, supporting the same set as normal S-mode (Svyrg) would be straight-forward.

>separate hypervisor generation CSR bit

This would be `vsstatus` as seen through `sstatus` I presume?

If YRG isn't supported then presumably it's assumed to be zero if YR=0 (strip loaded tag) or to match sstatus if YR=1 (normal operation).

tariqkurd-repo:
The next question is how you would structure these with extensions. The base+H only having Y and Svyrg adding the other bits would be regular, but we would presumably need a control bit in `hstatus` to control the interpretation of the PTE bits?

jrtc27:
Yes, I would expect H+Svyrg to define YR, YW and YD for both stages of translation, and YRG for VS-stage translation. G-stage translation[^1] YRG, if someone ever comes up with a useful meaning for it, could come as a later extension on top of H+Svyrg, but it has never been used, nor even built, so is not something to specify now.

[^1]: a truly terrible name, given it's the *host's* not the *guest's* translation being applied, but presumably so-called because it's translating the guest as a whole to make it a guest... why it's not just 1 and 2 like Arm I don't know

jrtc27:
> Jessica Clarke (@jrtc27) do you know how they are actually used on Morello? Certainly, supporting the same set as normal S-mode (Svyrg) would be straight-forward.
> 
> > separate hypervisor generation CSR bit
> 
> This would be `vsstatus` as seen through `sstatus` I presume?

No, because vsstatus is what the guest sees/controls, whereas G-stage translations are about what the guest doesn't see/control. As I understand the shape of the H spec, it would be something like hstatus.GYRG.

> If YRG isn't supported then presumably it's assumed to be zero if YR=0 (strip loaded tag) or to match sstatus if YR=1 (normal operation).

Hm, that's a good point, YRG for YR=0 isn't a generation bit, and YR isn't really the read bit. You'd end up in fact with YRG being defined but YR's only legal value being 0, as the "natural" way to remove the CSR bit(s), but that's quite weird.

Maybe the simplest thing *is* to just define hstatus.GYRG...

tariqkurd-repo:
The whole solution would be:

* add all RVY PTE bits to the G-stage PTEs
* add hstatus.GYRG to control the PTEs for G-stage translations to select Y or the set of 4 bits
* add UYRG, SYRG and YRGE into vsstatus (which are already specified)

...if Svyrg is implemented.

If it's not then you just get Y in G-stage PTEs.

That would give a very regular implementation.

avakar:
If you intend to define Svyrg in G-stage, what about `hstatus.YRGE`?

jrtc27:
Right, you would also need a separate enable bit for it, as the host and guest may disagree on whether they want to use it.

jrtc27:
> * add UYRG, SYRG and YRGE into vsstatus (which are already specified)

Is that not already the case, by virtue of vsstatus just being VS-mode's sstatus when in HS-mode?

tariqkurd-repo:
@jrtc27 we've had an internal discussion and I've added #1175 which seems to be simple to add and was clearly missing.
For the other fields (4-bit RVY field in G-stage PTEs), we're concerned that we'll have to add exception types to differentiate between normal G-stage faults and CHERI G-stage faults.
This may well be necessary, but it doesn't seem like it's the right time to do that.

Can we merge this one and come back to the G-stage RVY field in due course?

src/cheri/hypervisor-integration.adoc
```diff
@@ -14,6 +14,9 @@ generally orthogonal to CHERI; the main requirements, when integrating with
 for hypervisors are widened so that they can hold capabilities.
 The remainder of this chapter describes these changes in detail.
 
+NOTE: The {cheri_base_ext_name} field is not currently defined in G-stage PTEs and so remains reserved.
+ A future extension may define this field.
```
```suggestion
 A future extension may define this field to allow the hypervisor to control access to capabilities for guest OS translations.
```

tariqkurd-repo:
I've filed #1178 to record the discussion. I'll merge this PR as it as, as it reflects the current state of the specification.
Adding CHERI fields into G-stage translation is new functionality.

Issue #1172, opened 2026-07-27
[Clarification] - Does changing sstatus.YRGE take effect immediately?


avakar:
Changes to certain control bits that affect the interpretation of PTEs, like SUM or MXR, are explicitly specified to take effect immediately without requiring SFENCE.VMA.

Does changes to YRGE, which affects the interpretation of the RVY field in PTEs, require SFENCE.VMA? (cc @riscv/arc)

Related (merged privileged §12.2.1):
> Changes to the sstatus fields SUM and MXR take effect immediately, without the need to execute an SFENCE.VMA instruction. Changing satp.MODE from Bare to other modes and vice versa also takes effect immediately, without the need to execute an SFENCE.VMA instruction. Likewise, changes to satp.ASID take effect immediately.

avakar:
As a side note, requiring immediate effect is costly under H, as there's no way (at least not one obvious to me) to merge VS-stage and G-stage RVY fields and thus TLBs would have to store all 8 bits.

jrtc27:
Given it changes the interpretation of a few PTE bits, I would expect the better comparison to be ADUE or PBMTE. And unlike SUM/MXR, or satp.ASID, this is not something you ever expect to change dynamically (and if you do, you have to be very careful to ensure that you're not going to shoot yourself in the foot between the CSR write and SFENCE.VMA, including from taking an interrupt, since YW/YD/YRG are illegal to set when YRGE=0 but the moment your implementation decides to start translating with YRGE=1, which can legally be instantly, they are both legal and required for various common operations, in particular anything that wants to store a capability). I think the only question to think about is implications for initialisation code that might not know what the initial state of the hardware is, including any translation caches.

tariqkurd-repo:
It says that sfence.vma is required after changing PBMTE, due to the DTLB caching presumably, but says nothing about ADUE and so changing ADUE must take effect immediately.

We don't currently require an sfence.vma when changing sstatus.YRGE but maybe we should as the results of the translation are cached. But then if all 4 bits are always cached then sstatus.YRGE can be switched dynamically and it functions correctly, just the interpretation changes.

It could be safer to require an sfence.vma, after all, I don't expect YRGE to be changed often, so there's no performance impact to worry about.

tariqkurd-repo:
I've added #1175 which defined hstatus.YRGE. This means you'll presumably need to also store hstatus.YRGE in every PTE entry in the DTLB.
Unless we require sfence.vma.

[...]

PR #1176, opened 2026-07-28
require SFENCE.VMA after changing sstatus.YRGE

tariqkurd-repo:
Because changing sstatus.YRGE changes the interpretation of the RVY field which will be cached in DTLBs, changing the state of it requires an SFENCE.VMA

see discussion in #1172 

@avakar

### DIFF
--- a/src/cheri/cheri-pte-ext.adoc
+++ b/src/cheri/cheri-pte-ext.adoc
@@ -42,6 +42,31 @@ The 4-bit _pte_._{rv64y_pte4_field_name_lc}_  field is subdivided into four 1-bi
 |_pte_._{rv64y_pte4_field_name_lc}[3]_ | _pte_._{pte_cap_dirty_name_lc}_     | Capability dirty
 |===
 
+After changing `sstatus.{pte_cap_fc_enable}`, executing an SFENCE.VMA instruction with _rs1_=`x0` and _rs2_=`x0` suffices to synchronize address-translation caches with respect to the altered interpretation of _pte_._{rv64y_pte4_field_name_lc}_ fields.
+
+[NOTE]
+====
+Setting `sstatus.{pte_cap_fc_enable}` is only required once per OS instance.
+
+If any applications running under the OS require capability access without revocation, then the setting of `sstatus.{pte_cap_fc_enable}=0` and _pte_._{cheri_base_ext_name_lc}_=1 can be replicated with:
+
+* _pte_._{pte_cap_w_name_lc}_ = 1, and
+* _pte_._{pte_cap_dirty_name_lc}_ = 1
+
+And either:
+
+* _pte_._{pte_cap_r_name_lc}_ = 0, and
+* _pte_._{pte_cap_crg_pte_name_lc}_ = 1, and
+
+Or:
+
+* _pte_._{pte_cap_r_name_lc}_ = 1, and
+* _pte_._{pte_cap_crg_pte_name_lc}_ = `sstatus.xYRG`
+
+Therefore `sstatus.{pte_cap_fc_enable}` is not required to change dynamically and so requiring an SFENCE.VMA on a change should not cause a performance problem.
+
+====
+
 When all of the following are true, {cheri_priv_crg_ext} adds two related architectural features, {cheri_excep_name_pte_ld}s and Capability Dirty Tracking:
 
 . `sstatus.{pte_cap_fc_enable}` is set.


tariqkurd-repo:
Thinking about this more, maybe SFENCE.VMA isn't strictly required but I think it's very useful.

If sstatus.YRGE changes 0->1 and PTE.Y=1 across the transition, then the page mapping becomes useless for capability access afterwards.
Therefore software must have reprogrammed the PTE.RVY field before the transition to maintain capability access.
Therefore an SFENCE.VMA is required to make the DTLB coherent, so maybe we don't strictly need to require it for making sstatus.YRGE visible too.
However, it makes the implicit dependency easier to understand and the reality is that SFENCE.VMA still needs to only be issued once to make both changes visible.

@aswaterman do you have an opinion on this one?

aswaterman:
What are the microarchitectural implications of not requiring the SFENCE.VMA (i.e. if the update to YRGE is defined to take effect immediately)?

There are cases where changes to bits that change how permissions checks work are defined to take effect immediately (e.g. sstatus.SUM and sstatus.MXR).  If the microarchitecture implications are similar to that of those fields, then the SFENCE.VMA probably shouldn't be required.  If there's something deeper that makes it hard to implement or makes certain TLB uarches impractical, that's a different story.

tariqkurd-repo:
> What are the microarchitectural implications of not requiring the SFENCE.VMA (i.e. if the update to YRGE is defined to take effect immediately)?
> 
> There are cases where changes to bits that change how permissions checks work are defined to take effect immediately (e.g. sstatus.SUM and sstatus.MXR). If the microarchitecture implications are similar to that of those fields, then the SFENCE.VMA probably shouldn't be required. If there's something deeper that makes it hard to implement or makes certain TLB uarches impractical, that's a different story.

YRGE is not going to change dynamically very often unlike SUM/MXR. Once enabled for a given application to enable revocation, it is very likely to simply stay enabled, I can't think of any cases where it would be disabled again.

The puzzle to solve is the handling of the mapping of the 4-bit PTE.RVY field. If software has already enabled PTE.Y for the page to support capabilities and _then_ enabled YRGE the OS needs to maintain functionality. Therefore it may decide to program PTE.YW=1, YD=1, YR=0 YRG=1 so as not to lose access to capabilities. Essentially PTE.RVY needs to be updated.

Software is free to enable YRGE without reprogramming PTE.RVY but it's not useful.

Reprogramming PTE.RVY requires an SFENCE.VMA to make it visible, as normal, and so whether or not enabling YRGE also _explicitly_ requires it, there must be one. You certainly wouldn't want software to issue  _another_ one due to enabling YRGE.

In the future there may be other mappings of the PTE.RVY field for other revocation algorithms and then the same argument holds.

So I think the choice is how YRGE updates get synchronised in the pipeline.
Either:

1. we rely on software to have issued an SFENCE.VMA to synchronise YRGE, which is very likely to have been issued anyway due to an update to the PTE.RVY field, or
2. the microarchitecture tags issued instructions with YRGE (like MXR/SUM), or 
3. the microarchitecture detects changes to YRGE and internally synchronises

Solution 2 adds area/power.
Solution 3 is similar to changing privilege level, requiring global state to be updated and synchronised, so maybe isn't so expensive. It doesn't need the DTLB to be flushed, just the global value of YRGE to be distributed.
Solution 1 is probably the most elegant.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -23,6 +23,8 @@ The {cheri_priv_crg_ext} extension is enabled when the `sstatus.{pte_cap_fc_enab
 When `sstatus.{pte_cap_fc_enable}`=0, only _pte_._{rv64y_pte4_field_name_lc}[3]_ (_pte_._{pte_cap_rw_name_lc}_) has meaning: it completely enables or disables storing capabilities to the page, as defined in xref:section_pte_rvy[xrefstyle=full].
 When `sstatus.{pte_cap_fc_enable}`=1, the entire _pte_._{rv64y_pte4_field_name_lc}_ field is redefined as described in this chapter.
 
+After changing `sstatus.{pte_cap_fc_enable}`, executing an SFENCE.VMA instruction with _rs1_=`x0` and _rs2_=`x0` suffices to synchronize address-translation caches with respect to the altered interpretation of _pte_._{rv64y_pte4_field_name_lc}_ fields.
```
```suggestion
After changing `sstatus.{pte_cap_fc_enable}`, executing an SFENCE.VMA instruction with _rs1_=`x0` and _rs2_=`x0` suffices to synchronize address-translation caches with respect to the altered interpretation of _pte_._{rv64y_pte4_field_name_lc}_ fields.

NOTE: Changing the state of `sstatus.{pte_cap_fc_enable}` requires software to have already updated the _pte_._rvy_ field before making the change visible.
 Therefore the SFENCE.VMA can be delayed until `sstatus.{pte_cap_fc_enable}` has also changed state and so an additional SFENCE.VMA is _not_ required.

```

aswaterman:
> YRGE is not going to change dynamically very often unlike SUM/MXR. Once enabled for a given application to enable revocation, it is very likely to simply stay enabled, I can't think of any cases where it would be disabled again.

For a given application, sure, but this means we'd be changing YRGE on every context switch between a process with revocation and a process without revocation, right?  Or can you reduce the blast radius to the ASID of the process (assuming there are no Global mappings that are affected)?

> the microarchitecture tags issued instructions with YRGE (like MXR/SUM)

Simple designs (and even some complex ones) just flush the pipeline on sstatus writes so that MXR, SUM, etc. can be treated as global settings.

jrtc27:
> > YRGE is not going to change dynamically very often unlike SUM/MXR. Once enabled for a given application to enable revocation, it is very likely to simply stay enabled, I can't think of any cases where it would be disabled again.
> 
> For a given application, sure, but this means we'd be changing YRGE on every context switch between a process with revocation and a process without revocation, right? Or can you reduce the blast radius to the ASID of the process (assuming there are no Global mappings that are affected)?

No, processes that don't want revocation for whatever reason can just have the OS set the generation in the PTE and CSR to either constant 0 or constant 1, so will always match. The only time it would change would be switching OS (during boot from UEFI to kernel, or kexec, or in a hypervisor context when switching between guests).

aswaterman:
OK, if in practice this setting is changed only a constant number of times per OS instance, then requiring an SFENCE.VMA x0, x0 is fine.  But I guess the TLB uarch needs to associate the setting with the VMID so that an HFENCE.GVMA is not needed when the hypervisor performs a world switch.

tariqkurd-repo:
> OK, if in practice this setting is changed only a constant number of times per OS instance, then requiring an SFENCE.VMA x0, x0 is fine. But I guess the TLB uarch needs to associate the setting with the VMID so that an HFENCE.GVMA is not needed when the hypervisor performs a world switch.

Currently we don't have any G-stage RVY PTE bits defined - as mentioned here #1178 - I'll make a note of note in the issue.

I'll also improve the note in this PR to reflect the discussion.

arichardson (approved):
Looks good to me if everyone else is happy with this wording.

PR #1177, opened 2026-07-28
Make correction factors text easier to understand

arichardson:
Merge the two correction factors tables into one to make it more obvious why only 6 cases are valid.

@PRugg-Cap


nwf (approved):

PeterRugg (approved):
That's helpful: thank you!

src/cheri/rvy64-encoding.adoc
```diff
@@ -317,12 +317,12 @@ guaranteed that the top is larger than the base.
 [#section_cap_bounds_correction_factors]
 ===== Correction Factors
 
-The correction factors used in the bounds calculation, c~t~ and c~b~, are calculated as shown below using the
-definitions in xref:cap_encoding_ct[xrefstyle=short] and
-xref:cap_encoding_cb[xrefstyle=short].
+As described in xref:section_cap_bounds_format[xrefstyle=short] below, the base, _b_, and top, _t_, addresses are computed by substituting `B` and `T` for the upper `MW` bits of the capability address, leaving the address's own 2^E+MW^-aligned block unchanged.
+This substitution alone is only correct if `B`/`T` and the capability address lie in the same 2^E+MW^-aligned block.
+When they do not (address 0x20000 in <<capformat_correction_illustration>>), the correction factors c~t~ and c~b~ shift the substituted block up or down by one so that the computed bound falls in the block that actually contains `b`/`t`.
```
```suggestion
When they do not (address 0x20000 in <<capformat_correction_illustration>>), the correction factors c~t~ and c~b~ shift the substituted block up or down by one so that the computed bound falls in the block that actually contains _b_ or _t_.
```

src/cheri/rvy64-encoding.adoc
```diff
@@ -317,12 +317,12 @@ guaranteed that the top is larger than the base.
...
 
 `R` marks the lower boundary of the <<section_cap_representable_check,representable range>> (see <<cap_bounds_map>>).
-If `A`, `T` or `B` is less than `R`, then it lies in the 2^E+MW^ aligned region above `R`, and the corrections _c~t~_ and _c~b~_ adjust the upper bits of the decoded bounds accordingly.
+Comparing `A`, `T` and `B` against `R` tells us whether each lies in the 2^E+MW^-aligned block containing `R` (`space~L~` in <<cap_bounds_map>>) or the block above it (`space~U~`).
```
```suggestion
Comparing `A` (the bit range of the capability address which gives the offset within the 2^E+MW^-aligned block), `B` and `T` against `R` tells us whether each lies in the 2^E+MW^-aligned block containing `R` (`space~L~` in <<cap_bounds_map>>) or the block above it (`space~U~`).
```

src/cheri/rvy64-encoding.adoc
```diff
@@ -317,12 +317,12 @@ guaranteed that the top is larger than the base.
...
 
 As shown in xref:section_cap_bounds_format[xrefstyle=short] below, the upper bits of the bound are computed by substituting B or T into the upper bits of the capability address.
```
```suggestion
As shown in xref:section_cap_bounds_format[xrefstyle=short] below, the upper bits of the bound are computed by substituting `B` or `T` into the upper bits of the capability address.
```

src/cheri/rvy64-encoding.adoc
```diff
@@ -332,45 +332,39 @@ When they do not (address 0x20000 in <<capformat_correction_illustration>>), the
...
-| A    < R | B    < R | _c~b~_
-| false    | false    | 0
-| false    | true     | +1
-| true     | false    | -1
-| true     | true     | 0
-|==============================================================================
-
-xref:capformat_correction_cases[xrefstyle=short] shows examples of each of the possible cases that arise from this calculation.
+^1^ `B < R` requires `B[MW-1:MW-2] = 0`, which makes `R[MW-1:MW-2] = 3`.
+Since `T[MW-1:MW-2] = B[MW-1:MW-2] + LCout + LMSB` (xref:section_cap_bounds_decoding[xrefstyle=short]) is at most 2 in that case, `T < R` always holds when `B < R`, ruling out c~t~=-1,c~b~=0 and c~t~=0,c~b~=+1.
```
tariqkurd-repo: I think that this needs some human-style description, I'm sure the text is correct but it's difficult to follow in this form.

src/cheri/rvy64-encoding.adoc
```diff
@@ -419,23 +412,10 @@ include ranges which may not be present when the bounds are decoded:
...
-2^E+MW^/4 bytes below
-the base address and at least 2^E+MW^/4 bytes above the top address
-while still allowing the bounds to be correctly decoded.
-
-//[#cap_bounds_map] moved to intro chapter
-//.Memory address bounds encoded within a capability
-//image::../cheri/img/cap-bounds-map.png[width=80%,align=center]
+The compressed bounds encoding allows the address to roam over the _representable_ range, `space~R~`, while maintaining the original bounds  (see xref:section_cap_bounds_correction_factors[xrefstyle=short] above for how
+`R` and the corrections are calculated).
+The overall effect is that the  address can roam `2^E+MW^/4` bytes below the base address and at least  `2^E+MW^/4` bytes above the top address while still allowing the bounds to be
```
```suggestion
The overall effect is that the  address can roam at least `2^E+MW^/4` bytes below the base address and at least  `2^E+MW^/4` bytes above the top address while still allowing the bounds to be
```

tariqkurd-repo (commented):
The text is better but I'm worried that ARC will approve the existing text and then we'll say "we've rewritten it" and they'll have to review it again.

The previous correction factors PR aimed not to change the rest of the text so I'm really not sure we should be doing this.

src/cheri/rvy64-encoding.adoc
arichardson: I can just simplify it to say `B<R` implies `T<R` without these details?

arichardson:
I would classify this as trivial wording improvements (most if this is actually new from the previous PR we merged just last week!)

src/cheri/rvy64-encoding.adoc
tariqkurd-repo: I've made it easier to read.

tariqkurd-repo:
> I would classify this as trivial wording improvements (most if this is actually new from the previous PR we merged just last week!)

Again, the PR last week just added more text it didn't _change_ any.
This rewrites text. It's an improvement but we need to be careful with the process.
The other PRs are fixing errors and typos, not rewording.

A compromise would be to _add_ the combined table only and not do the rest of the rewording. That would be useful.

arichardson:
Thanks for the improved wording. I think we should merge this since the changes are minimal

src/cheri/rvy64-encoding.adoc
```diff
@@ -317,60 +318,57 @@ guaranteed that the top is larger than the base.
 [#section_cap_bounds_correction_factors]
 ===== Correction Factors
 
-The correction factors used in the bounds calculation, c~t~ and c~b~, are calculated as shown below using the
-definitions in xref:cap_encoding_ct[xrefstyle=short] and
-xref:cap_encoding_cb[xrefstyle=short].
+As described in xref:section_cap_bounds_format[xrefstyle=short] below, the base, _b_, and top, _t_, addresses are computed by substituting `B` and `T` for the upper `MW` bits of the capability address, leaving the address's own 2^E+MW^-aligned block unchanged.
```
tariqkurd-repo: >for the upper `MW` bits of the capability address

this isn't true - they go into bit positions [E+MW-1 : E]

src/cheri/rvy64-encoding.adoc
```diff
@@ -317,60 +318,57 @@ guaranteed that the top is larger than the base.
...
+This substitution alone is only correct if `B`/`T` and the capability address lie in the same 2^E+MW^-aligned block.
+When they do not (address 0x20000 in <<capformat_correction_illustration>>), the correction factors c~t~ and c~b~ shift the substituted block up or down by one so that the computed bound falls in the block that actually contains _b_ or _t_.
```
tariqkurd-repo: why address 0x20000? that's right on the boundary and isn't the value used for either b or t, which not pick an address which is actually used?

tariqkurd-repo: see #1184 the diagram is quite weird

tariqkurd-repo: this is an exsiting problem - not related to this PR

src/cheri/rvy64-encoding.adoc
tariqkurd-repo: shouldn't this be something like:

The upper bits of the capablity address determine the  2^E+MW^-aligned block, and then the values of `B` and `T` are substituted into the address to form the bounds.

src/cheri/rvy64-encoding.adoc
```diff
@@ -317,60 +318,57 @@ guaranteed that the top is larger than the base.
...
 
 `R` marks the lower boundary of the <<section_cap_representable_check,representable range>> (see <<cap_bounds_map>>).
-If `A`, `T` or `B` is less than `R`, then it lies in the 2^E+MW^ aligned region above `R`, and the corrections _c~t~_ and _c~b~_ adjust the upper bits of the decoded bounds accordingly.
+Comparing `A` (the bit range of the capability address which gives the offset within the 2^E+MW^-aligned block), `B` and `T` against `R` tells us whether each lies in the 2^E+MW^-aligned block containing `R` (`space~L~` in <<cap_bounds_map>>) or the block above it (`space~U~`).
```
tariqkurd-repo: or also the block below it as correction factors can be +1/0/-1 moving the block up or down

src/cheri/rvy64-encoding.adoc
```diff
@@ -317,60 +318,57 @@ guaranteed that the top is larger than the base.
...
 
-As shown in xref:section_cap_bounds_format[xrefstyle=short] below, the upper bits of the bound are computed by substituting B or T into the upper bits of the capability address.
+As shown in xref:section_cap_bounds_format[xrefstyle=short] below, the upper bits of the bound are computed by substituting `B` or `T` into the upper bits of the capability address.
```
tariqkurd-repo: this is almost a repeat of line 312?

src/cheri/rvy64-encoding.adoc
```diff
@@ -317,60 +318,57 @@ guaranteed that the top is larger than the base.
...
 This substitution is only correct if the memory bound and the capability address lie within the same 2^E+MW^-aligned region.
```
tariqkurd-repo: this is also repeated

src/cheri/rvy64-encoding.adoc
```diff
@@ -317,60 +318,57 @@ guaranteed that the top is larger than the base.
...
 When they do not (address 0x20000 in <<capformat_correction_illustration>>), the correction factors c~t~ and c~b~ adjust the computed upper bits to select the correct region.
```
tariqkurd-repo: repeated again

src/cheri/rvy64-encoding.adoc
tariqkurd-repo: I've just realised that this incorrect text was on line 327 before the edit so it's not a new problem.

tariqkurd-repo:
We need to fix #1184 and #1185 and I agree that the existing text isn't great as it doesn't introduce the concepts in a logical order, so improving this is definitely required. I think it'll take a few passes to get it just right so I've marked it as draft.

This is the most complex area of the spec to need to clearly explain to people, if it's not really good we're going to get questions for years to come about it so it's worth putting the effort in to really getting it good before freeze, after which point there's nothing we can do.

src/cheri/rvy64-encoding.adoc
tariqkurd-repo: filed as #1185

src/cheri/rvy64-encoding.adoc
```diff
@@ -293,17 +293,18 @@ T and B are both MW bits wide, and are formed of different bit ranges:
...
 
-The design rationale for the two EF cases is as follows:
+The design rationale for the two `EF` cases is as follows:
 
 1. `EF = 1`: The exponent is 0. When `enableL8=1`, `L~8~` encodes the MSB of the length,
-which can be used to derive T[MW-1:MW-2], forming a full MW-wide T field.
-2. `EF = 0`: The exponent is _internal_ with E stored in the lower bits of T and
-`B`, with `L~8~` used for the MSB of `E` when `enableL8=1`. E is chosen so that the most
-significant non-zero bit of the length of the region aligns with T[MW - 2] such
+which can be used to derive `T[MW-1:MW-2]`, forming a full MW-wide T field.
```
```suggestion
which can be used to derive `T[MW-1:MW-2]`, forming a full MW-wide `T` field.
```

src/cheri/rvy64-encoding.adoc
```diff
@@ -293,17 +293,18 @@ T and B are both MW bits wide, and are formed of different bit ranges:
...
+2. `EF = 0`: The exponent is _internal_ with E stored in the lower bits of `T` and
+`B`, with `L~8~` used for the MSB of `E` when `enableL8=1`. `E` is chosen so that the most
+significant non-zero bit of the length of the region aligns with `T[MW - 2]` such
 that this bit is implied by `E`.
 
 To save space in the encoding metadata, the top two bits of T are not stored directly but reconstituted as follows:
```
```suggestion
To save space in the encoding metadata, the top two bits of `T` are not stored directly but reconstituted as follows:
```

src/cheri/rvy64-encoding.adoc
```diff
@@ -317,60 +318,53 @@ guaranteed that the top is larger than the base.
...
 ===== Correction Factors
 
-The correction factors used in the bounds calculation, c~t~ and c~b~, are calculated as shown below using the
-definitions in xref:cap_encoding_ct[xrefstyle=short] and
-xref:cap_encoding_cb[xrefstyle=short].
+As described in xref:section_cap_bounds_format[xrefstyle=short] below, the `_base_` and `_top_` addresses are computed by substituting `B` and `T` for the upper `MW` bits of the capability address, leaving the address's own 2^E+MW^-aligned block unchanged.
+This substitution is only correct if `B`/`T` and the capability address lie in the same 2^E+MW^-aligned block.
+When they do not (as with the base in <<capformat_correction_illustration>>), the correction factors c~t~ and c~b~ shift the substituted block up or down by one so that the computed bound falls in the block that actually contains `_base_` or `_top_`.
 
 `R` marks the lower boundary of the <<section_cap_representable_check,representable range>> (see <<cap_bounds_map>>).
```
tariqkurd-repo: isn't this r_base in the diagram? Also no reason to refer back to figure 2 - it's also on the local figure.

arichardson: Oh yes that text also needs updating

src/cheri/rvy64-encoding.adoc
```diff
@@ -313,37 +313,81 @@ T[MW - 1:MW - 2] = B[MW - 1:MW - 2] + LCout + LMSB
...
+
+[#section_cap_bounds_correction_factors, reftext="Correction Factors"]
 ===== Correction Factors
 
-As described in xref:section_cap_bounds_format[xrefstyle=short] below, the `_base_` and `_top_` addresses are computed by substituting `B` and `T` for the upper `MW` bits of the capability address, leaving the address's own 2^E+MW^-aligned block unchanged.
-This substitution is only correct if `B`/`T` and the capability address lie in the same 2^E+MW^-aligned block.
-When they do not (as with the base in <<capformat_correction_illustration>>), the correction factors c~t~ and c~b~ shift the substituted block up or down by one so that the computed bound falls in the block that actually contains `_base_` or `_top_`.
+Substituting `B`/`T` into the capability address, as described in xref:section_cap_bounds_format[xrefstyle=short] above, only decodes to the correct _base_ or _top_ bound if `B`/`T` and the capability address both lie in the _same_ _s_=2^E+MW^-aligned block.
+In this case the relevant _correction factor_ is zero (_c~b~_ for the _base_ and _c~t~_ for the _top_).
```
arichardson: We haven't explained what a correction factor is at this point, I'd just drop this sentence.

tariqkurd-repo:
I've made a big update, and I think 3.1.5.3 and 3.1.5.4 are much easier to read now and are also accurate.

arichardson:
> I've made a big update, and I think 3.1.5.3 and 3.1.5.4 are much easier to read now and are also accurate.

Thanks, I'll take a proper look later but it looks much better IMO

tariqkurd-repo (approved):

src/cheri/rvy64-encoding.adoc
tariqkurd-repo: reworded - will commit shortly.

tariqkurd-repo (approved):

src/cheri/rvy64-encoding.adoc
tariqkurd-repo: I've fixed this.

src/cheri/img/correction-figs/capformat.tex
```diff
@@ -61,9 +67,11 @@
   \coordinate (srbracex) at (\srbracex,0);
     \draw[<-] (repbl |- addry) -- (srseprx |- addry) node [pos=1,anchor=base west] {\textbf{addr}};
   \ifthenelse{\equal{#2}{label}}{
-    \draw[damber] (srseplx |- repbl) -- (srseprx |- repbl) node [pos=1,anchor=base west] {$r_b$};
-    \draw[damber] (srseplx |- reptr) -- (srseprx |- reptr) node [pos=1,anchor=base west] {$r_t$};
-    \draw[<->,thick,damber] (srarrx |- repbl) -- (srarrx |- reptr) node [midway,left]{$s$};
+    \draw[damber] (srseplx |- repbl) -- (srseprx |- repbl) node [pos=1,anchor=base west] {$r_{base}$};
+    \draw[damber] (srseplx |- reptr) -- (srseprx |- reptr) node [pos=1,anchor=base west] {$r_{top}$};
```
nwf: These lines extend to the left of the "memory" for the "s"-annotated line below.  Since `\capannotationsfalse` can hide that, maybe you want to make these stop on the left edge of the "memory" as well in that case?  (Just use `(repbl)` and `(repbl |- reptr)` as the left coordinates, I think.)

src/cheri/rvy64-encoding.adoc
```diff
@@ -312,154 +313,137 @@ T[MW - 1:MW - 2] = B[MW - 1:MW - 2] + LCout + LMSB
...
-* If the top bound is below the alignment boundary and the capability address is above the alignment boundary, then c~t~=-1 to correct the top bound downwards.
-* If the top bound and the capability address are within the same alignment boundary, then c~t~=0, so the top bound is unchanged.
-* If the top bound is above the alignment boundary and the capability address is below the alignment boundary, then c~t~=1 to correct the top bound upwards.
+====== Correction Factor Calculations
 
+`R` marks the lower boundary of the <<section_cap_representable_check,representable range>> (shown as r~base~ in <<capformat_correction_illustration>>).
 
-[#section_cap_bounds_format]
-===== Format of the Bounds
+Comparing `A` (the bit range of the capability address which gives the offset within the 2^E+MW^-aligned block), `B` and `T` against `R` tells us whether each lies in the 2^E+MW^-aligned block containing `R` (`space~L~` in <<cap_bounds_map>>), the block above it (`space~U~`) or the block below (not shown in <<capformat_correction_illustration>>, all possible combinations are shown in <<capformat_correction_cases>>).
```
PeterRugg: I don't think this makes sense as is. There is no address block that "contains R": it's only an MW-bit thing and there are shadows of it in all the regions. The point is you have MW-width B and T in [0, 2^MW), and you want to map them to [R, 2^{MW-1}) (the upper portion) in a low region, and [0, R) (the lower portion) in a high region. Pondering suggested replacement text now.

PeterRugg:
```suggestion
If `B`, `T` or `A` is less than `R`, it is inferred that they lie in the space above the s-alignment boundary straddled by the representable region, labeled space~U~ in xref:capformat_correction_illustration[xrefstyle=short].
```

src/cheri/rvy64-encoding.adoc
```diff
@@ -312,154 +313,137 @@ T[MW - 1:MW - 2] = B[MW - 1:MW - 2] + LCout + LMSB
...
+.4+.^| true  | false | false | _c~t~_=-1, _c~b~_=-1
+             | false | true  | Impossible^1^
+             | true  | false | _c~t~_=0, _c~b~_=-1
+             | true  | true  | _c~t~_=0, _c~b~_=0
+|==============================================================================
 
-xref:top_bound_dec[xrefstyle=short] and xref:base_bound_dec[xrefstyle=short]
-include ranges which may not be present when the bounds are decoded:
+^1^ Since `B[MW-1:MW-2] = 0` in these cases, `R[MW-1:MW-2] = 3`. +
+As a result `T[MW-1:MW-2]` is at most 2 (according to this <<formula_top_bits_of_t,formula>>). +
```
```suggestion
^1^ Since `B[MW-1:MW-2] = 0` in these cases, `R[MW-1:MW-2] = 3` as the value has wrapped. +
As a result `T[MW-1:MW-2]` is at most 2 if LMSB and LCout are both 1 (according to this <<formula_top_bits_of_t,formula>>). +
```

src/cheri/img/correction-figs/capformat.tex
arichardson: Ah thank you! My tikz is very rusty so really appreciate this

PR #1181, opened 2026-07-31
Add combined correction factors table

tariqkurd-repo:
strictly additive subset of #1177 

add the combined correction factors table and the mathematical justification of the impossible cases without changing the surrounding freeze candidate text.


arichardson (changes requested):
I think this is strictly worse than just replacing it. I really struggled to understand the text without my change and this just adds redundancy which ARC really does not like. I still think we should merge #1177  instead

tariqkurd-repo:
closing this in favour of reworking #1177

PR #1187, opened 2026-08-10
Fix document titles

tariqkurd-repo:
They aren't extensions, they are base architectures, as pointed out by Andrew Lindsay


arichardson (approved):
This seems fine to me but we should check with others too

arichardson (changes requested):
Actually, let's not do this without an explicit request from ARC. This document title won't exist anywhere once ratified anyway so it adds churn

PR #1188, opened 2026-08-10
Consistent base/top bound address naming

tariqkurd-repo:
We still had a combination of "top address" and "top bound" so I've gone for "top bound address" everywhere.
Similarly for "base address bound"

### DIFF
--- a/src/cheri/attributes.adoc
+++ b/src/cheri/attributes.adoc
@@ -246,11 +246,11 @@ endif::support_varxlen[]
 
 :GCBASE:       YBASER
 :GCBASE_LC:    ybaser
-:GCBASE_DESC:  Read capability base address
+:GCBASE_DESC:  Read capability base bound address
 
 :GCTOP:       YTOPR
 :GCTOP_LC:    ytopr
-:GCTOP_DESC:  Read capability top address
+:GCTOP_DESC:  Read capability top bound address
 
 :GCTYPE:       YTYPER
 :GCTYPE_LC:    ytyper
--- a/src/cheri/attributes.adoc
+++ b/src/cheri/attributes.adoc
@@ -246,11 +246,11 @@ endif::support_varxlen[]
 
 :GCBASE:       YBASER
 :GCBASE_LC:    ybaser
-:GCBASE_DESC:  Read capability base bound address
+:GCBASE_DESC:  Read capability base address
 
 :GCTOP:       YTOPR
 :GCTOP_LC:    ytopr
-:GCTOP_DESC:  Read capability top bound address
+:GCTOP_DESC:  Read capability top address
 
 :GCTYPE:       YTYPER
 :GCTYPE_LC:    ytyper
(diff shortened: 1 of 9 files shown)


arichardson (commented):
This sounds a bit awkward to me, can we use "base address" everywhere instead?

tariqkurd-repo:
Base address is the base of a load/store though to my mind, so it's a bit ambiguous?

arichardson:
> Base address is the base of a load/store though to my mind, so it's a bit ambiguous?

Ah that's a good point. I guess in most cases it will be something like "the capability's base address" where it's okay.

arichardson:
I think if we are already saying something like "capability base" we don't need to add bound but adding it otherwise sounds good to me.

src/cheri/insns/scbnds_32bit.adoc
```diff
@@ -31,7 +31,7 @@ include::wavedrom/scbnds_32bit.adoc[]
 
 Description::
 Copy the capability from register `{cs1}` to register `{cd}`.
-Set the base address of its bounds to the value of `{cs1}.address` and
+Set the base bound address to the value of `{cs1}.address` and
```
arichardson:
```suggestion
Set the capability base address to the value of `{cs1}.address` and
```

src/cheri/insns/scbndsr_32bit.adoc
```diff
@@ -20,7 +20,7 @@ include::wavedrom/scbndsr_32bit.adoc[]
 
 Description::
 Copy the capability from register `{cs1}` to register `{cd}`.
-Set the base address of its bounds to the value of `{cs1}.address` and
+Set the base bound address to the value of `{cs1}.address` and
```
arichardson:
```suggestion
Set the capability base address to the value of `{cs1}.address` and
```

src/cheri/insns/scbndsrd_32bit.adoc
```diff
@@ -16,10 +16,10 @@ include::wavedrom/scbndsrd_32bit.adoc[]
 
 Description::
 Copy the capability from register `{cs1}` to register `{cd}`.
-Set the base address of its bounds to the value of `{cs1}.address`.
+Set the base bound address to the value of `{cs1}.address`.
```
arichardson:
```suggestion
Set the capability base address to the value of `{cs1}.address`.
```

src/cheri/attributes.adoc
```diff
@@ -246,11 +246,11 @@ endif::support_varxlen[]
 
 :GCBASE:       YBASER
 :GCBASE_LC:    ybaser
-:GCBASE_DESC:  Read capability base address
+:GCBASE_DESC:  Read capability base bound address
```
```suggestion
:GCBASE_DESC:  Read capability base address
```

src/cheri/attributes.adoc
```diff
@@ -246,11 +246,11 @@ endif::support_varxlen[]
...
 
 :GCTOP:       YTOPR
 :GCTOP_LC:    ytopr
-:GCTOP_DESC:  Read capability top address
+:GCTOP_DESC:  Read capability top bound address
```
```suggestion
:GCTOP_DESC:  Read capability top address
```

arichardson (approved):
LGTM with these changes

PR #1199, opened 2026-08-25
Fix Zca hint text and add into standalone spec - ready to merge - need to inform ARC

### LINKED ISSUE #1198: Zca HINTs are missing from standalone spec, and also inaccurate
The Zca hints from #695 have not made it into the standalone spec, so have not been submitted for ARC review.
Additionally this one is inaccurate:

The C.SLLI code points with _shamt_=0 or with _rd_=`x0` are not HINTs if the base architecture is _not_ {cheri_base_ext_name}.

if rd=x0 then it _is_ a hint!

suggested replacement wording:

The C.SLLI code points with with _rd_=`x0` are HINTs.
The C.SLLI code points with _shamt_=0 are HINTs  if the base architecture is _not_ {cheri_base_ext_name}, unless _rd_=`x0`.

tariqkurd-repo:
Fix #1198 

The standalone spec didn't have text about Zca HINTs and the text in the unpriv Zca section was inaccurate.
The unpriv Zca section has been rewritten and so the RVY text will need to be integrated again.


arichardson:
How is RVY different from the base RVI64/32 here? That is not clear from this text and I don't remember.

tariqkurd-repo:
Two encodings are no longer hints. I'll add the names tomorrow.

Sent from Outlook for Android<https://aka.ms/AAb9ysg>

arichardson:
I still don't see anywhere why these are no longer HINTs. What are they reserved for?

tariqkurd-repo:
> I still don't see anywhere why these are no longer HINTs. What are they reserved for?

It says it in the note:

`For non-{cheri_base_ext_name} bases, C.SRLI, C.SRAI and C.SLLI all have code points which are HINTs if _shamt_=0.`

The point is that 16-bit shifts with shamt=0 _are_ hints for non-CHERI, and _are not_ hints for CHERI.
The non-CHERI set is restricted to rd=x0 only.

Is that clear from the PR?

arichardson:
> > I still don't see anywhere why these are no longer HINTs. What are they reserved for?
> 
> It says it in the note:
> 
> `For non-{cheri_base_ext_name} bases, C.SRLI, C.SRAI and C.SLLI all have code points which are HINTs if _shamt_=0.`
> 
> The point is that 16-bit shifts with shamt=0 _are_ hints for non-CHERI, and _are not_ hints for CHERI. The non-CHERI set is restricted to rd=x0 only.
> 
> Is that clear from the PR?

Yes I see that, but if they are no longer HINTS, are they mapped to actual instructions or are they reserved for future instructions? If so what are they going to be allocated for?

tariqkurd-repo:
> Yes I see that, but if they are no longer HINTS, are they mapped to actual instructions or are they reserved for future instructions? If so what are they going to be allocated for?

They're just pointless instructions. All hints are where the natural operation of the instruction has no architectural effect and in these cases there is an effect for RVY but not for RVI. So they are just "uninteresting" as opposed to "hints". It just means we can't attach any additional hint behaviour to them for RVY, but they can for RVI.

arichardson:
> > Yes I see that, but if they are no longer HINTS, are they mapped to actual instructions or are they reserved for future instructions? If so what are they going to be allocated for?
> 
> They're just pointless instructions. All hints are where the natural operation of the instruction has no architectural effect and in these cases there is an effect for RVY but not for RVI. So they are just "uninteresting" as opposed to "hints". It just means we can't attach any additional hint behaviour to them for RVY, but they can for RVI.

Ah thanks for clarifying that I recall now. Can we add that as a clarification? Does the same happen for any other HINTs?

tariqkurd-repo:
It is just a clarification, and is the same as for all hints, we don't need to do anything special except that point out that some RVI hints are _not_ RVY hints.

arichardson:
I think we just need text here to say something like the following:

> For RVI there is no observeable side-effect when imm=0, so they can be used as HINTS.
> However, for RVY using those instructions actually invalidates the metadata and therefore is not a HINT.

Also looking at https://riscv.github.io/riscv-isa-manual/snapshot/spec/#rvc-t-hints, I think the same also applies to C.ADDI with imm=0?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -1298,6 +1298,22 @@ include::insns/jalr_16bit.adoc[]
...
+
+For {cheri_base_ext_name} _all_ arithmetic HINT instructions require _rd_=`x0`.
+
+[NOTE]
+====
+For non-{cheri_base_ext_name} bases, C.SRLI, C.SRAI and C.SLLI all have code points which are HINTs if _shamt_=0.
+
+As a result, for {cheri_base_ext_name} bases:
+
+ * No code points of C.SRLI and C.SRAI are HINTs.
```
```suggestion
 * No code points of C.ADDI, C.SRLI, and C.SRAI are HINTs.
```

tariqkurd-repo:
> I think we just need text here to say something like the following:
> 
> > For RVI there is no observeable side-effect when imm=0, so they can be used as HINTS.
> > However, for RVY using those instructions actually invalidates the metadata and therefore is not a HINT.
> 
> Also looking at https://riscv.github.io/riscv-isa-manual/snapshot/spec/#rvc-t-hints, I think the same also applies to C.ADDI with imm=0?

Yes I'd missed that one. Added now. Thanks!

arichardson:
ARC suggests we mark these as RESERVED for now. This will make it possible to align with future RVI usages of those opcodes.

arichardson (changes requested):
Need to change from HINT to RESERVED

tariqkurd-repo:
Also filed https://github.com/riscv/riscv-isa-manual/pull/3415 which I spotted during this.

arichardson (approved):
I pushed a bit of a simplification to the text, let me know if it looks good to you

PR #1200, opened 2026-08-25
Put VM translation process into standalone spec - ready to merge - need to inform ARC

### LINKED ISSUE #1201: VM translation process is missing from the standalone spec
#1200

tariqkurd-repo:
The VM translation process was only in the standalone spec and so had not been submitted for ARC review.
Fix #1201


src/cheri/cheri-pte-ext.adoc
```diff
@@ -197,3 +197,51 @@ Checking the stored {ctag} is less of a burden to the implementation than checki
 NOTE: Capability dirty tracking may be used to implement the store-barrier primitive from cite:[cornucopia-reloaded].
 
 NOTE: The minimum level of PTE support is to set _pte_._{pte_cap_r_name_lc}_=1, _pte_._{pte_cap_dirty_name_lc}_=1, _pte_._{pte_cap_w_name_lc}_=1 and _pte_._{pte_cap_crg_pte_name_lc}_=0 in all PTEs intended for storing capabilities (e.g., private anonymous mappings) and set `sstatus.U{pte_cap_crg_pte_name}`=0 and `sstatus.S{pte_cap_crg_pte_name}`=0 on all harts, which will enable capabilities to be loaded and stored successfully.
+
+[[sv32algorithm_rvy]]
+=== Virtual Address Translation Process
+
+NOTE: This section is copied and modified from the standard translation process.
```
arichardson, line 204: Instead of duplicating the whole text I think we should just do 'inserted etween base step 6 and 7', and the same for the final one? It's hard to see what's different otherwise.

tariqkurd-repo: Yeah maybe, or maybe I can put it in bold.
It's temporary but needs to be useful.

arichardson: Yes that would also work.

tariqkurd-repo: done

arichardson (approved):
LGTM

PR #1208, opened 2026-09-02
Clarify capability CSR WARL behaviour - need to inform ARC

### LINKED ISSUE #1207: specify what happens if an WARL CSR is extended (affects xtvec, xepc)
JALR can jump to a bad cap (e.g. reserved bit set, no X-perm, tag clear) and that value needs to be reported in Xepc.
Because we removed exception checks from the JALR itself, the exception is taken at the target, and so a redirect is taken to a bad PCC value which can't be executed.
This gives a loop from the execution stages to the frontend to handle the redirection to the bad target, and then to report the CHERI instruction fetch fault.

The identified problem is that the frontend now needs to see the entire PCC value including all 22(ish) reserved bits, and the X flag. In a small core this is fine, in a big core it's not good due to the number of PCC values in flight. We don't want multiple copies of the reserved bits and X-bit which are not needed for any other reason.

Xtvec already has the statement: `The fields in the metadata are WARL as many fields can be implemented as constants.` It doesn't _however_ specify what happens if a value is written which breaks that rule.

Therefore we need text along the lines of `It need not be capable of holding all possible invalid metadata values. Prior to writing XXX, implementations may convert an invalid metadata value into some other invalid metadata value that XXX is capable of holding.`

...and to apply that to Xtvec and Xepc.

That way the frontend has the option of just storing a single bit saying "faulty" on each PCC and report a single version of "faulty" metadata instead of preserving the whole thing.

Thoughts?

tariqkurd-repo:
fix #1207 

Make it explicit that if a WARL CSR is widened to a capability then the whole capability is WARL.

Also consistenty uses "widened CSR" instead of "extended CSR".

This is a clarification because xtvec, xepc are already WARL CSRs, and now the WARL behaviour explicitly applies to the metadata too instead of that needing to be inferred.

It also now specifies that the written tag must be cleared if the capability can't programmed without modification. This is simpler than the standard "require YADDRW semantics".

Making the metadata explictly WARL allows xepc, and so all PCC values within the core, not to need to store any reserved bits or the X-bit. This means that JALR to a capability with reserved bits sets or X-bit clear only needs to clear the tag and take a fault at the target, it doesn't need to report any reserved bits or the state of the X-bit if xepc has all reserved bits as read-only-zero and the X-bit as 1 as the specified WARL behaviour. For large cores this will give considerable area saving as every PCC value stores 23 bits fewer.

This is also useful for CHERIoT as they always set mepc.tag=0, which is WARL behaviour.


tariqkurd-repo:
I plan to convert this to a single statement about WARL capabilities in one PR (clarification) and a second one declaring mepc as WARL (spec change).

src/cheri/riscv-priv-integration.adoc
```diff
@@ -20,18 +20,15 @@ from which it can derive a capability to the address in this register.
...
 NOTE: <<mtvec_y>> exists in all CHERI implementations, and so may be used as a source of a <<root-rx-cap>> capability after reset.
 
 .Machine-mode trap-vector base-capability register
 include::img/mtveccreg.edn[]
 
-The fields in the metadata are WARL as many fields can be implemented as constants.
-
-NOTE: Examples of WARL behavior include always setting <<x_perm>> to 1 and setting the reserved fields to zero, otherwise the capability is unusable.
- Another example is to partially or fully restrict the bounds to constant values.
+NOTE: Implementations may use the WARL behavior to restrict the metadata to values valid for code pointers.
```
```suggestion
NOTE: Implementations may use the WARL behavior to restrict the metadata to only values valid for code pointers.
```

src/cheri/riscv-priv-integration.adoc
```diff
@@ -91,6 +88,11 @@ include::img/mepccreg.edn[]
 .  `mepc.address[1]=0` when IALIGN is fixed to 32
 .  `mepc.address[1]` reads as zero when IALIGN is programmable and is set to 32
 
+NOTE: Implementations may use the WARL behavior to restrict the metadata to values valid for code pointers.
```
```suggestion
NOTE: Implementations may use the WARL behavior to restrict the metadata to values only valid for code pointers.
```

src/cheri/riscv-priv-integration.adoc
```diff
@@ -79,7 +76,7 @@ include::img/mscratchcreg.edn[]
 [#mepc_y,reftext="mepc ({cheri_base_ext_name})"]
 ==== Machine Exception Program Counter Capability (mepc)
 
-The `mepc` register is widened to a capability.
+The WARL `mepc` register is widened to a capability.
```
arichardson, line 81: I guess this doesn't hurt, but not sure it's needed.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -822,38 +822,47 @@ All CSRs that can hold addresses are widened to YLEN bits.
...
-(e.g., _mtvec_ from the privileged specification).
+e.g., _xtvec_ and _xepc_ from the privileged specification.
 endif::[]
 ifndef::cheri_ratification_v1_only[]
 (e.g., _jvt_ from the "Zcmt" extension).
 +
 NOTE: Zcmt is not yet available for {cheri_base_ext_name}.
 endif::[]
++
+When a WARL CSR is widened it is possible that writing a capability to it will be corrupted, as some address bits may be modified on writing.
```
arichardson: THis all seems overly complicated, can't we just say the WARL behaviour also extends to the metadata and tag?

src/cheri/riscv-priv-integration.adoc
tariqkurd-repo: Yeah - it's reinforcement to point out that the metadata is WARL. It's not needed but useful.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -776,13 +776,21 @@ These do not contain pointers (e.g., _fcsr_ from the "F" extension).
...
 +
 NOTE: Zcmt is not yet available for {cheri_base_ext_name}.
 endif::[]
++
+When a WARL CSR is widened the metadata and {ctag} also become WARL.
+If the final written capability value cannot be written unmodified to the CSR due to WARL behavior then the capability value is considered to be _invalid_ and so must be written with its {ctag} set to zero.
++
+NOTE: Examples of WARL behavior for metadata are implementations requiring all fields to be in a valid (non-reserved) state, and also to fix specific permissions to certain values.
++
+WARNING: As _xepc_ CSRs are WARL, implementations should take care not to fix too many fields so that debug becomes difficult.
```
arichardson: I think this WARNING should be with xepc instead, and maybe merged with the NOTE there?

arichardson:
ARC did not object to this one. Let's give everyone a few more days to review this once conflicts are resolved and then merge before their next meeting.

PR #1209, opened 2026-09-03
Arc review feedback 20260902 - Non-normative changes only

arichardson:
This addresses most of the comments on https://drive.google.com/file/d/1O0UvR3vYf-R7vCZmqkrr-NfLlD-5mZQR/view?usp=sharing&resourcekey=0-yExBSDXWl2DmjsjQCYYeRA.

Should have addressed almost all PDF annotations. The following are open:

- naming of YSUNSEAL. ARC suggested YUNSEAL but that conflicts with the general one. Maybe we need a "S" suffix?
- Early introduction of Program Counter Capability (PCC)
- Names for privileged extensions
- Renaming YSENTRY->YSEALE. This is a normative change so needs to be a separate PR

### DIFF
--- a/src/cheri/app-hybrid-usage.adoc
+++ b/src/cheri/app-hybrid-usage.adoc
@@ -229,8 +229,7 @@ This allows a kernel to run non-CHERI-aware code in a strict integer environment
 Transitions between modes are performed using specific instructions:
 
 <<MODESW_INT>> and <<MODESW_CAP>>::
-These instructions are used to switch between modes without changing the control flow.
-They are typically used when entering or leaving a section of code that requires a different mode.
+These instructions are used to switch between modes without changing the control flow and are typically used when entering or leaving a section of code that requires a different mode.
 <<JALR_CHERI>>::
 When jumping to a target capability with the <<p_bit>> set differently from the current mode, the jump performs a mode switch.
 This is commonly used for function calls between compartments or libraries in different modes.
--- a/src/cheri/cheri-pte-ext.adoc
+++ b/src/cheri/cheri-pte-ext.adoc
@@ -2,7 +2,7 @@
 [#section_cheri_priv_crg_ext]
 == "{cheri_priv_crg_ext}" Extension, Version 1.0 for {cheri_base64_ext_name}
 
-This extension specifies use of the {rv64y_pte4_field_name} field in the Sv39, Sv48 and Sv57 PTE formats (e.g., see <<sv39pte>>).
+This extension specifies use of the {rv64y_pte4_field_name} field in the Sv39, Sv48, and Sv57 PTE formats (e.g., see <<sv39pte>>).
 
 ifdef::cheri_standalone_spec[]
 
@@ -70,7 +70,7 @@ Therefore `sstatus.{pte_cap_fc_enable}` is not required to change dynamically an
 When all of the following are true, {cheri_priv_crg_ext} adds two related architectural features, {cheri_excep_name_pte_ld}s and Capability Dirty Tracking:
 
 . `sstatus.{pte_cap_fc_enable}` is set.
-. The authorizing capability has <<c_perm>>.
+. The authorizing capability grants <<c_perm>>.
 . The PMA is set to _CHERI {ctag_title}_.
 
 NOTE: It may make sense to subdivide this into two extensions as _Capability Dirty Tracking_ is more generally useful than _{cheri_excep_name_pte_ld}s_.
@@ -196,7 +196,7 @@ Checking the stored {ctag} is less of a burden to the implementation than checki
 
 NOTE: Capability dirty tracking may be used to implement the store-barrier primitive from cite:[cornucopia-reloaded].
 
-NOTE: The minimum level of PTE support is to set _pte_._{pte_cap_r_name_lc}_=1, _pte_._{pte_cap_dirty_name_lc}_=1, _pte_._{pte_cap_w_name_lc}_=1 and _pte_._{pte_cap_crg_pte_name_lc}_=0 in all PTEs intended for storing capabilities (e.g., private anonymous mappings) and set `sstatus.U{pte_cap_crg_pte_name}`=0 and `sstatus.S{pte_cap_crg_pte_name}`=0 on all harts, which will enable capabilities to be loaded and stored successfully.
+NOTE: The minimum level of PTE support is to set _pte_._{pte_cap_r_name_lc}_=1, _pte_._{pte_cap_dirty_name_lc}_=1, _pte_._{pte_cap_w_name_lc}_=1, and _pte_._{pte_cap_crg_pte_name_lc}_=0 in all PTEs intended for storing capabilities (e.g., private anonymous mappings) and set `sstatus.U{pte_cap_crg_pte_name}`=0 and `sstatus.S{pte_cap_crg_pte_name}`=0 on all harts, which will enable capabilities to be loaded and stored successfully.
 
 [[sv32algorithm_rvy]]
 === Virtual Address Translation Process
--- a/src/cheri/debug-integration.adoc
+++ b/src/cheri/debug-integration.adoc
@@ -83,7 +83,7 @@ If `write=1` and `aamsize{ne}7` then {ctag}s are always written as zero.
 
 If a Program Buffer is implemented, this section shows how it can be configured and used to access the {ctag}s.
 
-The minimum Program Buffer configuration to support {ctag} access in general purpose registers is as follows:
+The minimum Program Buffer configuration to support {ctag} access in general-purpose registers is as follows:
 
 * `hartinfo` must be implemented.
 * `hartinfo.nscratch` is at least 1 and the <<dscratch0_y>> register is implemented.
@@ -92,7 +92,7 @@ The minimum Program Buffer configuration to support {ctag} access in general pur
 
 [NOTE]
 ====
-These requirements allow a debugger to read and write {ctag}s in general purpose registers without disturbing other registers.
+These requirements allow a debugger to read and write {ctag}s in general-purpose registers without disturbing other registers.
 These requirements may be relaxed if _Access Register_ abstract commands support full capability access.
 ====
 
@@ -147,7 +147,7 @@ no way to discover the former property.
 
 [NOTE]
 ====
-Accessing capabilities held in general purpose registers is a foundational operation that
+Accessing capabilities held in general-purpose registers is a foundational operation that
 allows for more complex debugger accesses.
 
 If the full capability access via _Access Register_ abstract command is not
@@ -158,7 +158,7 @@ through the Program Buffer.
 Analogously, if the full capability access is not available through the _Access Memory_ abstract command, the debugger may access capabilities in memory indirectly
 by executing `ly` and `sy` instructions via the Program Buffer.
 
-Any general purpose registers that serve as temporaries during these operations
+Any general-purpose registers that serve as temporaries during these operations
 must be saved by the debugger before use and restored afterwards.
 ====
 
--- a/src/cheri/debug-integration.adoc
+++ b/src/cheri/debug-integration.adoc
@@ -188,14 +188,12 @@ to this specification. The `pc` metadata has no architectural effect in debug
 mode. Therefore, `ASR-permission` is implicitly granted for access to all CSRs for
 instruction execution.
 
-On debug mode entry, <<dpc_y>> is updated with the
-capability in the `pc`, whose address field equals to the address of the next
-instruction to be executed upon debug mode exit as described in the _RISC-V Debug Specification_.
+On debug mode entry, <<dpc_y>> is updated with the capability in the `pc`, whose address field equals to the address of the next instruction to be executed upon debug mode exit as described in the _RISC-V Debug Specification_.
 
-When leaving debug mode, the value in <<dpc_y>> is unsealed if it is a {sentry_cap_uc} and is written to PC.
+When leaving debug mode, the value in <<dpc_y>> is unsealed if it is a {sentry_cap_uc} and is written to the `pc`.
 
-NOTE: A debugger may write <<dpc_y>> to change where the hart resumes and its mode, permissions, sealing or bounds.
- The value saved from the `pc` on debug mode entry cannot be sealed.
+NOTE: A debugger may write <<dpc_y>> to change where the hart resumes and its mode, permissions, sealing, or bounds.
+The value saved from the `pc` on debug mode entry cannot be sealed.
 
 The legalization of <<dpc_y>> follows the same rules described for `mepc`.
 
--- a/src/cheri/insns/dret.adoc
+++ b/src/cheri/insns/dret.adoc
@@ -18,9 +18,8 @@ It unseals <<dpc_y>> if it is sealed as a {sentry_cap_uc} and writes the result 
 It also restores `ddc` from <<dddc>>.
 
 NOTE: The `DRET` instruction is a documented method of exiting debug mode.
-However, it is a pseudoinstruction to return that technically does not execute
-from the program buffer or memory. It does not require the `pc` to
-grant `ASR-permission` so it never raises an exception.
+However, it is a pseudoinstruction that technically does not execute from the program buffer or memory.
+It does not require the `pc` to grant `ASR-permission`, so it never raises an exception.
 
 NOTE: The definition of `DRET` is not part of Sdext, and is suggested as an implementation choice in the _RISC-V Debug Specification_.
 This specification assumes the recommended specification including the recommended encoding.
--- a/src/cheri/introduction.adoc
+++ b/src/cheri/introduction.adoc
@@ -46,9 +46,9 @@ This section lists new extensions available to {cheri_base_ext_name}, and also e
 Almost all existing RISC-V extensions can be added to an implementation, but in most
 cases they will have some behavioral differences and/or new instructions operating on capabilities.
 
-==== Stable Extensions and Specifications
+==== Extensions and Specifications
 
-.Unprivileged stable {cheri_base_ext_name} base ISAs and extensions
+.Unprivileged {cheri_base_ext_name} base ISAs and extensions
 [#unpriv-extension-status]
 [options=header,align=center,width="90%"]
 |=============================================================================================================================================================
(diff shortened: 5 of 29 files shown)


src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -278,6 +278,8 @@ For example, a hypothetical capability with base 2^XLEN^ - 1 and top 2^XLEN^ + 1
 A capability with architectural top > 2^XLEN^ is impossible for a CHERI core to generate and is only possible due to a logic or memory fault (similar to malformed bounds; see xref:section_cap_malformed[xrefstyle=short]).
 Unlike malformed bounds, the top overflowing is not treated as a special case in the architecture: normal bounds check rules should be followed.
 
+NOTE: Requiring hardware to explicitly detect top > 2^XLEN^ values and raise a dedicated exception would add hardware complexity in critical paths for no practical benefit, as such values cannot be produced by valid capability operations.
```
arichardson: Not quite sure about this one. I assume we added this to avoid having to add extra logic to the bounds checking path? @PeterRugg @tariqkurd-repo 

The comment was:

"Why not just raise exception on bad top values? Shouldn't it be treated same as any malformed capability?"

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -331,6 +331,7 @@ They can establish a form of control-flow integrity between mutually distrusting
 Because they are immutable, the jump target cannot be modified, ensuring that control flow can only enter at the specific address encoded in the capability and not elsewhere in a function.
 
 A sealed entry point capability may be passed as the `{cs1}` argument to <<JALR_CHERI>>, and is _unsealed_ on dereference.
+The unsealed capability is written to the <<pcc>>, while the original `{cs1}` register is not changed.
```
tariqkurd-repo: this is a bit weird - the original register is _never_ changed on instruction execution - I guess I'm missing something here.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -201,7 +201,7 @@ NOTE: The _length_ and _top_ are (XLEN+1)-bit values, so <<GCLEN>> and <<GCTOP>>
 
 ==== Deriving New Bounds
 
-On system initialization, one or more <<root-cap>> capabilities are available.
+On initialization of the execution environment, the execution environment defines how the <<root-cap>> capabilities are made available to the execution context.
```
tariqkurd-repo: duplication of "the execution environment" is a bit strange

src/cheri/introduction.adoc
```diff
@@ -67,18 +67,18 @@ endif::[]
...
-|<<{rvy_i_mod_file_name},     {rvy_i_mod_ext_name}>>        | RVI instructions modified by {cheri_base_ext_name}
-|<<{rvy_zca_mod_file_name},   {rvy_zca_mod_ext_name}>>      | C instructions modified by {cheri_base_ext_name}
-|<<section_cheri_vector_integration, V ({cheri_base_ext_name} modified behavior)>> | V instructions modified by {cheri_base_ext_name}
-|<<{rvy_zicbom_mod_file_name},{rvy_zicbom_mod_ext_name}>>   | Zicbom instructions modified by {cheri_base_ext_name}
-|<<{rvy_zicbop_mod_file_name},{rvy_zicbop_mod_ext_name}>>   | Zicbop instructions modified by {cheri_base_ext_name}
-|<<{rvy_zicboz_mod_file_name},{rvy_zicboz_mod_ext_name}>>   | Zicboz instructions modified by {cheri_base_ext_name}
-|<<{rvy_zicsr_mod_file_name}, {rvy_zicsr_mod_ext_name}>>    | Zicsr instructions modified by {cheri_base_ext_name}
+| Base ISA or Extension | Differences
+|<<{rvy_i_mod_file_name},     I>>      | RVY redefines the behavior of control flow and RVI loads/stores
+|<<{rvy_zca_mod_file_name},   Zca>>    | Compressed control-flow and memory instructions operate on capabilities
```
tariqkurd-repo: also C.ADDI4SPN and C.ADDI16SP, so the comment isn't quite right.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -20,10 +20,12 @@ For example, in {rv64y_uni_base_name}:
 
 . `XLEN=64`
 . `L` denotes a long base name.
+. When `I` or `E` is omitted, the default is 32 integer registers (`I`); `E` specifies 16 integer registers (e.g., `RV64LEYA`).
 . `Y` denotes a CHERI base.
-. `A` denotes the capability encoding format.
+. `A` denotes the capability encoding format; the capability encoding format letter after `Y` is mandatory.
```
tariqkurd-repo: is it restricted to being a single letter?

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: I think this will need discussion - @PeterRugg and @jonwoodruff are the experts here.

src/cheri/cheri-pte-ext.adoc
```diff
@@ -142,7 +142,7 @@ NOTE: This is equivalent to the behavior of the _pte_._d_ bit except that _pte_.
 Capability dirty tracking behavior is enabled when, in addition to the rules above, all of the following are true:
 
 . A capability store or AMO instruction is executed.
-. The to-be-stored {ctag} is set.
+. The to-be-stored {ctag} is set to one.
 . _pte_._{pte_cap_w_name_lc}_ is set.
```
tariqkurd-repo: also needs to be "set to one"?

src/cheri/cheri-pte-ext.adoc
```diff
@@ -142,7 +142,7 @@ NOTE: This is equivalent to the behavior of the _pte_._d_ bit except that _pte_.
...
 . _pte_._{pte_cap_dirty_name_lc}_ is clear.
```
tariqkurd-repo: should be "set to zero"?

tariqkurd-repo:
Looks really good Alex! Only a few minor comments. Not sure about top-overflow though.

src/cheri/scripts/rvy_instruction_encodings.py
```diff
@@ -528,7 +528,7 @@ def next_r2type(name: str, rd: str, rs1: str, **kwargs):
         next_rtype("YSS", rs1="{cs1}", rs2="{cs2}", rd="rd"),
         next_rtype("YSUNSEAL", rs1="{cs1}", rs2="{cs2}", rd="{cd}"),
         next_rtype("YBLD", rs1="{cs1}", rs2="{cs2}", rd="{cd}"),
-        next_rtype("YSENTRY", rs1="{cs1}=0", rs2="{cs2}", rd="{cd}", rs1_label="src1=0"),
+        next_rtype("YSEALE", rs1="{cs1}=0", rs2="{cs2}", rd="{cd}", rs1_label="src1=0"),
```
tariqkurd-repo: I would prefer a separate PR for instruction renaming, to separate out normative changes from improved documentation.
That way it will be easier to identify the normative changes.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -19,11 +19,13 @@ The _capability encoding format_ gives full details of how CHERI capabilities ar
...
+. `L` denotes a long base name.
+. When `I` or `E` is omitted, the default is 32 integer registers (`I`); `E` specifies 16 integer registers (e.g., `RV64LEYA`).
 . `Y` denotes a CHERI base.
-. `A` denotes the capability encoding format.
+. `A` denotes the capability encoding format; the capability encoding format letter after `Y` is mandatory.
 
 NOTE: This naming scheme is based on the RISC-V https://lists.riscv.org/g/tech-unprivileged/topic/longer_base_name_proposal/116854896["longer base name" proposal], which allows base ISA names longer than two characters.
+Making the capability encoding format letter after `Y` mandatory simplifies future long base naming.
 
 NOTE: Future RV64 CHERI capability encoding formats could be named RV64LYB, RV64LYC, and so on, or they may have more descriptive names after the RV64LY prefix.
```
tariqkurd-repo, line 30: this note is incompatible with it being a single letter - I think we need to clarify this one.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -82,19 +83,21 @@ Specific capability encoding formats may support different ISA extensions and ad
...
+
+(draw-box "CT"    {:span 2})
+(draw-box ""      {:span 2 :borders {}})
+(draw-box "Metadata (bounds, permissions, etc.)"    {:span 32})
+(draw-box "Address"     {:span 32})
+
+(draw-box "1 bit" {:span 2 :borders {}})
+(draw-box ""      {:span 2 :borders {}})
+(draw-box "XLEN bits" {:span 32 :borders {}})
+(draw-box "XLEN bits" {:span 32 :borders {}})
```
tariqkurd-repo, line 97: I've imported this into #1212 and also replaced CT with CTag and applied the same change to all diagrams, so I think this edit can be reverted here.

src/cheri/riscv-unpriv-integration.adoc
```suggestion
. `A` denotes the capability encoding format; the capability encoding format after `Y` is mandatory.
```

No maybe better to just say format is required then we can have longer ones later.

src/cheri/riscv-unpriv-integration.adoc
arichardson: Yeah not sure if this is really needed. The comment was "This should be clarified to state that the unsealed capability is written to the PC. The original rs1 argument register is not changed."

Maybe the objection was just to the phrasing `is _unsealed_`.

arichardson: Maybe we just use "an unsealed version of this capability is written to the <<pcc>>"?

src/cheri/cheri-pte-ext.adoc
arichardson: yeah I just looked for `{ctag} is set`/`{ctag} set`. Let me fix that too.

src/cheri/scripts/rvy_instruction_encodings.py
arichardson: yes sounds good let me drop this. All others are non-normative.

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: fingers crossed - might be worth putting that in the email!

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: ok - it's exactly what Krste wrote

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: "of this capability value" - yes I think that would do it

src/cheri/introduction.adoc
```diff
@@ -67,18 +67,18 @@ endif::[]
...
-| Extension or Specification                                | Description
-|<<{rvy_i_mod_file_name},     {rvy_i_mod_ext_name}>>        | RVI instructions modified by {cheri_base_ext_name}
-|<<{rvy_zca_mod_file_name},   {rvy_zca_mod_ext_name}>>      | C instructions modified by {cheri_base_ext_name}
-|<<section_cheri_vector_integration, V ({cheri_base_ext_name} modified behavior)>> | V instructions modified by {cheri_base_ext_name}
-|<<{rvy_zicbom_mod_file_name},{rvy_zicbom_mod_ext_name}>>   | Zicbom instructions modified by {cheri_base_ext_name}
-|<<{rvy_zicbop_mod_file_name},{rvy_zicbop_mod_ext_name}>>   | Zicbop instructions modified by {cheri_base_ext_name}
-|<<{rvy_zicboz_mod_file_name},{rvy_zicboz_mod_ext_name}>>   | Zicboz instructions modified by {cheri_base_ext_name}
-|<<{rvy_zicsr_mod_file_name}, {rvy_zicsr_mod_ext_name}>>    | Zicsr instructions modified by {cheri_base_ext_name}
+| Base ISA or Extension | Differences
+|<<{rvy_i_mod_file_name},     I>>      | RVY redefines the behavior of control flow and RVI loads/stores
```
tariqkurd-repo:
```suggestion
|<<{rvy_i_mod_file_name},     I>>      | Control flow instructions and load/stores use capability values as address operands
```

src/cheri/introduction.adoc
tariqkurd-repo:
```suggestion
|<<{rvy_zca_mod_file_name},   Zca>>    | Compressed register move, stack pointer increments, jumps and load/stores operate on capability values
```

src/cheri/riscv-unpriv-integration.adoc
PeterRugg: Yeah, these are much harder to detect than the current malformed cases, since any exponent can lead to a top > 2^64 depending on the address. I think the main justifications behind the malformed behaviours we have are:
1. Define what happens in these cases for verification
2. Allow us to reclaim this space later without breaking backwards compatibility

For point 1: there's pretty clear behaviour we can define, which is to just do the 65-bit comparison (which you have to do anyway, because the top could be exactly 2**xlen and be completely legal).
For point 2: we wouldn't want to reclaim this space, since it depends on a shift and the address of the cap to detect.

src/cheri/riscv-unpriv-integration.adoc
arichardson: dropped it from this PR.

src/cheri/zabhlrsc.adoc
```diff
@@ -48,6 +48,6 @@ The Zabhlrsc extension provides the `LR.[B|H]` and `SC.[B|H]` instructions.
 ....
 
 LR.[B|H] behave analogously to LR.[W|D].
-They sign extend the result and write it to _rd_.
+These instructions sign-extend the result and write it to _rd_.
```
tariqkurd-repo: I think this file shouldn't be here any more as it's now a separate PR to the RISC-V ISA Manual - I'll raise another PR to delete it.

src/cheri/insns/dret.adoc
```diff
@@ -14,13 +14,12 @@ include::wavedrom/dret.adoc[]
...
-It unseals <<dpc_y>> if it is sealed as a {sentry_cap_uc} and writes the result into the `pc`.
+It unseals <<dpc_y>> if it is sealed as a {sentry_cap_uc} and writes the result into <<pcc>>.
 It also restores `ddc` from <<dddc>>.
 
 NOTE: The `DRET` instruction is a documented method of exiting debug mode.
-However, it is a pseudoinstruction to return that technically does not execute
-from the program buffer or memory. It does not require the `pc` to
-grant `ASR-permission` so it never raises an exception.
+However, it is a pseudoinstruction to return that technically does not execute from the program buffer or memory.
+It does not require <<pcc>> to grant <<asr_perm>>, so it never raises an exception.
```
```suggestion
It does not require the `pc`  to grant `ASR-permission`, so it never raises an exception.
```

tariqkurd-repo: <\<pcc>> and <<asr_perm>> don't resolve in this file

src/cheri/debug-integration.adoc
```diff
@@ -188,14 +188,12 @@ to this specification. The `pc` metadata has no architectural effect in debug
 mode. Therefore, `ASR-permission` is implicitly granted for access to all CSRs for
 instruction execution.
 
-On debug mode entry, <<dpc_y>> is updated with the
-capability in the `pc`, whose address field equals to the address of the next
-instruction to be executed upon debug mode exit as described in the _RISC-V Debug Specification_.
+On debug mode entry, <<dpc_y>> is updated with the capability in <<pcc>>, whose address field equals to the address of the next instruction to be executed upon debug mode exit as described in the _RISC-V Debug Specification_.
```
```suggestion
On debug mode entry, <<dpc_y>> is updated with the capability in the `pc`, whose address field equals to the address of the next instruction to be executed upon debug mode exit as described in the _RISC-V Debug Specification_.
```

src/cheri/debug-integration.adoc
```diff
@@ -188,14 +188,12 @@ to this specification. The `pc` metadata has no architectural effect in debug
...
 
-When leaving debug mode, the value in <<dpc_y>> is unsealed if it is a {sentry_cap_uc} and is written to PC.
+When leaving debug mode, the value in <<dpc_y>> is unsealed if it is a {sentry_cap_uc} and is written to <<pcc>>.
```
```suggestion
When leaving debug mode, the value in <<dpc_y>> is unsealed if it is a {sentry_cap_uc} and is written to the `pc`.
```

src/cheri/debug-integration.adoc
```diff
@@ -188,14 +188,12 @@ to this specification. The `pc` metadata has no architectural effect in debug
...
 
-NOTE: A debugger may write <<dpc_y>> to change where the hart resumes and its mode, permissions, sealing or bounds.
- The value saved from the `pc` on debug mode entry cannot be sealed.
+NOTE: A debugger may write <<dpc_y>> to change where the hart resumes and its mode, permissions, sealing, or bounds.
+The value saved from <<pcc>> on debug mode entry cannot be sealed.
```
```suggestion
The value saved from the `pc` on debug mode entry cannot be sealed.
```

src/cheri/zabhlrsc.adoc
tariqkurd-repo:
```suggestion
They sign extend the result and write it to _rd_.
```

src/cheri/insns/dret.adoc
```diff
@@ -18,9 +18,8 @@ It unseals <<dpc_y>> if it is sealed as a {sentry_cap_uc} and writes the result
 It also restores `ddc` from <<dddc>>.
 
 NOTE: The `DRET` instruction is a documented method of exiting debug mode.
-However, it is a pseudoinstruction to return that technically does not execute
-from the program buffer or memory. It does not require the `pc` to
-grant `ASR-permission` so it never raises an exception.
+However, it is a pseudoinstruction to return that technically does not execute from the program buffer or memory.
```
```suggestion
However, it is a pseudoinstruction that technically does not execute from the program buffer or memory.
```

tariqkurd-repo: existing wording is a bit weird

src/cheri/app-hybrid-usage.adoc
```diff
@@ -230,7 +230,7 @@ Transitions between modes are performed using specific instructions:
 
 <<MODESW_INT>> and <<MODESW_CAP>>::
 These instructions are used to switch between modes without changing the control flow.
-They are typically used when entering or leaving a section of code that requires a different mode.
+These instructions are typically used when entering or leaving a section of code that requires a different mode.
```
tariqkurd-repo: joining the two sentences would read better (it won't let me make a multi-line suggestion)

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -111,19 +109,18 @@ endif::[]
...
 All registers or memory locations able to hold a valid capability are YLEN bits wide and have an associated {ctag}.
-These are referred to as being _YLEN-bit_ in this specification.
 
 Not all memory in a real system will support {ctag}s.
 The Execution Environment Interface (EEI) will define what regions of the address space support them.
 
 The {ctag} cannot be directly set by software; it is _not_ a conventionally accessible bit of state.
 If the {ctag} is set to one then it shows that the capability has been derived correctly according to the principles listed above (_provenance_, _integrity_, _monotonicity_).
-If the rules are followed then the {ctag} will propagate through the instructions that modify, load or store the capability.
+If the rules are followed then the {ctag} will propagate through the instructions that modify, load, or store the capability value.
```
```suggestion
If the rules are followed then the {ctag} will propagate through the instructions that derive, load, or store the capability value.
```

best avoid "modify"

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -134,10 +131,10 @@ Therefore, whenever an instruction writes a capability with the {ctag} set to on
 * The instruction may have checked that none of the input capabilities were corrupted.
 ** This is the _integrity_ check. Not all instructions check all input operands for _integrity_.
 
-Capability load/store require the _provenance_ check:
+For capability loads and stores, provenance is preserved across memory:
 
-* Any store that wrote the capability to memory was correctly authorized.
-* Any load that read the capability from memory was correctly authorized.
+* Storing a capability to memory preserves the {ctag} in memory only if the store is performed using a valid capability that authorizes storing capabilities to that memory location.
```
```suggestion
* Storing a capability to memory preserves the {ctag} in memory only if the store is performed using a valid capability that authorizes storing capabilities to that memory location, by checking that every stored byte is within capability bounds and that suitable permissions are granted to write to memory.
```

tariqkurd-repo: like this to fully specify what `authorized` means? also for loads on the line below

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -134,10 +131,10 @@ Therefore, whenever an instruction writes a capability with the {ctag} set to on
...
+* Loading a capability from memory preserves the {ctag} in the destination register only if the load is performed using a valid capability that authorizes loading capabilities from that memory location.
```
```suggestion
* Loading a capability from memory preserves the {ctag} in the destination register only if the load is performed using a valid capability that authorizes loading capabilities from that memory location, by checking that every byte is within capability bounds and that suitable permissions are granted to read from memory.
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -189,7 +186,7 @@ In a typical load/store unit, the expansion of the bounds from `{cs1}` and bound
 
 A compressed format is used to encode the bounds with a scheme similar to floating-point using an exponent and a mantissa.
 Therefore, small exponents can allow byte granularity on the bounds, but larger exponents give coarser granularity.
-One bounds encoding format based upon cite:[woodruff2019cheri] is defined in <<section_cap_bounds>>.
+The standard bounds encoding format for {rv64y_uni_base_name} is based upon the compressed bounds encoding scheme cite:[woodruff2019cheri] and defined in <<section_cap_bounds>>.
```
```suggestion
The standard bounds encoding format for {rv64y_uni_base_name} is based upon the compressed bounds encoding scheme cite:[woodruff2019cheri] and is defined in <<section_cap_bounds>>.
```

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: I think that `top overflow` shouldn't be mentioned here at all as it's a function of the specific bounds encoding scheme. The statement should go entirely into the `malformed bounds` section so it's capability encoding format specific and the justification about excessive logic cost should be put there instead.

tariqkurd-repo: @PeterRugg

tariqkurd-repo: I've discussed this with Peter, we think ti makes sense to move the top overflow text and justificaton for it not being malformed into the malformed bounds section within the cap encoding format

arichardson: Okay I'll drop this from the PR and open a new one. This PR should just be the trivial fixes

arichardson:
Will address the remaining comments tomorrow morning and then merge since all the non trivial changes are now in different prs

src/cheri/riscv-unpriv-integration.adoc
arichardson: #1227

tariqkurd-repo: yep - handled in either #1227 or #1230

tariqkurd-repo (approved):

PR #1212, opened 2026-09-03
CT -> CTag in register diagrams - and specify endianness

### LINKED ISSUE #1210: CT used for both Tag and Type
Figure 1 appears to use CT for the tag whereas 2.3.9 and the rest of the document uses it for Capability Type.

tariqkurd-repo:
Fix #1210 - It shouldn't be CT in the register diagrams as that's the CT-field not the capability tag.
Also address ARC comment that the register diagrams don't show the endianness of the words. the text has come out quite small for this but at least the diagrams are accurate now.

### DIFF
--- a/src/cheri/riscv-unpriv-integration.adoc
+++ b/src/cheri/riscv-unpriv-integration.adoc
@@ -88,13 +88,13 @@ Specific capability encoding formats may support different ISA extensions and ad
 (def boxes-per-row 32)
 (draw-column-headers {:height 50 :font-size 18 :labels (reverse (concat ["0"] (repeat 28 "") ["XLEN-1" "" ""]))})
 
-(draw-box "CT"    {:span 1})
+(draw-box "CTag"    {:span 2})
 (draw-box "" {:span 1 :borders {}})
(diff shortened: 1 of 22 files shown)


src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -676,12 +678,12 @@ The <<gprs,XLEN-wide integer registers>> (e.g., `x1`, `x2`) are all widened to Y
 (def boxes-per-row 66)
 (draw-column-headers {:height 30 :font-size 25 :labels (concat ["YLEN" "" "" "" "" "2*XLEN-1 (YLEN-1)"] (repeat 28 "") ["XLEN XLEN-1"] (repeat 30 "") ["0"])})
```
```suggestion
(draw-column-headers {:height 30 :font-size 25 :labels (concat ["YLEN" "" "" "" "" "YLEN-1"] (repeat 28 "") ["XLEN XLEN-1"] (repeat 30 "") ["0"])})
```

Looks a bit awkward with `2*XLEN-1 (YLEN-1)`

arichardson (approved):
Much nicer, just one thing I noticed.

PR #1192, opened 2026-08-13
Minor updates - ready to merge

tariqkurd-repo:
Minor fixes and typos

Fix #1190

PR #1213, opened 2026-09-03
Clarify C.YMV expansion

### LINKED ISSUE #1190: C.YMV / YMV expansion is a bit strange
Image

but

Image

so we have some rs1/rs2 confusion on the expansion.

tariqkurd-repo:
fix #1190 

subset of #1192

### DIFF
--- a/src/cheri/insns/cmv_16bit.adoc
+++ b/src/cheri/insns/cmv_16bit.adoc
@@ -10,9 +10,12 @@ Mnemonic::
 `{C_MV_CAP} {cd}, {cs2}`
 
 Expansion::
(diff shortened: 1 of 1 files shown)


arichardson (commented):
I don't think this is important enough of a small "typo" fix that it warrants notifying ARC. It's just a consistency fix.

arichardson (approved):
I would argue the expansion was correct before, just inconsistent with c.mv.

arichardson:
In my last comment I meant that the commit title should really be "Clarify C.YMV expansion" and not "Fix". It wasn't broken just a bit odd.

PR #1214, opened 2026-09-04
Explicitly zero permissions on YPERMC on integrity check failure - need to ask ARC

### LINKED ISSUE #1202: Should YPERMC be consistent with YPERMR on integrity failures?
YPERMR is defined to report zeros in all allocated permissions on a cap that fails an integrity check: bad bounds, illegal perms combination, or reserved bits set. This is true even if the tag is clear, so this is architecturally visible even without memory corruption etc.

YPERMC doesn't seem to do the same on reading in the perms before clearing them: "Convert the [AP-field](https://riscv.github.io/riscv-cheri/#AP-field) and [SDP-field](https://riscv.github.io/riscv-cheri/#SDP-field) fields of capability rs1 into a bit field with the format shown in [Figure 6](https://riscv.github.io/riscv-cheri/#acperm_bit_field).". It may be confusing that YPERMC is happy to manipulate permissions that, according to YPERMR, you don't have. Might it be more consistent to have YPERMC follow the YPERMR rules?

I'm trying to think if any other instructions have a similar problem:
- YMODEW is similar. It currently says it only makes changes "if rs1 grants X-permission", which could already be read to mean that the integrity checks must pass, but this could maybe be clarified to "if rs1 passes integrity checks and grants X-permission" for the avoidance of doubt.
- Bounds manipulations don't take into account the existing bounds except to do the monotonicity check that will correctly fail if the integrity check fails, so these should be okay.

I don't think this is a major problem, just wanted to see if people think it's likely to cause confusion.

tariqkurd-repo:
This came from updating SAIL to the latest spec version.

YPERMC - if integrity checks fail then it's not clear what the resulting permissions should be, so just zero them. This is more relevant to RV32Y where they are encoded and the encoding could be a reserved one. Explicit zeroing wasn't included but should have been.

Fix #1202


PeterRugg:
I think this makes sense. Does it also clear the cap_level and mode implicitly?

tariqkurd-repo:
> I think this makes sense. Does it also clear the cap_level and mode implicitly?

For RV32Y it would clear mode as that's included in the AP field.
For RV64Y it won't as that's a separate bit.

cap_level is always separate so I wasn't planning to zero that.

The point is that if the AP field can't be interpreted we now have a defined method to produce a result.

arichardson:
> > I think this makes sense. Does it also clear the cap_level and mode implicitly?
> 
> For RV32Y it would clear mode as that's included in the AP field. For RV64Y it won't as that's a separate bit.
> 
> cap_level is always separate so I wasn't planning to zero that.
> 
> The point is that if the AP field can't be interpreted we now have a defined method to produce a result.

but if we zero all permissions we have to clear mode as well since that depends on X. I guess it doesn't matter since the result is untagged nonsense that could have that bit pattern anyway.

tariqkurd-repo:
> > > I think this makes sense. Does it also clear the cap_level and mode implicitly?
> > 
> > 
> > For RV32Y it would clear mode as that's included in the AP field. For RV64Y it won't as that's a separate bit.
> > cap_level is always separate so I wasn't planning to zero that.
> > The point is that if the AP field can't be interpreted we now have a defined method to produce a result.
> 
> but if we zero all permissions we have to clear mode as well since that depends on X. I guess it doesn't matter since the result is untagged nonsense that could have that bit pattern anyway.

good point - I've added that. Better to clean it all up.

arichardson (approved):
LGTM once ARC approves

arichardson (approved):
Approved by ARC in their latest meeting.

PR #1216, opened 2026-09-04
Add priv extension names - need to confirm names

tariqkurd-repo:
Add S / M hybrid extensions
Add SvyXX VM extensions


arichardson (changes requested):
I think https://github.com/riscv/riscv-cheri/pull/1221 is more in line with what Zicbom and Zicfss etc. do

arichardson:
ARC said we do need names and the omission of priv spec names from other extensions was a mistake and they will be introduced.

jrtc27:
Is that true even for *bases*?

tariqkurd-repo:
> Is that true even for _bases_?

there should be separate extension names per privilege mode.  The existing ISA examples quoted will be updated to have these names as their omission was a mistake.

so, yes!

arichardson:
I've tried to incorporate the latest ARC feedback into #1221. If you're happy with that we can close this PR.

PR #1219, opened 2026-09-07
Consistently use 'set to one/zero'

arichardson:
ARC pointed out that a plain set/clear can be ambiguous in some
cases. Replace all remaining instances of bare 'is set' and 'tag set'
with explicit 'set to one/zero' across all chapters (including ones
not submitted for review).

Also consistently use 'set to one/zero' instead of 'set to 1/0'. We
were using both, and this inconsistency was flagged.

Addresses ARC review comment on v0.9.9-ar20260810.


arichardson:
split out this mechanical update from #1209 to make that one easier to review

PR #1220, opened 2026-09-07
Clarify that in-bounds addresses are always representable

arichardson:
Make explicit that any address within the capability's bounds is
always representable and preserves the tag, and that tag clearing
only occurs when an address calculation exceeds the representable
range.

Addresses ARC review comment on v0.9.9-ar20260810.

### DIFF
--- a/src/cheri/riscv-unpriv-integration.adoc
+++ b/src/cheri/riscv-unpriv-integration.adoc
@@ -219,11 +219,7 @@ These bounds-reducing instructions all copy `{cs1}` to the output `{cd}`, set th
 ** If the requested bounds cannot be encoded exactly, the result uses the smallest representable bounds that contain the requested range and remain within the initial bounds.
 * <<CRAM>> can be used to calculate the nearest precisely encodable _length_ and _base_ values for a given size.
 
-The bounds are encoded using:
-
-* A range of upper bits of the address field, defined by the exponent.
-* Fields in the metadata.
-
+The bounds are encoded relative to the address field, using some upper bits of the address combined with bits from the capability metadata.
 The exact method for encoding the bounds varies by capability encoding format.
 See xref:section_cap_bounds[xrefstyle=short] for the method used for {rv64y_uni_base_name}.
 
@@ -232,16 +228,15 @@ See xref:section_cap_bounds[xrefstyle=short] for the method used for {rv64y_uni_
 
 Software sometimes uses out-of-bounds pointers, for example, to represent a one-past-the-end pointer to an array in C (`arr + size`).
 To support this pattern, CHERI must allow these out-of-bounds pointers to be held in capabilities while still protecting the bounds and integrity.
-Because the CHERI encoding scheme for memory bounds uses some of the upper bits of the address to calculate the bounds, only a limited range of out-of-bounds pointers can be represented for any given set of bounds.
+Because the standard bounds encoding scheme shares the upper bits of the address with the bounds, only a limited range of out-of-bounds pointers can be represented for any given set of bounds.
 This range is known as the *representable range*.
 
 Any address that is within the capability's bounds is always representable and preserves the {ctag}.
-
 The relationship between the bounds and the representable range is illustrated in <<cap_bounds_map>>.
 
 NOTE: The C standard permits forming a one-past-the-end pointer (but no more than one), as long as it is not dereferenced.
   Real-world software has been observed to temporarily create pointers that point multiple bytes past the end or even before the start.
-  To maximize software compatibility, the bounds representation is designed to allow for out-of-bounds pointers.
+  To maximize software compatibility, the bounds representation was designed to allow for out-of-bounds pointers.
 
 [#cap_bounds_map]
 .Memory address bounds and representable range encoded within a capability
@@ -250,13 +245,13 @@ image::../cheri/img/cap-bounds-map.png[width=80%,align=center]
 NOTE: The historic, uncompressed 64-bit CHERI-MIPS CPU had a 256-bit capability encoding with separate 64-bit values for the base and top memory bounds, and another for metadata fields.
  This scheme could represent all out-of-bounds pointers with a very high hardware cost.
 
-Deriving a new capability with an address outside the representable range changes the resulting bounds.
+Since the upper bits of the address are used to calculate the top and base bound addresses, deriving a new capability with a different address could change the resulting bounds.
 All instructions that derive a capability with a new address (such as <<SCADDR>>, <<CADD>>, and <<CADDI>>) must check that the new address does not change the bounds.
-If the bounds _do_ change then the {ctag} of the derived capability is set to zero, so that the capability becomes invalid.
+If an address calculation moves the address outside the representable range, the hardware detects that the bounds can no longer be represented correctly, and sets the {ctag} to zero so that the derived capability becomes invalid.
 
-NOTE: {cheri_base_ext_name} implementations that use a different encoding scheme to <<rv64y_cap_description>>, (e.g., for accelerators or specific micro-controllers) may specify an alternative to the representable range check.
-Such implementations may only allow the minimum one-past-the-end as required by the C standard.
-To accommodate this the capability encoding formats support the option of REP_RANGE being `rc0` and the ISA specification uses the phrase *represented exactly* for checking that the bounds have not been altered by deriving a new capability with an updated address field.
+NOTE: {cheri_base_ext_name} implementations that use a different encoding scheme to <<rv64y_cap_description>> (e.g., for accelerators or specific micro-controllers) may specify an alternative to the representable range check.
+Such implementations must still support one-past-the-end pointers as required by the C standard, but are not required to support a representable range larger than that.
+Therefore, the ISA specification uses the phrase *represented exactly* for this check.
 
 ==== Memory space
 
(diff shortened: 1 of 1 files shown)


arichardson:
I also dropped a bit of redundancy in that section while making that change.

tariqkurd-repo:
I think we need to be careful here - we're not forcing CHERI implementations to use CHERI Concentrate for all time - and I don't even know if CHERIoT uses it - so this intro chapter should probably be a bit more general by introducing the concepts and the actual reference to CHERI Concentrate is already (correctly) in the RV64LYA chapter.

I'll have a go at making the text more general.

tariqkurd-repo:
I've tried to make the text more specific by referring to the REP_RANGE parameter values rc1 and rc0.
I've also made it point out that the upper bits are shared between the address and the bounds only twice instead of three times!
Feedback welcome, I'm not sure it's good enough and @PeterRugg is looking at updating the text in the diagram to use the phrase "representable region" instead of "representable space" - see #1226

arichardson:
I think it might be easier to say that in some encoding formats the guaranteed rep range can be just the bounds themselves but generally it's larger (see section xxx for detail)? Then we don't need to reference the format variable

tariqkurd-repo:
Will fix that tomorrow, or feel free to do it 😉

tariqkurd-repo (approved):

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -220,41 +220,44 @@ They all copy `{cs1}` to the output `{cd}`, set the _base_ bound address of `{cd
...
+The exact method for encoding the bounds varies by capability encoding format.
+See xref:section_cap_bounds[xrefstyle=short] for the method used for {rv64y_uni_base_name}.
 
 [#section_rep_check_concept]
 ==== Representability and Updating the Address
 
 Software sometimes uses out-of-bounds pointers, for example, to represent a one-past-the-end pointer to an array in C (`arr + size`).
 To support this pattern, CHERI must allow these out-of-bounds pointers to be held in capabilities while still protecting the bounds and integrity.
-Because the CHERI Concentrate cite:[woodruff2019cheri] encoding scheme for memory bounds shares the upper bits of the address with the bounds, only a limited range of out-of-bounds pointers can be represented for any given set of bounds.
+Because the CHERI encoding scheme for memory bounds uses a range of the upper bits of the address to calculate the bounds, only a limited range of out-of-bounds pointers can be represented for any given set of bounds.
```
```suggestion
Because the CHERI encoding scheme for memory bounds uses some of the upper bits of the address to calculate the bounds, only a limited range of out-of-bounds pointers can be represented for any given set of bounds.
```
to avoid `range of` twice in the same sentence.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -605,15 +608,12 @@ NOTE: The default settings all match <<rv64y_cap_description>>.
...
 4+| *Other parameters*
 | AUIPC_SHIFT | integer values up to 12 | 12                | Shift distance of the immediate in AUIPC
-| REP_RANGE   | rc1/r0               | rc1                  | Whether out of bounds addresses are representable
 |==============================================================================
 
 Two parameters use non-numeric values:
 
 AP_ENC::
  Whether permissions in the <<AP-field>> are encoded using a _full_ (`full`) representation allowing all valid combinations of permissions, or a _partial_ (`part`) representation which does not represent all possible combinations and so uses fewer encoding bits.
-REP_RANGE::
```
arichardson, line 616: why did you delete this, we still have it in the encoding chapters and it doesn't do any harm having it here.

tariqkurd-repo: fixed

tariqkurd-repo:
This one is ready to merge now.

arichardson:
I've removed the duplicated paragraph and slightly simplified it. Let me know what you think.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -250,13 +245,13 @@ image::../cheri/img/cap-bounds-map.png[width=80%,align=center]
...
 
-NOTE: {cheri_base_ext_name} implementations that use a different encoding scheme to <<rv64y_cap_description>>, (e.g., for accelerators or specific micro-controllers) may specify an alternative to the representable range check.
-Such implementations may only allow the minimum one-past-the-end as required by the C standard.
-To accommodate this the capability encoding formats support the option of REP_RANGE being `rc0` and the ISA specification uses the phrase *represented exactly* for checking that the bounds have not been altered by deriving a new capability with an updated address field.
+NOTE: {cheri_base_ext_name} implementations that use a different encoding scheme to <<rv64y_cap_description>> (e.g., for accelerators or specific micro-controllers) may specify an alternative to the representable range check.
+Such implementations must still support one-past-the-end pointers as required by the C standard, but are not required to support a representable range larger than that.
+Therefore, the ISA specification uses the phrase *represented exactly* for this check.
```
tariqkurd-repo, line 254: you don't want to mention `rc0`? that's what it's for

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -232,16 +228,15 @@ See xref:section_cap_bounds[xrefstyle=short] for the method used for {rv64y_uni_
 
 Software sometimes uses out-of-bounds pointers, for example, to represent a one-past-the-end pointer to an array in C (`arr + size`).
 To support this pattern, CHERI must allow these out-of-bounds pointers to be held in capabilities while still protecting the bounds and integrity.
-Because the CHERI encoding scheme for memory bounds uses some of the upper bits of the address to calculate the bounds, only a limited range of out-of-bounds pointers can be represented for any given set of bounds.
+Because the standard bounds encoding scheme shares the upper bits of the address with the bounds, only a limited range of out-of-bounds pointers can be represented for any given set of bounds.
```
tariqkurd-repo: I won't let me add a suggestion.
The previous version of this sentence was more accurate as the bits aren't "shared with the bounds", they're "used by the bounds calculation"

arichardson: Yes let me fix that.

src/cheri/riscv-unpriv-integration.adoc
arichardson: I thought keeping it as generic text that doesn't mention these specific parameters is more future proof

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -239,20 +242,15 @@ NOTE: The C standard permits forming a one-past-the-end pointer (but no more tha
...
  This scheme could represent all out-of-bounds pointers with a very high hardware cost.
 
 Since the upper bits of the address are used to calculate the top and base bound addresses, deriving a new capability with a different address could change the resulting bounds.
-All instructions that derive a capability with a new address must check that the new address does not change the bounds.
-If the bounds _do_ change then the {ctag} of the derived capability is set to zero, so that the capability becomes invalid.
-
-Software can derive a capability with a new address using instructions such as <<SCADDR>>, <<CADD>> and <<CADDI>>.
-
-NOTE: <<SCADDR>> writes back a derived capability with a new address field and, if the {ctag} was previously set, sets the {ctag} of the derived capability to one if the resulting capability still has the same bounds.
+All instructions that derive a capability with a new address (such as <<SCADDR>>, <<CADD>>, and <<CADDI>>) must check that the new address does not change the bounds.
```
```suggestion
All instructions that derive a capability with a new address (such as <<SCADDR>>, <<CADD>>, and <<CADDI>>) must check that the new address does not change the decoding of the bounds.
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -229,6 +230,8 @@ Software sometimes uses out-of-bounds pointers, for example, to represent a one-
 To support this pattern, CHERI must allow these out-of-bounds pointers to be held in capabilities while still protecting the bounds and integrity.
 Because the standard bounds encoding scheme shares the upper bits of the address with the bounds, only a limited range of out-of-bounds pointers can be represented for any given set of bounds.
```
```suggestion
Because the standard bounds encoding scheme uses for the computation of the bounds, only a limited range of out-of-bounds pointers can be represented for any given set of bounds.
```

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: but that's eaxctly what `r0` says it's for above where REP_RANGE is defined. I'd happily take `r0` out of this version of the spec completely and that can be added by CHERIoT later.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -219,16 +219,19 @@ These bounds-reducing instructions all copy `{cs1}` to the output `{cd}`, set th
...
+The exact method for encoding the bounds varies by capability encoding format.
+See xref:section_cap_bounds[xrefstyle=short] for the method used for {rv64y_uni_base_name}.
 
 [#section_rep_check_concept]
 ==== Representability and Updating the Address
 
 Software sometimes uses out-of-bounds pointers, for example, to represent a one-past-the-end pointer to an array in C (`arr + size`).
 To support this pattern, CHERI must allow these out-of-bounds pointers to be held in capabilities while still protecting the bounds and integrity.
-Because the standard bounds encoding scheme shares the upper bits of the address with the bounds, only a limited range of out-of-bounds pointers can be represented for any given set of bounds.
+Because the standard bounds encoding scheme uses for the computation of the bounds, only a limited range of out-of-bounds pointers can be represented for any given set of bounds.
```
```suggestion
Because the standard bounds encoding scheme uses upper bits of the address for the computation of the bounds, only a limited range of out-of-bounds pointers can be represented for any given set of bounds.
```

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: done.

arichardson:
Made a final minor tweak, but should be ready to merge if you're happy @tariqkurd-repo @PeterRugg

tariqkurd-repo (approved):
I've resinstated r0 in the table, i twasn't complained about it, so I think that's ok
Happy to merge now

PR #1221, opened 2026-09-07
Restructure privileged extensions table

arichardson:
Split the privileged table into a table of architectural differences
under RVY/Zyhybrid and a table of actual extensions (only Svyrg).
This clarifies that most of these are just updated privileged
behaviours (analogous to Zicbom/Zicfiss/etc. allocating new privileged
bits) rather than separate extensions.

This should address the ARC review comment on v0.9.9-ar20260810 that
all entries in the table should have a name.

Alternative to https://github.com/riscv/riscv-cheri/pull/1216


tariqkurd-repo:
It's a fair point that Zicfiss has a privileged aspect and no extension name.
I think we need to ask them by email to clarify what we should be doing here.

arichardson (commented):
This was rejected, they want explicit names for it.

I think the table restructure might still make sense but maybe with the names from #1216 ?

arichardson:
Update to add explicit names:
image

arichardson:
ARC responded, said the names are find and said we can keep Sv39/etc. instead of Svy. So  I've incorporated that now.

PR #1223, opened 2026-09-07
extended CSRs -> widened CSRs

tariqkurd-repo:
Naming consistency, the previous move from extended to widened wasn't applied consistently

### DIFF
--- a/src/cheri/riscv-unpriv-integration.adoc
+++ b/src/cheri/riscv-unpriv-integration.adoc
@@ -825,7 +825,7 @@ All CSRs that can hold addresses are widened to YLEN bits.
 [[xlen_csrs, XLEN-bit CSRs]]XLEN-bit CSRs::
 These do not contain pointers (e.g., _fcsr_ from the "F" extension).
 
-[[widened_csrs, widened csrs]]Widened csrs::
+[[widened_csrs, Existing CSRs widened to YLEN bits]]Existing CSRs widened to YLEN bits::
 These are XLEN-bit CSRs widened to YLEN bits, which are able to contain pointers
 // No unprivileged pointer-type CSRs in v1.0, so let's use mtvec here.
 ifdef::cheri_ratification_v1_only[]
--- a/src/cheri/riscv-unpriv-integration.adoc
+++ b/src/cheri/riscv-unpriv-integration.adoc
@@ -764,7 +764,7 @@ Each thread (as scheduled by the kernel) can have its own context stored in priv
 <<utidc>> allows compartment-switching code to determine the currently running thread without a call into the operating system.
 This is read-only without <<asr_perm>> to prevent a malicious compartment from tricking the switching mechanism into accessing data corresponding to the wrong thread.
 While the RISC-V ABI includes a _thread pointer (tp)_ register, it is not usable for the purpose of reliably identifying the current software thread because the tp register is a general-purpose register and can be changed arbitrarily by untrusted code.
-Extending <<utidc>> to a capability allows a data structure to be shared (guarded by e.g., the subset sealing mechanism), which allows for example efficient recovery of a trusted stack, without requiring additional indirection.
+Widening <<utidc>> to a capability allows a data structure to be shared (guarded by e.g., the subset sealing mechanism), which allows for example efficient recovery of a trusted stack, without requiring additional indirection.
 =====
 
 ifdef::cheri_xycfg_csr[]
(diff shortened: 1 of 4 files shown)


jrtc27:
Why are we using the term "widened"? "Extended" is surely the better term? "Widened" implies it's merely just more bits, whereas "extended" better encompasses the added semantics.

jrtc27:
Right, #1070. But there we were very careful to include something like "to hold a capability". That important detail is missing here.

tariqkurd-repo:
> Right, #1070. But there we were very careful to include something like "to hold a capability". That important detail is missing here.

I just found that PR too - it took a while - and also #1063.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -824,7 +824,7 @@ All CSRs that can hold addresses are widened to YLEN bits.
 [[xlen_csrs, XLEN-bit CSRs]]XLEN-bit CSRs::
 These do not contain pointers (e.g., _fcsr_ from the "F" extension).
 
-[[extended_csrs, extended CSRs]]Extended CSRs::
+[[widened_csrs, widened csrs]]Widened csrs::
 These are XLEN-bit CSRs widened to YLEN bits, which are able to contain pointers
```
tariqkurd-repo, line 829: @jrtc27 is there an objection to this line?

tariqkurd-repo:
This is intended to be a non-controversial subset of #1208 to separate out the additional WARL text from the simple word substitution.

arichardson:
I'm not sure there is any confusion about extended CSRs vs widened. The previous issue was about avoiding English grammar that could be confusing to non native speakers but I think here we don't need to change it in these cases?

tariqkurd-repo:
the idea was to use one of the two consistently but some got missed.

arichardson (changes requested):
I prefer the previous wording, and keeping it avoids "churn", but if you feel strongly that we should do this I'm okay with making the change.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -1065,6 +1067,7 @@ If <<c_perm>> is not granted then:
 All capability data memory access instructions require naturally aligned addresses, and will take an access fault exception if this requirement is not met.
 Misaligned capability accesses cannot be emulated.
 
+NOTE: An access fault is raised instead of a misaligned exception because these instructions cannot be emulated since there is one hidden {ctag} per naturally aligned YLEN-bit memory region.
```
```suggestion
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -123,6 +124,7 @@ The Execution Environment Interface (EEI) will define what regions of the addres
 The {ctag} cannot be directly set by software; it is _not_ a conventionally accessible bit of state.
 If the {ctag} is set to one then it shows that the capability has been derived correctly according to the principles listed above (_provenance_, _integrity_, _monotonicity_).
 If the rules are followed then the {ctag} will propagate through the instructions that derive, load, or store the capability value.
+If the rules are followed then the {ctag} will propagate through the instructions that derive, load, or store the capability value.
```
```suggestion
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -113,6 +113,7 @@ endif::[]
 The {ctag} is a single additional bit associated with YLEN-bit memory locations and YLEN-bit registers.
 It is stored separately (i.e., is _out-of-band_ or _hidden_), and is _hardware-managed_.
 It indicates whether a YLEN-bit register or naturally aligned YLEN-bit memory location contains a valid capability.
+It indicates whether a YLEN-bit register or naturally aligned YLEN-bit memory location contains a valid capability.
```
```suggestion
```

tariqkurd-repo:
> I prefer the previous wording, and keeping it avoids "churn", but if you feel strongly that we should do this I'm okay with making the change.

Previously we had a mixture of extended and widened. I don't have a strong opinion about which we use but we need to consistently choose one so that's what this PR does. We'll get complaints if we have a mixture.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -825,7 +825,7 @@ All CSRs that can hold addresses are widened to YLEN bits.
 [[xlen_csrs, XLEN-bit CSRs]]XLEN-bit CSRs::
 These do not contain pointers (e.g., _fcsr_ from the "F" extension).
 
-[[extended_csrs, extended CSRs]]Extended CSRs::
+[[widened_csrs, widened csrs]]Widened csrs::
```
arichardson: Maybe 'YLEN-extended CSRs'? The current name doesn't really specify how they are related to RVY? That would also allow keeping the original anchor name

arichardson (commented):
The verb fixes look good to me but I think the change to the csrs category name sounds worse

src/cheri/riscv-unpriv-integration.adoc
```suggestion
[[widened_csrs, Standard CSRs widened to YLEN bits]]Widened csrs::
```

tariqkurd-repo: any better?

```suggestion
[[widened_csrs, Existing CSRs widened to YLEN bits]]Existing CSRs widened to YLEN bits::
```

How about this? All CSRs in the spec are standard

arichardson (approved):

PR #1231, opened 2026-09-08
remove outdated note

tariqkurd-repo:
This text must go in the integrated doc, and so this comment seems wrong.


arichardson (changes requested):
I'm not sure this introduction actually makes sense when it becomes part of the unpriv/priv specs, so I would argue the note is correct.

arichardson (approved):
Okay actually there is content here that would make sense so let's drop the NOTE.

PR #1238, opened 2026-09-10
Second round of ARC feedback

tariqkurd-repo:
This PR is for accumulating small non-normative changes

### DIFF
--- a/src/cheri/riscv-unpriv-integration.adoc
+++ b/src/cheri/riscv-unpriv-integration.adoc
@@ -219,16 +219,19 @@ These bounds-reducing instructions all copy `{cs1}` to the output `{cd}`, set th
 ** If the requested bounds cannot be encoded exactly, the result uses the smallest representable bounds that contain the requested range and remain within the initial bounds.
 * <<CRAM>> can be used to calculate the nearest precisely encodable _length_ and _base_ values for a given size.
 
-The bounds are encoded relative to the address field, sharing some upper bits of the address.
-The number of shared bits depends on the exponent, see xref:section_cap_bounds[xrefstyle=short].
+The bounds are encoded relative to the address field, using some upper bits of the address combined with bits from the capability metadata.
+The exact method for encoding the bounds varies by capability encoding format.
+See xref:section_cap_bounds[xrefstyle=short] for the method used for {rv64y_uni_base_name}.
 
 [#section_rep_check_concept]
 ==== Representability and Updating the Address
 
 Software sometimes uses out-of-bounds pointers, for example, to represent a one-past-the-end pointer to an array in C (`arr + size`).
 To support this pattern, CHERI must allow these out-of-bounds pointers to be held in capabilities while still protecting the bounds and integrity.
-Because the standard bounds encoding scheme shares the upper bits of the address with the bounds, only a limited range of out-of-bounds pointers can be represented for any given set of bounds.
+Because the standard bounds encoding scheme uses upper bits of the address for the computation of the bounds, only a limited range of out-of-bounds pointers can be represented for any given set of bounds.
 This range is known as the *representable range*.
+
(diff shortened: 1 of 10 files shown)


src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -605,10 +607,16 @@ NOTE: The default settings all match <<rv64y_cap_description>>.
 | AP_MAX      | integer values       | all ones             | Value of the <<AP-field>> giving maximum permissions
 | AP_ENC      | full<suffix>/ part<suffix>| full1             | Encoding scheme for the <<AP-field>>
 4+| *Other parameters*
-| AUIPC_SHIFT | integer values up to 12 | 12                | Shift distance of the immediate in AUIPC
 | REP_RANGE   | rc1/r0               | rc1                  | Whether out of bounds addresses are representable
```
arichardson: unrelated: looking at this table again, surely this should be part of the bounds encoding section. Then we can drop the other parameters section.

src/cheri/rvy64-encoding.adoc
```diff
@@ -54,16 +54,11 @@ NOTE: T[13:12], T[2:0] and B[2:0] are not specified by the metadata fields and a
 [#rvy64_uni_base_name_param_summary,width="100%",options=header,cols="1,1,4"]
 |==============================================================================
 | Parameter   | Default Value        | Comment
-| MW          | {cap_rv64_mw_width}  | Mantissa width for the bounds encoding
-| EW          | {cap_rv64_exp_width} | Exponent width for the bounds encoding
-| CAP_MAX_E   | {cap_rv64_exp_max}   | Maximum exponent value
-| enableL8    | 0                    | Whether the encoding format includes the `L~8~` bit
-| AP_MAX      | All ones             | Value of the <<AP-field>> giving maximum permissions
 | AP_ENC      | full1_rv64ya_1       | Encoding scheme for the <<AP-field>> - all combinations are representable
```
arichardson: shouldn't this be the default value in the other table? Then this table could be dropped entirely. Although I think it's probably best to keep this and just drop the AUIPC_SHIFT line.

src/cheri/rvy32-encoding.adoc
```diff
@@ -47,15 +47,15 @@ NOTE: T[9:8], T[1:0] and B[1:0] are not specified by the metadata fields and are
...
+| AP_MAX      | 0x8^1^              | Value of the <<AP-field>> giving maximum permissions
 | AP_ENC      | part1_rv32ya_1      | Encoding scheme for the <<AP-field>> - not all combinations are representable
-| AUIPC_SHIFT | 12                  | Shift distance of the immediate in AUIPC
-| REP_RANGE   | rc1                 | At least 1/4 of the capability length
 |==============================================================================
 
 ^1^If {cheri_default_ext_name} is supported, then the infinite capability must represent {cheri_int_mode_name} for compatibility with standard RISC-V code.
 See xref:cap_perms_encoding32_hybrid_zylevels1[xrefstyle=short].
 
+NOTE: Values are only shown if they don't match the defaults from <<cap_encoding_param_default_summary>> and <<rvy_base_isa_param_default_summary>>.
```
arichardson: I would drop this and just omit the AUIPC_SHIFT line since that is now a separate table. All values other than rep_range are different from the default, might as well keep that one as well to not have to go back to the original table.

arichardson (approved):
Thanks those all look good. I pushed a few small adjustments, let me know if those look good to you.

src/cheri/rvy64-encoding.adoc
tariqkurd-repo: yes it should and I'll note that every entry matches the defaults.

arichardson (approved):

tariqkurd-repo:
ready to merge when the checks pass

PR #1239, opened 2026-09-10
Reorder ISA  sections

tariqkurd-repo:
LY and SY appeared before the general description of loads and stores under CHERI which made the text awkward and it had duplicate text.
Introducing the effect on RVI instructions _first_ solves this by introducing basic load/store behaviour and _then_ LY/SY

**beware** this reorders text and so will conflict - best handle this one quickly

The ARC request was to remove the duplicate text from LY/SY, and this restructuring makes that feasible.

Subsequent commits are for ARC feedback which are affected by the reordering in the first commit, all non-normative.


arichardson (changes requested):
THanks for fixing this, I've cherry-picked some of these fixes to separate PRs (#1242, #1243, #1244) and created #1245 as a less invasive alternative to the big reodering. I think the one thing not yet split out is the title case fixes.

tariqkurd-repo:
replaced by other PRs

PR #1243, opened 2026-09-11
Remove xycfg text that is not part of the document

arichardson:
It has been ifdefd out for a long time and is unlikely to
be restored as is. Delete it to declutter the files.


arichardson:
Split out from #1239

tariqkurd-repo (approved):

PR #1244, opened 2026-09-11
Addresses ARC review comment on encoding map section

arichardson:
- There is no need to call this custom-0 as there is only one custom
major opcode - this is just 'custom".
- Should just call these by their name and give their major opcode encoding.
- This section is showing the major opcodes in this new base.  It should not be labelled "added instructions".
- This is implicit in the fact you're defining custom space for a new base, and shouldn't refer to other bases' custom extension opcodes as this will cause confusion.

Co-authored-by: Alexander Richardson <alexrichardson@google.com>
Signed-off-by: Alexander Richardson <alexrichardson@google.com>


arichardson:
split out from #1239

tariqkurd-repo (approved):

tariqkurd-repo:
happy to merge once the checks pass

PR #1245, opened 2026-09-11
Remove redundant generic load/store permissions in LY/SY section

arichardson:
Addresses ARC review comment on v0.9.9-ar20260810:
"This section is about LY and SY specifically. This text appears to
discuss all loads and stores, and is out of place in this section.
Avoid redundant specifications. If LY and SY behave differently than
other loads, should call out here, otherwise they follow the general
rule."


arichardson:
Created this as a less invasive alternative the reordering in #1239. The way I read the comment the complaint was just about the redundancy and the fact that we are talking about all kinds of loads again. So just reference the previous section that talks about requirements for all loads/stores.

arichardson (commented):
I think there is some more redundancy with the integer loads stores later on and I can try to clean that up more tomorrow, but since there is no complaint about that section maybe best to keep the changes minimal

tariqkurd-repo (approved):

tariqkurd-repo:
Happy to merge once the checks pass.

Issue #664, opened 2025-07-16
Leaving space for branch on tag [un]set?


davidchisnall:
I'm not sure if this would fly with ARC, but the RISC-V branch encodings for `bne` and `beq` waste a lot of space.  In particular, the operand order doesn't matter, which means that every encoding can be expressed in two ways.

I wonder if we could require, at least for `beqz` / `bnez` that the `$zero` operand is first?  This would make it possible (either initially or in an extension) to have a the other way around be a branch on tag clear/set.

We had this in CHERI MIPS and we see it in quite a few cases where we end up with the sequence of `cgettag` followed by `beqz` or `bnez`.  Being able to fold those together would be convenient.

This encoding space is already burned for RV{32,64}{I,E}, but in a new base architecture we could fix that mistake.  If we required the lower-numbered register in beq / bne to be first, we would then have the rest of that space available for other things.

The ordered comparisons do not have this problem.  They are less-than or greater-than-or-equal, with greater-than and less-than-or-equal being constructed by reversing the operand order.

arichardson:
I really like this approach of adding the tag check branches (at least for purecap only code, hybrid will need to use the existing semantics).

We should definitely suggest this to ARC

arichardson:
The RISC-V unpriviliged specification does not mention BNEZ/BEQZ other than the compressed opcodes, but both Clang and GCC use the `bne/beq <arg>, zero` ecoding: https://godbolt.org/z/zoxM5xKvo,  so we would need to use `bne/beq zero, <arg>` for byts/bytu (other some other name).

arichardson:
Looks like the asm-mannual defines the canonical bnez/beqz, so we don't have to rely on the compiler output: https://github.com/riscv-non-isa/riscv-asm-manual/blob/main/src/asm-manual.adoc#a-listing-of-standard-risc-v-pseudoinstructions

PR #1246, opened 2026-09-11
remove irregular behaviour of BNE, BEQ

### LINKED ISSUE #1235: Confirm that we want want irregular branch encodings with standard RVI
ARC review comment on the proposal to make BEQ/BNE reserved if rs1<=rs2 is:

_Do you really want to do this? This seems like an idea that will be regretted._

Longer term that seems likely, this could cause a lot of issues later on.....
Already having 16-bit FP-load/store encodings switching into cap load/store causes issues and confusion.
It would be cleaner and safer to simply add new encodings with any new functionality, we have all the reserved space after all.

making load/store via x0 _reserved_ falls into the same category, we have plenty of encoding space to add new encodings.

Both of these are unpopular with the design/verif teams.

tariqkurd-repo:
Fix #1235 

Make branch encodings regular with RVI.
Avoid future pain from haivn girregular branch encodings, as the encoding space restrictions have been eased by reallocating custom space.


arichardson:
This looks good to me but I'd like @davidchisnall to comment as well since I believe you proposed it first in https://github.com/riscv/riscv-cheri/issues/664

tariqkurd-repo:
We agreed to remove this in the TG meeting.

PR #1247, opened 2026-09-11
Capitalise headings, remove inline :leveloffset: directives

tariqkurd-repo:
Smarten up the document with consistent headings.
Easier maintenance without :leveloffset: directives so the text can be reordered if necessary.


tariqkurd-repo:
This one just tidies up the doc so merging quickly to allow others to be rebased

PR #1248, opened 2026-09-11
Combine widened CSRS and YLEN-bit CSRs

### LINKED ISSUE #1236: Combine YLEN CSRs with extended CSRs
The base ISA makes no distinction between YLEN CSRs (e.g. DDC) and extended CSRs (e.g. mtvec).
Zyhybrid _does_ though.
We should combine them both in the main specification and only distinguish in Zyhybrid.

tariqkurd-repo:
The difference between widened and YLEN-bit CSRs is restricted to the Zyhybrid mode description.
 
Fix #1236

### DIFF
--- a/src/cheri/insns/csrr_32bit.adoc
+++ b/src/cheri/insns/csrr_32bit.adoc
@@ -26,7 +26,7 @@ Encoding::
 include::wavedrom/csr-instr.adoc[]
 
 Description::
-These CSR instructions have extended functionality for accessing <<ylen_csrs>> and <<widened_csrs>>.
+These CSR instructions have extended functionality for accessing <<ylen_csrs>>.
 +
 Access to <<xlen_csrs>> is as defined in Zicsr.
 +
@@ -37,7 +37,7 @@ For the instructions on this page, all writes are XLEN bits only and use the sem
 +
 Read data from <<ylen_csrs>> is always YLEN bits.
 +
-Read data from <<widened_csrs>> is YLEN bits in {cheri_cap_mode_name} or XLEN bits in {cheri_int_mode_name}.
+If {cheri_default_ext_name} is implemented, and the current mode is {cheri_int_mode_name}, reads data from natively XLEN-bit CSRs is XLEN bits.
 
 Permissions::
 Accessing CSRs may require <<asr_perm>>.
--- a/src/cheri/insns/csrr_32bit.adoc
+++ b/src/cheri/insns/csrr_32bit.adoc
@@ -37,7 +37,7 @@ For the instructions on this page, all writes are XLEN bits only and use the sem
 +
 Read data from <<ylen_csrs>> is always YLEN bits.
 +
-If {cheri_default_ext_name} is implemented, and the current mode is {cheri_int_mode_name}, reads data from natively XLEN-bit CSRs is XLEN bits.
+If {cheri_default_ext_name} is implemented, and the current mode is {cheri_int_mode_name}, read data from natively XLEN-bit CSRs is XLEN bits.
 
 Permissions::
 Accessing CSRs may require <<asr_perm>>.
--- a/src/cheri/insns/csrrw_32bit.adoc
+++ b/src/cheri/insns/csrrw_32bit.adoc
@@ -24,17 +24,23 @@ Encoding::
 include::wavedrom/csrw-instr.adoc[]
 
 Description::
-CSRRW has extended functionality for accessing <<ylen_csrs>> and <<widened_csrs>>.
+CSRRW has extended functionality for accessing <<ylen_csrs>>.
 +
 Access to <<xlen_csrs>> is as defined in Zicsr.
 +
 CSRRW accesses to <<ylen_csrs>> read YLEN bits into `{cd}` and write YLEN bits of `{cs1}` into the CSR.
 +
-CSRRW accesses to <<widened_csrs>> read YLEN bits into `{cd}` and write YLEN bits of `{cs1}` into the CSR.
-However, if {cheri_default_ext_name} is supported and the current mode is {cheri_int_mode_name}, accesses to <<widened_csrs>> are XLEN-wide.
-The final write data for <<widened_csrs>> is determined using the semantics of the <<SCADDR>> instruction.
+When writing YLEN bits of `{cs1}`, if any <<section_cap_integrity,integrity>> checks fail then set the {ctag} to zero before writing to the CSR.
 +
-In all cases, when writing YLEN bits of `{cs1}`, if any <<section_cap_integrity,integrity>> checks fail then set the {ctag} to zero before writing to the CSR.
+NOTE: Any <<ylen_csrs, YLEN-bit>> CSRs may have additional rules defined to determine the final written value of the metadata and/or to write zero to the {ctag}.
+ifdef::cheri_standalone_spec[]
+These rules are shown in <<extended_CSR_writing_u>>, <<extended_CSR_writing_s>>, and <<extended_CSR_writing_m>>.
+endif::[]
+ifndef::cheri_standalone_spec[]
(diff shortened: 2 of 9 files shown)


src/cheri/insns/csrr_32bit.adoc
```diff
@@ -37,7 +37,7 @@ For the instructions on this page, all writes are XLEN bits only and use the sem
 +
 Read data from <<ylen_csrs>> is always YLEN bits.
 +
-Read data from <<widened_csrs>> is YLEN bits in {cheri_cap_mode_name} or XLEN bits in {cheri_int_mode_name}.
+If {cheri_default_ext_name} is implemented, and the current mode is {cheri_int_mode_name}, reads data from natively XLEN-bit CSRs is XLEN bits.
```
tariqkurd-repo:
```suggestion
If {cheri_default_ext_name} is implemented, and the current mode is {cheri_int_mode_name}, read data from natively XLEN-bit CSRs is XLEN bits.
```

src/cheri/riscv-hybrid-integration.adoc
```diff
@@ -105,7 +105,8 @@ Prefetches are also checked against the appropriate CSR (<<ddc>> for <<PREFETCH_
 +
 NOTE: Authorization of prefetches is useful to mitigate against side channels, e.g., preventing userspace from fetching kernel memory into a cache.
 +
-Reads to widened csrs are XLEN-wide, and writes use the semantics of the <<SCADDR>> instruction (see <<rvy_zicsr_mod_insn_table>>).
+Access to natively XLEN-bit revert to being XLEN-bit. +
```
tariqkurd-repo:
```suggestion
Access to natively XLEN-bit CSRs revert to being XLEN-bit. +
```

src/cheri/insns/csrr_32bit.adoc
```diff
@@ -37,7 +37,7 @@ For the instructions on this page, all writes are XLEN bits only and use the sem
 +
 Read data from <<ylen_csrs>> is always YLEN bits.
 +
-Read data from <<widened_csrs>> is YLEN bits in {cheri_cap_mode_name} or XLEN bits in {cheri_int_mode_name}.
+If {cheri_default_ext_name} is implemented, and the current mode is {cheri_int_mode_name}, read data from natively XLEN-bit CSRs is XLEN bits.
```
arichardson, line 40: We don't define natively xlen/YLEN csrs anywhere, I think we should add a linkable paragraph to make this clearer. 

I guess this would all be easier if intmode always did xlen accesses but we decided against that to avoid the need for modesw.

arichardson: I guess we already need the logic to see if you return XLen or YLEN in capmode, so having that still be present but slightly different in intmode shouldn't be too complex

tariqkurd-repo: it's true that non-extended CSRs return XLEN, extended + native YLEN return YLEN in cap mode and the extended XLEN change category in int mode.
I wasn't sure how to accurately describe XLEN ones which have been extended .

tariqkurd-repo: >I guess we already need the logic to see if you return XLen or YLEN in capmode, so having that still be present but slightly different in intmode shouldn't be too complex

we don't which is why we need to lose the distinction between native XLEN and widened XLEN CSRs.
what we _do_ have is reading XLEN from (e.g.) mtvec in int mode and YLEN from (e.g.) DDC which is where it gets awkward

It would be simpler if _all_ CSRs were accessed as XLEN in int mode - including DDC! So to program the whole capability do `YMODESWY; CSRRW DDC; WMODESWI`. I can't believe this would happen often, it would clean up the spec and it also removes the special case so makes it much easier to describe.

Thoughts?

tariqkurd-repo: it would be cleaner for the spec I think but best not propose a change like that at this late stage. I'll try to improve the wording.

tariqkurd-repo, line 40:
```suggestion
If {cheri_default_ext_name} is implemented, and the current mode is {cheri_int_mode_name}, read data from all CSRs declared by the RVI/RVE base architectures is XLEN bits.
```

src/cheri/insns/csrrw_32bit.adoc
```diff
@@ -24,17 +24,23 @@ Encoding::
...
-In all cases, when writing YLEN bits of `{cs1}`, if any <<section_cap_integrity,integrity>> checks fail then set the {ctag} to zero before writing to the CSR.
+NOTE: Any <<ylen_csrs, YLEN-bit>> CSRs may have additional rules defined to determine the final written value of the metadata and/or to write zero to the {ctag}.
+ifdef::cheri_standalone_spec[]
+These rules are shown in <<extended_CSR_writing_u>>, <<extended_CSR_writing_s>>, and <<extended_CSR_writing_m>>.
+endif::[]
+ifndef::cheri_standalone_spec[]
+These rules are shown in <<extended_CSR_writing_u>>, and there are further references in the _Privileged Specification_.
+endif::[]
++
+If {cheri_default_ext_name} is implemented, and the current mode is {cheri_int_mode_name}, writes to natively XLEN-bit CSRs are XLEN bits and the final write data is determined using the semantics of the <<SCADDR>> instruction.
```
tariqkurd-repo, line 43:
```suggestion
If {cheri_default_ext_name} is implemented, and the current mode is {cheri_int_mode_name}, writes to all CSRs declared by the RVI/RVE base architectures are XLEN bits and the final write data is determined using the semantics of the <<SCADDR>> instruction.
```

src/cheri/insns/csrr_32bit.adoc
arichardson: I agree that it would be simpler for the spec if ddc and *tidc were just xlen accesses in int mode and require modesw to get the full capability. I believe @jrtc27 was opposed to that though?

If we tell ARC that this simplifies the spec based on their request it's probably fine, but I agree that making such changes so late is bad (even though it only affects 2 CSRs).

src/cheri/insns/csrrw_32bit.adoc
arichardson: it's not just the RVI base architecture though, it's basically all CSRs except YLEN ones added by RVY + other RVY-based extensions.

tariqkurd-repo: in which case the original wording "natively" seems like the correct one, or we just keep all CSRs regular and read/write XLEN bits of DDC and xTIDC in int mode.

src/cheri/insns/csrr_32bit.adoc
jrtc27: I ask that people re-read everything in https://github.com/riscv/riscv-cheri/pull/354 and think about whether they still believe this is the right thing to do before we hash it out again.

tariqkurd-repo: I've reworded and believe that there's no need to change the spec

arichardson (commented):
Talking about "CSRs not added by {cheri_base_ext_name}" is almost more confusing that having the 3 categories. I think we should try to discuss this with ARC first?

tariqkurd-repo:
> Talking about "CSRs not added by {cheri_base_ext_name}" is almost more confusing that having the 3 categories. I think we should try to discuss this with ARC first?

* Natively XLEN CSRs
* CSRs declared as XLEN
* CSRs widened to YLEN
* RVI/RVE CSRs
* non-RVY CSRs
* declared in table X, Y, Z pointing to CSR tables in priv/unpriv
* list the extended CSRs by name, there aren't very many

any other options?

I've gone for "CSRs widened to YLEN-bit"

tariqkurd-repo:
This one is ready to merge

Issue #1237, opened 2026-09-10
ISA renaming and other queries


tariqkurd-repo:
YHIW / PACKY: Why do we have a second instruction name for the same operation?  Normally, pseudoinstructions are some special form of  an instruction with fewer explicit arguments - this looks identical to PACKY?  What would the disassembler print out when encountering this opcode?

YBASER / YLENR / YTOPR: Why have the R in these names?

SRLIY / YHIR: This consumes a lot of encoding space for a potential future enhancement that allows 128b values to be used as non-capability data bits.  Suggest only providing a way to read the upper XLEN bits now (i.e., explicitly provide YHIR only), using less encoding space, and worry about the other case later.

YEQ: This should be YSEQ to match other existing mnemonics for "set" on condition instructions (SLT,and friends).

YSS: For readability, suggest YSSUBSET, YSS is cryptic (and YSSUB is misleading)

YSENTRY: Mnemonic doesn't imply sealing - suggest YSEALE for the rs1=x0 case, and YSEAL for future general rs1 argument case.

YSUNSEAL: suggest YUNSEAL for the mnemonic (what was the additional S meant to connote?)

tariqkurd-repo:
#1232 #1250 #1251 

doesn't include: YBASER / YLENR / YTOPR: Why have the R in these names?

This are regular with e.g. YMODER, YMODEW for read and write so best confirm this one.

jrtc27:
Yes, please keep the R in those names. The lack of a W variant (which is handled by YBNDSW and friends) doesn't mean they shouldn't have an R.

jrtc27:
> YEQ: This should be YSEQ to match other existing mnemonics for "set" on condition instructions (SLT,and friends).

Are ARC aware of FEQ etc, which is what we tried to copy here?

jrtc27:
> YSS: For readability, suggest YSSUBSET, YSS is cryptic (and YSSUB is misleading)

Does this not contradict the prior point about aligning with other set-on-condition instructions which use a two character condition mnemonic?

arichardson:
[...]

PR #1250, opened 2026-09-12
Make YHIR a real instruction and remove SRLIY

arichardson:
Addresses ARC review feedback on SRLIY opcode space:
> This consumes a lot of encoding space for a potential future enhancement
> that allows 128b values to be used as non-capability data bits.  Suggest
> only providing a way to read the upper XLEN bits now (i.e., explicitly
> provide YHIR only), using less encoding space, and worry about the other
> case later.

Allocate YHIR as a 2-operand R-type instruction and remove SRLIY.

Partial fix for #1237


tariqkurd-repo:
Similarly for YHIW and PACKY, but those are cheap encodings as there's no shamt field.
Standard SRLI can have the immediate exended by 1 bit to allow for RV128 shifts, that bit appears unallocated.
I wonder if we could ever get a RV128 shift in the standard encoding space to replace SRLIY?

src/cheri/insns/wavedrom/gcmode.adoc
```diff
@@ -6,7 +6,7 @@
   {bits:  5, name: 'rd',         attr: ['5', 'dest'],             type: 2},
   {bits:  3, name: 'funct3',     attr: ['3', 'RVY-R=000'],        type: 8},
   {bits:  5, name: '{cs1}',      attr: ['5', 'src'],              type: 4},
-  {bits:  5, name: 'funct5',     attr: ['5', '{GCMODE}=00110'],   type: 3},
+  {bits:  5, name: 'funct5',     attr: ['5', '{GCMODE}=00111'],   type: 3},
```
tariqkurd-repo, line 9: Why move YMODER for YHIR?
I guess the encodings will all need updating when before we get approval but I'm unsure why moving YMODER is helping at this point - just more things changing when they don't need to (yet).

tariqkurd-repo (approved):
Just one minor comment about YMODER

src/cheri/insns/wavedrom/gcmode.adoc
arichardson: Since we have to change encodings anyway I thought it would make sense to keep the base ISA ones first then any extensions, but that does result in some extra changes.. @jrtc27 what do you think?

arichardson: You are right that it's probably better to just move it to the end?

arichardson:
> Similarly for YHIW and PACKY, but those are cheap encodings as there's no shamt field. Standard SRLI can have the immediate exended by 1 bit to allow for RV128 shifts, that bit appears unallocated. I wonder if we could ever get a RV128 shift in the standard encoding space to replace SRLIY?

For YHIW And PACKY I thought the request was just to drop the pseudo? We already encode it in the normal 3op range so no need for any changes there. 

I think the extra shift bit might end up being allocated for something else so probably better to just have a future YLEN arithmetic extension that allocated all these bits. Having a single 2 operand instruction that is technically the same as this shift shouldn't really matter, there are already tons of other redundant encodings

arichardson:
Moved it to the end to keep the differences minimal. The order of these encodings is arbitrary anyway.

PR #1251, opened 2026-09-13
Drop PACKY, just keep YHIW

arichardson:
Addresses ARC review comment:
"Why do we have a second instruction name for the same operation?
Normally, psuedoinstructions are some special form of an instruction with
fewer explicit arguments - this looks identical to PACKY? What would the
disassembler print out when encountering this opcode?"

Resolve this by using YHIW as the canonical instruction and dropping the
PACKY name so that all RVY instructions start with Y.
The internal attributes still use :SCHI: to minimize changes to the spec.

Partial fix for #1237


tariqkurd-repo:
merging with #1250 seems wise to remove both pseudos at once to avoid conflicts, otherwise looks good

arichardson:
> merging with #1250 seems wise to remove both pseudos at once to avoid conflicts, otherwise looks good

I'll land this one first since it's not an encoding change. And keep the YHIR separate.

tariqkurd-repo:
I think we should keep the YHIW name and not PACKY.
That wasn't clear to me from looking at the diffs, it's much more obvious from the actual doc.
Otherwise it looks very irregular.

arichardson:
> I think we should keep the YHIW name and not PACKY. That wasn't clear to me from looking at the diffs, it's much more obvious from the actual doc. Otherwise it looks very irregular.

It should have been obvious from the commit message :)

Sounds good to me, let's do that for now and then ask ARC for final confirmation

src/cheri/attributes.adoc
```diff
@@ -261,9 +261,9 @@ endif::support_varxlen[]
 :GCLEN_LC:     ylenr
 :GCLEN_DESC:   Read capability length
 
-:SCHI:         YHIW
-:SCHI_LC:      yhiw
-:SCHI_DESC:    Write capability metadata and clear {ctag} (pseudo)
+:SCHI:         PACKY
+:SCHI_LC:      packy
+:SCHI_DESC:    Pack Y register
```
tariqkurd-repo, line 266:
```suggestion
:SCHI:         YHIW
:SCHI_LC:      yhiw
:SCHI_DESC:    Write capability metadata and clear {ctag}
```

tariqkurd-repo (approved):

jrtc27 (approved):
I always hated PACKY as a name... :)

tariqkurd-repo:
@arichardson needs fixing as #1250 was merged

arichardson:
I am waiting for arc confirmation that we are ok with yhiw instead of packy before merging. but yes let me fix the conflicts

arichardson:
Using YHIW over PACKY was approved by ARC in their latest meeting. https://docs.google.com/document/d/1tnxDV9vV1K3Hoo7YLiTrublWOXO1OnPEMp5S1m_UfkQ/edit?resourcekey=0-WywAOL40tUzvQrWV3nsHTw&tab=t.0

PR #1252, opened 2026-09-14
Clarify that reserved metadata values depend on implemented extensions

### LINKED ISSUE #1240: Integrity check section comments
> 2. The capability metadata does not have any bits, or fields, set to a reserved value.

These rules are too strict and will break when future extensions make use of reserved encodings.  The text should make clearer that future extensions will make reserved encodings legal.

> All other integrity checks are optional as a valid capability can always be assumed to satisfy these integrity
rules, because it could only have been produced by a valid derivation. Consequently, an integrity check
should only change the result of an operation when it is performed on an invalid input capability (i.e.
whose capability tag is zero).

This paragraph (and this section) are confusing.  The intent is to convey that valid capabilities should never have non-integrity-check-passing values unless there are hardware faults or hardware design errors, so implementations are not strictly required to check them.  However, for hygiene, designs might do additional integrity checking.  Somewhere should have a table giving names to states of a capability (valid, invalid, integrity-checked, malformed bounds, etc.)

arichardson:
Addresses ARC review comment:
"These rules are too strict and will break when future extensions make
use of reserved encodings. The text should make clearer that future
extensions will make reserved encodings legal."

Add a note explicitly points out that future extensions may allocate values.

Partially fixes #1240 


tariqkurd-repo (approved):

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -631,9 +631,10 @@ NOTE: For {cheri_base64_ext_name}, the number of spare bits in the capability-en
 CHERI defines the following integrity check rules:
 
 . The bounds are not malformed (a capability encoding format-specific check; see xref:section_cap_malformed[xrefstyle=short] for {rv64y_uni_base_name}).
-. The capability metadata does not have any bits, or fields, set to a reserved value.
+. The capability metadata does not have any bits, or fields, set to a value reserved by the implemented extensions.
```
kasanovic, line 634: This is just saying the same thing which is a problem for future standard extensions. "The capability metadata does not have any bits, or fields, set to a value that is not valid for the implemented extensions"

PR #1254, opened 2026-09-15
Clean up definition of root capabilities

### LINKED ISSUE #1241: root capability comments
https://github.com/riscv/riscv-cheri/blob/87464822c547f58385fc099f55023e0bae73d22e/src/cheri/riscv-unpriv-integration.adoc?plain=1#L545

If the root capabilities aren't made availabe to software, then surely they don't exist?

https://github.com/riscv/riscv-cheri/blob/87464822c547f58385fc099f55023e0bae73d22e/src/cheri/riscv-unpriv-integration.adoc?plain=1#L551

Why should a root executable cover all addressses - would expect that normally it is constrained?

tariqkurd-repo:
I've tried to rationalise the usage of "root capabilities" so that they _only_ refer to capabilities which give maximum bounds /permissions and not any of the other uses - i.e. the ones available after reset or provisioned for usermode software.

Fix #1241


src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -540,34 +540,39 @@ NOTE: Software is completely free to define the usage of these bits.
...
-Upon initialization, the execution environment defines how root capabilities are made available to software.
-+
-NOTE: Implementations are not required to provide access to the entire address space to a single hart, the _root_ capabilities may be restricted to only cover a subset of the entire address space, or to remove certain permissions.
+===== Root Capabilities
+
+_Root_ capabilities are the capability encodings for the maximum set of bounds, covering the whole 2^XLEN^ address space, and that grant all permissions.
+
+Some capability-encoding formats like {rv64y_uni_base_name} provide the encoding for a single <<section_infinite_cap_encoding,Infinite>> capability which provides all the characteristics of root capabilities.
+
+Other capability-encoding formats may have a _set_ of root capabilities distinguishing between data and executable capabilities.
```
```suggestion
Other capability-encoding formats may have a _set_ of root capabilities.
For example, they could distinguish between data and executable root capabilities.
```

you could also have more - and CHERIoT also has sealing roots.

tariqkurd-repo:
```suggestion
Other capability-encoding formats may have a _set_ of root capabilities distinguishing between at least data and executable capabilities.

NOTE: Future extensions may add additional root capabilties for purposes such as sealing and unsealing.
```

we introduce those two exact types and use them, so they can't be "for example"

src/cheri/riscv-priv-integration.adoc
```diff
@@ -136,7 +134,7 @@ include::img/mtidcreg.edn[]
 When the base ISA is {cheri_base_ext_name}, the Y bit is one.
 
 `<<misa_y,misa>>.{CRE}` is a WARL field.
-If any of <<pcc>>, `__x__tvec`, or `__x__epc` is not currently configured as a root capability then zero is not a legal value for <<misa_y,misa>>.{CRE}.
+Zero is only a legal value for <<misa_y,misa>>.{CRE} if the capability metadata of <<pcc>>, `__x__tvec`, and `__x__epc`  are currently configured as the reset value.
```
```suggestion
Zero is only a legal value for <<misa_y,misa>>.{CRE} if the capability metadata of <<pcc>>, `__x__tvec`, and `__x__epc`  are currently configured as their respective reset values.
```
Maybe this?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -540,34 +540,41 @@ NOTE: Software is completely free to define the usage of these bits.
...
-in such systems, _root_ capabilities are often _Infinite_.
-+
-Other capability-encoding formats may have a set of _root_ capabilities distinguishing between data and executable capabilities as shown below.
-+
-Upon initialization, the execution environment defines how root capabilities are made available to software.
-+
-NOTE: Implementations are not required to provide access to the entire address space to a single hart, the _root_ capabilities may be restricted to only cover a subset of the entire address space, or to remove certain permissions.
+===== Root Capabilities
+
+_Root_ capabilities are the capability encodings for the maximum set of bounds, covering the whole 2^XLEN^ address space, and that grant all permissions.
```
arichardson: They don't necessarily grant all permissions. It's a maximum set of permissions and bounds for each given root. Not sure how we can word this cleanly.

tariqkurd-repo:
```suggestion
_Root_ capabilities are the capability encodings for the maximum set of bounds, covering the whole 2^XLEN^ address space, and that grant all encodable permissions.
```

like this?

nwf (approved):
Seems fine

src/cheri/riscv-priv-integration.adoc
```diff
@@ -925,7 +923,8 @@ The reset values are:
 * `<<menvcfg_y,menvcfg>>.{CRE}=0`
 * `<<senvcfg_y,senvcfg>>.{CRE}=0`
 * <<pcc>>.<<p_bit>>={INT_MODE_VALUE} (for {cheri_int_mode_name}).
-* <<ddc>>, <<pcc>>, `__x__tvec`, `__x__epc` are nominally <<root-cap,root capabilities>>.
+* <<ddc>> is nominally a <<root-rw-cap>> capability.
```
arichardson: I still find the nominally quite confusing, can we just drop it? Isn't this machine level reset, if the bootloader does something else that's fine too right?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -204,8 +204,8 @@ NOTE: The _length_ and _top_ are (XLEN+1)-bit values, so <<GCLEN>> and <<GCTOP>>
 
 ==== Deriving New Bounds
 
-Upon initialization, the execution environment defines how the <<root-cap>> capabilities are made available to the execution context.
-All capabilities with smaller bounds are derived from these.
+Upon initialization, the execution environment defines how capabilities are made available to the execution context.
```
arichardson, line 212:
```suggestion
Upon initialization, the execution environment defines how initial capabilities are made available to the execution context.
```

maybe add initial?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -540,34 +540,41 @@ NOTE: Software is completely free to define the usage of these bits.
...
-Other capability-encoding formats may have a set of _root_ capabilities distinguishing between data and executable capabilities as shown below.
-+
-Upon initialization, the execution environment defines how root capabilities are made available to software.
-+
-NOTE: Implementations are not required to provide access to the entire address space to a single hart, the _root_ capabilities may be restricted to only cover a subset of the entire address space, or to remove certain permissions.
+===== Root Capabilities
+
+_Root_ capabilities are the capability encodings for the maximum set of bounds, covering the whole 2^XLEN^ address space, and that grant all encodable permissions.
+
+Some capability-encoding formats like {rv64y_uni_base_name} provide the encoding for a single <<section_infinite_cap_encoding,Infinite>> capability which provides all the characteristics of root capabilities.
```
arichardson: I think this is a bit backwards now? We are careful in the rest of the spec to no longer use infinite capability except in the RV64 encoding chapter and instead always use root-rw or root-rx. A single NOTE: saying some capability encodings provide a single capability bit pattern that covers all root capabilities, and such a capability is reffered to as the infinite capability.

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: done but doesn't seem to be showing here?!

src/cheri/riscv-unpriv-integration.adoc
tariqkurd-repo: yep - done - I've removed "nominally" and sorted out the defintions of infinite bounds and the infinite cap

src/cheri/riscv-priv-integration.adoc
tariqkurd-repo: I think this is ok now - I've updated the text.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -540,34 +545,39 @@ NOTE: Software is completely free to define the usage of these bits.
...
+An unsealed capability that has <<infinite_bounds,infinite bounds>> and grants at least all of <<r_perm>>, <<w_perm>>, <<c_perm>>, and <<lm_perm>>.
+
+NOTE: Extensions introducing new permissions may require these to be provided by <<root-rx-cap>> and/or <<root-rw-cap>> capabilities.
+
+Upon initialization, the execution environment defines how the initial set of capabilities are made available to software.
+
+The definitions of <<root-rx-cap>> and <<root-rw-cap>> are the maximum possible encodable values.
+
+Implementations are not required to provide access to the entire address space to a single hart or even to provide all permissions.
+As such, where a root capability value is specified, harts may specify implementation defined values instead which reduce the bounds or permissions as required.
```
```suggestion
As such, where a root capability value is specified, harts may specify implementation defined values instead with reduced bounds and/or permissions as required.
```
maybe like this?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -540,34 +545,39 @@ NOTE: Software is completely free to define the usage of these bits.
...
+
+NOTE: The hart specific root capabilities may be the maximum that the hart can ever have access to, or may allow access to capabilities in memory with greater bounds or permissions.
```
arichardson: I know what you are trying to say here but I as it stands this is a bit confusing. I _think_ what you are trying to say is that the initially available capabilities might grant less privilege than some caps stored elsewhere in memory. I think maybe the problem is the use of root here. root is by definition the maximum
```suggestion
NOTE: The hart specific initial capabilities may be the maximum that the hart can ever have access to, but there could also be capabilities in memory with greater bounds or permissions.
```

arichardson (approved):
minor comments but LGTM.

PR #1255, opened 2026-09-15
confirm that a failed SC doesn't clear the tag

### LINKED ISSUE #1234: Specify want happens to the tag in memory on a failing SC
_What happens to the capability tag in memory when a regular SC fails? We'd expect it would leave it alone but should be spelled out explicitly.._

ARC feedback.
To my mind if the store fails then the memory is untouched so I propose saying exactly that, as ARC suggest.

Any comments?

tariqkurd-repo:
Fix #1234 
ARC requested clarification.


jrtc27 (changes requested):
I still believe this is the wrong change. If this is not unambiguously specified by the rest of the specification then we have done something wrong. A non-normative clarification *could* be ok, but only to augment / remind / reinforce what should already be unambiguously implied, not to fill a void.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -162,6 +162,8 @@ NOTE: Examples of such invalid operations include writing only the integer porti
 The implementation must set the {ctag} to zero if any byte in the naturally aligned YLEN-bit memory region is ever written using an operation other than a store of a capability operand.
 For example, the <<STORE_CAP>> instruction sets the {ctag} to one if all the instruction's prerequisites are met, including `{cs2}` 's {ctag} being set to one, but <<ldst,SW>> will always set the stored {ctag} to zero.
```
jrtc27: My (possibly old) copy of the ISA manual says the following for SC.W:

> SC.W conditionally writes a word in _rs2_ to the address in _rs1_: the SC.W succeeds only if the reservation is still valid and the reservation set contains the bytes being written. If the SC.W succeeds, the instruction writes the word in _rs2_ to memory, and it writes zero to _rd_. If the SC.W fails, the instruction does not write to memory, and it writes a nonzero value to _rd_.

I think that is pretty clear that, if the SC fails, there is no write, and therefore this does not apply. One could add SC as an *example* ("Since a failed SC does not perform a write, ..."), or as a "NOTE: A failed SC does not perform a write, and so ...", but adding it like an extra rule is wrong as it misleads, since it's not a new special case.

tariqkurd-repo: ok - try again - now saying "successful store" and downgraded to a NOTE

arichardson (approved):
New wording looks good to me, @jrtc27 does that sound good?

PR #1253, opened 2026-09-14
Simplify the optional integrity check wording

arichardson:
Addresses ARC review comment:
"This paragraph (and this section) are confusing.  The intent is to
convey that valid capabilities should never have non-integrity-check-
passing values unless there are hardware faults or hardware design
errors, so implementations are not strictly required to check them.
However, for hygiene, designs might do additional integrity checking.
Somewhere should have a table giving names to states of a capability
(valid, invalid, integrity-checked, malformed bounds, etc.)"

Try improving the commented-on section to make it less confusing. I
tried adding the requested table but it just made things worse IMO, so
hopefully this minor rewrite is sufficient.

Partially fixes #1240 

Signed-off-by: Alex Richardson <alexrichardson@google.com>

PR #675, opened 2025-07-28
Address latest round of unpriv ARC feedback

arichardson:
Added #ARC-COMMENT# TODOs for the parts I have not yet addressed.

PR #1256, opened 2026-09-16
Improve integrity check text.

tariqkurd-repo:
This paragraph (and this section) are confusing.  The intent is to convey that valid capabilities should never have non-integrity-check-passing values unless there are hardware faults or hardware design errors, so implementations are not strictly required to check them.  However, for hygiene, designs might do additional integrity checking.  Somewhere should have a table giving names to states of a capability (valid, invalid, integrity-checked, malformed bounds, etc.)

First step - define and use the term _invalid capability_.
Second step - add requested table.

Based on #1253


src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -141,10 +141,13 @@ For capability loads and stores, provenance is preserved across memory:
 When a memory access fails a check, either due to software error or malicious intent, the operation raises an exception.
 For other instructions a failed check will typically set the resulting {ctag} to zero, the exact behavior is defined by each instruction.
 
-For a capability to be valid, the {ctag} must be set to one, otherwise a capability is invalid.
+* Valid capabilities have the {ctag} set to one.
+** The term _validating a capability_ means deriving a new capability with the {ctag} set to one.
```
arichardson, line 145: Using this for setting tag to one is a bit odd since "validating" to me means checking if something is valid, not constructing a capability with `tag=1`.

arichardson: Also I don't think this is used anywhere so can just be dropped?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -141,10 +141,13 @@ For capability loads and stores, provenance is preserved across memory:
...
+* Invalid capabilities have the {ctag} set to zero.
+** The term _invalidating a capability_ means deriving a new capability with the {ctag} set to zero.
```
arichardson, line 147:
```suggestion
** The term _invalidating a capability_ means deriving a new capability with identical metadata and address but the {ctag} set to zero.
```

I'm not sure if we should be using this instead of set to zero. "set the {ctag} to zero" makes it obvious that the non-tag bits are preserved and only the tag bit changes, whereas "invalidate the capability" / "store invalidated data capability" could sound like the metadata might also be modified/zeroed and you have to look back at the original definition. the current verbose (and somewhat ugly) wording is at least obvious.

arichardson:
Using "valid/invalid capability" instead of "capability with its {ctag} set to one/zero" when referring to operands is a nice cleanup (and ARC said "valid capability" was fine, see #675, where we changed from valid/validity tag to capability tag).

I am not sure we should replace "set the {ctag} to zero" with "invalidate" or add the state table right now. We changed everything to "set the {ctag} to one/zero" in #1015 and #1219 because of ARC feedback, so maybe better to keep it that way?

For the table: I think it's better without a table, since valid (`tag=1`) and invalid (`tag=0`) are the only two architectural states. Integrity failures and malformed bounds are just valid capabilities that shouldn't really exist. Also the table is a bit self contradictory: we define "valid capability" as `tag=1` but the table says `Valid capability` means all integrity checks passed and makes `Integrity fail` (`tag=1`) a separate state. 

I'd say let's wait and see if ARC is happy with what we did in #1252 and #1253 first. And maybe just do the cleanup of using `valid/invalid capability` and defining those two terms, but without the extra changes?

src/cheri/riscv-unpriv-integration.adoc
```suggestion
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -580,7 +582,7 @@ NOTE: Future extensions may add additional root capabilities for purposes such a
 
 [#null-cap,reftext="NULL"]
 ===== NULL Capability
-A capability with all-zero metadata, its {ctag} set to zero, and an address of zero is referred to as the <<section_null_cap_encoding,NULL>> capability.
+A invalid capability with all-zero metadata and an address of zero is referred to as the <<section_null_cap_encoding,NULL>> capability.
```
```suggestion
An invalid capability with all-zero metadata and address is referred to as the <<section_null_cap_encoding,NULL>> capability.
```

tariqkurd-repo:
> Using "valid/invalid capability" instead of "capability with its {ctag} set to one/zero" when referring to operands is a nice cleanup (and ARC said "valid capability" was fine, see #675, where we changed from valid/validity tag to capability tag).
> 
> I am not sure we should replace "set the {ctag} to zero" with "invalidate" or add the state table right now. We changed everything to "set the {ctag} to one/zero" in #1015 and #1219 because of ARC feedback, so maybe better to keep it that way?
> 
> For the table: I think it's better without a table, since valid (`tag=1`) and invalid (`tag=0`) are the only two architectural states. Integrity failures and malformed bounds are just valid capabilities that shouldn't really exist. Also the table is a bit self contradictory: we define "valid capability" as `tag=1` but the table says `Valid capability` means all integrity checks passed and makes `Integrity fail` (`tag=1`) a separate state.
> 
> I'd say let's wait and see if ARC is happy with what we did in #1252 and #1253 first. And maybe just do the cleanup of using `valid/invalid capability` and defining those two terms, but without the extra changes?

My opinion is:

1. The ARC complaints (IIRC) were about "set the tag/clear the tag/tagged/untagged" and they wanted to have the value specified in all cases to make it unambiguous
2. Valid/invalid capability is so much easier to read than "a capability with its tag set to one/zero" every time, which also follows the convention set by Morello and of course CHERI v9
3. The table was explictly asked for, but I see your concern and have split it into architecturally valid and architecturally invalid states which hopefully is better
4. The process has gone on far too long so I'd like to shoot for (what I hope is) the best result, let's not purposefully add another iteration

We can certainly add replacing "with the tag set to zero" with "invalid capability" to the list of discussed points.

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -677,6 +679,22 @@ Implementations may choose to enforce optional integrity checks for a number of
...
+|Valid capability   | 1  | All integrity checks have passed.
+|Invalid capability | 0  | The YLEN-bit payload is not considered to be a capability unless authorized by <<CBLD>>.
+|=======================
+
+.Architecturally Invalid Capability States
+[#table_invalid_cap_states,%autowidth,options=header,align="center",cols="2,1,4"]
+|=======================
+|State              |{ctag_table} | Description
+|Top overflow       | 1  | All integrity checks have passed, but a logic fault has caused an undetected top overflow.
+|Integrity fail     | 1  | Any fields are in a reserved state, or bounds are <<section_cap_malformed,malformed>>.
```
```suggestion
|Integrity fail     | 1  | At least one metadata field is in a reserved state, or the bounds are <<section_cap_malformed,malformed>>, or both.
```

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -677,6 +679,22 @@ Implementations may choose to enforce optional integrity checks for a number of
 
 ====
 
+.Architecturally Valid Capability States
+[#table_valid_cap_states,%autowidth,options=header,align="center",cols="2,1,4"]
```
arichardson: Doesn't this duplicate the bullet points above?

src/cheri/riscv-unpriv-integration.adoc
```diff
@@ -649,34 +651,49 @@ CHERI defines the following integrity check rules:
...
 
-NOTE: In the {cheri_base_ext_name} ISA, this mandatory integrity check only applies to <<CBLD>>.
+NOTE: In the {cheri_base_ext_name} ISA, this mandatory integrity check only applies to the `{cs2}` input to <<CBLD>>.
 
 [NOTE]
 ====
 Implementations may choose to enforce optional integrity checks for a number of reasons:
 
-. Instruction level verification is simpler as the behavior is precisely specified for all 2^(YLEN+1)^ input values.
+. Instruction-level verification is simpler as deterministic behavior is specified for all 2^(YLEN+1)^ input values (similar to how the D extension defines deterministic behavior for improperly NaN-boxed register values).
```
arichardson: Not sure if the paren expression here should be a NOTE: ?

I just wanted to show ARC there is something a bit similar in the existing spec.

tariqkurd-repo:
I've got Claude to fix the signature problem from your commits.
I think we should merge this one now, we've tinkered with it enough.

arichardson (approved):

PR #1257, opened 2026-09-17
Bump NEXT_VERSION to v0.9.10

arichardson:
We have made semantic changes to the specification since the initial
0.9.9 version sent to ARC (for example, dropping the reserved irregular
branch encodings in #1246), so we need to bump the document version.

This version bump is safe for all toolchains as it will be emitted in
ISA strings as rv64y0p910, which compares greater than the prior 0p99.


arichardson:
As discussed in the last TG meeting.

PR #1259, opened 2026-09-17
Simplify pointer masking spec

### LINKED ISSUE #805: pointer masking and executable memory
We're concerned that pointer masking as defined means that a single pointer will mean different things to instruction memory and data memory.
The RISC-V pointer masking extensions apply to data memory only, not to instruction memory. Instruction fetch never masks the pointer.

The RVY spec currently says: 

`When bounds are encoded or decoded, a masked but _not_ sign extended address is used.`

and

`For the representable range check, both the original and new addresses are masked.`
`Bounds setting instructions also mask the address in the same way.`

So - to my mind - this is troubling. If I have an RWX capability where the upper PMLEN bits are 1's, then they'll be interpreted as 1's for the PC bounds check and 0's for the load/store bounds check. 

Therefore the same capability means more than one thing - surely this isn't correct?

It's got me thinking that if a capability has X permission then pointer masking should be disabled for all accesses using it. Note that setting MXR disables pointer masking in the same way.

But then.... do representable range checks for ADDY also need to check X permission? to decide whether to mask the pointer?

it seems to be that the spec isn't very clear.

tariqkurd-repo:
Fix #805 

It seems to me that the pointer masking spec should be simple.
Apply the ignore transformation to representable range checking and bounds decoding as currently specified by the pointer masking extensions.


arichardson:
There is a lot of subtlety here and it's easy to get it wrong (see Morello) so I'd need some more time to actually review this

tariqkurd-repo:
> There is a lot of subtlety here and it's easy to get it wrong (see Morello) so I'd need some more time to actually review this

this one is low priority but my feeling is that this spec is less broken than the existing one so it's an improvement.

src/cheri/ptrmask-integration.adoc
```diff
@@ -6,37 +6,65 @@ endif::[]
...
+This affects <<GCBASE>>, <<GCLEN>>, and <<GCTOP>>.
+
+=== Interaction With Capability Subsets
+
+The definition of <<cap_subset>> is extended to apply the transformation to both `cap1` and `cap2` before the bounds calculation.
+
+This affects <<CBLD>>, <<SCSS>> and <<YSUNSEAL>>.
+
+WARNING: Because dynamically changing the number of masked bits changes the interpretation of a capability,
+software must take the same care when sharing capabilities between address spaces with differing pointer masking settings as it generally must when sharing capabilities between address spaces with different page mappings.
```
tariqkurd-repo:
```suggestion
WARNING: Because dynamically changing the number of masked bits changes the interpretation of a capability, software must take the same care when sharing capabilities between address spaces with differing pointer masking settings as it generally must when sharing capabilities between address spaces with different page mappings.
```

src/cheri/ptrmask-integration.adoc
```diff
@@ -6,37 +6,65 @@ endif::[]
...
+
+WARNING: Because <<pcc>> bounds aren't masked but data memory access bounds are, if the PMLEN upper bits of a capability address are modified by the ignore transformation then the capability will only be usable for data memory access, as the address will be invalid for instruction fetch.
+
+WARNING: A capability with its address in the illegal VA space while PMM=0 will have its address and bounds move into the valid VA space if PMM is subsequently set to a nonzero value.
```
tariqkurd-repo:
```suggestion
WARNING: A capability with its address in the illegal VA space while PMM=0 may have its address and bounds move into the valid VA space if PMM is subsequently set to a nonzero value.
```

src/cheri/ptrmask-integration.adoc
```diff
@@ -6,37 +6,65 @@ endif::[]
...
+Such a capability is only possible if capabilities are available which have at least some of the gap between valid VA spaces in their representable ranges.
+It is recommended that software should not create such capabilities.
+
```
tariqkurd-repo:
```suggestion
It is recommended that software should not create such capabilities.
If PMLEN is less than the number upper address bits which represent the VA gap then enabling pointer masking can either cause bounds to move between invalid address ranges within the VA gap or from the VA gap into the valid space.
```

PR #1260, opened 2026-09-18
add note about [SH]FENCE.* and [SH]INVAL.*

### LINKED ISSUE #1258: Not all instructions taking an address as an operand take a capability
SFENCE.VMA
SFENCE.INVAL.IR
HINVAL.VVMA
SINVAL.VMA

...all have a VA in rs1.

This is currently unaffected by CHERI, so page table invalidations aren't capability controlled.
I don't think this is an oversight - afterall the PTW isn't capability controlled.

Is this something we're likely to want in the future? Is it worth having a NOTE: in the spec about it?

Because we haven't said they take capabilities then they're integer ops, so a capabilty _can_ be passed to them but without any checking.

tariqkurd-repo:
Important to note instructions which take an address operand but not a capability.
Fix #1258


tariqkurd-repo:
This one is ready to merge

